PR #1, opened 2026-08-06
Bench harness: cross-language serialize profiling + C++ baseline on two machines

rowan-claude:
The cross-language serialize profiling harness Glenn asked for this morning, plus the C++ baseline on two machines. Go/rust/cs runners plug in when the ports land today — the contract is written down and `run.sh` already detects them.

## What this is

- `bench/run.sh` — builds and runs whichever language runners exist, collects one CSV per host with compiler, exact flags, pinning and a noise label in the preamble. `make bench` wraps it.
- `bench/cpp/bench_main.cpp` — the reference runner: write and read paths, msgs/s and MB/s, over the pinned corpus instances (byte-checked against `testdata/wire/*.bin` and round-tripped before any number is produced) plus a 4096-message mixed batch through `WriteMessage`/`ReadMessage` for steady-state dispatch.
- `bench/{go,rust,cs}/README.md` — clearly-marked stubs with the exact runner contract (same pins, same LCG, same vary mappings, same CSV format).
- `bench/results/` — CSVs for both hosts and `2026-08-06-baseline.md` with the two tables, prediction scoring, and evidence-backed hotspots.

## Methodology

The const-params experiment's discipline, kept on purpose: escape barriers, per-iteration LCG variation (so the optimizer cannot fold the write loop), 64 variant read buffers, matched Release+Debug pairs. Changed from best-of-5 to median-of-7 with min/max/spread. One deliberate flag divergence from serialize's own bench: no `-ffast-math` (this repo pins wire determinism with `-ffp-contract=off`).

Hosts: Apple M2 (arm64, Apple clang 21, quiet, unpinned — macOS) and the space box, which is an **EPYC 9124** (Zen 4), not the Ryzen the morning notes said. The game server was running with one pinned thread per every isolated core 1-15, so per the guest rules the bench ran `taskset -c 0` on the shared core and all EPYC numbers are labelled **NOISY**; g++ 13.3 primary, clang 18.1 beside it.

## Headline numbers (Release medians)

- M2 / clang 21: writes 17-1040 M msg/s across the corpus; batch dispatch 74 M msg/s write, 30 M msg/s read.
- EPYC / g++ 13: writes 9-139 M msg/s; batch 65 write / 20 read.
- EPYC / clang 18: up to **4x faster than g++ on the same machine** for tiny-message writes (test 505 vs 139 M msg/s, probe_header 496 vs 124) — g++ leaves `BitWriter::WriteBits`/`FlushBits`/`serialize_double_internal` outlined per field (perf-annotate evidence in the results doc; the const-params inliner-spill mechanism at emitted-code scale).

Predictions were committed before each measurement and scored in the results doc; the biggest refutation: reads do NOT beat writes across the board for generated code — validation, branch zeroing and Message-union zeroing sit on the read side.

## Hotspots seeded for the optimize pass (nothing optimized here)

1. `ReadMessage` zeroes the Block-sized union (~2 KB) per ~25 B message — 60.6% of batch-read self-cycles are `rep stos` on EPYC; `_platform_memset` second-overall in the M2 sample. Batch read is 2.4-3.2x slower than write everywhere because of it.
2. The g++ outlining above — forced inlining of hot stream ops or the const-params compile-time surface is the lever; this harness measures it either way.
3. Byte/string paths: `fixed_bytes` rides `write_bits(...,8)` per byte; string reads re-walk for the interior-null scan. testdata is the slowest benchmark on every host.

## Findings in passing

- The generated C++ is **not g++ -Werror clean**: `-Wclass-memaccess` (branch zeroing memsets non-trivial structs, `Types.h` ReadRigidBody) and `-Wtype-limits` (`read_int64` on a uint32 full-range field, `Wire.h` ReadProbeBits). `run.sh` suppresses both for g++ with a comment; emitter fix tracked separately. Apple clang flags neither, which is why `make` never caught it.
- All 11 wire goldens verify byte-identical on x86_64 Linux under both compilers — first cross-OS confirmation of the corpus pins outside macOS.

## Notes

- GitHub Actions is in a major outage today — no CI on this PR; everything above ran locally on both hosts (commands and full CSVs in `bench/results/`).
- First commit's message notes a re-commit: the shared checkout collided with the fixed-uint128 session mid-flight; their branch was restored untouched and this branch carries bench files only, built from a worktree since.


gafferongames:
2568056 extends this PR with the Go, Rust, and C# runners and the first four-language comparison.

- All three ports implement the full runner contract: golden + round-trip self-check gates every runner (a corpus mismatch refuses to bench), same pins / LCG / vary mappings / batch builder as the C++ reference, warmup + 7 runs + median-of-7, shared CSV schema. run.sh's SKIP lines retire; its preamble now records go/rust/dotnet versions.
- Same-sitting four-language runs on both hosts (M2 quiet; EPYC 9124 core-0 NOISY, user-local toolchains under ~/rowan-bench): every golden gate held on both. Tables + predictions-vs-outcomes: bench/results/2026-08-06-four-language.md.
- GitHub Actions is still in a major outage — this, like the rest of the PR, was built, gated, and verified locally on both hosts; CI has seen none of it. Re-check the corpus gates in CI once Actions recovers.

rowan-claude:
Two harness fixes measured and landed today, plus the re-run: `be0ec4f` + `8abe38c`, tables in `bench/results/2026-08-06-four-language-v2.md` (both host CSVs beside it; v1 files untouched).

## Fix 1 — the read loops were benchmarking the allocator

Every C++-reference-derived read loop declared a fresh decode instance per iteration while the write loops hoisted theirs. In go that instance escapes through the opaque read function value and is heap-allocated + zeroed EVERY message — `runtime.mallocgc` + GC were ~27% cumulative of the v1 go read profile. In cpp it is a constructor-zeroed stack temporary (a full ~2 KB `Message` memset per dispatch read in the batch); in rust a stack-zeroed `T::default()` per iteration. cs already decoded into one reused instance — v1's only honest read column.

All four read loops now hoist ONE reused decode instance (the write loops' existing discipline). Proof is printed by the runners themselves per bench — one extra untimed 256-op pass per path: **go `ReadMemStats` Mallocs delta = 0 write / 0 read on every bench and `message_batch`, both hosts; cs `GC.GetAllocatedBytesForCurrentThread` delta = 0 bytes both paths, both hosts**; rust and cpp are heap-free by construction (loop shape verified — no per-iteration clone, no per-iteration ctor). Reuse decodes identically: structure fields are fixed across variants and the cpp dispatch read re-establishes the selected arm itself at selection.

**v2 read numbers are not comparable to v1 reads.** Writes are the same-sitting control: 0.94–1.12x of v1, mostly within a couple percent.

## Headline ranking changes

- **The batch-read crown was a harness artifact — cpp takes it on both hosts.** M2: cpp 67.24 > cs 35.60 > go 31.15 > rust 20.61 M msg/s (v1: cs > go > cpp > rust). EPYC: cpp 41.64 > go 31.78 > cs 21.55 > rust 16.15 (v1: go > cs > cpp > rust). v1 charged cpp a ~2 KB Message construction per read while go/cs read into reused storage. cs stays the fastest *managed* batch reader on the M2; go keeps that title on the EPYC. rust is unchanged by design — `read_message` still returns the enum by value; the `read_message_into` experiment stays the open seed.
- **Go's read deficit was mostly the allocator**: reads rose 1.14–3.24x (M2) and 1.07–2.43x (EPYC) — chat read 3.24x. Go now reads faster than it writes on chat/test/shipcreate/ship_shallow/probebits/batch; it still reads slower on rigidbody (0.74–0.78x), inputpacket (0.65–0.68x), probearray, testdata — float-heavy/nested shapes, real read-path work, the profile-first seed for the go port.
- **cpp reads** rose where the decode instance is big/non-trivial (inputpacket 1.29–1.62x, EPYC chat 2.23x, batch ~2.1x); on the EPYC every cpp read now beats its write. **rust ~flat** (0.87–1.09x — only stack zeroing removed, as predicted by construction). **cs is the control**: 0.87–1.00x with its writes drifting identically.

## Fix 2 — the never-selected-arm poke (owed from schema PR #3's report)

`check_message_stream_golden` in the cpp runner poked fields of a never-selected union arm on a fresh `Message`. Now it selects arms by assignment (`m.chat = Chat{};` / `m.test = Test{};`) first — required under PR #3's contract (construction initializes only the tag; an arm is established zeroed when selected), equally valid on current main (ctor memsets the union). Wire bytes identical under both; the `message_stream` golden held on both hosts. Same two lines as `e22ce86` on the local measure branch — PR #1 is now merge-ready against #3's semantics.

## Notes

- GitHub Actions was in its major outage (recovering) — both hosts ran and verified locally; the golden + round-trip gates held everywhere.
- The EPYC had several concurrent sessions pinning benchmarks to core 0 today; a contaminated first attempt (spreads to 65%) was discarded and the final run launched behind a process-inspection quiet gate. Write-control confirms clean conditions.
- Heads-up for the README relative-performance table (086ee2d..93f936b): it is derived from the v1 M2 numbers — the read-side ratios are superseded by v2 (e.g. the batch-read column inverts). Left untouched here since this change is bench/-only.

PR #4, opened 2026-08-06
PROPOSAL — DO NOT MERGE: tail-only union-arm zeroing, priced per SPEC §5

rowan-claude:
**PROPOSAL — DO NOT MERGE. DECISION IS GLENN'S.** No code changes on this branch: one document, `working/tail-only-zeroing-proposal.md`.

## Why this exists

PR #3 (`optimize-read-zeroing`) moved union zeroing from construction to arm selection and took batch read 1.80–2.11x. It left one residual, flagged in its own body: **selection still zeroes `sizeof(arm)`** — Chat's 264 B for a ~13 B message, Block's 2004 B for a 64–191 B payload. SPEC §5 explicitly prices tail-clearing as its own change ("if tail-clearing on read is ever wanted instead, it is a generated-code change, priced then"). This PR is that pricing, for the union-arm case.

## What the document contains

- **The exact SPEC wording change** (two edits: the §6.1 union-surface sentence PR #3 rewrote, and a §5 scope addition), keeping whole-arm zeroing conforming and adding tail-only as an equally conforming implementation with an identical read-success post-condition.
- **The safety argument**: a compile-time field-coverage partition of every arm's bytes (written unconditionally / written-or-zeroed conditionally / padding), the read-failure story (§5 unchanged; recommended: re-zero the arm on the cold failure path so the no-stale-exposure property stays real), and the requirement that PR #3's pinned stale-leak test stays green unchanged.
- **The expected win, bounded by PR #3's measurements — honestly small for the motivating case**: ~10% less zeroing traffic on the measured batch mix, an estimated **~1–2% on batch read**, because §5 requires buffer tails zeroed on success anyway (~235 of Chat's 264 B). The full wins are total-coverage arms and near-capacity buffers. If the conclusion is "not worth the emitter surface", that conclusion is the document working as intended.
- **Four open questions for Glenn**: padding/memcmp semantics, the failure path, whether ~1–2% justifies the machinery (and naming — without recommending — where the rest of the cycles live: the semantics-changing used-prefix variant that PR #3's stale-leak test deliberately forbids), and cross-surface scope.

## Notes

- Builds on PR #3 merging first; branch is off `origin/main` and carries no dependency on any code branch.
- GitHub Actions outage today: nothing rode CI (documentation-only branch anyway).

### DIFF
--- a/working/tail-only-zeroing-proposal.md
+++ b/working/tail-only-zeroing-proposal.md
@@ -1,0 +1,182 @@
+# Tail-only union-arm zeroing — SPEC proposal
+
+**STATUS: PROPOSAL — DO NOT MERGE. The decision is Glenn's.** No code changes ride this
+branch; this document is the exact wording change, the safety argument, the honestly
+bounded win, and the open questions. It assumes PR #3 (`optimize-read-zeroing`) merges
+first — everything below builds on its rule that an arm's storage is zero-established at
+selection.
+
+## 1. The residual this prices
+
+PR #3 removed the whole-union memset (2016 B per `Message` construction) and moved zeroing
+to arm selection: `message.chat = Chat{};` before `ReadChat`. That was the #1 hotspot
+(60.6% of batch-read self-cycles in `rep stos` on the EPYC; batch read improved 1.80-2.11x
+across the three host/compiler pairs).
+
+The residual: **selection still zeroes `sizeof(arm)` regardless of how little the decode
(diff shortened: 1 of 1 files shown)


gafferongames:
Priced and **declined at this time** — the document's own anticipated conclusion. ~1–2% on batch read for the measured mix doesn't justify the emitter surface, and the table layer has since absorbed the config-shaped workloads, leaving the dense wire purely realtime where the win is smallest. The pricing document is merged to main as a banked record (working/tail-only-zeroing-proposal.md); revisit only if a total-coverage-arm workload appears.

PR #38, opened 2026-08-15
bench: the per-port inline-budget gate — CI fails when the serialize chain stops inlining

### LINKED ISSUE #25: harness: everything depends on the serialize chain inlining end to end, nothing warns when it stops, and self-reported green shipped bugs four times in two days
Severity: **high** · Kind: harness · Call: ours-to-fix · Affected repos: all six (bench/verdict machinery lives here in schema; the gate would guard every port)

## What is wrong

Every runtime's throughput depends on the serialize chain inlining end to end, and nothing warns when it stops. The same per-packet function gets four different inline verdicts from one compiler at one level (GCC 13.3 -O2: C write NO, C read YES, C++ write YES, C++ read NO — driven by max-inline-insns-auto=15 vs max-inline-insns-single=70), and Go's SerializeBits and tryReadBits sit at EXACTLY the cost-80 cliff with zero headroom. One added line to a Go hot function silently halves throughput and the suite stays green.

Second, documented because it is the pattern behind four shipped defects in two days: every stage that self-reported complete and green contained at least one real defect (four for four), and each was found by an adversary or by an instrument built hours earlier for another purpose — never by more tests written by the same worker.

## How it manifests

- schema 4dc9b62: Rust verdicts published `full` while packet read held 11 out-of-line calls (awk namespace bug in the transitive join).
- schema 34b7ed8 then 7db74ef: C partial:N unroll-inflated twice (packet read 12→3 per op).
- schema 2e26eaa: partial:N counted static sites, penalising the correct #[cold] pattern — four sightings before it was fixed.
- The candidate pass returned pre-fix rust rows while its preamble recorded the fix branch.

## Evidence

> Mechanism proven: GCC 13.3 -O2 reaches four different inline verdicts on the same per-packet function (C write NO under max-inline-insns-auto=15, C read YES, C++ write YES under max-inline-insns-single=70, C++ read NO); bench_c.s has 6 calls to fields_write and ZERO to fields_read while bench_cpp.s has 2 to Serialize<WriteStream> and 12 to Serialize<ReadStream>. Go's SerializeBits and tryReadBits sit at EXACTLY the cost-80 cliff with zero headroom. Self-reported-green instances: schema 4dc9b62 (Rust verdicts published `full` while packet read held 11 out-of-line calls — awk namespace bug in the transitive join); schema 34b7ed8 then 7db74ef (C partial:N unroll-inflated twice, packet read 12→3 per op); schema 2e26eaa (partial:N counted static sites, penalising the correct #[cold] pattern — four sightings before it was fixed); the candidate pass returning pre-fix rust rows while the preamble recorded ee5bd52.

## Fix direction

1. A per-port inline-budget gate in CI — accepting that it will also fire on compiler-version changes and that telling those apart is manual. Method note that must ride with the fix: Go's first inlining scan printed zero warnings because of BUILD CACHING and nearly filed a false clean bill — `-a` is required.
2. Adversarial verification as a required stage before any measured pass, not an optional one; the refusal tool needs adversarial fixtures as a permanent test class (three of its four holes were found only by someone actively attacking it).

For the ledger's other side: the strongest attacks found nothing wrong with the nine §5.3 refusal rules, the C++ verdicts (confirmed exactly by manual disassembly), the C/C++ published rates (reconciled with independent wall-clock to 0.1%), or corpus_id.

Not blocked on any DECISION issue.

---
Part of the 2026-08-15 estate known-issues register (index issue on mas-bandwidth/schema).

rowan-claude:
Closes #25. Stacked on the era-markers branch — merge that first. 238 budgeted legs; a 15-fixture selftest attacks the real machinery; the go-build -a false-clean refusal is live.


rowan-claude:
The ubuntu-vs-macOS selftest split is fixed in 1d6c558. Root cause: the counting machinery passed the runtime-target regex into awk as `-v RT=...`, and `RT` is a built-in variable in gawk (the record terminator, overwritten on every record read) — ubuntu-latest runs gawk as `awk`, so the regex was clobbered to a newline after the first input line and every runtime call counted zero (hot=0), which is exactly the false-full direction the adversarial selftest exists to refuse. mawk and BSD awk have no `RT` builtin, which is why macOS passed.

Fix, POSIX awk only: the awk-side variable is `RTPAT` everywhere, and backslash escapes no longer cross the `-v` boundary (POSIX leaves them undefined there; gawk strips `\.` to a bare `.` with a warning) — `COLD_SPLIT` and the go/cs regexes spell the literal dot as `[.]`.

Evidence, both ways:
- before, awk=gawk 5.4.1 via PATH shim: 7 selftest failures — the six ubuntu shapes plus the e2e fixture ubuntu skips for lack of otool
- after: selftest exit 0 under gawk 5.4.1, mawk 1.3.4-20240123 (built from source — the ubuntu 24.04 vintage), mawk 1.3.4-20260302, and macOS system awk 20200816; 0 gawk warnings remain
- real go leg end to end under the gawk shim: inline-gate PASS, 34 legs within budget, 0 unguarded
- make test exit 0 after merging main in (17e388f); generated/ tree current; gofmt/vet/tidy clean

rowan-claude:
CI is fully green as of e462119 (run 31910940142: test on both OSes, lint, vuln, windows — 5/5). Three commits since the deferral:

1. **1d6c558** — the deferral's root cause: `RT` is gawk's built-in record terminator, so the `-v RT=<runtime regex>` the counting machinery relied on was clobbered to a newline after the first input line on ubuntu (whose `awk` is gawk), and every hot count read zero. Renamed to `RTPAT`, plus `[.]` instead of `\.` in every regex crossing `-v` (POSIX leaves those escapes undefined). Selftest proven under gawk 5.4.1, mawk 1.3.4-20240123 and -20260302, and BSD awk 20200816.
2. **921bbd2** — bench/rust's hand-written rt family updated to serialize.rs 2.0.0's canonical `Result<(), S::Error>` signature (the 2.0.0 release landed between CI runs; this branch's rust leg is the first thing that compiles bench/rust in CI). Oracle gate passed: wire byte-identical, corpus_id 6b118770d85af4f6.
3. **e462119** — ledger edit: cs chat write pinned 3 -> 4. The gate fired on serialize.cs 1.3.0's wstring rework (byte-identical cs bench code, pre/post-1.3.0 runs measured 3 then 4 on the same runner) — a runtime-version cost, told apart manually per issue #25 and accepted deliberately.

Also merged main in (17e388f): the inline-gate steps compose with the lint/vuln/windows jobs; make test exit 0, generated/ current, gofmt/vet/tidy clean.

PR #75, opened 2026-08-17
js-final-2: the closing measurement — the flat tier is the JavaScript number

rowan-claude:
The race's closing measurement, under the ruling (Glenn, verbatim): **"Whichever correct implementation is fastest is the one we use for JavaScript."**

3 interleaved rounds on the M2 Air (era `js-final-2`, quick tier), cpp/c/js legs, control window OK (3.0%), all 11 legs golden-gated on corpus `7a7c343f1a446f4c`, runtimes pinned to the sixlang-air-1 tags. Measured at `f1cbee8`; code-identical to main's `60d0d92` (the diff is #73's doc rewrite only).

## One table — blended time vs C++ = 100 (22 comparable gen rows, best-of), + real_packet absolutes

| implementation | blend | real_packet write M/s | real_packet read M/s |
|---|---:|---:|---:|
| C++ (generated + serialize v1.10.0) | **100** | 21.15 | 26.71 |
| C (generated + serialize.c v1.4.0) | **102** | 25.62 | 30.16 |
| **JS flat — SHIPS** (generated, zero-dependency) | **1684** | 1.45 | **6.05** |
| serialize.js hand-port v1.1.0 (compat/reference) | *3427*† | — | — |
| JS generated runtime-call (old shape, for the record) | **7662** | 0.52 | 0.61 |

† rt-family basis (8 hand-written-runtime rows vs the same C++ rows) — like-for-like, but not the gen basis.

**Notes:** (1) The flat tier ships as THE JavaScript path: fastest correct on all 22 rows, 4.55x the runtime-call tier blended; real_packet read 4.4x off C++ — inside the 4–8x band — and 9.9x over the old shape; writes are vary-bound (14.6x, flat ≈ wasm parity there, per the probe). (2) The runtime-call tier stays as labeled supplementary rows; serialize.js stays the compat/reference tier; ship_shallow + message_batch remain runtime-only by construction. (3) Browser caveat: node 26/V8 numbers only; JSC/SpiderMonkey unmeasured. (4) Owed: certified twin pass for the flat codec (§2.6.1 new-binary rule; quick tier here per sixlang precedent); two one-round js codec=runtime collapses (probearray/read 40x, testdata/write 69x, round 2 only, recovered) named in the era doc — the §2.6.1 class, V8 specimen, unattributed. (5) js legs ran at nice 5 (launcher artifact, adjudicated no-correction in the era doc: runtime rows reproduce sixlang within noise).

Archive: `bench/results/js-final-2/` (era doc, pass CSV, pass log, boundaries).


rowan-claude:
Superseded: the era archive landed in #76 (merged, green-verified), and the final js-final-2 verdict — flat codec ~17x blended, real_packet read 4.4x off C++ — is recorded in the campaign record (cairn 7974800c, addenda 39-40) with the archive carrying the tables. This PR's snapshot is retained in its branch history; nothing here is lost by closing.

Issue #28, opened 2026-08-14
perf: compressedFloatParams is recomputed per field in all five runtimes — 30.8% of Go's read profile — and Go and C# never got a diagnosis lane


rowan-claude:
Severity: **medium** · Kind: performance · Call: ours-to-fix · Affected repos: all five runtimes; schema can emit the fix

## What is wrong

compressedFloatParams — a float32 divide, a clamp, a Ceil and a BitsRequired — is recomputed per field in all five runtimes, measured at 30.8% of Go's read profile and worth 4.6% read / 13.8% write to precompute. The parameters depend only on (min, max, resolution), which are compile-time constants at nearly every call site.

Second half of the finding: five languages were measured to the same contract and only three (C, C++, Rust) were ever given a diagnosis lane. Go and C# show the same shape the other ports had before their fixes — double-digit runtime calls inside the timed loop — and C# has the worst inline verdicts in the family with no work in flight.

## How it manifests

Go: bench_packet read 34.25 M msg/s partial:12 vs cpp 111.10 full; bench_ints read 39.81 M partial:10 vs cpp 165.82 full. C#: bench_packet read 33.57 M (always,default,partial:17), bench_ints read 50.23 M partial:17, writes partial:9.

Two caveats that must ride with any follow-up: (1) every cs-vs-cpp ratio is checked-vs-unchecked by construction (serialize.cs has no check-removal mode; schema 05d806b corrected the rows to checks=always) and needs that caption. (2) The Go-side lead was proven NOT to transfer to the other languages (it constant-folds once inlined) but was never tested as a Go-side fix, and the mechanism found for C/C++/Rust (fallible chains decaying cold) is LLVM-specific and will not apply to gc — Go needs its own investigation, not a transplant. Honest first question before optimising C#: whether the JIT verdict extraction (linkage=asm) measures what the other ports' verdicts measure.

## Evidence

[...]

PR #79, opened 2026-08-20
Generated Go carries its own compressed-float constants: the runtime's per-field derivation folds into the call site

rowan-claude:
Partial fix for #28: the schema-emitter half, Go backend.

## Route A (emitter only), and the code says it is the only right one

`compressedFloatParams` depends solely on `(min, max, resolution)`, which are
compile-time constants at every generated call site. The Go backend now folds
them at generation time, printing step count, wire width and delta as float32
literals, and does the quantization inline against an ordinary `SerializeBits`.
**No runtime API is added or changed**, so the five-runtime contract question
Route B would have raised does not arise.

This is not a new pattern in the repo. It is the pattern:

- the Go backend already folds ranged-int bounds, enum bounds, string lengths
  and fixed-point ranges the same way (`// the runtime range refusal, folded
  (SPEC §5)`), and the generated output reads identically to its neighbors;
- `internal/pack/encode.go` and `internal/codegen/js/flat.go` **already** derive
  compressed-float parameters from `ir.CompressedFloatParams` at generation
  time. This change makes the Go backend the third caller of the same IR helper;
- SPEC §6.1's IR preservation invariant already sanctions it verbatim:
  *"compressed floats keep `(min, max, resolution)` with derived step count"*.
  No spec change.

## Why Go, and only Go, in this sitting

The issue warned that the Go mechanism needed its own investigation rather than
a transplant. It does, and the inline ledger is the diagnosis:

| | verdict on `probearray` (the compressed-float-carrying gen bench) |
|---|---|
| **go** | **partial:12**. gc **refuses** `SerializeCompressedFloat32`: read cost **199**, write **187**, against budget **80** |
| cpp | full (0 hot runtime calls per op), at both O2 and O3 |
| c | full (0), at both O2 and O3 |
| rust | full (0) |
| cs | partial:3 |

C, C++ and Rust already measure **zero hot runtime calls per op** on that
benchmark. Their compilers have inlined the call and constant-folded the
derivation already, exactly as #28's caveat predicted, so folding in those
emitters would be churn carrying wire risk for no measured gain. C# is the only
other candidate, and CLAUDE.md already records this class of fold measuring
**0.97-1.04x** there because the JIT had inlined and lzcnt-folded it, which
makes it a measure-first proposition rather than a mechanical drop-in. Go is the
outlier at partial:12, and it is the one fixed here.

## Wire compatibility: unchanged, and proven four ways

1. **`testdata/wire/` is untouched** (`git status --porcelain testdata/wire` is
   empty). The C++ leg writes those pins; the Go and go-ludicrous legs
   byte-compare them and pass.
2. **The gate-7 FMA-discriminating vectors ride the Go leg.** `compressed_probe`
   pins `0.005` over `[0, 10]` at `0.01` (step 1 unfused, 0 fused) and `-4.8585`
   over `[-5, 5]` at `0.001` (142 vs 141), and the C++ leg asserts the
   discrimination property itself so the vectors cannot quietly stop
   discriminating.
3. **`internal/pack`'s independent encoder** packs the same pinned instances and
   byte-compares. Green.
4. **A dedicated differential**: the *generated* fold against the runtime call it
   replaced, over 16 declarations (every compressed-float declaration in
   `examples/`, `bench/corpus/RealWorld.schema` and serialize.go's own packet
   bench, plus degenerate and million-step ones). **1,120,178 checks**, wire
   bytes and decoded float32 bit patterns identical on every one. Inputs: dense
   in-range sweeps with overshoot, all specials (NaN, ±Inf, subnormals,
   ±MaxFloat32, exact bounds, ±resolution/2), 400k uniform random float32 **bit
   patterns**, 400k in-range randoms, and every representable quantized code
   including the bit headroom for each narrow field.
   The differential was **mutation-verified**. Dropping the writer's `float32()`
   round makes it fail with different wire bytes on the documented `0.005` case;
   dropping the reader's round fails on decode; dropping the headroom refusal
   fails on the read sweep.

`make test` (all six language legs) is green. `go vet`, `make generated-current`,
fuzz seeds, `make check`, fmt-drift and the inline gate (`selftest` plus
`leg go`, 34 legs within budget) are all green. Only the Go **source** golden is
re-pinned.

Both `float32()` conversions are carried across verbatim and are load bearing in
both directions, because Go fuses a multiply into an add unless a conversion
forces the intermediate rounding, and arm64 takes that permission. `uint32()`
truncates where the runtime floors, and they agree because the clamp leaves the
product non-negative.

## Numbers

**Caveat, loudly: these are Air numbers.** arm64 MacBook Air (M2), shared
desktop core, no pinning, idle apps resident. This is **not** the certification
box, and per `bench/README.md` a bare `bench/run.sh` invocation is **not a
publishable pass** (a publishable pass is a driver pass). These are **relative
signals from two independent same-sitting A/B pairs**, offered as conviction
rather than as publishable rows. Each row is the runner's own median of 7 after
a warmup.

    bench/run.sh --only go --out <before|after>.csv     # x4, sequential

Compressed-float-carrying rows, each pair normalized by that pair's own control
median (28 control rows), so baseline drift between sittings divides out:

| row | pair 1 | pair 2 |
|---|---|---|
| `probearray` write | **+12.0%** | **+12.0%** |
| `probearray` read | **+8.0%** | **+8.0%** |
| `testdata` read | +2.0% | +1.8% |
| `testdata` write | +0.9% | -0.1% |

Raw medians (M msg/s), pair 1 then pair 2: `probearray` write 20.53 to 22.88,
and 20.41 to 22.82; `probearray` read 17.60 to 18.91, and 17.49 to 18.85.

`probearray` carries 2 compressed floats out of ~26 field ops. `testdata`
carries 1 out of ~25 and is bulk/string-dominated, which is why its signal is
small. The 28 control rows sat at median ratio 0.995 and 0.998.

**Banked predictions, and one refutation.** Predicted before measuring:
(1) hot-call verdict drops 12 to 10 on `probearray`; (2) `probearray` +4-10%;
(3) `testdata` 0-3%; (4) no systematic control movement.

**Prediction 1 is REFUTED.** The verdict did not move; it is still partial:12.
`SerializeBits` *does* inline (cost 73/80), but its body calls
`readBits`/`writeBits`, which do not. So the fold trades one out-of-line call
for another and deletes the arithmetic between them, rather than removing a
call. Throughput moved anyway, and above the predicted band on write.
**Consequence: `bench/inline-budget.txt` needs no edit, and the gate passes
unchanged.**

**Two unattributed movements, filed rather than claimed**, per CLAUDE.md's
instrument-honesty rule:

- `probe_header read` moved **+20.5% and +18.3%**, reproducibly in both pairs.
  `ProbeHeader` contains **no compressed float** (`const`, `bits`, `reserved`,
  `align`, `uint64`), so this cannot be the change's semantics. It is almost
  certainly code layout, since the fold shrinks `Wire.go`'s generated code and
  shifts alignment for everything after it. It reproduces because it is a
  property of the two binaries rather than of the sitting. **Not claimed as part
  of this fix.**
- `message_batch` moved -2.1%/-2.8% in pair 1 and -0.8%/+0.7% in pair 2. It does
  not reproduce, which is consistent with the documented ±20% layout noise on
  that row. Filed, not claimed.

## What remains for #28

This closes the Go emitter half only. Left open, with what was measured:

- **C, C++, Rust**: no work indicated. Measured full (0 hot calls per op) on
  `probearray`, so already inlined and constant-folded. Folding would be churn
  with wire risk for no gain.
- **C#**: the only other candidate (partial:3). Needs its own measurement first,
  given the recorded 0.97-1.04x for this class of fold there.
- **JS**: already folded (`js/flat.go`), and flat is THE js path.
- **The runtimes themselves are untouched.** #28's headline number, 30.8% of
  Go's read profile, is a **serialize.go-side** profile of that repo's own
  packet benchmark (7 compressed floats in a 133-byte packet). Nothing in this
  PR can move it, because a hand-written caller of `SerializeCompressedFloat32`
  still pays the per-field derivation. Whether the runtime should gain a
  precomputed-params entry point is the five-runtime family decision #28
  describes, and is deliberately **not** taken here.
- **A measurement gap worth its own issue.** schema's Go, Rust and C# bench
  runners have **no `real_packet` row**; only cpp, c and js do, even though
  `generated/bench/go/realworld/` is generated and rides `make test`.
  `real_packet` is the compressed-float-dense workload (7 fields), so the
  backend with the largest compressed-float cost is measured on the corpus with
  the least of it. `probearray` (2 fields) is the densest Go row available today.
- **Diagnosis lanes for Go and C#**, #28's second half, are not addressed here.

Refs #28 (partial: Go emitter only).

### DIFF
--- a/generated/bench/go/realworld/RealWorld.go
+++ b/generated/bench/go/realworld/RealWorld.go
@@ -212,8 +212,14 @@ func WriteRealPacket(stream *serialize.WriteStream, value *RealPacket) error {
 		stream.SerializeBits(&offsetValue, 21)
 	}
 	{
-		compressedValue := value.F004Cf32
-		stream.SerializeCompressedFloat32(&compressedValue, 0.0, 2000.0, 0.1)
+		normalizedValue := value.F004Cf32 / 2000.0
+		if !(normalizedValue >= 0) { // the runtime's clamp form — it forces NaN into range too
+			normalizedValue = 0
+		} else if !(normalizedValue <= 1) {
+			normalizedValue = 1
+		}
+		integerValue := uint32(float32(normalizedValue*20000.0) + 0.5)
+		stream.SerializeBits(&integerValue, 15)
 	}
 	{
 		rangeValue := int32(value.F005Uint)
@@ -312,8 +318,14 @@ func WriteRealPacket(stream *serialize.WriteStream, value *RealPacket) error {
 	}
 	stream.SerializeBits(&value.F026Bits, 9)
 	{
-		compressedValue := value.F027Cf32
-		stream.SerializeCompressedFloat32(&compressedValue, -2.0, 2.0, 0.25)
+		normalizedValue := (value.F027Cf32 - (-2.0)) / 4.0
+		if !(normalizedValue >= 0) { // the runtime's clamp form — it forces NaN into range too
+			normalizedValue = 0
+		} else if !(normalizedValue <= 1) {
+			normalizedValue = 1
+		}
+		integerValue := uint32(float32(normalizedValue*16.0) + 0.5)
+		stream.SerializeBits(&integerValue, 5)
 	}
 	stream.SerializeBits(&value.F028Bits, 4)
 	{
@@ -458,8 +470,14 @@ func WriteRealPacket(stream *serialize.WriteStream, value *RealPacket) error {
 	stream.SerializeFloat64(&value.F059F64)
 	stream.SerializeBits(&value.F060Bits, 8)
 	{
-		compressedValue := value.F061Cf32
-		stream.SerializeCompressedFloat32(&compressedValue, -90.0, 90.0, 0.5)
+		normalizedValue := (value.F061Cf32 - (-90.0)) / 180.0
+		if !(normalizedValue >= 0) { // the runtime's clamp form — it forces NaN into range too
+			normalizedValue = 0
+		} else if !(normalizedValue <= 1) {
+			normalizedValue = 1
+		}
+		integerValue := uint32(float32(normalizedValue*360.0) + 0.5)
+		stream.SerializeBits(&integerValue, 9)
 	}
 	{
 		rangeValue := int32(value.F062Uint)
@@ -486,20 +504,38 @@ func WriteRealPacket(stream *serialize.WriteStream, value *RealPacket) error {
 		}
 	}
 	{
-		compressedValue := value.F065Cf32
-		stream.SerializeCompressedFloat32(&compressedValue, 0.0, 30.0, 0.5)
+		normalizedValue := value.F065Cf32 / 30.0
+		if !(normalizedValue >= 0) { // the runtime's clamp form — it forces NaN into range too
+			normalizedValue = 0
+		} else if !(normalizedValue <= 1) {
+			normalizedValue = 1
+		}
+		integerValue := uint32(float32(normalizedValue*60.0) + 0.5)
+		stream.SerializeBits(&integerValue, 6)
 	}
 	{
 		fixedValue := int64(value.F066Ufixed)
 		stream.SerializeFixed64(&fixedValue, 2, 14, 0, 2)
 	}
 	{
-		compressedValue := value.F067Cf32
-		stream.SerializeCompressedFloat32(&compressedValue, -100.0, 100.0, 0.25)
+		normalizedValue := (value.F067Cf32 - (-100.0)) / 200.0
+		if !(normalizedValue >= 0) { // the runtime's clamp form — it forces NaN into range too
+			normalizedValue = 0
+		} else if !(normalizedValue <= 1) {
+			normalizedValue = 1
+		}
+		integerValue := uint32(float32(normalizedValue*800.0) + 0.5)
+		stream.SerializeBits(&integerValue, 10)
 	}
 	{
-		compressedValue := value.F068Cf32
-		stream.SerializeCompressedFloat32(&compressedValue, 0.0, 2000.0, 1.0)
+		normalizedValue := value.F068Cf32 / 2000.0
+		if !(normalizedValue >= 0) { // the runtime's clamp form — it forces NaN into range too
+			normalizedValue = 0
+		} else if !(normalizedValue <= 1) {
+			normalizedValue = 1
+		}
+		integerValue := uint32(float32(normalizedValue*2000.0) + 0.5)
+		stream.SerializeBits(&integerValue, 11)
 	}
 	stream.SerializeBits(&value.F069Bits, 11)
 	{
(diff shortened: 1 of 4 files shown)


rowan-claude:
## Studio (M3 Ultra) A/B — `236bb45` vs its parent `5a42804`

Paired re-measurement of this merge on a quiet Mac Studio (Apple M3 Ultra, 32 cores, macOS 26.5.2, go1.26.5). **Strong quiet-machine signal, not certification — the space box remains the certification bench.**

**Protocol** (`bench/BENCH-STANDARD.md` §2): 7 measured rounds per side after a discarded build round; the two sides are **interleaved round by round**, alternating which goes first by round parity, so a load spike lands on both legs rather than one. Headline is best-of-7 with median and spread beside it (§2.2); a row with >40% spread is refused rather than published (§2.3). Sibling runtimes pinned to the CI tags — serialize v1.10.0, serialize.go v1.11.0. `corpus_id 6b118770d85af4f6` on both sides.

**Two independent passes** were run end to end. Both are shown, because a single pass cannot tell a real effect from a lucky window.

### Headline rows — M msg/s, best-of-7

| row | pre | post | Δ pass 1 | Δ pass 2 |
|---|---|---|---|---|
| `probearray` read | 22.02 | 25.93 | **+17.5%** | **+17.8%** |
| `probearray` write | 24.81 | 26.72 | +7.5% ¹ | **+7.7%** |
| `testdata` read | 14.89 | 14.94 | +0.5% | +0.3% |
| `testdata` write | 11.64 | 11.65 | +1.4% | +0.1% |
| `bench_packet` read | 47.10 | 47.68 | +1.0% | +1.2% |
| `bench_packet` write | 33.39 | 33.14 | −2.0% | −0.7% |
| `bench_ints` read | 53.91 | 53.90 | +0.1% | −0.0% |
| `bench_ints` write | 56.95 | 56.87 | +0.0% | −0.1% |

¹ pass 1's `probearray` write row carried 44.2% spread on the post side and was **refused under §2.3**; pass 2 reproduced it clean at 1.3/1.7% spread. The pass-2 figure is the one to read.

*(pre/post columns are pass 2, whose spreads are ≤5% on both sides of every row above.)*

### Controls — rows the emitter did not touch

`chat` +0.3/−0.8, `test` −0.1/−0.8, `message_batch` −0.2/−0.6, `probebits` −0.0/+0.8, `ship_shallow` −0.1/−0.1, `shipcreate` −0.0/+0.4, `rigidbody_moving` −0.1/+0.7, `rigidbody_at_rest` −0.0/−2.0 (read/write, pass 2). Everything is inside ±1% except `rigidbody_at_rest` write at −2.0%, which pass 1 put at −0.4% with 18% spread — it is the noisiest control in the set and moves in the wrong direction to flatter this change. The pair stands.

### What this confirms and what it corrects

- **`probearray` is confirmed, and larger here than signalled.** The Air's +12 read / +8 write becomes **+17.6% read / +7.7% write** on the Studio, reproduced across two independent passes. Write matches almost exactly; read is materially bigger on this machine.
- **`testdata` is flat.** Under 1.5% on every measurement, either direction. Whatever this change buys, it does not show up there.
- **The runtime-API family (`bench_*`) stays put.** Those are hand-written call sites the emitter never touched. They move within ±2% across both passes with no consistent direction, which is what a null control is supposed to look like.

### The `probe_header` read mystery — **it does not reproduce here**

The PR recorded an unattributed but reproducible `probe_header` read movement of **+20.5% and +18.3%** across both Air pairs, on a struct with no compressed float in it. On the Studio the same row moves **+0.6% (pass 1) and +1.2% (pass 2)**, with post-side spread of 1.2% and 0.6% — i.e. it is not there at all.

That makes it very likely what it looked like: **a code-layout artifact of one machine and toolchain, not a property of this change.** No profile was run, and none is owed — an effect that vanishes on a second machine does not need attribution on the first. Recorded here so the number in the PR body has its counter-measurement attached rather than standing alone.

*Deviation from the standard, stated: §2.6's C++ control legs were not run. The pre/post interleave plus the untouched-row controls above are what certifies this window.*

PR #83, opened 2026-08-20
C emitter: bulk [N]uint8 paths at statically byte-aligned positions, read and write

### LINKED ISSUE #63: Candidate A: C emitter bulk [N]uint8 paths, both directions
From the 2026-08-17 C/C++ real-world parity campaign (cairn 7974800c, eras cfixes-air → realworld2-space). Filed so it isn't forgotten; no commitment implied.

Candidate A: the C emitter emits per-byte loops for [N]uint8 (generated/c/WireWire.h:772 and :985 at 976f942) where the C++ emitter emits a bulk call. Fix both directions (read and write).

Receipt: the testdata rows — C++ led C's read 83 in era cfixes-air. Runtime dependencies (mas-bandwidth/serialize.c#26) are all merged, so this is unblocked. A brief exists in cairn 7974800c, the ~10:10 entry.

Receipts: era CSVs at space:~/rowan-bench-quick/results/ and on schema branches; cairn: rowan-new cairns/7974800c-*.md.

rowan-claude:
The C emitter emitted a per-byte loop for fixed `[N]uint8` arrays where the C++ emitter has emitted a bulk call since ce1e888 (`generated/c/WireWire.h:772` and `:985` at 976f942, both directions). This closes that gap, read and write.

## The bulk API, and its version pin

`serialize_write_bytes( stream, data, bytes )` and `serialize_read_bytes( stream, data, bytes )`, from **serialize.c v1.4.0**. That is the tag `.github/workflows/ci.yml` pins as `SERIALIZE_C_TAG`, and the version this branch was built and tested against locally. Both functions were reworked in mas-bandwidth/serialize.c#26, now merged, into the shape that makes them worth calling from generated code: the write half moves the partial scratch word as one 8-byte store and then the whole payload as one bulk copy, rather than pushing the block's head and tail through the packer a byte at a time, and the read half is an align plus one copy on the read spine.

The generated C now reads like its C++ sibling:

```c
/* C, this branch */
if ( !serialize_write_bytes( stream, value->fixed_bytes, 17 ) /* byte-aligned [N]uint8 — bulk copy, wire-identical to the per-byte loop */ )
{
    return 0;
}
```
```cpp
// C++, generated/cpp/WireWire.h:433
write_bytes( stream, value.fixed_bytes, 17 ); // byte-aligned [N]uint8 — bulk copy, wire-identical to the per-byte loop
```

The comment rides inside the call the way this emitter's `const` and `reserved` items already carry theirs (`generated/c/WireWire.h:165`, `:173`), so the line matches neighboring generated C as well as the C++ twin.

There is no new analysis. The C backend consults the same `ir.AlignedFixedByteArrays` the C++ backend consults, loaded per struct in `emitWriteFunc`/`emitReadFunc` and cleared for object views. A view's field set is assembled from the object declaration rather than walked as struct items, so no view field's position is statically proven, and the per-byte loop stays there exactly as it does in C++.

## Wire-compat proof

The wire decision is ce1e888's, unchanged. `serialize_write_bytes` and `serialize_read_bytes` align first, and that align contributes zero bits at a byte boundary, so N unranged 8-bit operations and one aligned bulk copy are the same bytes (SPEC §4.3 pins bit *i* of the stream to byte *i*/8). Only statically-provable positions switch. Everywhere else the loop stays, because there the bulk path's align *would* insert padding the pinned encoding does not have.

Failure semantics match too, which is the other half of "identical". On write, capacity is the writer's contract under `serialize_assert` in both `serialize_write_bits` and `serialize_write_bytes`, following serialize.c#52's ruling that "C should match C++ and have no checks on write at all (except assert)"; neither shape has a runtime capacity check to lose. On read, `serialize_read_bytes` refuses past `bits_limit` via `serialize_read_fail` exactly where the per-byte `serialize_read_bits` would, and the align's zero-padding check has nothing to check at a boundary.

Receipts, all on this branch:

* No wire golden moved. `git diff HEAD~2 -- testdata/wire testdata/table testdata/golden/id.txt` is empty. The protocol id did not move either.
* The C leg proves both directions against the C++-pinned bytes. `test/c/main.c` fills `fixed_bytes[i] = i * 3` in `fill_test_data`, writes `TestData`, byte-compares the result against `testdata/wire/testdata.bin` (the same file the C++, C#, Go, Rust and JS legs are held to), and only then reads those verified bytes back and `memcmp`s `fixed_bytes`. Write direction against the golden, read direction against bytes already proven equal to it.
* Only the C source golden re-pinned: `testdata/golden/c/WireWire.h`, two hunks, the write and read halves of `write_test_data` and `read_test_data`.
* `make test` green end to end across C++, C, Go, Rust, C#, JS, both corpora, plus `go test ./...`. `make generated-current` green. `gofmt`, `golangci-lint` under the repo's v2.12.2 config, and CI's `modernize` analyzer all clean.

Corpus effect is two generated functions each way: `TestData.fixed_bytes[17]` in the main corpus, declared after an explicit `align`, and `BenchPacket.blob[17]` in the bench corpus, the same shape.

## The new guard

`internal/codegen/c/c_test.go` is red before this change and green after, verified by checking the three emitter files back to `HEAD~1` and re-running. `TestBulkBytesEmission` requires the marked positions (after `align`, and after a `string` whose wire ends aligned by §4.7) to take the bulk call and to no longer hold the loop, and requires the unmarked ones (at struct entry, off a byte boundary, ranged `[min,max]`, counted `[<= N]`) to keep it, because bulk-copying those moves the wire. `TestBulkBytesMatchesAnalysis` then requires that for every struct in the test corpus, the set of fields the C wire functions bulk-copy equals `ir.AlignedFixedByteArrays` field for field. That second one is what matters long-term: it pins C to the shared analysis rather than to a predicate of its own, which is what keeps C and C++ on one golden. `internal/ir/align_test.go` already pins the analysis itself; this is the emission half of the same property, now held on both backends.

## BENCH PENDING AT REVIEW

No benchmark was run on this branch, because siblings were building concurrently on the same machine. The A/B belongs on a quiet box at review.

One note for reading the numbers. The change also shrinks the two affected generated bodies by a 17-iteration loop each, which lowers their cost at the call site, and `internal/codegen/c/spineinline.go` records the remark evidence that prices `read_test_data` at 2345 and `write_test_data` at 2910 against a hot-callsite inline threshold of 250. So a `testdata` delta has two plausible sources, the bulk copy itself and a cheaper body to inline, and the controls below are what separate either of them from noise.

Rows to run, C against C++, both directions:

* `testdata`, the row this change exists for, and the only generated message row in the bench harness whose type carries a fixed `[N]uint8`. This is the row where C++ led C's read by 83% in era cfixes-air (`space:~/rowan-bench-quick/results/`). Read is the headline; write should move too.
* `chat`, a bulk-heavy control that must *not* move: its payload is a `string`, which already took `serialize_write_bytes` and `serialize_read_bytes` before this change. If `chat` moves, the run is measuring the machine rather than the patch.
* `real_packet`, the §1.7 realistic snapshot, at 0% bulk share by bits by construction (`bench/corpus/RealWorld.schema` carries zero string, bytes, wstring or `[N]uint8` fields). Flat control.
* `probe_header` and `probebits`, small bit-packed generated rows with no byte arrays. Flat controls.
* The family rt rows (`bench_packet`, `bench_ints`, `bench_bits`, `bench_mixed`), which are hand-written runtime calls rather than generated code, so an emitter change cannot touch them at all. These are the null control, and they calibrate how much of any `testdata` delta is real.

Closes #63

### DIFF
--- a/generated/bench/c/BenchWire.h
+++ b/generated/bench/c/BenchWire.h
@@ -109,15 +109,9 @@ static SCHEMA_UNUSED SCHEMA_C_WRITE_INLINE int write_bench_packet( serialize_wri
     {
         return 0;
     }
+    if ( !serialize_write_bytes( stream, value->blob, 17 ) /* byte-aligned [N]uint8 — bulk copy, wire-identical to the per-byte loop */ )
     {
-        int32_t i;
-        for ( i = 0; i < 17; i++ )
-        {
-            if ( !serialize_write_bits( stream, (serialize_uint32_t) value->blob[i], 8 ) )
-            {
-                return 0;
-            }
-        }
+        return 0;
     }
     return 1;
 }
@@ -220,19 +214,9 @@ static SCHEMA_UNUSED SCHEMA_C_READ_INLINE int read_bench_packet( serialize_read_
     {
         return 0;
     }
+    if ( !serialize_read_bytes( stream, value->blob, 17 ) /* byte-aligned [N]uint8 — bulk copy, wire-identical to the per-byte loop */ )
     {
-        int32_t i;
-        for ( i = 0; i < 17; i++ )
-        {
-            {
-                serialize_uint32_t raw = 0;
-                if ( !serialize_read_bits( stream, &raw, 8 ) )
-                {
-                    return 0;
-                }
-                value->blob[i] = (uint8_t) raw;
-            }
-        }
+        return 0;
     }
     return 1;
 }
--- a/generated/c/WireWire.h
+++ b/generated/c/WireWire.h
@@ -767,15 +767,9 @@ static SCHEMA_UNUSED SCHEMA_C_WRITE_INLINE int write_test_data( serialize_write_
     {
         return 0;
     }
+    if ( !serialize_write_bytes( stream, value->fixed_bytes, 17 ) /* byte-aligned [N]uint8 — bulk copy, wire-identical to the per-byte loop */ )
     {
-        int32_t i;
-        for ( i = 0; i < 17; i++ )
-        {
-            if ( !serialize_write_bits( stream, (serialize_uint32_t) value->fixed_bytes[i], 8 ) )
-            {
-                return 0;
-            }
-        }
+        return 0;
     }
     serialize_assert( schema_utf8_valid_( (const serialize_uint8_t *) value->text, value->text_length ) );
     if ( !serialize_write_int( stream, value->text_length, 0, 255 ) )
@@ -978,19 +972,9 @@ static SCHEMA_UNUSED SCHEMA_C_READ_INLINE int read_test_data( serialize_read_str
     {
         return 0;
     }
+    if ( !serialize_read_bytes( stream, value->fixed_bytes, 17 ) /* byte-aligned [N]uint8 — bulk copy, wire-identical to the per-byte loop */ )
     {
-        int32_t i;
-        for ( i = 0; i < 17; i++ )
-        {
-            {
-                serialize_uint32_t raw = 0;
-                if ( !serialize_read_bits( stream, &raw, 8 ) )
-                {
-                    return 0;
-                }
-                value->fixed_bytes[i] = (uint8_t) raw;
-            }
-        }
+        return 0;
     }
     if ( !serialize_read_int( stream, &value->text_length, 0, 255 ) )
     {
--- a/internal/codegen/c/c.go
+++ b/internal/codegen/c/c.go
@@ -121,6 +121,12 @@ type gen struct {
 	deps                []string
 	wire                bool
 	body                strings.Builder
+
+	// fixed [N]uint8 arrays of the struct being emitted whose element bytes
+	// are statically byte-aligned (ir.AlignedFixedByteArrays) — the same map
+	// the C++ backend consults, reloaded per function so it never outlives
+	// its struct
+	bulkBytes map[*ir.Field]bool
 }
 
 func (g *gen) pf(format string, args ...any) {
@@ -488,6 +494,12 @@ func (g *gen) emitWireHeader() {
 }
 
 func (g *gen) emitWriteFunc(st *ir.Struct) {
+	// fixed [N]uint8 arrays at statically byte-aligned positions take
+	// serialize.c's bulk-bytes path instead of a per-byte loop — the bulk
+	// call's internal align contributes zero bits at a boundary, so the wire
+	// is byte-identical (the same switch, off the same analysis, as the C++
+	// backend)
+	g.bulkBytes = ir.AlignedFixedByteArrays(st)
 	g.pf("/* Writes %s. Returns 1 on success, 0 on failure — the stream latches the\n", st.Name)
 	g.pf("   error, so a caller may check once at the end of a message. */\n")
 	g.pf("static SCHEMA_UNUSED SCHEMA_C_WRITE_INLINE int write_%s( serialize_write_stream_t * stream, const %s * value )\n{\n", snake(st.Name), st.Name)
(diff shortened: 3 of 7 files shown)


rowan-claude:
## Studio (M3 Ultra) A/B — `b8ddb87` vs its parent `1976999`

Paired re-measurement of this merge on a quiet Mac Studio (Apple M3 Ultra, 32 cores, macOS 26.5.2, Apple clang 21.0.0). **Strong quiet-machine signal, not certification — the space box remains the certification bench.**

**Protocol** (`bench/BENCH-STANDARD.md` §2): 7 measured rounds per side after a discarded build round; the two sides are **interleaved round by round**, alternating which goes first by round parity, so a load spike lands on both legs rather than one. Headline is best-of-7 with median and spread beside it (§2.2); rows above 40% spread are refused rather than published (§2.3). Sibling runtime pinned to the CI tag, serialize.c v1.4.0. `corpus_id 7a7c343f1a446f4c` on both sides.

**Two independent passes** were run end to end, because a single pass cannot tell a real effect from a lucky window. Pass 1's window was the dirtier of the two — four rows exceeded 40% spread on the post side and were refused; pass 2 had exactly one row above 15% and none above 40%. Both are reported.

### Headline rows — M msg/s, best-of-7

| row | pre | post | Δ pass 1 | Δ pass 2 |
|---|---|---|---|---|
| `testdata` read | 78.38 | 90.85 | **+15.8%** | **+15.9%** |
| `testdata` write | 45.05 | 54.84 | **+20.2%** | **+21.7%** ¹ |

¹ pass 2's `testdata` write row carried 38.5% spread on the **pre** side — under §2.3's invalidation bar but marked noisy. Pass 1's pre side was clean at 3.6%, and the two passes agree to 1.5 points, so the effect is not resting on the noisy row. *(pre/post columns are pass 2.)*

### The refutation, reported as loudly as a confirmation would be

**The `era-cfixes-air` run reported ~83% on `testdata` read. That does not hold here: the Studio measures +15.8% and +15.9% on two independent passes.**

The effect is **real, clean and worth having** — roughly +16% read and +21% write, tight spreads on both sides, reproduced. But it is about a fifth of the signalled magnitude, and **the number that gets published should be the smaller one** until the space box says otherwise. A five-fold disagreement between two machines on the same commit pair is itself a finding: it says this row is far more sensitive to machine and toolchain than a bulk-copy path ought to be, and anyone quoting the 83% should know that before they do.

### Controls

**Named controls** (Δ pass 1 / pass 2): `chat` read −0.0/+0.3, write +0.1/+0.1 · `real_packet` read −0.0/+0.0, write −0.1/+0.1 · `probebits` read +0.0/−0.2, write +0.0/−0.0. Flat on both passes.

**Null control — the hand-written family-`rt` rows**, which this emitter change cannot reach: `bench_ints` −0.2/+0.2 read, +0.1/+0.2 write · `bench_mixed` −0.2/+0.6 read, −0.2/−0.3 write · `bench_packet` +0.2/−0.9 read, −0.3/−0.3 write · `bench_bits` −0.9/**+3.5** read, −0.5/−0.7 write.

**One null-control row moved, and it moved in the direction that flatters this change.** `bench_bits` read came back at +3.5% in pass 2 (spreads 0.9% pre, 3.9% post) after −0.9% in pass 1, where that row was one of the four refused for >40% post spread. Fifteen of the sixteen null-control measurements sit inside ±1%; this one does not, and it has no mechanism — the emitter does not touch those hand-written call sites. **Read it as the residual noise floor of this bench pair on this machine, which is therefore about 3 points rather than the 1 point the other rows suggest.** The headline effect is 5× that floor and survives it; the small secondary movements below are inside it.

### Secondary movements — reproducible, mechanism unchecked

Four rows moved consistently in the same direction across both passes, all small: `probe_header` write **+3.1/+5.2**, `ship_shallow` read +2.3/+1.9, `probearray` write +2.5/+2.4, `message_batch` write +1.6/+1.2. These are generated structs, so a bulk `[N]uint8` path landing in them is a plausible mechanism — **but no generated diff was inspected and no profile was run, so this is recorded as an observation and not a claim.** `probe_header` write is the one worth a look: it is the largest, it grew between passes, and its spreads were 0.6–10%.

*Deviation from the standard, stated: §2.6's C++ control legs were not run. The pre/post interleave plus the untouched-row controls above are what certifies this window.*

PR #84, opened 2026-08-20
cpp: the write-spine demand stops at the dispatch arms

### LINKED ISSUE #62: message_batch write: the dispatch-body lever — arm per-message spines without flattening them into WriteMessage
From the 2026-08-17 C/C++ real-world parity campaign (cairn 7974800c, eras cfixes-air → realworld2-space). Filed so it isn't forgotten; no commitment implied.

The accepted residual from #60: with the per-message spines' always_inline demanded, message_batch write sits +7-16% vs pre-demand off on both arches (era space-55-scoped). It is also the row where C leads C++ 2.1x on Air (era realworld-air, 211.4).

One candidate lever: keep the spines always_inline but mark WriteMessage noinline per-call — or restructure the dispatch so the spine bodies don't flatten into WriteMessage and fatten the batch loop.

Receipts: era CSVs at space:~/rowan-bench-quick/results/ and on schema branches; cairn: rowan-new cairns/7974800c-*.md.

rowan-claude:
The accepted residual from #60, taken at the mechanism it was attributed to.

## The lever, and why this one

#60 exempted `WriteMessage` from the write-spine demand so the batch build loop
would keep a call boundary. It does. But **an inlining demand is transitive
across a call**: `always_inline` on every `Write<Message>` meant the dispatch
still dragged every message body into its own body, whatever the cost model
said. The boundary moved; the fat did not. That is the +7-16% `message_batch`
write residual, and it is a property of the dispatch body, not the loop.

The fix is one forwarder per message arm, spelled **plain `inline`**:

```cpp
inline bool WriteMessageArmTest( serialize::WriteStream & stream, const Test & value )
{
    return WriteTest( stream, value );
}
```

and the dispatch case calls the arm instead of the spine. The demand still
flattens the spine into its arm — nothing is weakened anywhere — but the arm is
then an ordinary inline candidate the compiler judges **on cost at the dispatch
site, exactly as it judged the spine itself before the demand existed**. Both
representations (union and variant) share one set of arms.

Two candidates were built and measured before choosing:

| | dispatch body | arms outlined |
|---|---|---|
| `noinline` on every arm (a hard boundary) | 844 B | all 6, including an **8-byte** `Heartbeat` arm — pure call overhead |
| **plain `inline` forwarder (this PR)** | **1288 B** | the 2 the compiler had actually declined |

The hard boundary over-corrects: it forces out arms the compiler wanted inline
even in the pre-demand state. The forwarder restores the *decision*, not a
verdict — which is why it reaches the pre-demand number rather than overshooting
it.

No new macro, no compiler-specific attribute: `inline` is the whole mechanism.

## Mechanism proof

Apple clang 21.0.0, arm64, `-O3`, the `bench/run.sh` Release flags, one TU
(`bench/cpp/bench_main.cpp`), `examples/` corpus. Three builds: **off** = the
pre-demand state, reproduced with no code change by
`-DSCHEMA_WRITE_INLINE_DEFINED -DSCHEMA_WRITE_INLINE=inline`; **main**;
**branch**. `nm -n` sizes of the `example::` write symbols:

| symbol | off (pre-demand) | main | branch |
|---|---|---|---|
| `WriteMessage` | **1288** | **1832** | **1288** |
| `WriteTest` / arm | 324 | *(forced in)* | 324 |
| `WriteTimescale` / arm | 332 | *(forced in)* | 332 |

**1832 − 1288 = 544 = 324 + 332 − 112**, the two call sequences the flattening
removed. The residual's mechanism is not inferred here; it is the arithmetic.

The branch reproduces the pre-demand dispatch body **to the byte**, and the two
spines the compiler had declined on cost are outlined again at exactly their
pre-demand sizes. Disassembly shows the dispatch reaches them by **tail call**,
so the restored boundary costs no call frame:

```
example::WriteMessage:
  ...
  0000000100019160  b   example::WriteMessageArmTimescale(serialize::WriteStream&, example::Timescale const&)
  00000001000190c4  b   example::WriteMessageArmTest(serialize::WriteStream&, example::Test const&)
```

**No call was added to the loop.** `bl` sites into `example::WriteMessage`
across the whole binary: **7 before, 7 after.** The batch loop still makes
exactly one call per message; only the callee's body changed, and the two arms
it now declines are reached by tail branch from inside it.

**The per-message win is untouched.** In the *off* build 12 `example::Write*`
spines stand outlined beside `WriteMessage` — the compiler refusing them at
cost, which is the whole reason the demand exists. Ten of them are nothing to do
with the dispatch: `WriteTestData` 2444, `WriteShipData_Shallow` 1144,
`WriteProbeSample` 992, `WriteInput` 984, `WriteShipCreate` 944,
`WriteProbeBits` 588, `WriteQuat` 576, `WriteInputPacket` 568, `WriteVec3` 448,
`WriteQuantizedRotation` 328. In **both** main and this branch all ten are
absent — still flattened into their direct callers by the demand. The only
symbols this PR adds to the binary are the two arms the dispatch declined.

## What the inline gate says

Not run as a leg: `inline-gate.sh leg cpp` runs a bench round, which belongs on
the quiet machine at review. What was run, statically:

- **`bench/tools/inline-gate.sh selftest` — green**, all 15 adversarial fixtures
  ("the machinery can fail where it must").
- The **real** counting machinery (`count_calls`, sourced from
  `inline-verdict.sh` under `INLINE_VERDICT_LIB=1`, with the cpp `RT`/`HELPER`
  regexes) over all three disassemblies: `example::WriteMessage` reports
  **transitive hot runtime calls = 0** in off, main and branch alike, and both
  new arms report **0**. The arms contain no `serialize::` call for the
  attribution to lose.
- Whole-binary total transitive runtime calls (hot + cold), every symbol:
  **27 before, 27 after.** The out-of-line runtime-call structure is unchanged;
  only the dispatch body's shape moved.

So `cpp message_batch write O3 2` (and its `O2 2` twin) should re-measure
unchanged. Confirming that against a real leg is part of the review below.

*Noted, not fixed:* `count_calls` follows `bl`/`blr` and not a tail-branch `b`
into a generated helper. It does not bite here (the arms hold zero runtime
calls), and it is not new — the pre-demand state tail-branched to
`WriteTest`/`WriteTimescale` the same way, which is the shape the budget was
seeded under on 2026-08-15. Flagging it as a real blind spot in the machinery
for a separate decision.

## C is deliberately untouched — and it complicates the story, honestly

The same shape does not exist there. C's `write_message` carries
`SCHEMA_C_WRITE_INLINE` **itself** (`internal/codegen/c/dispatch.go:174`) — the
dispatcher is demanded, not exempted, so there is no outlined dispatch body for
spines to fatten. Checked, not assumed: the C bench binary built at the same
`-O3` has **zero** outlined generated `write_*`/`read_*` symbols, `write_message`
included. The whole dispatch is inline in the batch loop. `bench/inline-budget.txt`
agrees: `c message_batch write O3 0`. There is no lever here to pull.

That is worth stating plainly rather than burying, because it cuts against a
naive reading of this PR: C is the leg that **leads C++ 2.1x** on this very row
(era realworld-air, 211.4), and C does it with the fully-flattened shape. So
"flattened dispatch is bad" is *not* what the evidence says. Three distinct
shapes are in play:

| | calls per message | dispatch body |
|---|---|---|
| C | **0** — dispatch inline in the loop | in the loop, specialized per call site |
| C++ main | 1 | outlined, **1832 B**, every arm forced in |
| C++ this branch | 1 (+ tail branch) | outlined, **1288 B**, arms by cost |

C++ cannot simply copy C's shape: #55/#60 measured the blanket demand
(dispatch inlined into the loop, C's shape) at **21% worse** on Zen 4 — the
generated C++ entries are `linkonce_odr` header functions with no
last-call-to-static bonus, unlike C's `static` entries, so the two languages'
inliners are not making the same trade. What this PR claims is narrower and
testable: **within** the C++ shape, the dispatch body should not be fattened by
force. Whether closing that closes the whole C-vs-C++ gap on this row is a
separate question, and the issue's own comment says the base Air deficit is
unattributed.

## Wire proof

`testdata/wire/` and `testdata/table/`: **zero changes.** The non-comment diff
in the generated headers is exactly the six arm forwarders plus the six dispatch
cases that now call them, in both representations. Everything else is comment.

## Gates green locally

`make test` whole: `schema_test`, `schema_test_variant`, `schema_test_random`,
`schema_test_ludicrous`, `schema_test_c`, `schema_test_c_ludicrous`,
`schema_test_bench`, `schema_test_bench_c`, the Go / Rust / C# / JS legs and
their ludicrous twins, `go test ./...`. Plus `make generated-current`
("generated/ tree is current") and the inline-gate selftest.

The multi-file dispatch test now pins the arms the way it pins the rest of the
dispatch surface: one arm per message, plain `inline`, called by its dispatch
case, and **never** spelled `SCHEMA_WRITE_INLINE` — that last assertion is the
one that keeps this from being undone by accident.

`WriteMessageArm<Name>` is registered in the checker's claimed-name registry
beside the other per-message generated symbols, so a user declaration that
collides is a compile error naming the mechanism, not a broken header.

---

# BENCH PENDING AT REVIEW

**Not run here** — sibling children were building concurrently on this box and
would contaminate the numbers. Every claim above is static (symbol sizes,
disassembly, call counts); nothing in this PR asserts a timing.

The three-way A/B needs no code edits — the pre-demand arm is a build-flag
override:

| arm | how |
|---|---|
| **off** (pre-demand) | any commit + `-DSCHEMA_WRITE_INLINE_DEFINED -DSCHEMA_WRITE_INLINE=inline` |
| **main** | `main` as it stands |
| **branch** | this branch |

### The rows that decide it

| lang | bench | path | opt | arch | expectation |
|---|---|---|---|---|---|
| cpp | `message_batch` | **write** | O3 | Air (Apple clang, arm64) | branch ≈ **off**; the +7-16% vs off closes |
| cpp | `message_batch` | **write** | O3 | space (Zen 4, gcc) | same |
| cpp | `message_batch` | **write** | O2 | both | same (the budget pins O2 too) |

### Controls that must NOT move

| lang | bench | path | why |
|---|---|---|---|
| cpp | `message_batch` | **read** | `ReadMessage` is untouched — any movement is noise or box |
| cpp | every per-message **write** row (`shipcreate`, `ship_shallow`, `inputpacket`, `probebits`, `probearray`, `probe_header`, `testdata`, `chat`, `test`, `rigidbody_moving`, `rigidbody_at_rest`) | write | generated code **byte-identical** to main; branch must equal main, and both must beat off. This is the half #60 bought and this PR must not spend |
| c | `message_batch` | write | untouched language — the cross-check that the box is quiet |

### Reading the result

- branch ≈ off **and** branch ≈ main on the per-message rows → the residual is
  closed and the lever is the whole story.
- branch ≈ main on `message_batch` write → the dispatch body was not the cost;
  the mechanism above is still real, but the time is elsewhere, and the issue's
  own comment names the suspects (I-cache thrash across cycling message types,
  clang/arm64 dispatch shape, or measurement).

**Air caveat, from the issue's comment:** `message_batch` write on Air is the
noisiest row in the suite — 25% same-binary spread measured, era wander
173→211 with no code change — and its **base** deficit is unattributed and
separate from anything this PR touches. Tame the spread (twin gates / §2.6.1)
before believing an Air delta either way. The Zen 4 leg is the one that can
resolve a 7-16% effect cleanly.

Closes #62

### DIFF
--- a/generated/bench/cpp/BenchWire.h
+++ b/generated/bench/cpp/BenchWire.h
@@ -17,7 +17,8 @@ namespace bench {
 #ifndef SCHEMA_WRITE_INLINE_DEFINED
 #define SCHEMA_WRITE_INLINE_DEFINED
 // SCHEMA_WRITE_INLINE — how every generated Write function is spelled,
-// EXCEPT the WriteMessage dispatch surface (see the comment on it): an
+// EXCEPT the message dispatch surface: WriteMessage and its per-message arm
+// forwarders (see the comments on them). It is an
 // inlining DEMAND (always_inline / __forceinline), the serialize family's
 // remedy for LLVM refusing the linkonce_odr Write entries into their call
 // sites at cost over threshold — no last-call-to-static bonus exists for a
--- a/generated/bench/cpp/RealWorldWire.h
+++ b/generated/bench/cpp/RealWorldWire.h
@@ -17,7 +17,8 @@ namespace realworld {
 #ifndef SCHEMA_WRITE_INLINE_DEFINED
 #define SCHEMA_WRITE_INLINE_DEFINED
 // SCHEMA_WRITE_INLINE — how every generated Write function is spelled,
-// EXCEPT the WriteMessage dispatch surface (see the comment on it): an
+// EXCEPT the message dispatch surface: WriteMessage and its per-message arm
+// forwarders (see the comments on them). It is an
 // inlining DEMAND (always_inline / __forceinline), the serialize family's
 // remedy for LLVM refusing the linkonce_odr Write entries into their call
 // sites at cost over threshold — no last-call-to-static bonus exists for a
--- a/generated/cpp/MessagesWire.h
+++ b/generated/cpp/MessagesWire.h
@@ -18,7 +18,8 @@ namespace example {
 #ifndef SCHEMA_WRITE_INLINE_DEFINED
 #define SCHEMA_WRITE_INLINE_DEFINED
 // SCHEMA_WRITE_INLINE — how every generated Write function is spelled,
-// EXCEPT the WriteMessage dispatch surface (see the comment on it): an
+// EXCEPT the message dispatch surface: WriteMessage and its per-message arm
+// forwarders (see the comments on them). It is an
 // inlining DEMAND (always_inline / __forceinline), the serialize family's
 // remedy for LLVM refusing the linkonce_odr Write entries into their call
 // sites at cost over threshold — no last-call-to-static bonus exists for a
@@ -307,12 +308,54 @@ SCHEMA_READ_INLINE bool ReadMessageType( serialize::ReadStream & stream, Message
     return true;
 }
 
+// The dispatch arms: one forwarder per message, standing between WriteMessage
+// and that message's Write spine. Spelled plain `inline`, NOT
+// SCHEMA_WRITE_INLINE, and that is the whole point — an inlining demand is
+// transitive across a call, so a dispatch calling the spines directly would
+// drag every message body into its own body whatever the cost, which is what
+// made the batch build loop pay for the demand. Behind a plain-inline
+// forwarder the demand still flattens the spine into the arm, and the
+// compiler then decides per arm — on cost, the way it decided before the
+// demand existed — whether that arm lands inside the dispatch body or stays
+// a call. Do not spell these SCHEMA_WRITE_INLINE; that is the shape this
+// exists to avoid.
+inline bool WriteMessageArmBlock( serialize::WriteStream & stream, const Block & value )
+{
+    return WriteBlock( stream, value );
+}
+
+inline bool WriteMessageArmChat( serialize::WriteStream & stream, const Chat & value )
+{
+    return WriteChat( stream, value );
+}
+
+inline bool WriteMessageArmHeartbeat( serialize::WriteStream & stream, const Heartbeat & value )
+{
+    return WriteHeartbeat( stream, value );
+}
+
+inline bool WriteMessageArmSynchronize( serialize::WriteStream & stream, const Synchronize & value )
+{
+    return WriteSynchronize( stream, value );
+}
+
+inline bool WriteMessageArmTest( serialize::WriteStream & stream, const Test & value )
+{
+    return WriteTest( stream, value );
+}
+
+inline bool WriteMessageArmTimescale( serialize::WriteStream & stream, const Timescale & value )
+{
+    return WriteTimescale( stream, value );
+}
+
 // WriteMessage is deliberately OUTSIDE the write-spine demand: plain `inline`,
-// not SCHEMA_WRITE_INLINE. Its callees (every per-message Write) flatten into
-// its body, but the dispatch switch itself stays a call boundary — demanding
-// it into a batch build loop was measured to slow that loop ~21% while every
-// per-message row kept its win with the boundary in place. The compiler
-// remains free to inline it where that pays.
+// not SCHEMA_WRITE_INLINE — demanding it into a batch build loop was measured
+// to slow that loop ~21% while every per-message row kept its win with the
+// boundary in place. It reaches each message through the WriteMessageArm*
+// forwarders above rather than calling the demanded spines directly, so no
+// spine body is forced into this one. The compiler remains free to inline
+// this — and any arm — where that pays.
 inline bool WriteMessage( serialize::WriteStream & stream, const Message & message )
 {
     switch ( message.type )
@@ -324,37 +367,37 @@ inline bool WriteMessage( serialize::WriteStream & stream, const Message & messa
             {
                 return false;
             }
-            return WriteBlock( stream, message.block );
+            return WriteMessageArmBlock( stream, message.block );
         case MessageType::Chat:
             if ( !WriteMessageType( stream, MessageType::Chat ) )
             {
                 return false;
             }
-            return WriteChat( stream, message.chat );
+            return WriteMessageArmChat( stream, message.chat );
         case MessageType::Heartbeat:
             if ( !WriteMessageType( stream, MessageType::Heartbeat ) )
             {
                 return false;
             }
-            return WriteHeartbeat( stream, message.heartbeat );
+            return WriteMessageArmHeartbeat( stream, message.heartbeat );
         case MessageType::Synchronize:
             if ( !WriteMessageType( stream, MessageType::Synchronize ) )
             {
                 return false;
             }
-            return WriteSynchronize( stream, message.synchronize );
+            return WriteMessageArmSynchronize( stream, message.synchronize );
         case MessageType::Test:
             if ( !WriteMessageType( stream, MessageType::Test ) )
             {
                 return false;
             }
-            return WriteTest( stream, message.test );
+            return WriteMessageArmTest( stream, message.test );
         case MessageType::Timescale:
             if ( !WriteMessageType( stream, MessageType::Timescale ) )
             {
                 return false;
             }
-            return WriteTimescale( stream, message.timescale );
+            return WriteMessageArmTimescale( stream, message.timescale );
     }
     return false; // not a message type; nothing was written
 }
--- a/generated/cpp/ObjectsWire.h
+++ b/generated/cpp/ObjectsWire.h
@@ -20,7 +20,8 @@ namespace example {
 #ifndef SCHEMA_WRITE_INLINE_DEFINED
 #define SCHEMA_WRITE_INLINE_DEFINED
 // SCHEMA_WRITE_INLINE — how every generated Write function is spelled,
-// EXCEPT the WriteMessage dispatch surface (see the comment on it): an
+// EXCEPT the message dispatch surface: WriteMessage and its per-message arm
+// forwarders (see the comments on them). It is an
 // inlining DEMAND (always_inline / __forceinline), the serialize family's
 // remedy for LLVM refusing the linkonce_odr Write entries into their call
 // sites at cost over threshold — no last-call-to-static bonus exists for a
--- a/generated/cpp/RenderWire.h
+++ b/generated/cpp/RenderWire.h
@@ -18,7 +18,8 @@ namespace example {
 #ifndef SCHEMA_WRITE_INLINE_DEFINED
 #define SCHEMA_WRITE_INLINE_DEFINED
 // SCHEMA_WRITE_INLINE — how every generated Write function is spelled,
-// EXCEPT the WriteMessage dispatch surface (see the comment on it): an
+// EXCEPT the message dispatch surface: WriteMessage and its per-message arm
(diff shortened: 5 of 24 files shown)


rowan-claude:
## Studio (M3 Ultra) — the row moves, but the lever does not move it

Measured under `bench/BENCH-STANDARD.md`, 7 interleaved rounds, C and C++ legs, at **both**
optimization levels per §3.3. **The space box remains the certification bench.**

**Window certification:** `load_max` 2.29–2.9 across passes, `foreign_procs 3`, unpinned.
`corpus_id 7a7c343f1a446f4c`, `checks=removed`, `linkage=hdr`, `inline=full` on every row
reported below.

### The three-way, which is what this campaign was built for

The pass carried a third arm — **`off`**, the branch with the lever disabled — and it is
the arm that decides the question. `message_batch` write, O3, round 2:

| comparison | ratio | what it isolates |
|---|---:|---|
| main → **off** (lever OFF) | **1.04x** | the branch, without the lever |
| main → **branch** (lever ON) | **1.04x** | the branch, with the lever |
| **off → branch** | **1.00x** | **the lever alone** |

**The +4% is fully present with the lever switched off.** Lever-on against lever-off is
1.00x, spreads 5.1 and 1.7. Whatever is making this branch faster than main on this row, it
is not the mechanism this PR is about.

### O2 says the same thing louder

| comparison | O3 r2 | O2 r1 |
|---|---:|---:|
| off → branch (`message_batch` write) | 1.00x | **0.99x** |
| main → branch (`message_batch` write) | 1.04x | **1.01x** |

The branch-wide effect largely **collapses at O2** while the lever stays at unity in both.
A delta that appears at one optimization level, is absent at the other, and does not track
the switch that is supposed to cause it is the signature §0 of the standard already
describes — this is the shape of **code layout**, not of an algorithmic win.

`bench_ints` write shows the same pattern from the other side: `off → branch` 1.01x at O3
and 1.06x at O2, while `main → off` is 1.05x at O3. Inconsistent in direction across levels,
which is what layout noise looks like and what a real lever does not.

### What I am NOT claiming

I am not claiming the lever is worthless. I am claiming **this pass cannot attribute the
movement to it**, and that the Air's +7–16% signal does not reproduce here as a lever
effect — the closest thing to it on this machine is a branch-wide +4% that survives turning
the lever off.

The estate's own rule applies: *an isolated win must re-prove itself in composition*, and a
wrong-magnitude prediction is a refutation. Reporting it plainly rather than quietly
picking the flattering pair.

### Rows excluded, and exactly why

Disclosure at the point of use, per §4.2's principle:

- **`real_packet` write and read, C and C++ — 4 rows.** `inline=unknown`; §5.3 rule 7 makes
  them un-ratioable. They were dropped from the tables above rather than published
  unverdicted. **A pass wanting these rows needs `run.sh --inline`.**
- **`cpp,probe_header,read` in `branch-O3-r1` — spread 57.87%** (min 833M against max
  1957M), past §2.3's 40% invalidity line. `relative.go` refused the whole O3-r1 table on
  it, correctly, and no number from that pairing appears here.

That second one may interest whoever is chasing the unattributed `probe_header` read
movement: on this machine that row is **bistable within a single 7-round pass**, in a window
whose control legs agreed to 0.03%. That is §2.6.1's state-selective interference, not
ordinary noise.

*Campaign CSVs at `studio-bench-20260820/campaign/pr84/`; nine passes, main/branch/off × O2/O3.*

rowan-claude:
Gate outcome: not merged. This draft was held pending Studio numbers, with the gate armed as "merge if the numbers support the mechanism," and they refute it instead. The three-way arm decides the question cleanly: off to branch is 1.00x at O3 and 0.99x at O2, so the lever alone does nothing, while the branch-wide +4% survives with the lever switched off and largely collapses at O2. That is the shape of code layout, not of the mechanism this PR claims, and the Air's +7 to 16 percent does not reproduce here as a lever effect.

Disposition: the PR stays open for exactly one question, the same three-way (main, off, branch, both optimization levels) on the space box, which remains the certification bench and needs its per-run blessing. If the certification bench agrees with the Studio isolation, this closes as refuted, with the branch preserved as the instrument. No merge under the armed gate either way until then.

The bistable cpp probe_header read observation (57.87% spread inside a 0.03% control window, the state-selective interference shape) belongs to the probe_header lane rather than this PR, and is noted there.

rowan-claude:
Closing on Glenn's word, 2026-08-20. The Studio three-way isolated the lever at 1.00x across both optimization levels while the branch-wide movement survived lever-off and collapsed at O2, which is the code-layout shape, and the certification-bench rerun was judged not worth its cost: the space box is the estate's one gcc/x86 surface where the lever could in principle behave differently, and that question reopens this PR only if x86 dispatch ever shows up hot in a real profile. The branch stays as the instrument, with the off arm that decided the question. The bistable probe_header observation from the same campaign lives in the probe_header lane, not here.

rowan-claude:
Reopened: the close was my misreading of Glenn's answer. His ruling is the opposite disposition — run the three-way on the space box, low cost, worth a try. The certification run is now the plan of record: main, off, branch, both optimization levels, under the box's bench rules. This PR holds open for that verdict.

rowan-claude:
The certification bench has spoken, and it agrees with the Studio while going one further: on gcc/x86-64 the lever is not merely inert, it is a regression on the headline row.

Method: six passes on the space box (AMD EPYC 9124 Zen 4, pinned core 15, server stopped for the run and restarted with verification by the campaign's exit trap), gcc/g++ 13.3, three arms per BENCH-STANDARD (main at bf6b22c, the PR branch at 79e3add, and issue-62-lever-off-epyc at c864949, the branch with the seven dispatch call sites reverted to direct spine calls). All three O3 windows certified clean with twin gates OK. Both O2 lever-arm passes self-invalidated on control-leg disagreement above the 5 percent bar (10.1 and 10.5 percent) and their ratios are refused per the standard, so O2 contributes nothing here and the verdict rests on O3 alone.

The O3 numbers, medians with spreads 1.5 to 2.6 percent: cpp message_batch write runs main to off at 1.007x, main to branch at 0.890x, and off to branch at 0.884x. The lever alone costs 11.6 percent on the exact row it was built to speed up. The other three write rows (cpp bench_ints, c message_batch, c bench_ints) hold off to branch at 0.999x to 1.001x, lever-inert. One disclosure: the inline column reads unknown on every row because the verdict pass could not run on the box (its Go predates go.mod; the bench tools were cross-compiled per arm), which is symmetric across arms and does not touch the isolation.

So the cross-bench picture is complete. Studio, clang/arm64: lever isolated at 1.00x, the movement was layout. EPYC, gcc/x86-64: lever isolated at 0.884x on the headline row, a real cost. The Air's original +7 to 16 percent signal is refuted as a lever effect on both architectures that matter.

Closing as refuted under Glenn's standing disposition for this PR (the certification run was the one question holding it open). The branch and the off arm stay as the instrument; the durable record with all six CSVs and the campaign log lives on the box at bench-epyc-20260820, README inside.

Issue #86, opened 2026-08-20
inline-verdict: count_calls follows bl/blr but not tail-branch b into generated helpers


rowan-claude:
Found during #84 (the #62 dispatch-body lever) and flagged there for separate decision: the inline gate's count_calls follows bl/blr call sites but not a tail-branch b into a generated helper, so a generated function reached by tail call is invisible to the transitive hot-call count.

It does not bite #84 — the forwarder arms hold zero runtime calls, and the pre-demand dispatch tail-branched the same way, which is the shape the budgets were seeded under — but the blind spot is real: a future generated helper reached by tail call could carry hot runtime calls the gate would not see, and its green would be a partial green wearing a full one's face.

Fix direction, either honest: teach count_calls to follow b targets that resolve to generated symbols (the checker already owns a claimed-name registry to resolve against), or add the limitation to the gate's own declared-misses output so every run states it, the way rowan-family instruments declare what they cannot see.

Kind: instrument blind spot · Found by: the #84 sitting, 2026-08-20 · The gate's selftest fixtures should gain a tail-call specimen either way, red-first.

PR #88, opened 2026-08-20
rust: the generated write spines demand inlining — the write half of #60's shape, measured in and the read half measured out

### LINKED ISSUE #77: Rust backend: emit #[inline(always)] on generated spines — the hint-vs-demand gap behind the 4x
From the 2026-08-18 attribution (cairn 7974800c addendum 40; full remarks in the session record). Era sixlang-air-1 measured generated Rust at ~4x C++ blended (rigidbody_moving write 14.7x). Mechanism, receipted: rust/functions.go:40,49 emits #[inline] (the HINT) where cpp/writeinline.go emits always_inline (the DEMAND); LLVM declines at cost 2285 vs the hint-boosted threshold 325 — intra-crate, so LTO is not the lever and the reference itself runs no-LTO. The runtime's own #[inline(always)] works (zero hot runtime calls); the outlining is in the GENERATED helpers, which the §4 inline-verdict metric cannot see (it counts runtime calls only — instrument blind spot, worth its own fix). CANDIDATE FIX: mirror #60's scoped shape exactly — #[inline(always)] on generated write/read spines, dispatch surfaces stay #[inline]. Known accepted cost per the C++ precedent: code size/compile time. Related: C# carries the shallower half of the class (per-message entry call + checks=always rows; cores already AggressiveInlining); Go's is structural (gc has no demand mechanism — different levers, Glenn-decision-sized). Filed so it isn't forgotten; no commitment implied.

rowan-claude:
The generated Rust **write** spines now demand inlining instead of hinting it,
which buys `rigidbody_moving` write 1.19x and `rigidbody_at_rest` write 1.09x
with no regressed read row. The **read** half deliberately keeps the plain hint,
because the C++ shape was ported here exactly, measured, and refused by the
evidence.

## The taxonomy this emits

| surface | spelling |
|---|---|
| `write_<t>`, object Deep/Shallow view writers, `write_message_type`, `write_object_type` | `#[inline(always)]` (DEMAND) |
| every read surface: per-type spines, view readers, tag readers, and both dispatch surfaces `read_message` / `read_message_into` | `#[inline]` (HINT) |
| `write_message`, the write dispatch surface | `#[inline]` (HINT) |
| `TableWriter` / `TableReader` reflection helpers | `#[inline]` (unchanged) |

54 write surfaces demand; 60 read and dispatch surfaces hint. The emitter carries
the line in `internal/codegen/rust/inline.go`, and the dispatch-once gate pins it
the way #60 made it pin the C++ one.

`write_message` keeps the hint for #60's own reason, with the rationale comment
emitted beside it as the twin of `emitWriteDispatchComment`: its callees flatten
into its body while the batch loop keeps a call boundary.

## Why the write demand

`#[inline]` is a HINT. It raises LLVM's threshold for the callee (275 to 325 in
these builds) and LLVM still declines when the callee prices over it. The
generated spines price far over: the `rigidbody_moving` write spine at cost
2285, refused, leaving a real call per serialize that the C++ twin does not
have (issue #77).

The lever is not LTO. The refused calls are intra-crate, generated spine to
generated spine and generated spine to a runtime that already demands inlining
on its own hot path, and the reference C++/C legs run no-LTO themselves.

## Why the read half diverges from C++

The C++ backend demands on its read spines too (`SCHEMA_READ_INLINE`, blanket,
`ReadMessage` included). **That shape was ported here first, surface for
surface, and measured.** It collapsed two rows:

| bench | path | main | C++ shape ported | B/A | spread A/B |
|---|---|---:|---:|---:|---|
| probearray | read | 63.09 | 33.54 | **0.53x** | 3.0% / 5.7% |
| shipcreate | read | 44.51 | 31.54 | **0.71x** | 1.4% / 3.2% |
| ship_shallow | read | 34.10 | 32.50 | 0.95x | 1.9% / 3.4% |
| probebits | read | 145.16 | 136.51 | 0.94x | 3.7% / 2.3% |

Twin-gate-OK pass, spreads 1.4 to 5.7%, while the hand-written rt-family control
rows held at 1.00 to 1.03x. An isolating pass then returned every collapsed row
to 1.00x by hinting the read spines alone, with every write win intact. That
isolating configuration is what this PR emits.

This is the same failure class #60 attributed on the C++ write side, a large
body forced whole into a timed loop, arriving on Rust's read side. It is also
why the C++ write demand itself stayed default-off until its own regressions
were attributed.

So the divergence is evidence-driven and per-backend, which is the rule #28
already banked when it refused to carry a C/C++/Rust mechanism into Go
sight-unseen. That caveat named Go; here the same rule bit inside the LLVM
family. Raising the read half to the demand again is a measurement, not an edit,
and `inline.go` says so.

## Confirmation A/B on the shipped shape

**Air-relative signals, not certification numbers.** Host `macbook` (Apple M2
MacBook Air), unpinned by construction (§7), interactive session. cargo/rustc
1.97.1, release, no LTO. serialize.rs `71a22cd`. Two driver passes,
`--rounds 5 --langs rust --twins --inline`, back to back in one sitting, each in
its own isolated git worktree. BEFORE = main `d309b75`, AFTER = this branch.
**Both passes `window: OK` (control delta 1.2% and 1.4%) and `twin_gate: OK`.**

| bench | path | main | branch | B/A |
|---|---|---:|---:|---:|
| rigidbody_moving | write | 42.44 | 50.70 | **1.19x** |
| rigidbody_at_rest | write | 75.68 | 82.44 | **1.09x** |
| probearray | read | 62.90 | 62.90 | 1.00x |
| shipcreate | read | 44.07 | 44.50 | 1.01x |
| ship_shallow | read | 34.13 | 34.47 | 1.01x |
| probebits | read | 144.75 | 145.51 | 1.01x |
| message_batch | write | 78.57 | 75.64 | 0.96x |
| message_batch | read | 48.07 | 47.86 | 1.00x |

**No collapsed rows.** Every other generated row sits at 0.99 to 1.01x, and the
rt-family control rows at 0.99 to 1.01x. `message_batch` write at 0.96x is the
one residual, the same shape and direction as the residual #60 accepted on the
C++ side, and it is bought by the per-message write rows.

This reproduces the earlier isolating pass (1.19x / 1.08x / 1.00x / 1.00x /
0.96x on the same rows) across two sittings and two different main baselines.

## Wire proof

Inline attributes and one emitted comment. **Zero diff under `testdata/wire/`
and `testdata/table/`.** Full `make test`, all six legs, green on this branch at
its rebased head: the four C++ binaries, both C binaries, then go / rust / cs /
js and their ludicrous twins, then `go test ./...`.

## What the inline gate says

`bench/tools/inline-gate.sh leg rust`: **PASS, 34 legs within budget, 0
unguarded.** No budget was edited. The per-symbol ledger is byte-identical to
main's, verdict for verdict.

That is worth stating because the ported C++ shape **failed** this gate, and not
by exceeding a budget. Every hot count was identical to main's there too; what
happened was that `read_message_into` under the demand had no out-of-line symbol
left, so the §4 machinery lost the anchor it counts against and correctly
refused to guess:

```
inline-gate FAIL (rust O3): budgeted leg message_batch read produced NO verdict
```

The instrument measures runtime calls inside a named out-of-line generated
helper. It cannot see what generated-helper inlining did to the code in either
direction. The sharper receipt is in the refuted pass: `probearray read` and
`shipcreate read` were `inline=full`, zero hot runtime calls, in **both** passes
while their throughput halved. This is the same class as #86 (the gate follows
`bl`/`blr` but not a tail-branch `b` into a generated helper): a blind spot in
the §4 verdict machinery, not a budget question. This PR does not trip it,
because the read surfaces keep their symbols.

## Banked predictions vs outcomes

Written before any measurement, against the ported C++ shape.

| # | prediction | outcome |
|---|---|---|
| P1 | wire bytes unchanged | **HIT** |
| P2 | write rows improve substantially; at least one 2x or better | **REFUTED.** Best generated row is 1.19x |
| P3 | read rows improve, less than writes | **REFUTED, badly.** Two read rows collapsed (0.53x, 0.71x) |
| P4 | message_batch write flat to 10% down | **HIT.** 0.96x |
| P5 | rt-family rows flat | **HIT.** 0.99 to 1.03x across every pass |
| P6 | rust-vs-C++ blended gap closes from ~4x to under 2x | **REFUTED.** See below |
| P7 | no rust hot count rises; at least one falls | **HIT on the first half, worse on the second.** None rose, none fell, and under the ported shape one verdict vanished |
| P8 | binary 5 to 25% bigger | **REFUTED.** +2.6% |
| P9 | compile time 10 to 50% longer | **REFUTED.** No measurable change |
| P10 | cpp control unchanged | **HIT.** Control delta 1.2 to 1.4% |

**The central refutation (P2/P6).** #77 attributed era sixlang-air-1's
Rust-vs-C++ blend (405, with `rigidbody_moving write` at 14.67x C++ time) to
this hint-vs-demand gap. Closing the gap moves that row by **1.19x**. A 1.19x
gain cannot account for a 14.67x row gap. The mechanism the issue names is real,
and the two rigidbody rows prove it changes what LLVM emits, but it is a small
term rather than the term. **The 4x remains unexplained.**

## Cost

Measured A/B/A/B/A/B across the two worktrees, `cargo clean` before every build,
three rounds each, bit-stable throughout.

| | main | branch | delta |
|---|---:|---:|---:|
| `generated/rust` rlib | 1,237,592 B | 1,237,528 B | −64 B |
| `bench/rust` release binary | 632,880 B | 649,600 B | **+16,720 B (+2.64%)** |
| binary `__TEXT` | 458,752 B | 475,136 B | +16,384 B (+3.57%) |
| generated-crate clean build | 0.72 s | 0.72 s | none measurable |
| bench-binary clean build | 2.00 s | 1.98 s | none measurable |

Read `__TEXT` with care: Mach-O quantizes it to 16K pages, and +16,384 is
exactly one page, so that row is a resolution floor rather than a precise
measure. The binary figure is the honest one. The rlib barely moves because
`#[inline(always)]` bodies are not codegen'd into it; the cost shows up in the
consumer.

## This does NOT close #77

#77 should stay open, holding two things this work did not settle:

- **the unexplained 4x.** The mechanism #77 names is confirmed real and worth
  1.19x on its biggest row. Whatever accounts for the rest of the Rust-vs-C++
  blend has not been found.
- **the instrument blind spot.** The §4 inline verdict counts runtime calls and
  cannot see generated-helper outlining, in either direction. Same class as #86.
  Until that is fixed, a green rust inline gate is not evidence that generated
  Rust inlined well, and a red one is not evidence that it inlined badly.

What this PR settles is narrower and worth having on its own: the write half of
the demand is a measured win, the read half is a measured loss on this backend,
and the emitter now records which is which and why.


rowan-claude:
## Studio (M3 Ultra) — the write win reproduces, and it is specific

Measured on the quiet machine under `bench/BENCH-STANDARD.md`, 7 interleaved rounds,
`--langs rust`. **The space box remains the certification bench**; this is the strong
quiet-machine signal and is offered as that.

**Window certification:** control legs `367,206,013` → `367,093,856` median msgs/s, a
**0.03%** delta against §2.6's 5% bar. `load_max 2.29`, `foreign_procs 3`, unpinned
(Darwin, no taskset). `corpus_id 6b118770d85af4f6`, `inline=full`, `checks=always`,
`linkage=crate`, `opt=O3` on both sides — ratioable under §5.3 with no escape flags.

### The rows that move, both rounds

| bench | path | main | PR #88 | ratio r1 | ratio r2 |
|---|---|---:|---:|---:|---:|
| rigidbody_moving | write | 51.44 | 60.07 | **1.17x** | **1.17x** |
| rigidbody_at_rest | write | 93.67 | 102.77 | **1.10x** | **1.06x** |
| rigidbody_moving | read | 62.00 | 60.32 | **0.97x** | **0.97x** |

Rates are best-of-7 M msgs/s (§2.2), round 1. Spreads on the moving rows: 5.6/6.2 (r1),
5.0/5.5 (r2) — comfortably inside §2.3's 15% band.

**`rigidbody_moving` write lands on 1.17x in both rounds independently.** That is the
cleanest signal in the pass.

### The rest is flat, and that is the useful half

All 31 other row-pairs sit at 1.00x ± 0.01 — `chat`, `test`, `inputpacket`, `shipcreate`,
`ship_shallow`, `probe_header`, `probebits`, `probearray`, `testdata`, `message_batch`,
the four `bench_*` rt rows and the bitpacker. **The hand-written family-`rt` rows do not
move**, which is the null control behaving: this change is in the generated spines, and the
runtime API called by hand should be untouched. It is.

So the effect is **specific to the generated write spines**, exactly where the PR claims it.

### The one cost, and I would not wave it through

`rigidbody_moving` **read** is `0.97x` in **both** rounds, spreads 2.2/3.3 and 3.4/3.9.
A consistent 3% read regression on the same shape the write win lands on, reproducing
independently, is a finding rather than noise. Worth a look before merge: whether the
`inline(always)` demand on the write spine is pushing the read path out of a budget, or
displacing it in i-cache.

### Rows excluded, named

None in this pass. Two rows carried spreads over §2.3's 15% mark and are flagged rather
than dropped: `probearray write` main r1 at 21.8%, `bench_ints read` r2 at 13.2/10.1 —
neither is a headline row here and neither approaches the 40% invalidity line.

*Numbers collected by the keeper on the Studio; the campaign CSVs are at
`studio-bench-20260820/campaign/pr88/`.*

rowan-claude:
Gate outcome: held, not merged yet. The armed gate was "merge if the Studio numbers support," and they half do. The write win reproduces exactly where the PR claims it: rigidbody_moving write at 1.17x in both rounds independently, rigidbody_at_rest at 1.10x and 1.06x, and all 31 other row-pairs flat at 1.00x including every hand-written family-rt row. The null control behaving is what makes the effect specific to the generated write spines.

But rigidbody_moving read at 0.97x in both rounds on the same shape is a named, reproducing cost, and the gate does not wave a consistent regression through on the strength of a win beside it.

Two roads resolve this. The first is diagnosing the displacement on the quiet bench: whether the write-spine inline demand pushes the read path out of an inlining budget or displaces it in layout, and the O2 leg of the same rows would help separate those. The second is taking the trade explicitly as an owner decision, since +17% write against -3% read on the moving shape is a workload judgment rather than a measurement, and it is not mine to make silently. A certification pass on the space box would speak to both. Worth saying that the write-side-only reshape is doing its job — after the full mirror collapsed reads at 0.53x and 0.71x, the remaining read cost is 3%, so the open question is small and specific.

rowan-claude:
Both legs of the diagnosis I named when the gate held, run on the Studio (M3 Ultra). The
short version is that nothing on the read path was pushed out of an inlining budget,
because nothing on the read path changed.

The O2 pass first. Same A/B/A/B shape as the morning campaign, `--rounds 7 --langs rust
--twins --inline`, all four arms `window: OK` and `twin gate: OK`, the only difference
`CARGO_PROFILE_RELEASE_OPT_LEVEL=2`. Best-of-7 per §2.2.

| bench | path | O3 r1/r2 | O2 r1 | O2 r2 |
|---|---|---:|---:|---:|
| rigidbody_moving | write | 1.17x / 1.17x | **1.198x** | **1.202x** |
| rigidbody_at_rest | write | 1.10x / 1.06x | 1.089x | 1.084x |
| rigidbody_moving | read | **0.97x / 0.97x** | 0.996x | 0.994x |

The write win holds at O2 and is slightly larger there. The read cost does not survive: the
null control, main r1 against main r2 at O2, puts run-to-run noise on these rows between
0.991 and 1.011, so 0.994 sits inside the noise where 0.97 did not. One caveat rides with
every O2 number, which is that the rust runner hardcodes `O3` in its CSV suffix. Those rows
say O3 and are O2, and they ratio against each other and against nothing else.

Then the disassembly, both arms rebuilt clean at O3, `otool -tV`, addresses and crate hashes
normalized and immediates kept. `read_rigid_body` is 275 instructions with zero differences
between the arms. `read_vec3` is 197 instructions with zero differences, still called out of
line three times per read on both sides. `write_vec3` exists as an out-of-line symbol on main
and is absent on the branch, so the three calls per serialize are inlined away, which is the
write win standing in the binary rather than inferred from a timer.

That refutes the budget theory directly, because a read path pushed out of a budget emits
different code and this one emits identical code. What differs is placement. `__TEXT` grows
from 458,752 to 475,136 bytes, and the two read functions land at different alignments (mod
64 of 40 against 48, and 20 against 44) and a different distance apart. It also overrides my
own pre-registered reading of the O2 leg, which said a cost vanishing at O2 would point at
budget. The direct instrument beats the inference, and the two agree once scale is included:
at O2 the branch binary is 224 bytes larger than main, at O3 it is 16,736 bytes larger, and
the regression shows up only in the second case.

For the trade, and this is my reading rather than a measurement: a 3% cost living in
placement is not a property of this PR's design in the way the 17% win is. It is real on this
binary and this compiler, and an unrelated change to the crate can move it in either
direction. The write win is mechanical and will not move.

Three things named rather than chased. Two O2-only rows moved in both rounds and are not part
of this question, `message_batch` write at 0.980 and 0.982 and `testdata` read at 1.055 and
1.084, neither appearing at O3. The inline ledger reports `hot=0` for `rigidbody_moving` write
on both arms at both opt levels, so the gate's own instrument does not resolve the row this
question is about, which wants its own issue. And no cycle profile was run: once the emitted
code proved identical, a profile could only say where the cycles went rather than whether
codegen changed, and the cheaper instrument had already answered.

CSVs are at `studio-bench-20260820/campaign/pr88-o2/`, four arms logged there, and the
disassembly comes from clean O3 rebuilds of both worktrees.

rowan-claude:
Gate lifted: merging on the diagnosis above. The held condition was the 0.97x read, and the disassembly resolves it. read_rigid_body and read_vec3 compile byte-identical on both arms, so the cost was never this PR's design; it was placement at O3 scale (sixteen kilobytes of growth moving alignments), the kind of number an unrelated crate change moves in either direction. The write win is the opposite kind of fact: write_vec3's out-of-line symbol exists on main and is gone on the branch, 1.17x at O3 and 1.20x at O2, mechanical and in the binary. Glenn asked whether more information was coming before the trade call; this is it, and it dissolves the trade rather than weighing it. A placement artifact against a mechanical win is not a trade.

One step before the button: the branch is being updated against current main so CI runs the combination — main's inline gate now counts tail branches (#97) and carries honest budgets of 2 and 3 on the rust message_batch legs, and this PR's demand is exactly the mechanism that may inline those tails away. The combined run is the first measurement of that. Merging on green.

The keeper's third named thing, the inline ledger reporting hot=0 for rigidbody_moving write on both arms while main's binary demonstrably carries three out-of-line write_vec3 calls per op, is an instrument gap in its own right and gets its own issue.

Issue #158, opened 2026-08-29
table: versioned content declarations — the archive contract beside the wire contract


rowan-claude:
Add `table` to the language: a record shape under the archive's contract, beside `type` under the moment's contract. The goal is one type language serving both a game's wire and the tooling around the game, with no opinions about content structure.

The two nouns, two contracts:

- `type`: same-or-refuse, one protocol id, naked bitpacked wire, freely editable (any edit moves the id; both ends redeploy together). The reader and writer are the same build by contract.
- `table`: versioned, append-only lineage, header'd wire (version + length per row), compiler-enforced evolution. The reader may be a different build than the writer; content outlives builds.

A table is not a type missing wire. It carries the other wire: rows that must be readable by builds that did not write them.

Candidate evolution model, for the spec to adjudicate: **append-only with positional identity.** Fields are only appended, never removed, renumbered or retyped; the compiler enforces this with version fences in the declaration and refuses edits below the newest fence. An old reader takes the prefix it knows and skips the tail by length; a new reader zero-defaults the missing tail (both ends already agree on defaults via the projection). No per-field tags, no TLV, no vtables: the problem tags exist to solve (skipping unknown fields mid-stream) disappears when unknown fields can only be a suffix, which also means row interiors can stay bitpacked within each version's known prefix.

Composition rule that gives the "one language" property: tables may use declared types as field groups. The container decides the contract; the type decides only the shape. The same type is bitpacked naked in a packet and encoded under the table regime in a row. Types do NOT implicitly become tables: the explicit `table` declaration is what carries the retention contract and licenses the compiler to enforce append-only exactly where archives exist, while types stay freely editable.

Relationship to the view (see the companion issue): a table's rows are read through retained descriptors, i.e. the view mechanism pointed at old bytes. The projection machinery extends naturally: each table version hashes like everything else, so lineage is printable and diffable.

Open questions for the spec's front page:
1. Embed the descriptor in stored data (self-contained archives) or reference it by hash (leaner files, lineage must be on hand)?
2. Version fence syntax, and how the compiler checks append-only (against in-file fences, against the previous projection, or both).
3. The unfreeze is formal: the projection carries `table=false` FROZEN tokens per SPEC 3.1/4.11, and dropping one is a ProjectionVersion bump, taken deliberately.

rowan-claude:
[...]

Issue #157, opened 2026-08-29
The view: a generated reflective descriptor surface for every declaration


rowan-claude:
Every declaration gets a generated **view**: a reflective surface over the shape (field names, kinds, bounds, order) that tooling walks at runtime. Implicit, zero opt-in, no language change, no wire change.

The mechanism is a promotion, not an invention. The projection (`ir.WireProjection`) is already the canonical reified shape: "a canonical text listing exactly the facts that determine the bytes on the wire," printable, diffable, and hashed into the protocol id. The view work is specifying that artifact as a generated, runtime-loadable descriptor and emitting per-type static metadata plus a generic walker in each target language.

Design frame that motivates it (one artifact, three tenses):
- The `type` wire ERASES the descriptor. Both ends hold it at compile time, so only its hash (the protocol id) remains. Same-or-refuse is descriptor equality.
- A view LOADS the descriptor at runtime. Inspectors, editors, dumpers and exporters walk values generically without hand-written glue.
- A `table` (separate issue) RETAINS descriptors across time: versioned rows readable through old shapes.

Scope notes:
- Views are metadata about the wire, never new bytes on it. The hot path stays the generated monomorphic code; the view is the general path for tooling, and that split is deliberate. The serialize.dart and serialize.java performance phases measured monomorphic generated code at 2-5x faster than dispatched access, which is exactly why both surfaces should exist.
- The descriptor format is specced once, per SPEC discipline, before any emission is written.

gafferongames:
Status: SPEC-TABLES.md no longer specifies the view — its 2026-09-01 rewrite (PR #223) scopes that document to tables, whose own reflection descriptors (§6 there) now serve the walk/print/edit need for data. This issue remains the home for a general view over type declarations, unspecified today; any future design starts fresh.

gafferongames:
Owner, 2026-09-02 (verbatim): "We should revisit the 'view' concept. Does a type have a view generated for it, alongside it, and a table has the view built in?"

Where it stands (main): a TABLE carries its descriptors built in, always (SPEC-TABLES.md §8 — TableFieldInfo/TableTypeInfo, the vocabulary columns, keyed/optional/block columns); every TYPE a table reaches gets them too (a walk must enter nested types); a type no table reaches gets none — the packet-only path stays zero-cost.

[...]

PR #159, opened 2026-08-29
SPEC-TABLES.md: the view and table — draft specification for adjudication

rowan-claude:
Draft specification for the two features of #157 and #158, written in SPEC.md's register. **For adjudication and language review — do not merge.** Five commits: the draft, the fold of the closing #158 rulings, and three cold-read repair rounds (5 → 4 → 2 blocks, trajectory falling; reader three: READY after these). The spec now parks for the owner's review.

## Two rulings to see first

- **Table declarations do not enter the unit's wire shape projection** (§3.3). The protocol id covers the type wire alone; each table's shape hashes into its own lineage, version by version; the `table=false` FROZEN-token drop remains a ProjectionVersion bump as a rendering change. Carried through §2.1/§2.2 and the §15 landing list.
- **Stored tables are uniform-version** (§4.1, new this round). Document validation refuses a mixed-version stored table: every specced pipeline (parallel pack at the current fence, whole-table hot swap, whole-table patch splice) produces uniform tables, so the mixed case is legal-but-unproducible and is refused rather than given semantics. Standalone rows (§6) keep per-row version headers and mix freely between peers — skew lives BETWEEN builds, never within one stored table. The JSON document shape carries one `version` per table object accordingly.

## Why a separate SPEC-TABLES.md

SPEC.md opens with "where the compiler and this document disagree, one of them is a bug" — a contract over the shipped compiler, which refuses `table` by name today (§4.11). This draft lives beside it; §15 lists the landing edits (scope/non-goals narrowing to the type wire, the §3.1 tables-stay-out statement, the §4.2 attribute-vocabulary growth — `capacity`, `key`, `ref` legal on type-body fields — the §4.11 unfreeze, the §6.1 measure exception, the `table` library family, and conformance gates including the canonical-JSON round trip), and SPEC-TABLES.md retires into SPEC.md when ruled and implemented.

## The rulings, recorded (§14)

Provenance: **q1–q3 were delegated and are boarded with their reasoning in issue #158's final comment**, alongside the process ruling (the spec grows to cover the full table source language and returns for review; first implementation C++ only); **q4 was ruled directly on the record.**

1. **Descriptors embedded by default**, with a strip flag for pipelines that keep lineage on hand.
2. **In-file `version N` fence markers, enforced against the retained lineage** — a mid-row edit refuses at the very next compile.
3. **Optional declared key** — a `key` attribute on one field; keys when declared, bare index otherwise.
4. **Composition from day one.**

What remains PROPOSED is spelling, not design, held for the §10 language review: the `table` declaration syntax and grammar productions, row-header widths, the deletion rule, the `capacity`/`key`/`ref` surfaces, the directory rendering, and the canonical-JSON taste calls.

## The review deliverable: §10, the table source language

The complete grammar for table source in a `.schema` file, worked example per construct: the `table` declaration hosting the inherited field grammar; composition; `key`; `capacity`; fence markers (integer literals only — a lineage ordinal, not a tunable, which also keeps fields named `version` legal) with a complete two-fence file; every reference form via wrapper types so SPEC §4.8 needs no edit; `ref` legal on integer fields anywhere and inert in pure packet use; and documents are never declared in source — a document is an assembly artifact, tables are the source-level unit.

## Round-three repairs, for the record

- Uniform-version stored tables (the ruling above), with the JSON document shape gaining `"version"` per table object.
- Canonical JSON's totality scoped honestly: total over contract-honoring wire-legal values, with one named refusal — dump refuses a `string(N)` payload violating the writer's UTF-8 contract, surfacing a writer-contract violation rather than inventing a spelling for corrupt data.
- The zeros-vs-construction divergence is now named in §7: untaken branches materialize zeros on a wire read (SPEC §5) and construction values on JSON import — byte-invisible, since untaken branches pack no bytes and the dump omits them.
- §6's lineage-derivability sentence restated: versions at or below head derive from the fence markers together with the current declarations, sound because the fences froze exactly those contributions.
- Spelling residues pinned: base64 is RFC 4648 standard alphabet with padding; string-spelled integers are decimal, no leading zeros, minus only when negative; compressed floats dump by the finite-float rule over the decoded value, byte identity resting on the quantizer's encode-decode idempotence.

Reader three's aside about SPEC.md §3.1's untaken-branch parenthetical is filed as its own issue on the schema board — pre-existing wording, tracked separately, not this PR's problem.

Earlier rounds' repairs (worked-bytes correction, prefix-relative row validation with version-0 refusal, legal fence-refusal remedies, integer-literal fences, the non-finite bit-pattern spelling) remain as recorded in the commit history.

Closes nothing; #157 and #158 remain the design record.

### DIFF
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -1,0 +1,686 @@
+# schema — views and tables (draft specification)
+
+**Status: DRAFT, for adjudication.** This document specifies two features that
+are not in the language: the **view** (a generated reflective descriptor
+surface — [#157](https://github.com/mas-bandwidth/schema/issues/157)) and
+**`table`** (versioned content declarations —
+[#158](https://github.com/mas-bandwidth/schema/issues/158)). It is written in
+SPEC.md's register and to its rigor, but it is not yet normative: SPEC.md
+promises that where the compiler and the document disagree one of them is a
+bug, and the compiler refuses `table` by name today (SPEC §4.11). §9 lists the
+open questions this draft deliberately does not decide, and §10 lists the
+SPEC.md text that changes when this lands. Syntax shown in examples is marked
+**PROPOSED** wherever it is not already in the language.
+
+## 1. One descriptor, three tenses
+
+Every schema unit already has one canonical reified shape: **the projection**
+(`ir.WireProjection`, SPEC §3.1) — a canonical text listing exactly the facts
+that determine the bytes on the wire, printable, diffable, and hashed into the
+protocol id. Everything in this document is a promotion of that artifact, not
+an invention beside it. One descriptor, three tenses:
+
+- **The `type` wire ERASES the descriptor.** Both ends hold it at compile
+  time, so only its hash — the protocol id — remains. Same-or-refuse is
+  descriptor equality (SPEC §3).
+- **The view LOADS the descriptor at runtime.** Inspectors, editors, dumpers
+  and exporters walk values generically, against generated metadata, without
+  hand-written glue (§2).
+- **The `table` RETAINS descriptors across time.** Rows written under old
+  shapes stay readable by new builds, because the shape they were written
+  under is kept, versioned, and enforced (§3).
+
+Types are essentially fixed; **tables are the sanctioned evolution surface of
+the language.** The wire contract of SPEC §3 — one id, same-or-refuse, no
+evolution machinery — is untouched by everything below: nothing here adds a
+byte, a tag, or a version field to any `type`'s wire.
+
+The one-line model underneath all of it: **types are value semantics; tables
+are pointer semantics.** A type embeds by value — it flattens inline, copies
+as bytes, has no identity, cannot be null, cannot contain itself. A table is
+reached by reference — it has directory identity, is shared, is nullable, is
+swappable behind its index, and may participate in cycles. The versioning
+split is a consequence, not a rule: values are timeless shapes, so types
+never version; identities persist across time, so tables must. Types are
+values, tables are objects, the directory (§4.1) is the heap — and it is a
+heap you can memcpy.
+
+## 2. The view
+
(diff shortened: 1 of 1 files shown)


rowan-claude:
Review inputs from the first implementation. Per the one-language-first process ruling, the C++ storage core now exists at mas-bandwidth/table (0.1.0, sixteen suites green in Debug, Release and ASan+UBSan, the worked row pinned byte-exact, relocatability proven by unaligned memcpy, parallel build byte-identical across worker counts), and building it against this draft surfaced seven things the ratified spec should settle. First, 3.4 states pad-to-byte as a write rule but never names the read-side check; the implementation validates pad-zero for versions at or below the fence and never judges pad in the skip case, and the spec should pin both. Second, a field with a positive minimum and no specified default materializes zero into a validated read when a tail is absent; the implementation treats materialized tails as construction semantics exempt from range validation, which the spec should state. Third, a real hazard for 3.6: serialize's flush stores whole 64-bit words, so packing directly into a shared slab writes up to seven bytes past a row's end, a data race between adjacent workers; the implementation packs rows into worker-local scratch and copies exactly length bytes, and the spec should note the constraint. Fourth, refs inside a skipped tail cannot be closed-world validated by an older reader, so 4.1's no-dangling claim is scoped to what the fence can see; one sentence reconciles it. Fifth, 4.1 does not rule whether directory extents may overlap. Sixth, the directory has no magic or container-format-version field; the row header can never version but the directory rendering could, worth deciding before ratification (the implementation adds a magic, marked proposed). Seventh, an observation: with fixed-width kinds every row of a version measures identically, so the mixed-version refusal is nearly unreachable behind the offset-index check, which affects how it should be tested, not whether it should exist. The content hash is pinned as proposed: the low 64 bits of SHA-256, mirroring the protocol id's truncation.

Issue #155, opened 2026-08-29
Dart backend: emit monomorphic constant-width code (measured directives from the serialize.dart port)


rowan-claude:
The serialize.dart performance phase (mas-bandwidth/serialize.dart, commits 7f858c9..df54871) measured where Dart AOT loses to the C++ reference, and the findings are directives for the schema Dart backend. Final library numbers on the bench that produced them (MacBook Air, Apple Silicon, iteration numbers): bitpacker 724/1002 MB/s write/read, stream 15.4/11.0 Mpps — about 5–6x off the C++ reference, and the gap is structural to library-shaped code, not fixable inside the library.

What the generated code must do differently, each backed by a measurement in that repo:

1. **Emit literal constant bit widths, not width variables.** Dart cannot constant-fold widths across calls the way C++ templates do; the library evaluates masks and shifts per call. Generated code knows every width at generation time — emit them as literals so the compiler folds the masks.

2. **Emit separate read and write functions per message, monomorphic against one stream type.** The family's unified serialize pattern costs real dispatch on Dart; the measure row prices dispatch at ~27ns/packet. Generated code has no reason to pay it.

3. **Inline the packer into the generated body** (or use `@pragma('vm:prefer-inline')` deliberately). The port's biggest single win (+44/+55% on the bitpacker rows) came from prefer-inline plus a derived bit index and branchless masks — generated flat code gets all of that by construction.

4. **AOT is the target that matters** (Flutter release ships AOT). The port measured a JIT-only pathology (bitpacker read ~185 MB/s JIT vs ~1000 AOT, a JIT inliner artifact) — benchmark generated code under AOT, and do not tune for JIT at AOT's expense.

5. **Bounds-check hoisting**: one check per multi-group operation instead of per group measurably helped the read leg. Generated readers know the total bit budget per message — hoist further to one check per message where refusal semantics allow.

The port's bench (bench/bench.dart, golden-gated per family discipline) is the regression instrument: generated-code throughput should be compared against the library rows there, on the same machine, and the working expectation from the C++ evidence is that constant-width monomorphic generated code closes most of the 5–6x.

rowan-claude:
[...]

Issue #153, opened 2026-08-27
Generate Dart from schema


rowan-claude:
Queued, not started: Glenn's addition to the work set, 2026-08-27.

A Dart backend for the schema compiler, alongside the existing generators, per the codegen philosophy: generate for language independence AND for performance -- allocation-free in the generated path where the language allows, and where Dart's model does not allow it, the closest honest shape rather than a pretense. Interpreters stay in tooling; shipped code is generated.

Depends on serialize.dart (mas-bandwidth/serialize's Dart port issue) as the runtime the generated code targets.

rowan-claude:
Performance directives for this backend, measured on the serialize.dart port, now live at #155 — constant literal widths, monomorphic per-message read/write functions, prefer-inline, AOT as the target. The port itself (mas-bandwidth/serialize.dart) is phase-1 complete and wire-verified, with the family bench in-repo as the regression instrument.

rowan-claude:
Implemented and merged in PR #161. The backend generates monomorphic constant-width Dart per issue #155, covered by the same goldens and cross-language byte-identity gates as every other language, with four test legs (JIT with asserts and AOT, normal and ludicrous corpora), analyzer and formatter as refusers, and the family benchmark golden-gated in bench/dart. The USAGE and README register now reads seven languages.

PR #161, opened 2026-08-30
Dart backend: monomorphic constant-width generated code (#153, #155)

rowan-claude:
The seventh backend: `--lang dart` emits Dart 3 for the VM and AOT (Flutter release ships AOT — the deployment the backend is tuned for), built to issue #155's measured directives from the serialize.dart port.

## The backend

`internal/codegen/dart` (dart.go storage/declarations, functions.go wire emission, int128.go the emitted pair):

- **Self-contained output.** Generated code imports `dart:typed_data` and its sibling generated files, never a runtime package. The serialize.dart bitpacker's proven-fast shapes (64-bit scratch, derived bit index, branchless masks, `@pragma('vm:prefer-inline')` helpers, the no-slack tail-window reader) are inlined at every field with **literal constant widths and masks** (#155 items 1 and 3).
- **Monomorphic functions per type** (#155 item 2): `write<Name>(value, view) -> bytes`, `read<Name>(value, view, numBits) -> bool`, `measure<Name>(value) -> bits`, `zero<Name>(value)`. Nested types inline into the caller's body: zero dispatch on the wire path. `measure` folds static runs to generation-time literals; a fully static type folds to a single `return`.
- **Reads validate everything, never throw on hostile bytes.** Bounds checks fuse per maximal static run and hoist toward message granularity where refusal semantics allow (#155 item 5); ranges, wire constants, reserved and align padding, union tags, interior nulls all refuse as a `false` verdict.
- **Writes trust, with asserts** that compile out without `--enable-asserts`: the C++ stance, in Dart's own correctness idiom.
- **Value domain:** every integer width through 64 rides bit-transparently in `int`; fixed point stores the raw scaled integer; the 128-bit widths ride an emitted `Int128`/`UInt128` pair (`Int128.dart`, written beside a unit exactly when it has 128-bit storage). Typed-data lists back scalar arrays. Member names are the first-letter-lowered `ir.GoExportName`, a bijection on it, so the checker's collision registry covers Dart without a second registry; a spelling that collides with a Dart reserved word is refused, never mangled.
- Analyzer-clean and `dart format`-clean, both gated in `make test`. Full corpus coverage: every IR construct the other backends emit (types, enums with headroom, flags with widening, unions, constants, fixed/ufixed at all storage widths, 128-bit integers, arrays, string/bytes, defaults, branches, const/reserved/align, degenerate ranges, empty types).

## Conformance

Wired exactly like the sibling targets: source goldens for both corpora, `generated/{dart,dart-ludicrous}` and `generated/bench/dart` committed and held current, and `test/dart` + `test/dart-ludicrous` byte-comparing the same C++-pinned `testdata/wire` goldens as every other leg, each run twice: JIT `--enable-asserts` and AOT release. The Dart SDK is pinned (3.13.2) via the Makefile `DART` variable; CI installs the same version.

```
cd test/dart && dart --enable-asserts main.dart            OK
cd test/dart && dart compile exe … && schema_test_dart     OK   (AOT)
cd test/dart-ludicrous && dart --enable-asserts main.dart  OK
cd test/dart-ludicrous && … schema_test_dart_ludicrous     OK   (AOT)
```

`make test` is green end to end across all seven languages (C++, C, Go, Rust, C#, JS both modes, Dart both modes, goldens, fuzz seeds, staleness).

**Negative control:** with the bool width broken from 1 to 2 bits in the emitter, the leg goes red on the first pinned instance and stays red across the corpus; restored, green:

```
FAILED: shipcreate_flags: Dart bytes == the C++-pinned bytes
FAILED: inputpacket: wrote 63 bytes, golden has 61
FAILED: inputpacket: measure 485 bits vs 63 bytes written
…
```

## The prediction test (#155)

`bench/dart` is the instrument: the generated codecs over the four family `rt` shapes, measured with serialize.dart's own bench methodology (same LCG, same vary mappings, 64 read variants, best-of-five, 1M packets) and **golden-gated**: every pinned instance byte-compares against `testdata/wire` and all 64 variants of every shape round-trip field-perfect before any number (§1.5). Rates are linear at 2x iterations.

Same sitting, MacBook Air (Apple Silicon), AOT release, quiet machine. These are iteration numbers; read them as ratios:

| row | op | C++ (serialize bench.cpp, -O3) | serialize.dart library (AOT) | **generated Dart (AOT)** | generated vs library | generated gap to C++ |
|---|---|---:|---:|---:|---:|---:|
| packet (12-op stream shape, 49 B) | write | 98.6 Mpps | 15.6 Mpps | **17.6 Mpps** | 1.13x | 5.6x |
| packet | read | 141.6 Mpps | 11.2 Mpps | **24.5 Mpps** | 2.19x | 5.8x |
| int packet | write | 177.3 Mpps | 21.4 Mpps | **59.4 Mpps** | 2.8x | 3.0x |
| int packet | read | 548.3 Mpps | 16.0 Mpps | **88.0 Mpps** | 5.5x | 6.2x |
| bits packet | write | 526.5 Mpps | 31.9 Mpps | **58.1 Mpps** | 1.8x | 9.1x |
| bits packet | read | 779.3 Mpps | 21.3 Mpps | **95.7 Mpps** | 4.5x | 8.1x |
| mixed packet | write | 150.5 Mpps | 20.8 Mpps | **53.8 Mpps** | 2.6x | 2.8x |
| mixed packet | read | 573.3 Mpps | 14.0 Mpps | **84.9 Mpps** | 3.9x | 6.8x |

(Library rows this sitting: bitpacker 737.6/1001.5 MB/s; the library's same-sitting gaps to C++ were packet 6.3x/12.6x and shapes 8–37x. `measure` on fully static shapes folds to a literal, 107 Mpps vs the library's 37.8; the row is priced by the vary call.)

**Verdict on the prediction** ("constant-width monomorphic generated code closes most of the 5–6x"):

- **Read side: confirmed.** 12.6x → 5.8x on the packet row; 34–41x → 6–8x on the pure-field shapes. Generated reads run 2.2–5.5x the library.
- **Write side: confirmed on pure-field shapes** (8–16x → 2.8–9.1x, 1.8–2.8x over the library), **refuted on the packet row**, which moved only +13% (6.3x → 5.6x). The residue there is not dispatch or width arithmetic (those are gone by construction); it is the `[17]uint8` blob riding the per-byte merge loop where the library (and C++) use a fused word-store bulk copy, plus the float32 conversion scratch. Both are named follow-on passes below, each with these rows as its baseline.

Follow-on passes, ruled ≠ now: (1) the `ir.AlignedFixedByteArrays` bulk-bytes path for Dart; (2) bench/run.sh + CSV-preamble integration per bench/README.md; (3) adding the refusal-capable emitters (js, dart — both can return errors) to the fuzz backends table, which currently treats a generate error as fatal.

## Rides along: a js flat-tier wire-refusal fix

Transliterating the flat kernels convicted a latent defect in `internal/codegen/js/flat.go`: the read-side run fusing counted a union field at `MaxBitsUnion` (tag + largest arm), so the fused bound overshot any valid stream carrying a smaller arm: `ReadProbeColliderFlat` refused the runtime tier's own bytes whenever every arm was None. First commit fixes it (unions are dynamically sized; the tag read carries its own bound), re-pins the `WireFlat.js` source golden (wire unchanged), and adds the convicting all-None shape to `test/js` as a permanent regression.


rowan-claude:
Branch rebased and force-pushed. Five SPEC-TABLES.md commits were riding this PR by accident (the branch was cut from a local checkout sitting on spec-view-table) and are removed; that specification ships only through its own adjudication PR, #159. Also folded: the seven-language register sweep the adversarial review required (README, USAGE, the CLI usage string, SECURITY), with SECURITY.md now naming the fuzzed backends truthfully: C, C++, C#, Go and Rust. Extending the fuzz table to the refusal-capable emitters stays a named follow-on. Full suite green after both changes.

Issue #170, opened 2026-08-30
The generated-code optimization round: forensics-corrected baselines, then per-emitter profiling passes


rowan-claude:
The generated-code optimization round, ruled due once the Java and Elixir backends merge. Standing routing per the same conversation: all performance work happens at the schema level, in the emitters, never in the serialize runtime libraries, which are complete as reference rows.

Order of work. First the bench forensics that make every later number trustworthy: a generated-C++ bench leg as the honest reference column (the current tables compare generated code against the serialize library's bench.cpp, which conflates generated-versus-library with language-versus-language), consumption discipline audited across every bench runner so no JIT can elide timed work (the JVM measure row already demonstrates sinking), and the packet-write process-bimodality mechanism pinned or at least bounded. Second, correctness before speed: the Dart fixed-point default double-scaling (already filed). Third, the named residues: the Dart bulk-bytes path for aligned byte arrays (already filed), and whatever the Elixir backend's first bench names as its residue. Fourth, a profiling pass per generated leg, applying each port's proven idioms wherever an emitter lacks them, with per-row before and after numbers against the forensics-corrected baseline and the family's golden-gated discipline throughout. Wire bytes never move; every optimization merges only with the full cross-language conformance gate green.

rowan-claude:
Refinement from the PR #169 forensics and the owner: the per-emitter profiling pass is a technique matrix, not an open-ended profile. Enumerate every optimization technique any emitter carries (fused bulk copy for aligned byte runs, hoisted bounds checks, literal-width folding, inline discipline, per-language fast idioms like VarHandle word access, typed-array overlays, unboxed integer windows), cross against all backends, and fill the holes: Dart lacks bulk copy (the measured residue of its refuted packet-write row), C# is unaudited, and the new backends get audited on arrival. Every fill is measured against the generated-C++ reference leg the review built, on the corrected baselines, which also separates technique holes from language physics: how much of generated Java at a third to a half of generated C++ is closable is exactly what the matrix answers.

rowan-claude:
[...]

Issue #168, opened 2026-08-30
Dart backend double-scales fixed-point defaults (DefInt is already raw)


rowan-claude:
Found while building the Java backend, cross-checking storage-default semantics against the C++ golden.

`ir.Field.DefInt` for a fixed-point default is ALREADY the raw scaled integer — the C++ target pins this:

```
testdata/golden/ludicrous/cpp/Ludicrous.h:123
    int32_t w = 1073741824; // fixed(2, 30) — Q2.30, raw value scaled by 2^30
```

The Dart emitter scales it again (`internal/codegen/dart/dart.go`, `scalarStorage`, both the `is128` fixed branch and the `TFixed` branch call `shiftedRaw(f.DefInt, FracBits)`), so the pinned Dart golden carries a double-scaled default:

```
testdata/golden/dart/../ludicrous/dart/Ludicrous.dart:2123
  int w = 1152921504606846976;   // 2^60 — should be 1073741824 (2^30)
```

Latent because nothing serializes a default-constructed `FixedQuat`: the value is out of the declared raw range, so the first `--enable-asserts` write of a default-constructed instance would trip the writer contract, and a release write would put a wrong (masked) value on the wire.

Fix is one-line-per-branch (use `f.DefInt` directly) plus a golden re-pin (storage text only — the wire encodings are untouched). The Java backend already uses `DefInt` directly; a regression check that exercises a defaulted fixed field end to end would cover the class for every backend.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rowan-claude:
Independently confirmed by the PR #169 adversarial review, with citations: the C++ golden pins w = 1073741824 (raw, scaled by 2^30; cpp.go documents DefInt as raw) while the merged Dart golden carries 1152921504606846976, which is 2^60, produced by shiftedRaw over the already-raw DefInt at internal/codegen/dart/dart.go lines 653 and 664. The Java backend applies shiftedRaw only to whole-unit bounds and pins the correct value. Latency analysis in the filing also confirmed: no test serializes a default-constructed FixedQuat, so the defect is latent until one does. Scheduled inside the generated-code optimization round as the correctness-before-speed item.

Issue #167, opened 2026-08-30
Elixir backend: unboxed-width monomorphic generated code (measured directives from the serialize.elixir port)


rowan-claude:
Measured directives for the Elixir backend, from the serialize.elixir performance phase (mas-bandwidth/serialize.elixir, commits 5fbaa83..0fdd46f), the way #155 and #156 carried them for Dart and Java. Final library numbers on the bench that produced them (MacBook Air, Apple Silicon, iteration numbers): bitpacker 46/60 MB/s, packet rows 1.1-2.7 M packets/s, roughly 2.7-5x behind production JavaScript and two orders behind C++.

What generated code must do differently, each backed by a measurement in that repo:

1. Keep every intermediate under the small-integer boundary. The library's single biggest wins came from packing in 32-bit words (the 64-bit scratch boxed at or above 2 to the 59th on most writes, +41% when fixed) and decoding through 40-bit windows (+91%). Generated code knows every width at generation time, so it can prove statically that no intermediate boxes.

2. The allocation ceiling is the API, and generated code does not have to pay it. The library's immutable per-op surface allocates a result tuple plus a new stream struct per field, and identical arithmetic threaded flat ran 319 versus 47 MB/s, a 7x gap that is entirely heap traffic. Generated readers and writers are single functions per type: thread the offset and scratch as plain accumulator arguments through the whole body and allocate exactly once, building the result struct at the end. No stream struct, no tuple per field.

3. Function-head dispatch and per-type monomorphism, never a module table: the library measured +10-21% from head dispatch alone, and generated code starts monomorphic by construction.

4. Validation is always on. The BEAM has no compile-out assert, so there is no release shape; generated reads validate everything as the family requires, and the write side's contract checks are the O(1) ones only, matching the library's stance.

5. The fairest context rows are the JS backend's generated numbers on the same machine, not C++.

rowan-claude:
[...]

PR #171, opened 2026-08-30
The Elixir backend: the ninth language

rowan-claude:
The eighth backend: `--lang elixir`, built to the measured directives of issue #167 (the serialize.elixir port's performance phase). Branched from main; merges independently of #169 — see the collision note at the end.

## The backend

`internal/codegen/elixir` (`elixir.go` + `functions.go`), registered in `compiler/builtin.go`. Generated code is self-contained (no serialize.elixir dependency, no hex packages) and mix-format-clean by construction (`mix format --check-formatted` rides `make test` as a refuser; the goldens pin the formatted text).

The issue #167 directives, realized:

- **Every intermediate stays under the BEAM small-integer boundary.** The write side merges in at-most-32-bit groups through a byte-granular scratch — after every merge the whole bytes flush to the output binary, so the scratch never exceeds 7+32 bits and nothing approaches 2^40. The read side is the port's 40-bit window decode (`rd/3`, `@compile {:inline, ...}` so literal widths reach the mask arithmetic).
- **Single functions per type, plain accumulator threading, one allocation.** `write_<name>(value)` threads `data`/`scratch`/`scratch_bits` as rebound locals; `read_<name>(data, num_bits)` threads `bits_read` and builds the result struct once at the end. No stream struct; no per-field tuple (the merge is branchless single-variable rebinds precisely because an Elixir conditional cannot rebind several variables without one). Arrays are tail-recursive function-head loops — one returned tuple per loop, never per element; wire branches and union arms cost one tuple per construct, the only way values cross an Elixir scope boundary.
- **Function-head dispatch, monomorphic by construction** — loop helpers dispatch on `[]` / `[e | rest]`, per type, per field.
- **Validation always on.** Reads validate everything — fused static-run bounds, ranges, wire constants, reserved and align padding, union tags, string content — and never raise on hostile bytes: the body throws `:invalid` to a catch at the surface (an early exit with no per-op result tuple), surfaced as `{:ok, value} | :error`. Writer contracts are the O(1) checks only, raising `ArgumentError` — the port's misuse convention; the BEAM has no compile-out assert, so there is no release twin.

Coverage: every IR construct the other backends emit — enums (headroom and empty), flags with widening, unions (empty/None arms, arrays of unions), constants rendered symbolically (cross-file references fully qualified; `div`/`rem` carry schema's truncating semantics exactly), fixed/ufixed at 16/32/48/112-bit storage plus degenerate ranges, int128/uint128 as plain BEAM integers (arbitrary precision makes the (hi, lo) pair emulation of bounded-word targets unnecessary), arrays, string/bytes as binaries, transitive defaults, branches with SPEC §5 zero semantics, const/reserved/align, empty types.

Named deviations, each with its reason:

- **Refusal surface** `{:ok, value} | :error`: the port's own ok/error convention with no stream to return; reads are pure functions of the bytes.
- **Non-finite floats travel as `{:nonfinite, bits}`** exactly like the port — BEAM float terms cannot hold NaN or the infinities; round-trips reproduce the transmitted pattern bit-exactly (leg-tested both widths).
- **Strings validate UTF-8 on read** in addition to the family's interior-null refusal — the read side is the trust boundary and a BEAM binary claiming to be a String should be one. This refuses only hostile wire no other leg pins as accepted.
- **defstruct defaults fold to literals** (other backends render defaults symbolically): defstruct defaults evaluate at module compile time, where a symbolic cross-module call would impose an ordering the runtime surface never has. Constants themselves stay symbolic (`def max_position_units, do: max_world_meters() * position_units()`) — resolution is at runtime, so declaration order never binds.
- **Fixed-point defaults use `ir.Field.DefInt` directly** — it is already the raw scaled integer. The Dart backend's double-scaling (issue #168) is not reproduced here.
- **Counted arrays and strings carry no separate count/length field**: the list length and `byte_size` are the count, the storage principle in a value-semantics language.

## Conformance proof

`make test` green end to end (all eight backends; the Elixir legs and the formatter refuser included), `make generated-current` green, `golangci-lint run ./...` 0 issues. The Elixir legs, run exactly as the Makefile runs them:

```
$ mix format --check-formatted generated/elixir/*.ex generated/elixir-ludicrous/*.ex \
    generated/bench/elixir/*.ex generated/bench/elixir/realworld/*.ex
(clean)

$ cd test/elixir && elixir main.exs
OK
$ cd test/elixir-ludicrous && elixir main.exs
OK
```

Those legs byte-compare every pinned instance against the C++-pinned `testdata/wire` goldens: main corpus (shipcreate_flags, rigidbody_moving/at_rest, chat, probe_header, inputpacket, testdata, compressed_probe, probebits, probecollider, probearray), the ludicrous unit (ludicrous_state, ludicrous_state_untargeted, degenerate_probe, unsigned_probe), and the bench corpus (bench_packet/ints/bits/mixed, real_packet at its pinned 1629 bits) — plus round-trips to identical bytes, measure held to written bytes, §5 branch zeroing, transitive defaults, the refusal vectors (corrupt const, nonzero reserved, smuggled offsets, out-of-range tags/counts/lengths, interior null, invalid UTF-8, truncation, oversized num_bits), and the always-on writer contracts.

Never-raise under noise, beyond the pinned vectors: 20,000 random buffers of random sizes through the three deepest readers:

```
hostile noise verdicts (no raise = this line prints): %{error: 55196, ok: 4804}
```

## Negative control

Widened every ranged write by one bit (`ir.BitsRequired(min, max) + 1` in `emitWriteOffset`), regenerated:

```
FAILED: shipcreate_flags: wrote 29 bytes, golden has 28
FAILED: shipcreate_flags: Elixir bytes == the C++-pinned bytes
FAILED: shipcreate_flags: measure 219 bits vs 29 bytes written
FAILED: shipcreate_flags: round-trips to identical bytes
...
FAIL	github.com/mas-bandwidth/schema/v2/internal/goldens	0.423s
```

Restored: both legs `OK`, goldens `ok`, tree clean.

## The bench

`bench/elixir/main.exs` — the minimal-form golden-gated runner (§1.5), serialize.elixir's own methodology: same LCG, same vary-function field mappings, 64 read variants each decoded and field-verified before any row is timed, best of five trials, 1M packets per row. Consumption discipline for every number below: every loop's work flows into a sink accumulated across trials and printed under `SERIALIZE_BENCH_SINK` (this run: `sink: 404020968750`), and the LCG varies every written packet, so no loop's work can be folded away.

MacBook Air, Apple Silicon (M2), Erlang/OTP 29.0.5 + Elixir 1.20.4 (BeamAsm), Node 20 for the JS rows — iteration numbers on a shared laptop, not certification numbers. Two of the earlier JS attempts stalled under macOS App Nap; the rows below come from a caffeinate run. Library and generated rows measured back to back, same machine, same session.

| row | serialize.elixir library (streams) | generated Elixir | generated / library |
|---|---|---|---|
| packet write | 41.8 MB/s (0.9 M pkt/s) | 82.6 MB/s (1.8 M pkt/s) | 2.0x |
| packet read | 68.9 MB/s (1.5 M pkt/s) | 123.6 MB/s (2.6 M pkt/s) | 1.8x |
| packet measure | 1.9 M pkt/s | 6.0 M pkt/s | 3.2x |
| ints write / read | 1.3 / 1.8 M pkt/s | 2.4 / 10.6 M pkt/s | 1.8x / 5.9x |
| bits write / read | 1.5 / 2.7 M pkt/s | 3.2 / 13.8 M pkt/s | 2.1x / 5.1x |
| mixed write / read | 1.1 / 1.7 M pkt/s | 2.3 / 9.1 M pkt/s | 2.1x / 5.4x |

The fair context rows (issue #167 directive 5) — the generated-JS `rt` rows on the same machine:

```
bench_packet       write       3.32 M msg/s      154.9 MB/s   (min 3.24, max 3.34, spread 3.1%)
bench_packet       read        4.01 M msg/s      187.3 MB/s   (min 3.95, max 4.04, spread 2.2%)
bench_ints         write       5.62 M msg/s       75.0 MB/s   (min 5.45, max 5.69, spread 4.3%)
bench_ints         read        6.37 M msg/s       85.0 MB/s   (min 6.35, max 6.38, spread 0.4%)
```

(Node 26.7.0, production mode, the runner's own golden gate passed. The bench_bits and bench_mixed JS rows were not captured — the ~25-minute JS run was stopped after bench_ints; the two rows above are the context that matters, bench_packet being the stream-comparable shape.) Generated Elixir measures roughly 0.4-0.55x generated JS on the packet shape (write 1.8 vs 3.3 M pkt/s, read 2.6 vs 4.0) and 0.4x/1.7x on ints (write 2.4 vs 5.6, read 10.6 vs 6.4 — the Elixir read of a pure-ranged-int shape outruns the JS flat read on this machine). Against C++ the family's usual two orders stand; per issue #167 directive 5, JS is the fair context and these rows sit inside it.

**The honest verdict on the 7x-dodge prediction.** The prediction was that single-function accumulator threading dodges the port's measured 7x allocation ceiling. The read side substantially delivers: 5.1-5.9x over the library streams on the pure-field shapes — threading one integer through an inlined window decode with a single struct allocation at the end is exactly the flat shape the port's probe measured, and the residue from 7x is the try/catch frame, the per-shape struct build and the harness's own vary/LCG. The write side does not: 1.8-2.1x, not 7x. The write accumulator is a binary, and the byte-granular flush performs an append BIF call per field where the port's word writer appends once per 32 bits — the price of the only merge shape that is both boxing-free and tuple-free under Elixir's scoping rules. A branch-guarded append (skip when no whole byte completed) was tried and measured indistinguishable, so the simpler branchless form ships. The write side's next step — coalescing statically-adjacent fields into one masked word merge at generation time, which the emitter can prove — is #170's kind of work, not this landing's.

## The sweep (seven -> eight)

README (the tagline and the feature bullet, the `--lang` line, the backends-through-the-interface line, the language lists, the self-contained note), USAGE (`--lang` list, the slack table — Elixir none — the write-idiom paragraph, both count sites, the `Targets()` example — which also gains the `dart` entry it was missing — the emitters count, and a per-language **Elixir** note), FAQ (every count site and language list; the five-backend fuzz sentence untouched), SECURITY (the two count sites; the fuzz-backends sentence untouched), CONTRIBUTING (toolchain list with the OTP/Elixir pins, both count sites), CLAUDE.md (the what-make-proves paragraph), cmd/schema/main.go (both usage strings). Repo description left to the orchestrator.

**Count collision with #169 (Java):** both PRs take the same seven-to-eight sites (README/USAGE/FAQ/SECURITY/CONTRIBUTING/CLAUDE.md, the ci.yml "all seven/eight languages" comments, `TestGoldenLudicrousSource`'s target list and its ALL-SEVEN comment, and the `make test` invocation lines in ci.yml). Whichever merges second rebases into nine-language wording and a combined target list — flagging it here so the orchestrator resolves it at merge time rather than discovering it.

Toolchain: Erlang/OTP 29.0.5 + Elixir 1.20.4 pinned per the Makefile's dist/ pattern (urls + sha256 in the Makefile note); CI uses erlef/setup-beam (macOS supported for OTP 25-29) in the test job and every inline-gate leg.


rowan-claude:
Branch rebased over the Java merge and force-pushed by the coordinator. The count collision resolved to nine-language wording across every swept site (README, USAGE incl. the slack row and write-idiom paragraphs and the per-language note, FAQ, SECURITY count sites, CONTRIBUTING, CLAUDE.md, both CLI usage strings, the goldens harness and its ludicrous list, ci.yml with both setup-java and setup-beam and one merged make-test override line). One additional repair made during the rebase: commit ee0a58d had accidentally committed the 6.4MB compiled schema binary; it is stripped from history on this branch and /schema is gitignored. Full nine-language suite green on the rebased head by the coordinator's own run, 26 legs.

Issue #175, opened 2026-08-30
Bench estate finishing pass: complete the ninelang sweep record, the elixir runner header, the quick table's family column


rowan-claude:
The finishing pass for the one-contract bench estate, sized for one sitting on any bench. First, complete the ninelang full-sweep record: bench/results/2026-08-30-ninelang-air-1.csv holds the cpp, c, go, rust and cs legs; run the remaining js, dart, java and elixir legs under the full standard and land them either as an appended block whose provenance is recorded (the existing preamble pins an earlier commit; the bench code is byte-identical through the merge, but the appended legs must say what ran them) or as a completed re-run of the whole sweep. Second, the review's two riding notes: bench/elixir/main.exs still carries its pre-conformance header (best-of-five wording, a BENCH_STREAM_PACKETS override that no longer exists, and a named follow-on that this merge delivered) — rewrite present-state; and the quick mode's blended table prints no family column while its rows mix rt-family natives with gen-family newcomers — print the family beside the language so the table cannot be misread from a screenshot. Two cosmetics at will: the doubled java compile-path documentation, and the js runner's missing CSV header line under --only js, which predates the merge.

rowan-claude:
Correction from the owner's question, and it reprioritizes this issue and the tie-breaker. The blended quick table is mixed-subject: the c, cpp, go, rust and cs rows measure the hand-written serialize libraries' runtime stream paths on the family shapes, while the java, dart, js-flat and elixir rows measure schema-generated code. The PR #172 review named this (its note two) and the conformance pass unified only the contract, not the subject. Consequence one: the Java-at-100-percent row compares generated Java against library C++, and the earlier forensics measured generated C++ at roughly 3.4 times the library's rate on these shapes, so a subject-unified table should restore the native tier to first by around three times over Java. Consequence two: the tie-breaker issue's premise is half wrong; subject unification comes before PGO, which is a second-order term on top of it. The corrected priority for the table work: generated-Bench legs for the five native languages first (the follow-on the review already named), family and subject printed in every table, then the PGO and sink-equalization pass on the unified subjects. Until that ships, the honest reading of the current table is: within the generated subject, java, dart, js-flat and elixir rank as shown; within the library subject, c, cpp, rust, cs and go rank as shown; and the two halves must not be ranked against each other.

gafferongames:
[...]

Issue #176, opened 2026-08-30
Native-tier tie-breaker: PGO plus mcpu-native C/C++ legs, equalized sinks, three-way rerun


rowan-claude:
The native-tier tie-breaker, promoted from #170's comments at the owner's ask: the blended quick table shows Java at 100 percent with C++ at 105 and C at 106, a statistical tie whose three known terms are C2's free profile-guided layout (the C and C++ legs build without PGO), exact-core targeting (bounded at five percent or less by an earlier mcpu-native run), and the consumption asymmetry the PR #172 review named (the C++ leg escapes the full buffer per iteration while the Java leg publishes a narrower sink).

The experiment, apples to apples: build the C and C++ bench legs two-pass PGO (clang -fprofile-instr-generate, one --quick --bare profile run with LLVM_PROFILE_FILE set, xcrun llvm-profdata merge, rebuild with -fprofile-instr-use and -mcpu=native) and equalize the consumption discipline across the native-tier legs, either full-buffer escape everywhere or the narrow sink everywhere, stated in the CSV. Then three same-sitting quick runs of c, cpp and java. If the tie is machinery, C and C++ retake five to ten percent and the tier settles native-first; if the tie survives equalized measurement, that is the published finding.

One recipe note from a first attempt: the bench COMMON_FLAGS reference the driver's SERIALIZE variable for the include path, so a hand build outside run.sh must set SERIALIZE to the sibling serialize checkout (or pass -I to it explicitly) or the generated headers fail to find serialize.h. Cleanest is probably a --pgo flag on run.sh itself so the two-pass build lives in the driver where the flags already resolve.

gafferongames:
DEFERRED by owner lock (2026-08-31, ruling on #170): no changes to C/C++ during the optimization round, and PGO/-mcpu=native changes the reference's build. This experiment re-arms when the round closes and the lock lifts. The sink-equalization half of this issue's premise proceeds separately in the #175 pass, which does not touch the c/cpp legs.

PR #178, opened 2026-08-30
Generated bench legs for the five native languages: the schema table becomes single-subject

### LINKED ISSUE #177: Generated bench legs for the five native languages: the schema table's missing subject
The schema table's missing subject, and the precondition for publishing it again as one table: generated-code bench legs for the five native languages. Today the c, cpp, go, rust and cs rows in the blended table measure the hand-written serialize libraries' runtime paths, while java, dart, js-flat and elixir measure generated code, so the halves must not be ranked against each other. The work: a generated-Bench leg per native language over generated/bench code, under the BENCH-STANDARD contract, each language in its generated best case per the profiling doctrine on #170, with subject and family printed in every row and in the quick table. The earlier forensics measured generated C++ at roughly 3.4 times the library's rate on these shapes, so expect the native tier to retake the top of the unified table by around three times over generated Java; if it does not, that is a finding, not a smoothing target. After this ships, the PGO and sink-equalization pass (#175) runs on the unified subjects, and the per-language pairs (generated best case against hand-written usage) become the published compiler-value numbers.

gafferongames:
The blended quick table is now single-subject. Every language's headline row measures schema-GENERATED code, the family prints in every row, and the hand-written-runtime blend rides as a second labeled section that never ranks against it. From the first single-subject sweep (M2 MacBook Air, one caffeinated sitting, 2026-08-31, `bench/results/2026-08-31-gennative-quick-arm64-macbook.csv`):

```
quick mode — iteration instrument, not certification (bench_mixed only, blended write+read)
subject: schema-GENERATED code (family gen) — what the compiler delivers
  lang     family         ns/msg  % of best
  cpp      gen              1.73       100%
  c        gen              1.88       108%
  java     gen              5.70       329%
  rust     gen             10.19       588%
  cs       gen             13.85       799%
  dart     gen             16.04       925%
  go       gen             26.82      1546%
  js       gen             72.96      4207%
  elixir   gen            282.25     16275%

subject: hand-written runtime usage (family rt) — what the serialize libraries deliver by hand; never ranked against gen
  lang     family         ns/msg  % of best
  cpp      rt               5.94       100%
  c        rt               6.07       102%
  rust     rt               9.91       167%
  cs       rt              22.54       380%
  go       rt              25.18       424%
```

The #170 forecast holds at the top: generated C++ retakes first at 3.3x over generated Java, right on the predicted ~3x, with generated C beside it. Below the top it does NOT hold uniformly, and that is a finding, not a smoothing target: generated Java outranks generated Rust (1.8x), C#, and Go. The per-language compiler-value pairs (gen vs rt, full 7-run medians on bench_mixed / bench_packet, M msg/s write+read) say why:

| lang | gen mixed w/r | rt mixed w/r | gen packet w/r | rt packet w/r |
|---|---|---|---|---|
| cpp  | 595 / 550 | 148 / 181 | 499 / 390 | 92 / 128 |
| c    | 596 / 474 | 158 / 172 | 495 / 384 | 117 / 140 |
| rust | 97 / 102  | 102 / 100 | 66 / 76   | 77 / 90 |
| cs   | 76 / 71   | 48 / 42   | 35 / 54   | 32 / 34 |
| go   | 40 / 34   | 43 / 36   | 22 / 28   | 27 / 28 |

C and C++ get ~3–5x from generation (the always-inline monomorphic emit); C# gets ~1.6x from the JIT inlining the generated statics; Rust and Go generation sits at parity or a few percent BELOW hand-written — both already monomorphize the same runtime, and the generated form buys nothing there yet. That gap is exactly what the #170 technique matrix exists to attack, on these now-existing baselines.

What this PR does:

- Adds family-`gen` rows for `bench_packet` / `bench_ints` / `bench_bits` / `bench_mixed` to the five native runners (`bench/{cpp,c,go,rust,cs}`), over `generated/bench/<lang>` (verified current against `bin/schema` regeneration — zero diff). Same goldens, same pinned instances, same LCG vary mappings, same iteration counts, same driver discipline as each leg's existing gen rows. The existing `rt` rows are untouched.
- `--quick` in each native leg now runs the gen `bench_mixed` (golden gate intact, 3 measured runs) beside the rt row; the run.sh blended table selects family gen per language (js: codec=flat, never runtime), prints the family per row (part of #175's quick-table item), and prints the rt blend as its own section.
- Rust: `benchcorpus` (generated/bench/rust) joins the bench crate graph; the §3.5 runtime-override grows a matching manifest, and all four override manifests now carry the `serialize-official` package rename they predated (they would not have resolved when the override fired).
- The inline-verdict backfill becomes family-guarded: rt-loop verdicts stamp only rt/bits rows, corpus attributions only gen rows. The new gen Bench rows have no verdict source yet and stay `inline=unknown` (correctly un-ratioable) — teaching the §4 pass to attribute them is a named follow-on.

What is and is not measured: generated best case per the #170 doctrine — plain optimized release builds, no PGO (that is #175, which now runs on unified subjects); quick numbers are the iteration instrument, never certification; the full 7-run legs for all five languages are recorded in `bench/results/2026-08-31-gennative-{cpp,c,go,rust,cs}-arm64-macbook.csv`; no driver pass (interleaved rounds, window gate) was run for these, so absolutes carry the usual single-sitting caveat.

Negative control: doctored one byte of `testdata/wire/bench_mixed.bin`; the cpp, c, and cs quick legs each refused with `WIRE GOLDEN MISMATCH: bench_mixed ... refusing to bench code that does not match the corpus` twice (gen gate first, then rt), emitted zero CSV rows, and exited 1 with the corpus_id visibly changed; restored the golden and confirmed all three green at the pinned corpus_id.

Closes #177. Part of #175 (the quick-table family column).


gafferongames:
Adversarial review verdict: MERGE, dual-reproduced. Receipts in brief:

- The Java-above-Rust/C#/Go ordinal finding SURVIVES adversarial verification — same LCG, same vary mappings token-for-token, same iters, bytes_per_op identical in every language's rows, corpus_id identical (goldens from pinned disk files; a doctored byte changes the id, proving the oracle never shares the transformation). Mechanism confirmed real: generated Java is a self-contained VarHandle word codec with dormant asserts; generated Rust carries live per-field range branches plus runtime stream calls.
- One MAJOR measured caveat, attached to any publication of the Java magnitude: the Java read sink is the narrowest in the estate (publishes one field; cpp/c clobber the whole struct). Widening it to all 11 decoded fields moves Java read 175.4 -> 135.5 M msg/s (-23%), blended 5.67 -> 6.56 ns — published 329% is ~380% at the widest sink (upper bound; some delta is the sink adds themselves). The ranking holds at the widest sink (Java still 1.55x ahead of generated Rust). The sink/checks equalization pass (#175) inherits this as MEASURED at -23% read / ~16% blended, not the "few percent" #176 estimated.
- Negative control reproduced independently; blend math recomputed by hand from the committed CSVs, all rows match; contract fidelity verified across all five legs; rt rows byte-untouched; regen zero-diff; the rust override-manifest breakage confirmed real (probe manifest: unrenamed form fails cargo metadata) and the fix minimal.

Caption to carry when this table republishes: rows within family gen are ranked under one contract and one corpus; sink discipline and checks mode are equalized in the #175 pass, and until then the Java magnitude carries the measured sink bound above.

PR #180, opened 2026-08-31
NEGATIVE CONTROL: cpp-lock must refuse this PR (do not merge)

gafferongames:
One appended newline to bench/cpp/bench_main.cpp — a locked path under bench/LOCK. This PR exists to prove the cpp-lock guard goes red on the class it guards; it closes unmerged once the red is observed. A guard never seen to fail is decoration.


gafferongames:
Red observed: cpp-lock FAILURE with the locked-path refusal naming bench/cpp/bench_main.cpp. The guard fires on its class. Closing unmerged — the control did its job.

PR #181, opened 2026-08-31
Bench equalization pass: full-struct sinks across the nine legs, loud skips, the caption — the blended table republished (#175)

gafferongames:
The bench equalization pass (#175): every read loop in the estate now observes the FULL decoded struct per iteration, the quick table carries its caption, a skipped leg is loud, and the blended table republishes measured under the equalized discipline — one contract, one subject, one corpus, one machine, one caffeinated sitting.

## The equalized blended quick table

`bench/run.sh --quick`, nine languages, M2 Air, 2026-08-31, corpus `457b96dfc5a0ffcd`, CSV: `bench/results/2026-08-31-equalized-quick-arm64-macbook.csv`.

subject: schema-GENERATED code (family gen) — what the compiler delivers

| lang | family | checks | ns/msg | % of best |
|---|---|---|---:|---:|
| cpp | gen | removed | 1.72 | 100% |
| c | gen | removed | 1.85 | 107% |
| java | gen | contract | 6.49 | 377% |
| rust | gen | always | 10.22 | 594% |
| cs | gen | always | 13.59 | 790% |
| dart | gen | contract | 16.37 | 952% |
| go | gen | always | 26.93 | 1566% |
| js | gen | contract | 75.53 | 4391% |
| elixir | gen | contract | 288.68 | 16782% |

subject: hand-written runtime usage (family rt) — never ranked against gen

| lang | family | checks | ns/msg | % of best |
|---|---|---|---:|---:|
| cpp | rt | removed | 5.97 | 100% |
| c | rt | removed | 6.04 | 101% |
| rust | rt | always | 9.96 | 167% |
| cs | rt | always | 22.45 | 376% |
| go | rt | always | 25.19 | 422% |

**Caption (now printed by the driver above every quick table):** held constant — contract (BENCH-STANDARD §2.8 quick: bench_mixed, blended write+read over max rates), corpus (id per CSV row), machine, one sitting; sink discipline equalized to full-struct observation (#175). NOT equalized: checks mode — printed per row; a cross-language checks ruling is deferred to the owner (no repo-recorded ruling names a fastest-correct checks form per language, and this pass never silently changes what a leg measures).

## The sink effect, per language (bench_mixed quick, max rates, frozen #178 baseline sitting → this sitting)

| lang | read M msg/s | Δ read | blended ns/msg | Δ blend | % of best | leg changed? |
|---|---|---:|---|---:|---|---|
| cpp | 554.5 → 553.0 | -0.3% | 1.73 → 1.72 | -0.8% | 100% → 100% | NO (frozen reference) |
| c | 475.4 → 481.5 | +1.3% | 1.88 → 1.85 | -1.6% | 108% → 107% | NO (frozen reference) |
| java | 171.9 → 136.0 | **-20.9%** | 5.70 → 6.49 | **+13.8%** | 329% → 377% | YES — one field → full fold |
| rust | 99.9 → 99.5 | -0.5% | 10.19 → 10.22 | +0.3% | 588% → 594% | NO (verified conformant) |
| cs | 69.9 → 70.9 | +1.5% | 13.85 → 13.59 | -1.9% | 799% → 790% | NO (verified conformant) |
| dart | 79.3 → 74.3 | **-6.4%** | 16.04 → 16.37 | +2.1% | 925% → 952% | YES — one field → full fold |
| go | 34.7 → 34.5 | -0.5% | 26.82 → 26.93 | +0.4% | 1546% → 1566% | NO (verified conformant) |
| js | 53.0 → 38.3 | **-27.7%** | 72.96 → 75.53 | +3.5% | 4207% → 4391% | YES — sink+1 → full fold |
| elixir | 9.0 → 6.7 | **-26.1%** | 282.3 → 288.68 | +2.3% | 16275% → 16782% | YES — one field → full fold |

The java movement lands on the PR #178 review's measured prediction (-23% read / ~16% blend at the widest sink, upper bound including the sink adds): measured here -20.9% read / +13.8% blend, published 329% → sink-equalized 377% (the review predicted ~380%). Every unchanged leg moved ≤1.9% across the two sittings — the window is clean. **No ranking changed anywhere.**

## What equalization means, per leg

One bar (BENCH-STANDARD §2.7, added by this PR): every decoded field store materializes in memory every iteration and cannot be elided. Two mechanisms:

- **cpp / c — FROZEN, untouched** per the owner's C/C++ lock during the optimization round (#170): their empty-asm whole-struct clobber is the reference discipline the other legs were equalized TO. Re-run for same-sitting comparability only.
- **rust — already conformant, verified in the emitted asm**: the timed read loop carries one out-of-line `bl read_bench_mixed` per iteration with `&out` in a register, plus `black_box(&out)` (a zero-cost memory barrier). No change.
- **go — already conformant, verified in `go tool objdump`**: the gen loop's `readFn` is an opaque indirect `CALL (R2)` through the function value per iteration; the rt loops carry a direct out-of-line `CALL main.readRtMixed(SB)`; `runtime.KeepAlive(&out)` pins liveness. Materialization is forced by the call ABI. No change.
- **cs — already conformant, verified in JitDisasm**: the shared-generic BenchMessage invokes `readFn` as a delegate (`ldr x3, [delegate, #0x18]; blr x3`) into a long-lived heap decode target; RyuJIT does no cross-iteration heap dead-store elimination. No change.
- **java — widened (the review's measured offender)**: each of the four read loops published ONE field; now a full-struct fold (floats via `floatToRawIntBits` bitcast, booleans 0/1, blob element-wise).
- **dart — widened**: same single-field → full-fold change (doubles truncated via `toInt()`).
- **js — widened everywhere**: the gen rows (flat AND runtime tiers — all corpus shapes, `real_packet`'s 97 fields included) and the rt rows sank `+1`; now full-struct folds. BigInt fields are observed through an allocation-free comparison — a per-iteration BigInt add allocates and would measure the allocator, not the observation. Every `sinkOf` is checked finite at the golden gate, so a typo'd field name (`undefined` → NaN) refuses to bench instead of silently poisoning the sink.
- **elixir — widened**: single field → full pattern-match fold (destructuring measured 19% faster than dot access — the fold ships in its fastest correct form; the remaining -26% read is the BEAM's honest full-observation price, an upper bound carrying the fold's own adds).

**Write side — unchanged everywhere, audited**: cpp/c/rust escape the whole buffer per iteration through zero-cost barriers; go/cs/js/java/dart write through opaque calls or globally-reachable heap buffers and sink the per-iteration byte count (§2.7's named barriers); elixir's write constructs the binary itself. Every leg already met the bar.

The fold legs' read numbers are upper bounds — the fold is real per-iteration work the clobber languages get free. The caption and §2.7 say so; nothing is smoothed.

## #176 — DEFERRED by owner ruling

C/C++ is locked during the optimization round (owner ruling on #170): no PGO, no `-mcpu=native`, no build-flag changes to the c/cpp legs — they are the frozen reference. The PGO/native tie-breaker experiment #176 describes is deferred until the lock lifts. Note the tie it was opened on is already dissolved: it compared library-C++ (rt) against generated Java — a mixed-subject artifact; under one subject and equalized sinks, generated C++ leads Java 3.8x.

## #175 items landed here

1. **Sink equalization** (the MEASURED inheritance from the #178 review) — done, republished, numbers above.
2. **Loud skip**: `--quick --only <lang>` with a missing toolchain now prints an ABSENT row with the reason in the headline table, and any `--only` run that produced zero data rows exits non-zero (was: SKIP + empty table + exit 0).
3. **BENCH-STANDARD §2.8** rewritten present-state to the two-section family-aware quick output; §2.7 gains the read-side sink-discipline clause (the bar, the two mechanisms, the measured fold costs).
4. **The caption**: the printed quick table names what is held constant and prints the checks mode per row.
5. **bench/elixir/main.exs** stale pre-conformance header (best-of-five wording, a `BENCH_STREAM_PACKETS` override that no longer exists, a delivered follow-on named as pending) rewritten present-state.
6. **js CSV header** now emitted under `--only js` (the #175 cosmetic).
7. **Full-standard legs for java, dart, js, elixir** recorded under the equalized sinks (`bench/results/2026-08-31-equalized-full-{java,dart,js,elixir}-arm64-macbook.csv`) — the four legs the ninelang full-sweep record was missing, with their provenance in each preamble.

Still open on #175: the doubled java compile-path documentation cosmetic.

**Negative control** (run before the sweeps): one byte doctored in `testdata/wire/bench_mixed.bin` →
```
GOLDEN GATE FAILED: bench_mixed pinned instance vs testdata/wire/bench_mixed.bin
reporting nothing.
```
exit 1; golden restored byte-identical (git-clean), gate green, corpus_id `457b96dfc5a0ffcd` unchanged through the whole pass — wire bytes never moved. No emitter or runtime changes anywhere in this PR.

### DIFF
--- a/bench/BENCH-STANDARD.md
+++ b/bench/BENCH-STANDARD.md
@@ -642,6 +642,24 @@ stride clause was added 2026-08-15, with the measurement that demanded it):
 - **Escape barriers.** Empty-asm memory clobber (C/C++), `std::hint::black_box`
   (Rust), `runtime.KeepAlive` + package sink (Go), `GC.KeepAlive` + static sink (C#).
   Write barriers observe the **buffer**, not `buffer[0]`.
+- **Read-side sink discipline (#175): every read loop observes the FULL
+  decoded struct per iteration, by one of two mechanisms, each meeting one
+  bar — every decoded field store materializes in memory every iteration
+  and cannot be elided.** The native legs get it free: the C/C++ clobber
+  and Rust's `black_box(&out)` are zero-cost memory barriers over the whole
+  struct, and the Go and C# read calls are opaque out-of-line calls taking
+  the decode target's pointer (verified in the emitted assembly — an
+  indirect call through a function value / delegate per iteration), with
+  KeepAlive pinning liveness. The JIT/managed legs whose decode call the
+  compiler inlines (java, dart, js, elixir) have no free barrier, so their
+  idiom is a per-iteration fold of every decoded field into the sink —
+  numbers added, booleans as 0/1, arrays element-by-element over the
+  decoded extent, floats bitcast or truncated, JS BigInt observed by
+  allocation-free comparison. The fold is real work the barrier languages
+  do not pay; those legs' read numbers are therefore **upper bounds**
+  carrying the observation cost (measured on bench_mixed, M2: java -20%,
+  js-flat -29%, dart -5%, elixir -28% read rate against the pre-#175
+  narrow sinks).
 - **The decode target is hoisted out of the timed loop** and reused, in every
   language. Constructing a fresh zeroed message per iteration is harness overhead
   charged to the library. `bench/cpp/bench_main.cpp:250-256` already documents this;
@@ -659,9 +677,24 @@ comparison costs minutes, not an evening; nothing it prints publishes.
   cannot hold quick mode's one-minute-per-leg bound: elixir runs 8,000,000
   (the BEAM is ~2 orders behind the native legs). Every count is recorded
   in the `iters` column as always.
-- The driver prints the blended table: per-language per-message time
-  `t = (1/w + 1/r) / 2` over the §2.2 headline (max) rates, sorted
-  ascending, fastest = 100%, every other language as its time multiple.
+- The driver prints **two blended sections, one per subject, never ranked
+  against each other** (the profiling doctrine's apples-to-apples rule,
+  enacted for the table by #177/#178): the headline section is family
+  `gen` — every language's schema-GENERATED code, the native legs included
+  since their generated Bench legs landed — and the second labeled section
+  is family `rt`, the serialize runtime API called by hand. Within a
+  section: per-language per-message time `t = (1/w + 1/r) / 2` over the
+  §2.2 headline (max) rates, sorted ascending, fastest = 100%, every other
+  language as its time multiple. Family and `checks` mode print per row
+  (checks is a recorded, NOT equalized property — §3.4), the js blend takes
+  the `codec=flat` tier only (THE js path; `codec=runtime` rows never
+  blend), and a caption above the sections names what is held constant:
+  contract, corpus, machine, sitting, and the equalized full-struct sink
+  discipline (#175).
+- A language whose leg cannot run prints as an **ABSENT row with the
+  reason** in the headline section, and an `--only` invocation that
+  produced zero data rows **exits non-zero** — a skipped leg is a refusal,
+  never a quietly green empty table (#175).
 - Warmup stays adequate for the JIT legs — the discarded run is tens of
   millions of iterations, which carries the JVM to C2/OSR; a quick leg that
   shortened warmup below that would measure the interpreter and be wrong
--- a/bench/README.md
+++ b/bench/README.md
@@ -250,6 +250,10 @@ A runner is a standalone program in `bench/<lang>/` that:
    the batch builder exactly);
 4. uses the same discipline: escape barriers (or the language's equivalent,
    e.g. `runtime.KeepAlive` / `std::hint::black_box` / `GC.KeepAlive`),
+   the read-side full-struct sink (BENCH-STANDARD §2.7: every read loop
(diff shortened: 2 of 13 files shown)


gafferongames:
Adversarial review verdict: MERGE — every headline number recomputed to the digit from the committed CSVs (the driver's own awk reproduced the published table byte for byte), the lock verified clean independently of CI, sink completeness proven per leg (js via a recursive recording-Proxy over all 18 shapes; go via independent disassembly confirming the opaque-call ABI argument), both controls reproduced (loud skip both directions; doctored golden refused with the corpus_id shifting as a second detector), checks-mode confirmed not smuggled. Five non-blocking follow-ons filed on #175: strengthen the js sink guard past Number.isFinite (it catches 1 of 4 typo classes — the current code is proven clean, the guard is just weak); name the js BigInt one-bit observation as a per-language deviation rather than equalized; correct §2.7's four percentages to the max-rate figures (they reconcile with no single statistic as written); note elixir's unexplained +6.1% cross-sitting write drift before its BLENDED figure is cited as a sink measurement (the -26% read result is untouched); optionally render ABSENT rows in the rt section too.

Issue #191, opened 2026-08-31
The data-driven bench: variant wire data + serialize-blind drivers around the generated codecs


gafferongames:
Owner, 2026-08-31, naming the day's structural inefficiency: "the obvious way to do this, since we have *a language that generates per-language serialize implementations* is to do this once in schema." — the bench_mixed redefinition cost hours because nine harnesses' shape-dependent code was hand-rewritten per language, which is precisely the work an emitter exists to do.

THE WORK: a bench-emitter mode (schema bench) that generates, per language, from bench/corpus/Bench.schema: the pinned instance initializer, the LCG vary function over value fields (structure fields pinned, the §2.2 scheme), the full-struct §2.7 sink fold in each leg's ruled idiom, and the golden byte-check. What stays hand-written, once, shape-independent: each leg's timed-loop driver, escape barriers, CSV plumbing — the parts that did not change today.

PAYOFF: a shape change becomes edit-the-schema-and-regen — minutes, not hours (first customer: wstring joining bench_mixed when #188 lands). It also deletes the hand-transcription error class the reviews keep having to catch, and the two hand-derivations of these idioms (the #178 gen legs, the #184 redefinition) are the proven reference implementations to extract from.

SEQUENCING: after #184's PR lands; before the next shape change; does not block the laggard queue (emitter work re-runs benches but never re-shapes them).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
Owner ruling, 2026-08-31, verbatim — the trust doctrine this issue exists to serve: "To be 100% clear. I don't trust hand generated serialize code per-language that is written by hand by AI. It will be slightly different per-language so profile results are probably suspect."

Consequences, effective now: (1) Any cross-language profile table whose harness shape code is hand-written per language is INTERIM — published with that label, never canonical. That includes the #184 redefinition's first sweep: its wire correctness is proven by the shared goldens, but its harness uniformity rests on review, and review catches what it looks at; generation makes the divergence class impossible by construction. (2) The canonical table publishes only from THIS issue's emitted harnesses — same generator, same field walk, same discipline in every language, uniform by construction. (3) The hand-written legs' role is exactly and only the correctness oracle for the emitter's equivalence gates. This is the profiling law's deepest form: apples to apples enforced by a compiler rather than promised by a reviewer.

gafferongames:
[...]

Issue #189, opened 2026-08-31
C/C++ emitter defect: string helpers' include guard is TU-wide, definition namespace-local — two generated units in one TU break


gafferongames:
Named by the #184 redefinition worker (commit b0a9790) and worked around leg-locally in bench/cpp because the emitter was mechanically locked at the time: the emitted string helpers guard against re-emission with a TU-wide include guard while the helpers themselves are namespace-local, so the SECOND generated unit included into one translation unit loses its helpers. Exposed the moment the redefined bench_mixed put string fields into the bench TU, which includes both the Bench and RealWorld units.

Owner ruling, 2026-08-31, live: "We can unlock the C/C++ emitters to fix the emitter defect found by the worker." — a ONE-SHOT scoped carve, sequenced in the lock's own protocol:
1. PR A (single-file bench/LOCK edit): suspend the emitter and generated-c/cpp prefixes, citing this ruling, naming this issue as the sole licensed change.
2. PR B (the fix): correct the guard scoping in internal/codegen/{c,cpp}, regenerate (the diff must show exactly the guard/namespace correction across the c/cpp outputs and NOTHING else), remove the bench/cpp leg-local workaround, wire bytes byte-unchanged everywhere, and the reproduction control — the frozen shapes' c/cpp bench rows re-run in-sitting against the equalized reference, because a source-structure change to the reference legs must prove the reference numbers did not move.
3. PR C (single-file): restore the lock's prefixes. Separate from B by mechanical necessity: the guard reads bench/LOCK from the PR's own tree, so a PR that both restored the prefixes and touched emitter paths would refuse itself.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
Fixed and the protocol is closed: PR B (#220) landed the per-package guards (`SCHEMA_<PKG>_UTF8_VALID_DEFINED` / `SCHEMA_<PKG>_INTERIOR_NULL_DEFINED`), with a two-package one-TU regression leg (`test/guard/`) proven red on the old emitter and green on the fix, the generated diff guard-lines-only, wire bytes untouched, and the c/cpp reproduction control within ±2.9% of the sitting-2 reference. The C emitter was confirmed NOT defective (file-scope statics; TU-wide guard correct there). PR C (#222) restored the lock's prefixes — the one-shot carve is spent. Shipped in v2.2.1.

PR #195, opened 2026-08-31
schema bench: the harness emitter — nine legs' shape code generated from the corpus (#191)

gafferongames:
Draft — landing language by language; the go leg is on the emitter, the rest follow in this PR.

## What this is

`schema bench --lang <lang> --out <dir> --wire-dir testdata/wire bench/corpus/Bench.schema`
emits the SHAPE-DEPENDENT half of a bench leg. Issue #191: the bench_mixed
redefinition cost hours because nine harnesses' shape code was hand-rewritten
per language — precisely the work an emitter exists to do.

Emitted, per shape, per language:
- the pinned instance
- the §2.2 LCG vary mapping over value fields
- the §2.7 full-struct sink fold (legs with no free memory barrier)
- the decoded-field check (legs that gate variants field-by-field)

Hand-written, untouched: every leg's timed loops, escape barriers, buffer
discipline, CSV plumbing — the shape-INDEPENDENT half, which is exactly the
half the redefinition did not have to touch.

## The pinned values have one source: the goldens

`testdata/wire/<shape>.bin` is the §1.5 oracle every leg is already gated
against. The emitter DECODES it and emits the values it read, so the pin
cannot disagree with the golden by construction, and the hand-transcription
error class the reviews keep catching is gone rather than centralised.

## The vary mapping is derived, not transcribed

From each field's declared wire type: a masked draw of the largest
power-of-two sub-range that fits inside the field's bounds, offset by its
minimum, out of one LCG word. In range by construction. Structure — array
counts, string/bytes used lengths, the union tag, a branch gate — never
varies, so bytes/op is constant and the §1.5 residual holds. Array elements
vary at stride 2 through their value fields: §2.7's "representative subset"
made mechanical, and cost-proportional to the hand mapping it replaces.

The mapping's CONSTANTS therefore differ from the hand-written ones (a
derived rule cannot reproduce nine ad-hoc tables). The wire does not: same
fields, same widths, same bit counts, and every runner still asserts
bytes/op did not move under variation.

## Locks

- `internal/codegen/**` byte-unchanged: `git diff --stat main -- internal/codegen` is empty.
- The emitter lives in `internal/benchgen`, a new sibling package.
- The #189 carve licenses only the string-helper guard fix; nothing here touches it.

## Equivalence receipts

Per leg, filled in as each lands: golden gate green, corpus_id unchanged,
rates in the committed neighborhood.

| leg | golden gate | corpus_id | rates |
|-----|-------------|-----------|-------|
| go  | green | 409f416533f1a995 unchanged | pending |


gafferongames:
Closed unmerged: this is the harness-emitter design (#191's original framing), which the owner ruled against — "Expanding the compiler seems like a heavy lift. Baggage that everybody gets that maybe they don't want. Consider the user." The chosen design (option C, data-driven variant corpus + shape-blind runners) shipped instead in #197 and is now on main. Kept in history; nothing here is wanted.

Issue #196, opened 2026-08-31
The excision: remove all per-serialize-repo profiling code, and schema's own rt rows with it


gafferongames:
Owner ruling, 2026-08-31, verbatim: "Does this mean we clean up and remove any profiling that exists per-serialize repo from now on? I think we should. Any old profiling code will re-activate in future and drift so is suspect. we should excise it."

THE RULING: every serialize-family repo's bench/profiling code is deleted — serialize (bench.cpp and kin), serialize.c, serialize.go, serialize.rs, serialize.cs, serialize.js, serialize.dart, serialize.java, serialize.ex — bench sources, bench CI legs, bench README/USAGE sections. Deletion, not archival: a disabled bench is a re-activation surface, and git history is the archive. schema's own family-rt rows and their hand-written shape code go in the same sweep — same suspect class, different address. Named consequence, chosen deliberately: the hand-written-experience product retires; the generated subject via the data-driven schema bench (#191) becomes the estate's only benchmark. If the compiler-value pair (generated vs hand-written) is ever wanted again, it is remeasured deliberately from scratch, never revived from dormant code.

SEQUENCING: after #191's data-driven bench lands and produces the canonical table — the new instrument exists before the old ones are removed. One PR per repo, pure deletions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
Owner ruling, 2026-08-31, the policy this excision enacts: "We do all profiling inside schema now." — profiling has ONE home, this repo. No serialize-family repo, and no consumer repo, carries benchmark or timing code touching serialize/schema shapes; the sanctioned instrument is bench/corpus/Bench.schema generating per-language codecs, measured by hand-coded but SHAPE-BLIND runners (timing/CSV/loop only — hand-coded runners are fine, hand-coded measurement is not). Everything else is deleted, and a CI gate in this repo is being built to keep it true rather than to promise it.

gafferongames:
THE LOCK IS IN. Shape gate merged (aa6c62e). On the rebased head the ledger's ratchet refused with nine stale counts and seven entries matching nothing — because #199, #204 and #201 had deleted exactly the class it guards. Name hits across the nine runner legs fell **1946 to 29**, and not one of the 29 is hand-coded shape measurement: every remaining hit is a comment, a generated-symbol callsite, or an English word that is also a corpus field name. The register's 19 entries each name what they exempt and why.

[...]

Issue #194, opened 2026-08-31
The absolute ledger: per-language ns/msg time-series on the canonical shape — the anti-churn instrument


gafferongames:
Owner, 2026-08-31, the fear named verbatim: "I've seen this pattern over and over so many times, you get C++ faster, and then everything is slow, and then other languages catch up and now C++ is slow, and how do I know that we're making *absolute* progress in terms of C++ being faster vs. just spinning wheels up/down." / "Shuffling the profiling results like work is getting done but we're just going around in circles."

A relative table cannot answer this by construction — it re-normalizes on every movement. The instrument: a committed ABSOLUTE ledger — one row per sitting per language, ns/msg on the canonical bench_mixed (frozen shape, one machine class per ledger, schema-bench harnesses once #191 lands), appended at every sitting that runs the sweep, plotted as per-language curves over time. The C++ row is the headline: absolute progress = that curve going down or holding; ANY C++ regression beyond the recorded noise floor goes RED (a perf regression goes red, it does not drift — BENCH-STANDARD's own law, finally given its time axis). Every other language's curve separates catching up (their curve falling) from churn (cpp's curve rising) at a glance.

Backfill: today's sittings from the committed CSVs, labeled by shape era (the 21-byte era and the 438-byte era never share an axis — the shape is part of the question). Kept as a committed CSV + a tiny generator in bench/tools; the two-column relative table remains the per-sitting report, the ledger is the across-time truth.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
Owner confirmation, verbatim: "Yes, a time series over time showing both the total amount of time for a given benchmark (unchanging) reducing over time week to week of optimization." — the metric is ABSOLUTE TOTAL TIME on the unchanging canonical benchmark, one curve per language, sampled week to week as optimization lands. The benchmark being unchanging is the axis's precondition: the 438-byte shape freezes on #192's merge, and any future shape change starts a new ledger era rather than bending this one.

gafferongames:
Owner, completing the two-view doctrine: "The display being relative % is just a good way to understand relative performance of each language impl." — the two instruments, each with its one job: the relative two-column table answers WHERE EACH LANGUAGE STANDS (per sitting, C++ the denominator); the absolute ledger answers WHETHER WE ARE GETTING FASTER (across weeks, unchanging benchmark). Neither substitutes for the other; churn is what happens when only the first exists.

gafferongames:
[...]

Issue #184, opened 2026-08-31
bench_mixed redefined: every serialize method in one representative message, integers dominating the bytes


gafferongames:
Owner directives, 2026-08-31, live, verbatim across three beats: "Let's extend it to support all serialization_* methods" / "But in such a way that most of the bytes serialized are integers, and while we can have strings and wstrings and byte arrays, they should be relatively short and should not dominate. If strings and arrays dominate, then just put in 10X more of the other types per-message." / "No need for a new bench. Just change bench mixed as per-rec." — and, on continuity: "I'm not impressed with bench_mixed so it's nothing to hold on to tightly."

THE WORK: redefine BenchMixed in bench/corpus/Bench.schema as one representative message covering every construct the schema language expresses (enumerate the inventory from SPEC.md; anything serialize.h supports that schema cannot express is NAMED in the record as not-expressible, never silently skipped). THE WEIGHTING LAW governs the shape: integers (ranged ints, bit windows, full-width 32/64, 128-bit) carry most of the wire bytes; one short string, one short wstring, one small byte array, small arrays; float32/64, compressed float, fixed-point, bool, enum, nested struct, align all present. If the non-integer constructs would dominate the byte budget, add 10x more integer fields per message instead of shrinking them below realism.

CONSEQUENCES, handled deliberately: bench_mixed goldens re-pin from the C++ reference (wire FORMAT rules unmoved — the corpus content changes, the format does not); all nine legs' bench_mixed pin/vary/sink/loops rewritten under the equalized §2.7 discipline (string/wstring/bytes sinks defined per leg); the blended table RE-BASELINES on the new shape in one caffeinated sweep — the round's prior numbers stay in the record as the old-shape history, and laggard progress measures against the new baseline from then on. Fuzz/sabotage vectors extended to the new constructs.

LOCK INTERPLAY (ruled with the directive): the redefinition reaches bench/{c,cpp} and generated/bench/{c,cpp} through BYTE-UNCHANGED emitters — which is the real protection. bench/LOCK narrows to the emitters and the non-bench generated c/cpp outputs (a tripwire, since nothing else regenerates them); the extension PR's control is that the UNCHANGED shapes' c/cpp rows (packet/ints/bits) reproduce in-sitting.

Sequenced behind the rust-read archaeology verdict (bench box discipline); ahead of laggard 2 (Go), so later laggards measure against the type-complete blend.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #156, opened 2026-08-29
Java backend: separate monomorphic read/write functions, literal widths, inline-threshold discipline (measured directives from the serialize.java port)


rowan-claude:
The serialize.java performance phase (mas-bandwidth/serialize.java, commits 5c2c12c..b22d582) measured the JVM's specific failure modes against the C++ reference, and each is a directive for the schema Java backend. Final library numbers (MacBook Air, Apple Silicon, iteration numbers, default JVM flags): bitpacker 1987/1657 MB/s, stream 63.0/60.4 Mpps; small-packet reads sit 9–12x off the C++ reference and the causes are named below with the probes that measured them.

What the generated code must do differently:

1. **Emit separate read and write functions per message, each monomorphic against one stream type.** A probe whose serialize function only ever saw ReadStream ran 111 Mpps vs 58 in the family bench — the unified three-stream pattern costs ~2x on the JVM by itself. Generated code should never dispatch through a stream interface.

2. **Emit literal constant bit widths.** C2 cannot fold a `static final int[]` of widths the way clang folds a `static const` array; a literal-width probe ran +40% write / +85% read over the array-driven rows. Generation time knows every width — emit literals.

3. **Keep generated method bytecode small — C2's inlining thresholds are the ceiling.** Two hard-won facts: (a) dormant `assert` bodies count against inline thresholds even when disabled at runtime — the library's writeBits was 202 bytecode bytes and refused inlining until assert bodies moved to `assert checkX(...)` predicate calls (identical -ea semantics); generated code with asserts must use the predicate form. (b) A method that inlines both write and read paths exceeds InlineSmallCode (2500 default) and then nothing inlines it — "already compiled into a big method". Separate read/write generation (directive 1) also solves this; if a generated body grows large, split per field group.

4. **Zero allocation in steady state is achievable and verified** — JFR allocation sampling over a full bench run showed no library allocations (holder cells + reset() reuse). Generated code should hold the same line: no boxing, no per-message allocation on the hot path.

5. **Benchmark discipline**: per-shape timed loops, or type-profile pooling across message shapes turns call sites megamorphic and produces bimodal numbers (measured: 53.6 vs 101.6 Mpps run-to-run on one row until the harness was fixed). Applies to any generated-code benchmark harness too.

The port's bench (bench/, `make bench`, golden-gated) is the regression instrument for generated-code comparisons on the same machine.

rowan-claude:
[...]

PR #197, opened 2026-08-31
The data-driven bench across all nine languages: variant data + serialize-blind drivers, §2.9 ratified (#191)

gafferongames:
The data-driven bench across all nine languages. #191's chosen design (option C), fanned out from the C++ tracer: one committed variant file, nine serialize-blind drivers, **zero field names in any of them**. Emitters and `generated/` are byte-unchanged — `git diff main -- internal/ generated/` is empty.

## The table — the first sitting under the trusted contract

`bench/run.sh --quick`, all nine legs, one sitting, 2026-08-31, M2 Air. Generated C++ = 100% as the **denominator**, so a language beating C++ prints below 100% (#184, owner: *"I want nothing but a table with two columns: language, %"*).

```
subject: schema-GENERATED code (family gen) — what the compiler delivers; C++ = 100%
  language        %
  cpp          100%
  c            113%
  java         175%
  rust         189%
  dart         436%
  cs           448%
  js           582%
  go           712%
  elixir      5075%
  §2.3 NOISY: cs spread 20.5% > 15% — judge this row against the noise, not the digit
```

`corpus_id 6b213fbfa1a03a99` on **every row of every leg** — the same 64 committed wire records defined the work inside every timed loop, in all nine languages. That is the uniformity proof the design was built for, and it is now a mechanical fact rather than a review promise.

**This is `--quick`: the iteration instrument, never the certification instrument (§2.8).** The window was dirty (shared laptop, a node process at 100%, load avg ~3.0 at start; recorded in the CSV preamble). See "The window refuses the cross-sitting comparison" below.

<details><summary>The CSV's fuller columns (§2.9: everything but language and % is CSV-only)</summary>

| lang | write M msg/s | spread | round_trip M msg/s | spread | ns/msg (round_trip) | % of C++ |
|---|---:|---:|---:|---:|---:|---:|
| cpp | 7.340 | 6.6% | 3.585 | 3.3% | 279.0 | 100% |
| c | 6.422 | 3.0% | 3.185 | 9.0% | 314.0 | 113% |
| java | 4.724 | 2.3% | 2.050 | 1.3% | 487.9 | 175% |
| rust | 4.940 | 11.1% | 1.894 | 2.5% | 528.0 | 189% |
| dart | 1.548 | 0.1% | 0.822 | 1.8% | 1217.1 | 436% |
| cs | 1.689 | 4.0% | 0.800 | 20.5% | 1250.2 | 448% |
| js | 1.077 | 1.4% | 0.616 | 3.0% | 1622.8 | 582% |
| go | 1.149 | 1.1% | 0.503 | 0.4% | 1987.2 | 712% |
| elixir | 0.117 | 2.0% | 0.071 | 3.1% | 14156.1 | 5075% |

`bench/results/2026-08-31-arm64-macbook-datadriven-quick.csv`
</details>

## THE RANKINGS MOVED — reported, not smoothed

Against the committed `2026-08-31-mixedfull` quick sweep (the last sitting under the hand-written harnesses):

| | old blend rank | new round-trip rank |
|---|---|---|
| | rust, c, cpp, java, cs, dart, js, go, elixir | **cpp, c, java, rust**, dart, cs, js, go, elixir |

**rust fell 1st → 4th; cpp rose 3rd → 1st; java rose 4th → 3rd.** Two mechanisms, and this PR does not claim to have separated them:

1. **F1's harness-inflation removal was uniform in principle and wildly non-uniform in magnitude.** Old write row → new write row, per language: cpp **+119.5%**, elixir **+112.3%**, c **+76.5%**, go **+36.2%**, dart **+32.9%**, js **+27.9%**, java **+24.8%**, rust **−1.1%**, cs **−7.0%**. Every leg carried a `vary_*` function inside its timer; what each one COST differed by two orders of ratio. Rust's cost almost nothing (LLVM had it well in hand); C++'s cost more than half the row.
2. **The statistic changed.** The old blend was `(write + read)/2`; the new one is the round-trip, which pays the real decode. Rust's old read row was its weak half (3.66 M/s against a 5.00 M/s write) and the round-trip charges it in full; C++'s old read was its strong half.

### The window refuses the cross-sitting comparison

The `rt` rows are **identical code in both sittings** (untouched by this PR) and are the control leg. They moved:

| lang | rt write | rt read |
|---|---:|---:|
| cpp | **+40.8%** | +32.3% |
| c | +17.5% | +23.0% |
| rust | **−19.8%** | −13.4% |
| cs | −14.1% | −17.4% |
| go | −9.1% | −15.0% |

**±40% on untouched code, and split by direction.** So the per-language write-row deltas above CANNOT be attributed to the contract change from this data — the two windows are not comparable, and saying otherwise would be exactly the smoothing the profiling doctrine forbids. What the sweep DOES establish is the **within-sitting** table at the top: one window, one corpus_id, nine legs, one contract.

Settling the attribution needs a quiet-box interleaved sitting (§2.4/§2.6). Named, not done here.

## What landed

**1. The eight remaining legs converted** — `c`, `go`, `rust`, `cs`, `java`, `js`, `dart`, `elixir` — to the tracer's exact contract. Each leg's hand-written `BenchMixed` pin, vary, field-check and sink fold is **deleted**. Each driver loads the 64 variant records at a fixed stride derived as `filesize/64`, byte-compares variant 0 against the wire golden, gates every variant decode→re-encode→byte-compare, then times **WRITE** (64 pre-decoded instances round-robin, byte-fold sink) and **ROUND_TRIP** (decode + re-encode, byte-fold sink). READ derives to stderr and is never a CSV row.

**THE PROPERTY HELD IN ALL NINE.** No language forced a field name. What each one needed instead:

| leg | how the shape is named | notes |
|---|---|---|
| cpp, go, cs | one explicit type argument | `bench_datadriven<bench::BenchMixed>`, `benchDataDriven[bench.BenchMixed]`, `BenchDataDriven<Bench.BenchMixed>` |
| c | `#define BM_TYPE` | a second include template, `bench_datadriven.inc`, mirroring `bench_message.inc` with **no `BM_VARY`** — there is nothing left to vary by hand |
| rust | a turbofish | `bench_datadriven::<benchgen::BenchMixed, _, _, _, _>` |
| dart, js | a constructor passed as a value | `BenchMixed.new`, `bench.BenchMixed` |
| elixir | nothing at all | the two generated function references carry it |
| java | the type, in declarations | see below |

**Java is the one leg with a design note.** Its driver is deliberately **monomorphic** rather than generic: this leg's own JVM discipline (#156 item 5) is one timed loop method per shape and direction, because a shared generic loop pools its type profile across shapes and turns the timings bimodal. It still names zero fields. A second data-driven shape gets its own copy of those ~40 lines — still O(1) in shape CHANGES, which is what the design buys.

**The sink folds are gone, and with them §2.7's whole read-side problem.** java's `sinkOfBenchMixed`, dart's `sinkOfBenchMixed`, elixir's `sink_of_bench_mixed`, js's `sinkOfBenchMixedGen` — deleted. The round-trip's decode is observed by its own re-encode, in every language, with nothing to audit. §2.7's named per-language deviation (JS observing a BigInt as one bit, making the js read row a FLOOR) and the "upper bound" caveats on java/dart/elixir have nothing left to apply to. elixir's `strip_lossy` `BenchMixed` clause went too: a byte gate has no compressed-float round-trip problem to exclude.

The js leg **keeps its cross-tier oracle** (flat vs runtime: bytes, fields and verdicts on all 64 variants) — `deepEqual` walks the decoded objects by their own keys, so it names no field either.

`rt` rows and `BenchPacket`/`BenchInts`/`BenchBits` are untouched (the rt excision is #196).

**2. §2.9 RATIFIED** in `BENCH-STANDARD.md`, normative for the gen family's `bench_mixed` rows: write + round_trip, read derived and never a row, the published blend is the round-trip. It records the comparability statement (old blend ≈ `round_trip/2`, **in form only** — the factor of two cancels in a ratio, which is why the §2.8 percentages are the same statistic either way), **F1's harness-inflation finding**, and **F3's residual**: the gate is codec self-consistency, not data integrity — a flipped bit in a plain VALUE field of variant k>0 decodes to a different in-range value and re-encodes to the same bytes, and only `corpus_id` catches it. That is why the variant file is registered as corpus in all nine legs. §2.8 is amended to match; §5.1 documents `round_trip` as a `path` value.

**3. F4 FIXED.** `bench/tools` grows one `paths` list — `write`, `read`, `round_trip` — driving `aggregate`, `twingate`, `bands` and the absolute table, so a path added to a runner can no longer make rows silently vanish (`aggregate`'s straggler refusal now names path stragglers too). `run.sh --quick` prints the owner's ruled headline: **two columns, language and %, cpp as the denominator**, everything else CSV-only. **An empty headline section now exits non-zero** — the pre-fix behaviour was a leg that produced rows looking identical to a leg that produced none, at exit 0, which is precisely what #175 legislated against.

§2.3 is enforced in the two-column form **without adding a column**: a row over the 40% INVALID threshold prints no number, and every NOISY row over 15% is named in a note below the table. Publishing a noisy digit with nowhere to say so is what the ruled form would otherwise have done.

## For #194's ledger: this sitting opens the data-driven sub-era

F1's consequence, restated for the ledger: **every language's `bench_mixed` write curve steps at this boundary, and not because anything got faster.** The 438-byte era needs a harness **sub-era** boundary here, or the ledger reads a measurement change as absolute progress — the exact churn #194 exists to detect. The step is per-language and ranges from **−7% to +120%**, so a single scaling factor will not paper over it. Result CSVs from this sub-era carry `datadriven` in the filename; this PR's is `bench/results/2026-08-31-arm64-macbook-datadriven-quick.csv`.

## Verification (pasted, 2026-08-31)

**Every leg's gate green, one corpus_id:**
```
c      write/round_trip  6b213fbfa1a03a99
cpp    write/round_trip  6b213fbfa1a03a99
cs     write/round_trip  6b213fbfa1a03a99
dart   write/round_trip  6b213fbfa1a03a99
elixir write/round_trip  6b213fbfa1a03a99
go     write/round_trip  6b213fbfa1a03a99
java   write/round_trip  6b213fbfa1a03a99
js     write/round_trip  6b213fbfa1a03a99
rust   write/round_trip  6b213fbfa1a03a99
```
Distinct `corpus_id` values across all 28 data rows of the sweep: **one**.

**Negative control** — byte 0 of variant 0 flipped, run in cpp and in a managed leg, then restored:
```
=== (a) cpp ===
WIRE GOLDEN MISMATCH: bench_mixed (438 golden vs 438 actual bytes) — refusing to bench code that does not match the corpus
refusing to emit CSV rows from a failing run
BENCH FAILED (corpus_id b500dd0f2c13678a)
EXIT=1        rows in the CSV: 0

=== (b) java ===
GOLDEN GATE FAILED: bench_mixed variant 0 vs testdata/wire/bench_mixed.bin
reporting nothing.
EXIT=1        rows in the CSV: 0

--- restored: sha256 b26c23e9a5e2af82f3f18b987e4bc1c171f7f58f044047b9ab0f8ef3ffcb4078, tree clean, all nine legs green again
```

**The F4 refusal**, driven through `run.sh`'s own awk block verbatim against a CSV carrying rt rows and no gen `round_trip` row — the exact pre-fix state:
```
REFUSED (#175/§2.9): the gen headline section has ZERO rows — no leg reported a
bench_mixed family-gen round_trip row. Nothing was measured; printing an empty
table at exit 0 is the defect this refusal exists to stop.
EXIT=3   (run.sh turns this into exit 1)
```
A leg that refuses outright still aborts earlier, under `set -e`, before the headline runs — that path is unchanged and shown in the negative control above.

**The property, checked mechanically rather than by eye.** Every `BenchMixed` field name (58 of them, both naming conventions) matched against each of the nine drivers' text:
```
bench/cpp/bench_main.cpp         driver text   8278 chars  field names found: NONE
bench/c/bench_datadriven.inc     driver text   6994 chars  field names found: NONE
bench/go/main.go                 driver text   8058 chars  field names found: NONE
bench/rust/src/main.rs           driver text   6768 chars  field names found: NONE
bench/cs/src/Program.cs          driver text  10232 chars  field names found: NONE
bench/java/Main.java             driver text   4968 chars  field names found: NONE
bench/js/main.mjs                driver text   7266 chars  field names found: NONE
bench/dart/main.dart             driver text   7046 chars  field names found: NONE
bench/elixir/runner.exs          driver text   5958 chars  field names found: NONE
```
(The raw run flagged `drift` in all nine; every hit is the English word, in the comment *"so this driver cannot drift from another language's driver"*. The remaining real `.drift` accesses in `bench_main.cpp` and `main.mjs` are family `rt` — hand-written runtime usage is exactly what family `rt` measures, and it stays.)

**Emitters and `generated/` byte-unchanged:** `git diff main --stat -- internal/ generated/` is empty.

**`make test`** green (it caught one real thing: `dart analyze` failed on the now-dead `Int128` import the deleted pin/vary was the only user of — removed).

## Known follow-on, surfaced by this work

**F7 — the §4 inline verdict pass no longer attributes the `bench_mixed` timed loops.** `bench/tools/inline-verdict.sh` finds its timed line ranges by grepping the drivers for `write path: 1 warmup` / `read path: 1 warmup` markers. The data-driven loops carry `WRITE:` / `ROUND-TRIP:` comments instead, so their call sites fall outside the marked ranges and land as untimed/unattributed. It does not break the pass (the markers still exist for `bench_message`), and `--inline` is not part of the canonical `--quick` sweep, so `bench_mixed` rows simply stay `inline=unknown` — which correctly refuses their ratios rather than publishing a wrong one. Naming it rather than expanding this PR's scope; it wants its own pass alongside #196.

**F8 — the property wants a CI gate, not a reviewer.** The check pasted above is a script I ran by hand. Right now "zero field names in every driver" is enforced by review, which is the very thing #191's design exists to stop relying on: the owner's trust doctrine is *apples to apples enforced by a compiler rather than promised by a reviewer.* A ~15-line CI job matching the shape's field names against the nine driver bodies would make the property mechanical and permanent. Named rather than built — landing it here would expand a PR that must not merge on my judgment, and it belongs beside #196's rt excision.

Refs #191, #184, #194, #175.


gafferongames:
Adversarial review verdict: MERGE, after the one required fix (now pushed: bench/README §5.1's write/round_trip paragraph documented the pre-ratification tracer state — C++-only, PROPOSED, empty-section-at-exit-0 — three claims this branch had already changed).

The reviewer verified rather than accepted: the load-bearing property holds in all nine legs, each timed loop confirmed to call the generated symbol with nothing between timer and codec but stream setup and the byte-fold sink, no cached buffer or elided re-encode anywhere; the zero-field-names property re-checked with its own extraction (53 field names, 133 case forms, word-boundary matched — every hit adjudicated to English prose or the run-statistics helper, zero field accesses, and no hardcoded 438/3504/field-count/offset table in any driver); the gate confirmed to run before timing in every leg; the negative control reproduced independently (variant-0 byte flip -> refusal, zero rows, exit 1, restored to the stated sha); F3's residual reproduced exactly as documented (a value-bit flip inside the stats block passes the gate and is caught only by corpus_id); §2.9's text checked line-for-line against the code with its comparability claim correctly hedged; and F4's fix confirmed by executing the awk (two columns, cpp denominator, exit 3 -> run.sh exit 1 on an empty headline, table reproduced byte for byte).

Two one-line residues recorded rather than fixed here: elixir's strip_lossy is now an identity clause under a comment describing an exclusion that no longer excludes (scaffolding to remove), and run.sh's render() awk has an unused local plus a leaked global. Both ride the #196 sweep.

PR #198, opened 2026-08-31
go: the flat word codec — 712% to 248% of generated C/C++

gafferongames:
Laggard 1 of the re-ranked queue (#170): generated **Go**, the biggest prize on the trusted table.

**Go moves from 712% to 248% of generated C/C++.** Measured as a paired same-sitting A/B — main's Go emitter output against this branch's, nothing else changed between the halves — that is **2.35x on run.sh's own statistic**, with **1.98x write** and **2.36x round_trip** on the leg's own rates and a **0.9940x** negative control over 16 unchanged rows. Both CSVs are committed. Wire bytes never move; `checks=always` is preserved; `bench/LOCK` carries no active prefixes today (the #189 carve window) and nothing under its suspended prefixes moved except a protocol id.

## What changed since the block

The adversarial review found three CONFIRMED correctness regressions with one root cause. All three are fixed, all three now have corpus coverage in nine languages, and every number in this body is re-derived against the trusted baseline the review named.

### The root cause, and the fix

PR #183's Rust template maps one item to exactly one piece, so its fallback — re-emit each of a run's pieces — IS re-emitting each of its items, safe by construction. This emitter's levers 3–6 made `flatPiecesOf` return 1..N pieces per item, with pieces carrying items that belong to a NESTED struct at a different base expression, and the fallback was inherited verbatim. The count-based "flatten only where it reduces calls" policy is correctly implemented and is exactly what routes execution into that fallback.

| # | defect | what it did |
|---|---|---|
| 1 | **silent wire corruption** | `type Pair { values [2]float64 }` emitted the element loop TWICE on write and read — 32 bytes where Rust and main emit 16. Go-to-Go round trips passed, so nothing went red. |
| 2 | **generated Go does not compile** | a nested struct split by the run cap emitted `value.C` on the OUTER type. Under a name collision it would have serialized the WRONG field silently. |
| 3 | **generated Go does not compile** | a bare `type Vec2 { x float64; y float64 }` carried an unused `math` import — `g.needsMath` was set during SPECULATIVE classification and never cleared when the run fell back. |

A run now accumulates whole ITEMS as `flatGroup`s and splits only on item boundaries, so a fallback re-emits each item against the base that item owns. And **`flatPiece` no longer carries an `ir.Item` at all**: a piece is a bit-placement recipe and nothing else, which makes re-emitting one *structurally impossible* rather than merely avoided. That is the choice between the two repairs the review offered — restoring the 1:1 invariant would have given up levers 4–6, so the invariant is restored one level up, where the fallback actually operates, and the 1:N classification inside a run is kept. `needsMath` moves into the float pieces' `emit`/`read` closures, which run at emission.

**Generated output for the existing corpus is byte-identical before and after this fix** — none of the three shapes exists in it. That is the finding, not the reassurance, and it is issue #203. It also means **the measured win is untouched by the correctness repair**: `generated/bench/go/{Bench.go,realworld/RealWorld.go}` are byte-for-byte what the blocked branch emitted, so the numbers below measure the same codec the review reproduced at 2.35x blended.

### The corpus blind spot (#203) — mandatory, and it caught a second defect

`examples/Degenerate.schema` carries twelve shapes the corpus's realism masked: a fixed scalar array as essentially the whole message (`[2]float64`, `[2]uint64`, `[2]int64`, `[1]T`, `[N]T` landing exactly on a chunk boundary, an array plus a trailing field, two arrays back to back); a bare two-float unit with no other math consumer in the file; and a nested struct as the ONLY field, as the FIRST field, and straddling the run cap.

They are pinned as one C++-authored wire golden, `testdata/wire/degenerate.bin` (209 bytes), and **all nine legs are held to it on every push**: the stream legs (C++, C, Go, Rust, C#, JS) write the twelve into one stream; the whole-message legs (Dart, Java, Elixir) concatenate their twelve buffers. That equality is why every type in the file is a whole number of bytes wide — the file's header states it, because it is load bearing. `internal/codegen/golang` states the same three properties directly against `maxRunBits`, so re-tuning that constant cannot quietly retire the coverage.

**RED FIRST.** On the pre-fix emitter (997c553), against the new corpus file:

```
$ go build -o /tmp/schema-prefix ./cmd/schema      # at 997c553
$ /tmp/schema-prefix generate --lang go --out /tmp/dg-red examples
$ cd /tmp/dg-red && go build ./...
# example
./Degenerate.go:10:2: "math" imported and not used
./Degenerate.go:447:30: value.C undefined (type *TrioStraddle has no field or method C)
./Degenerate.go:485:30: value.C undefined (type *TrioStraddle has no field or method C)
```

Defect 1 does not fail a compile — that is the whole problem — so it is measured directly. A `SpanF64`-only unit, written through the generated Go writer:

```
before (997c553, unused import removed by hand so it links):
    SpanF64 bits=256 bytes=32 (SpanF64MaxBits=128)
after:
    SpanF64 bits=128 bytes=16 (SpanF64MaxBits=128)
```

The emitter test goes red on all four assertions at 997c553 and green after:

```
$ go test ./internal/codegen/golang/ -run TestFlatFallback     # emitter at 997c553
--- FAIL: TestFlatFallbackReEmitsItemsNotPieces
    [2]float64 emitted 4 element loops, want 2 (one write, one read) — a fallback over
      pieces emits the loop once per element and doubles the wire
    a nested struct split by the 384-bit run cap named "value.C" on the OUTER type in
      func WriteOuter( — ... the fallback must re-emit the ITEM against its own base
    a nested struct split by the 384-bit run cap named "value.C" on the OUTER type in
      func ReadOuter( — ...
    a bare two-float unit imported math 1 times — nothing in it reaches math, so the
      import does not compile
FAIL

$ go test ./internal/codegen/golang/ -run TestFlatFallback     # this branch
ok      github.com/mas-bandwidth/schema/v2/internal/codegen/golang      0.173s
```

**The new file caught a SECOND emitter defect on sight, the same class as the Go one.** The Dart backend declared all four conversion-scratch views whenever any was needed, so a file converting only float64 carried an unreferenced `_u32` and `dart analyze` refused it (exit 2). No corpus file had ever been float64-only. The views are now emitted per conversion. That is the corpus paying for itself before it was even committed.

Existing wire goldens are **unmoved** — `git status testdata/wire` shows only the new `degenerate.bin`. The corpus protocol id moves, as adding a file must (`0x0bde7acdd36abc6a` → `0xa89dd9e036603208`), and is re-pinned; every other generated-source diff in this PR is that id and one JS tier banner relocating to the unit's new alphabetically-first flat module.

**Bench `corpus_id` is unchanged at `6b213fbfa1a03a99`** in every row of both new CSVs: the runner hashes the goldens it loads, and it does not load the conformance corpus.

## The measurement, receipted

`bench/run.sh --quick`, one sitting, back to back, M2 MacBook Air, generated subject, run.sh's own printed statistic (family gen, `bench_mixed`, round_trip over max rates). Nothing changed between the halves but `generated/bench/go`. Taken on the tree as it stands, after #204 rewrote the runners.

- `bench/results/2026-09-01-gocodec-before-arm64-macbook.csv`
- `bench/results/2026-09-01-gocodec-after-arm64-macbook.csv`

| | before | after | |
|---|---:|---:|---|
| write | 1.35 M msg/s | **2.67** | **1.98x** |
| round_trip | 0.60 M msg/s | **1.41** | **2.36x** |
| headline (max) | 583% | **248%** | **2.35x** |
| headline (median) | 582% | **243%** | 2.39x |

**The negative control, in this sitting, from the committed files.** Every other leg is unchanged code compiled identically in both halves:

| lang | write | round_trip |
|---|---:|---:|
| cpp | 1.0087x | 1.0024x |
| c | 1.0103x | 1.0017x |
| rust | 0.9700x | 0.9764x |
| cs | 0.9795x | 1.0022x |
| js | 0.9979x | 1.0339x |
| java | 1.0205x | 0.9868x |
| dart | 0.9888x | 0.9426x |
| elixir | 0.9963x | 0.9893x |

**Geomean 0.9940x over 16 control rows, min 0.9426, max 1.0339** — no row over §2.3's 40% INVALID threshold, so all sixteen count. The generated Go rows moved because the generated Go changed; nothing else did.

The box was shared throughout (other Claude sessions live on a fanless M2 Air); the noise line in both preambles says so, and `--quick` is the iteration instrument, never certification (§2.8). The one row to judge against the noise rather than the digit is go's after-half **write**, at 7.36% spread — its max rate gives 1.98x and its median 1.88x. The round_trip rows, which are the published statistic, carry 0.36% and 1.89% and agree on both statistics.

**Replication, named but not offered as a receipt:** the same paired A/B run before this branch was rebased onto #204 gave 602% → 251% (2.40x) with a 1.0050x control over 14 rows. Its CSVs are not committed, because #204 replaced the runner they were taken with, and a receipt should name an instrument that still exists. The committed pair above is the one taken on this tree.

### The two-column table, after

| language | % of cpp@100 |
|---|---:|
| c | 98% |
| cpp | 100% |
| java | 149% |
| rust | 159% |
| **go** | **248%** |
| cs | 365% |
| dart | 400% |
| js | 475% |
| elixir | 3800% |

Go was the worst laggard but elixir. It is now **fifth of nine**, below cs, dart and js, and within 1.56x of rust.

## The baseline, corrected

**The PR said 592%. That number appears nowhere in the repo.** The trusted table — #170's re-ranking, from main's committed `bench/results/2026-08-31-arm64-macbook-datadriven-quick.csv` — puts generated Go at **712%** (max statistic, which is the one run.sh prints) and **711%** (median). Against 712%, this branch's 248% is a **2.87x** improvement: the direction is favourable and the true win is larger than the PR claimed.

The two baselines do not agree, and the disagreement belongs in the open:

| | trusted table (#170) | this sitting's before half |
|---|---:|---:|
| go round_trip, max | 0.503 M msg/s | 0.601 M msg/s |
| cpp round_trip, max | 3.585 M msg/s | 3.501 M msg/s |
| headline | 712% | 583% |

The **denominators agree to 2.4%**; the go leg alone ran **19% slower** in the trusted sitting, whose own noise line reads *"NOISY: shared laptop, one node process at 100% and Safari live"*. So the causal claim is the paired one — **2.35x, one sitting, controls at 0.9940x** — and 712% → 248% is the same movement read off the published statistic across sittings, which is what the re-ranked queue tracks. The PR's earlier "580% and 592% agree to 2%" reconciliation is withdrawn; it reconciled a number that does not exist against a sitting whose CSV was not committed.

## The spread the earlier CSV carried

Recording what the review recorded. In `bench/results/2026-08-31-gocodec-quick-arm64-macbook.csv` (committed on this branch earlier), the go round_trip row has **10.44% spread** against 0.40–1.73% for every other row in that file. That sits just under run.sh's 15% NOISY threshold, so no note printed. **On the median statistic that CSV's headline is 275%, not 255%** — a 20-point gap between the two statistics on one row.

The re-run resolves it rather than arguing about it: in the new sitting the go round_trip rows carry **0.36%** (before) and **1.89%** (after) spread, and the two statistics agree to within 2% — **248%** on max, **243%** on median. The earlier CSV stays in the tree as the record of the sitting it came from; the paired files are the evidence.

## Claims withdrawn as UNRECEIPTED

- **"geomean 1.40x over 30 cells, best rows `bench_ints` read 2.57x and `shipcreate` read 2.05x, no cell below 0.99x"** — the per-shape leg. No CSV was committed, and #199 retired those shapes, so it cannot be re-run on main. **Withdrawn.** The paired bench_mixed files above are what stands.
- **"the control rows moved 0.9977x over 10 rows (min 0.987, max 1.007)"** — same leg, same reason. **Withdrawn and replaced** by the 1.0050x-over-14-rows control measured in this sitting, from committed CSVs.
- **The per-lever table below is retained as the development record, not as receipted measurement**: each lever's number was taken on the retired per-shape leg during development. The composed result is what the two committed CSVs measure.
- **"`bench/go` is untouched"** — **false, and dropped.** See below.

## `bench/go` IS touched

A `--cpuprofile` flag and a `runtime/pprof` import were added to `bench/go/main.go`. **The flag stays**, and this PR now states it in the runner's usage line and makes it refuse instead of panicking. It is how this round's conviction was obtained — `go tool pprof` on the baseline put **~86% of generated-codec time inside the runtime's per-field bit calls** against ~2% in the generated code itself — and the next round will want it. It is an iteration instrument: it changes nothing about what is measured, and no timed row in this PR was taken under it.

## Where the time went, measured first

`-gcflags=-m -m` gave the mechanism (go1.27, inline budget 80):

```
write_stream.go:99:  cannot inline (*WriteStream).writeBits:      cost  92 exceeds budget 80
read_stream.go:55:   cannot inline (*ReadStream).readBits:        cost 106 exceeds budget 80
read_stream.go:100:  cannot inline (*ReadStream).SerializeInt:    cost 168 exceeds budget 80
```

`SerializeBits` *is* inlinable, but the bodies underneath it are not — so **every field cost one real call into the runtime, on both paths**. Issue fewer calls, and the rest follows.

## The levers (development record — see the withdrawal note above)

| # | lever | write | round_trip |
|---|---|---:|---:|
| 1 | Ranged-integer **read** fold — kills a runtime `BitsRequired` per field and drops off the non-inlinable ranged entry points | +1.5% | +2.6% |
| 2 | **The flat word codec** — runs gathered, values OR'd into chunk locals at literal shifts, one chunk per call, checks fused per run. `WriteMixedEntity`: 14 calls → 3 | 1.50x | 1.63x |
| 3 | Nested struct bodies **placed inline** instead of called per element — 88 calls and 88 sticky-error loads per message | +9% | +10% |
| 4+5 | Nested struct **fields** join the parent run; small fixed arrays unroll into it | +2% | +4.5% |
| 6 | **Consecutive array elements share chunks** — eighty 18-bit stats carried 18 of a possible 64 bits each; 80 calls → 25 | +14% | +21% |
| 7 | **Run cap measured, not inherited** — 256→2.66/1.40, **384→2.78/1.49**, 512→2.75/1.46, 1024→2.62/1.43 | +4% | +6% |

Two results worth naming because they went against prediction:

- **Chunk width was a tie, then wasn't.** I predicted 64-bit chunks would beat Rust's 32 by halving calls. First measurement: a **wash** (`SerializeBits64` is itself non-inlinable at cost 278, so one bigger call replaced two small ones). Re-measured after the later levers, 64 wins clearly on reads — the balance shifted once call count fell. Kept at 64, on the second measurement, not the first.
- **A regression the per-shape leg caught.** `rigidbody_*` write went 0.91x: `RigidBody` is all float64, so every piece is exactly one whole chunk — flattening removed no call while adding a local per field, where the per-field form passed the struct field's own address. The policy became *flatten only where it reduces the call count*. That policy is correct and is kept — and it is also precisely what routes execution into the fallback the review found broken, which is why the fallback had to be made safe rather than the policy relaxed.

## Two deliberate deviations from the Rust template (#183)

1. **Chunks are 64 bits, not 32** — measured above.
2. **Every piece is masked to its width.** serialize.go's write path *masks* a too-wide value; the Rust runtime only `debug_assert`s, so #183 had to *refuse*. Masking is what keeps the Go form observably identical to what it replaces.

And one coverage win Rust could not have: **compressed floats do not break a run here.** This emitter already folds quantization into generated arithmetic ending in a plain bit write, so quantized fields are ordinary pieces — where Rust's live in the runtime, which is why #183's compressed-float shapes barely moved.

## The floor, with its mechanism measured

The remaining overhead is **Go's fixed inline budget at the runtime boundary**. Building the leg with **Go PGO** (a labeled diagnostic, **not shipped**) reached **3.28 M/s write and 1.72 round_trip**, a further 1.18x/1.15x; `-gcflags=-m` confirms why: under PGO, `writeBits`, `WriteStream.SerializeBits64` and `ReadStream.SerializeBits64` all become inlinable; without it, none are. (Development-record number, same status as the lever table.)

Two independent lifts, both outside this PR's remit:

- **B1 — bring `writeBits` (cost 92) and `readBits` (106) under the 80-unit budget in serialize.go.** They miss by 12 and 26. A routing question, not a workaround: #170 routes this round to schema-level changes only. The runtime's own comments already *claim* these bodies stay within the budget; at go1.27 that claim is false.
- **B2 — Go PGO for the leg.** A build change; the C/C++ reference is locked and builds without PGO, so this is a cross-leg equalization decision for the owner.

**B3 — `unsafe` pokes at the stream's private scratch from generated code** was enumerated and **rejected on sight**: it reimplements the runtime inside generated output against unexported fields.

## Wire and semantics

- Existing wire goldens **unmoved** (`testdata/wire` clean apart from the new `degenerate.bin`); bench `corpus_id 6b213fbfa1a03a99` unchanged in both new CSVs.
- Regenerating every target with this branch's own `bin/schema` build over a cleaned `generated/` is **zero-diff** against what is committed.
- Rebased onto current main (#200's shape gate, #204's runner retirement, #205's LOCK amendment); `go run ./bench/tools/shapegate` reports **clean, 19 files on the ledger, no new debt**.
- **`make test` is green in full on this box**, including the java, dart and elixir conformance legs — the pinned `dist/` toolchains were available here, so this PR is not relying on CI for them: `schema_test`, `_random`, `_ludicrous`, the C and C-ludicrous legs, `schema_test_bench`, `test/{go,rust,cs,js,dart,java,elixir}` and their ludicrous twins, `dart analyze` + `dart format --set-exit-if-changed`, `mix format --check-formatted`, the goldens, the fuzz leg, and `go test ./...`. `make check` passes. `cpp-lock` is inactive (bench/LOCK carries no active prefixes during the #189 carve).
- **Zero allocations** and no heap escapes on the Go path: `-gcflags=-m` shows no chunk local escaping, and the after-half run's own alloc note reads `bench_mixed one pass (256 ops/path): write 0 allocs, round_trip 0 allocs` / `write 0 bytes, round_trip 0 bytes`.
- `checks=always` preserved: the same values are refused, with the same error.

Two behaviours move, both named rather than smoothed, and both the same class the Java/Dart/JS-flat/Rust fusion already carries:

1. **Refusal timing on write.** A run's (and a group's) refusals all run before it packs, so a refused write has emitted fewer bits than before. Which values are refused, and with which error, is identical.
2. **Error kind on a doubly-bad read.** A stream both truncated inside a run *and* carrying an out-of-range value before the truncation now surfaces the stream's overflow error where the per-field form surfaced the range refusal. Both refuse the packet; **the set of accepted streams is unchanged**, which is the property that matters at the trust boundary.

One further delta, from folding the ranged read: the refusal is returned rather than latched on the stream. The write folds already shipped with exactly this property, and generated callers return immediately on error.

## Named follow-ons, not held for

- **Fixed-point (`SerializeFixed64`) still breaks every run it sits in** — foldable in principle, the same shape as the compressed-float fold.
- **String/bytes length could join the preceding run**; the bytes themselves must stay on the bulk path.
- **The 128-bit family** stays on the runtime path.
- **A per-shape leg to replace the one #199 retired**, if the owner wants per-shape evidence in future optimization PRs — right now `bench_mixed` is the one shape, by ruling.
- **The same fallback audit for the other flat emitters** (Rust, JS-flat, Java, Dart): #183's 1:1 invariant is intact in Rust, but any backend that later grows a 1:N classification inherits this bug. `Degenerate.schema` now guards all of them on the wire; the structural guarantee — a piece that cannot name an item — is Go's alone so far.


gafferongames:
Adversarial review verdict: **BLOCK** — three CONFIRMED correctness regressions, one root cause. The performance claim is real and reproduced independently (2.06x write / 2.50x round_trip / **2.35x blended**, i.e. better than the PR's 2.24x), wire identity holds on the current corpus (36 generate invocations re-run: zero diff; test/go byte-compares against the C++-pinned goldens and passes), and the lock and scope are clean. But:

1. **Silent wire corruption.** `type Pair { values [2]float64 }` emits the element loop TWICE on both write and read — 32 bytes where Rust and main emit 16. Go-to-Go round-trips pass, so nothing goes red. Triggers on any fixed scalar array (`[2]float64`, `[2]uint64`, `[2]int64`, and multiples).
2. **Generated Go does not compile** when a nested struct is split by the run cap: emits `value.A` on the outer type. Where a name collision exists this would silently serialize the WRONG field instead of failing.
3. **Generated Go does not compile** for a bare two-float unit: `needsMath` is set during speculative classification and never cleared when the run falls back — unused import. Any Vec2/Vec3/Quat-shaped file with no other math user.

ROOT CAUSE: #183's Rust template keeps a strict 1:1 item-to-piece mapping, so its fallback (re-emit each piece directly) is safe by construction. Levers 3-6 made `flatPiecesOf` return 1..N pieces per item, with pieces carrying items from nested structs at a different base expression — and the fallback was inherited verbatim. The count-based flattening policy is implemented correctly and is exactly what routes execution into the now-unsafe fallback. Ports never invent contracts; a port must re-derive the invariant it inherits.

TO UNBLOCK: fix the fallback to re-emit whole ITEMS against the correct base (or restore the invariant); move `needsMath` to emission; ADD CORPUS COVERAGE for all three blind spots (one-line schemas — see #203); correct the baseline (the PR says 592%, the trusted table says **712%** — the number 592 appears nowhere in the repo, so the 'movement transfers' reconciliation needs re-deriving; direction is favourable, the true win is larger); drop the '`bench/go` untouched' claim (a `--cpuprofile` flag was added there); and either commit the negative control and per-shape CSVs or restate those claims as unreceipted — the 0.9977x control over 10 rows is the load-bearing number and it can no longer be re-run on main since #199 retired those shapes.

Also recorded: go's round_trip spread is 10.44% in the committed CSV against 0.40-1.73% elsewhere, sitting just under the 15% NOISY threshold so no note printed; on the median statistic the headline is 275%, not 255%.

gafferongames:
Unblock response — every item in the verdict, with where it landed. The body is rewritten around it.

**1. Silent wire corruption / 2. nested field on the outer type / 3. unused `math`** — fixed in `fb8d6f1`, one repair. A run now accumulates whole ITEMS as `flatGroup`s and splits only on item boundaries, so the fallback re-emits items against the base each item owns. Beyond that, **`flatPiece` no longer carries an `ir.Item` at all** — a piece is a bit-placement recipe and nothing else, so re-emitting one is structurally impossible rather than merely avoided. That is the choice between your two options: restoring #183's strict 1:1 would have cost levers 4–6, so the invariant is restored one level up, where the fallback actually operates. `needsMath` moves into the float pieces' `emit`/`read` closures, which run at emission.

**Corpus coverage (#203)** — `17faefa`. `examples/Degenerate.schema`, twelve types, one C++-pinned golden, all nine legs on every push, plus the three properties stated against `maxRunBits` in `internal/codegen/golang` so re-tuning the cap cannot retire them. Verified red first at `997c553`: four failing assertions and generated Go that does not compile. It caught a **second** emitter defect on sight — Dart declared all four conversion-scratch views whenever any was needed, so a float64-only file carried an unreferenced `_u32` and `dart analyze` refused it. No corpus file had ever been float64-only.

**Baseline** — corrected to **712%** throughout, re-derived from main's committed `2026-08-31-arm64-macbook-datadriven-quick.csv` (cpp round_trip max 3.585 M/s over go's 0.503). Against 712% this branch's 248% is **2.87x**. The old "580% and 592% agree to 2%" reconciliation is withdrawn. The two baselines disagree by 19% on the go leg alone — denominators agree to 2.4% — and the body says so rather than smoothing it.

**`bench/go` untouched** — false, dropped. The `--cpuprofile` flag stays (it is how the ~86%-in-the-runtime conviction was obtained); `a3faf94` states it in the usage line and makes it refuse instead of panicking.

**The unreceipted claims** — the per-shape geomean and the 0.9977x control are **withdrawn**, not restated: #199 retired those shapes and no CSV was committed. Replaced by a paired same-sitting A/B on the shape that survives, both halves committed (`0e930d5`): **583% → 248%, 2.35x**, controls **geomean 0.9940x over 16 rows**. Taken after rebasing onto #204, so the CSVs name an instrument that still exists.

**The 10.44% spread / 275% median** — recorded in the body verbatim, and resolved by the re-run: go's round_trip rows now carry 0.36% and 1.89%, and the two statistics agree at 248% and 243%.

Two things worth flagging for the reviewer:

- `generated/bench/go/{Bench.go,realworld/RealWorld.go}` are **byte-for-byte what the blocked branch emitted**. The correctness repair changes no generated output on any existing corpus, so the numbers measure the same codec you reproduced at 2.35x blended.
- `make test` is green **in full on this box**, java/dart/elixir conformance included — the pinned `dist/` toolchains were available here, so this is not leaning on CI for those three legs. `go run ./bench/tools/shapegate` is clean at 19 ledger entries, no new debt.

Not merging — yours to call.

gafferongames:
Re-review verdict: **MERGE**. Every wire question answered by the reviewer's own commands in its own clone, not by reading claims — which is what a PR that already shipped one silent wire defect earns.

THE REPAIR, verified structurally AND empirically. Structurally: `flatPiece` genuinely has no `ir.Item` field, no expression in the package can turn a piece back into an item, the two branches of `flush()` are mutually exclusive, and every other run call site has NO fallback at all (`!ok` routes to the per-field emitter before a run is built). So there is no second path. Empirically: a differential oracle against origin/main's per-field emitter (no flat codec at all), same schema through both compilers, identical reflection driver, 24 seeds per type — **4176 type/seed combinations across four adversarial corpora, ZERO byte differences, zero round-trip mismatches**. Adv1 swept nested structs across the 384-bit cap including an outer type declaring `a/b/c` to collide with its nested field names (the case that would have silently serialized the WRONG field under the old bug — byte-identical). Adv2 covered fixed scalar arrays at every width and the 128-bit unroll cap, arrays of nested structs at bounds 1/2/3/7/8/9/16/17 hitting both the group loop and the remainder loop, arrays straddling the cap. Adv4 covered mixed-kind fallbacks (runs whose groups differ in kind, falling back as a whole).

RED-FIRST REPRODUCED VERBATIM at 997c553's parent — all four assertions fail, and compiling the pre-fix generated Go reproduced the PR's error text INCLUDING ITS LINE NUMBERS. The silent-corruption assertion is dispositive at source level: pre-fix `WriteSpanF64` contains the element loop twice (256 bits against a declared MaxBits of 128); at HEAD it appears once.

WIRE: reviewer built bin/schema itself, `rm -rf generated/`, re-ran all 36 generate invocations — zero diff. `git diff 997c553 HEAD -- generated/bench/go` empty, so the repair costs none of the win. Only `degenerate.bin` added under testdata/wire, nothing else moved; it is genuinely C++-pinned (written only by the C++ leg under SCHEMA_UPDATE_WIRE_GOLDENS), and all nine legs really byte-compare it — each assertion read individually, not taken on the claim, with the bit total 1672 = 209 bytes matching the golden exactly. `make test` exit 0 in full on that box, all nine legs.

THE DART FIX traced through all four reachable scratch states, correct and minimal. And a genuinely useful negative: C#/Java/Elixir use function-local scratch so they cannot carry this class, while JS-flat DOES emit `const SC` unconditionally — the same shape, but with no analyze/lint gate to fail on it. Named as the follow-on this PR already lists rather than a blocker.

NUMBERS recomputed from both committed CSVs: 583%→248% (2.35x), round_trip 2.3603x max / 2.3640x median (agrees both ways), controls geomean 0.9940x with min 0.9426 max 1.0339 over n=16 — exact. Baseline 712.4% max / 711.2% median; 712→248 = 2.871x.

Three findings recorded, none blocking: (a) the headline sentence says '1.98x write' unqualified where that is the max statistic and median is 1.88x — disclosed at line 104 and the table row is labelled, so it wants one word, not a block; (b) both new CSVs stamp a schema commit that does not resolve on origin (it is the pre-rebase SHA of what is now 1f47f56, so the measurement was taken on this branch's content, but a re-runner cannot reconstruct the tree from the PR alone); (c) a 'denominators agree to 2.4%' line that did not reproduce on any statistic tried — immaterial, the conclusion routes through the paired sitting.

Also noted mechanically: `generated/js/RenderFlat.js` lost its 21-line tier explainer to `DegenerateFlat.js` because the JS flat emitter attaches that doc to whichever flat file sorts first, and Degenerate now sorts before Render. Verified mechanical; a note for the JS emitter's owner, since the doc will migrate again on the next alphabetically-early schema.

Issue #174, opened 2026-08-30
Elixir backend performance: aggressive round — generation-time bit-syntax swizzling is the lead


rowan-claude:
Owner directive, standing: the Elixir position (roughly 50x blended on the one-contract sweep of 2026-08-31, unchanged by the conformance pass) is not acceptable for the backend; investigate aggressively. This issue promotes the plan from #170's comments to its own board home so any bench can pick it up cold.

The leading hypothesis is generation-time bit-syntax swizzling. The emitter currently packs by manual integer arithmetic because the wire's LSB-first order inside little-endian 64-bit words fights bit syntax's MSB-first grain at runtime. But the generator knows every field width statically, so it can compute each word's segment layout in advance, pre-swizzled so MSB-first segments produce LSB-first wire order, and emit one single binary construction expression per word, with one literal-segment match per word on read. That hands the whole pack to the ERTS bit-syntax engine in C: one allocation per word instead of an append BIF per field (the measured write tax), and it replaces window arithmetic on reads.

Secondary avenues if the primary disappoints: whole-row single-expression construction where alignment permits; iodata shapes revisited under literal widths; and an honest floor statement. The target is closing the gap to the generated-JS flat tier, not to native. If the BEAM's honest floor after this work is five to fifteen times the fastest language, publish that number as the floor with its mechanism named, and the README says what the port is for: the tier that holds a million connections while validating every hostile byte, where per-message cost was never the binding constraint. Every variant golden-gated; the conformant harness from PR #172 is the measuring instrument.

rowan-claude:
A second tier to hold in reserve if the pure-BEAM floor comes in short of acceptable: a NIF backend. The generator could emit small C pack and unpack kernels per unit, compiled as a NIF, with the pure-Elixir implementation kept as the portable fallback exactly the way the JS backend carries flat beside runtime. That would put Elixir near native speed at the cost of a real deployment-story change: per-platform native compilation, and a NIF fault takes down the VM, so the kernels would need the same never-throw discipline enforced in C. This is a design decision for the owner, not an optimization to take silently; it goes on the table only after the swizzle round reports its honest floor.

gafferongames:
The aggressive round landed as PR #202, ready for review: **3624% → 1451% of generated C/C++** on the §2.9 blend (write 2.74x, round_trip 2.50x), measured before and after in one sitting with the eight other languages as the in-sitting control — none of them moves more than 2%.

[...]

PR #202, opened 2026-08-31
elixir #174: the aggressive round — 3743% to 1457% of generated C/C++

gafferongames:
Issue #174. The Elixir position — thousands of percent of generated C/C++ on
the §2.9 blend — worked with the method the owner set: profile first,
enumerate every lever, rank by win/effort, take them one commit and one
measurement at a time, and record the ones that fail as results.

**3743% → 1457% of generated C/C++.** Write 2.79x, round_trip 2.51x, both
halves measured in one sitting after the rebase and **both committed**:
`bench/results/2026-08-31-elixirswizzle-{before,after}-quick-arm64-macbook.csv`.

| | lever | measured |
|---|---|---|
| A | one binary append per group, not per field | write 1.41x |
| C | one window decode per group, not per field | read 1.49x |
| B | one map read per scope, not per field | write 1.63x |
| E | the float32 step, and a declaration constant folded | ~1.03x |
| D | the static offsets the generator already knew | **~0 — at the noise floor** |
| J | a scalar array element sized its own read window | **~0 on this shape** |
| K | one append for a whole clause of elements | write 1.17x |
| L | one window decode for a whole clause of elements | read 1.14x |

The per-lever figures are each commit's own measurement, recorded as the
lever landed. The certification pair below is the committed sweep.

## The certification pair, and what the controls actually did

`bench/run.sh --quick`, both halves three minutes apart on the same fanless
M2 Air, corpus_id `6b213fbfa1a03a99` on every row. BEFORE stamps main at
`52691a0`; AFTER stamps this branch at `fef86ed` — the last code commit, so
both stamps resolve on origin and reconstruct the tree that was measured.

| language | before | after | control drift |
|---|---:|---:|---:|
| c | 98% | 99% | +0.71% |
| cpp | **100%** | **100%** | — |
| java | 163% | 147% | **−9.28%** |
| rust | 153% | 154% | +0.33% |
| go | 238% | 240% | +0.71% |
| cs | 360% | 363% | +0.69% |
| dart | 367% | 371% | +1.30% |
| js | 468% | 482% | **+3.05%** |
| **elixir** | **3743%** | **1457%** | −61.06% |

Elixir's own rates: write 135,696 → 378,044 msg/s, round_trip 95,413 →
239,723 msg/s. Both elixir rows sit far inside §2.3's 15% noise gate
(spreads 0.22% and 0.38%), so neither is a row `bench/tools/relative.go`
would exclude.

**Five of the seven non-cpp controls hold inside 1.3%. Two do not**, and
saying otherwise would be false: js +3.05% and java −9.28%, the two JIT
legs, and both had the noisier half in BEFORE (js spread 3.3%, java 1.9%,
against 0.4% and 0.2% after). Read them as JIT variance, not as signal.

The five tight controls all moved the SAME way, +0.3% to +1.3%, which means
cpp was marginally faster in the AFTER sitting. That drift runs **against**
this headline, not with it: a relatively faster cpp inflates every other
language's percentage, so 1457% is if anything slightly high and the
improvement slightly understated. The result is conservative.

An earlier revision of this PR published **3624% → 1451%** from an
uncommitted sweep. That pair is withdrawn. Its elixir round_trip row had
come in at 15.3% spread against the 15.0% `spreadNoisy` threshold in
`bench/tools/relative.go:86` — the repo's own tool would have excluded it,
and because the headline uses the max statistic, a noisier row carries a
luckier max. It was also the most flattering of the three figures then in
hand (1491%, 1456%, 1451%). The pair above replaces it with clean rows and
committed data; it moves the substance not at all.

## What lever D does and does not retire (issue #174)

The issue proposed one binary construction per word with pre-swizzled
segments. **That is two claims, and only one of them is dead.**

- **The arithmetic half — literal segment widths let ERTS pack better — is
  retired.** Lever D built it exactly as specified and measured ~0. The micro
  isolating it settles why: twelve appends of one 32-bit segment cost 103.6
  ns, and the same twelve appends with a DYNAMIC segment width cost 103.7 —
  identical. The literal widths were never the term that mattered.
- **The construction half — several registers riding ONE construction rather
  than one append each — is live, and it is lever M.** Its micro is direct
  evidence for the issue's own mechanism: one append of twelve segments is
  73.3 ns against 103.6 for twelve appends of one, **1.41x**. M is
  prototyped, byte-identical in output, and measured at 1.14x on the stats
  loop.

So #174's hypothesis is **not** retired. What the round established is that
append COUNT dominates — the same insight the hypothesis was reaching for by
a different route. (This matches the correction already posted on #174.)

## The floor, with its mechanism

At 1457% the port is 14.6x the fastest language — inside the five-to-fifteen
band #174 asked to have published rather than smoothed, and still ~3x short
of the generated-JS tier. The shipped stats clause, decomposed a layer at a
time: append 7.96 of 13.19 ns/stat, merge 2.03, contract checks 0.76, and
2.44 ns/stat just walking a list of maps and reading two fields before any
bit is packed. The BEAM's 60-bit fixnum caps a scratch register at 59 bits,
which is why the write group budget is 52 and the read window 49, and which
caps how much of the append cost ANY grouping can remove.

Named and not built: **lever M** (above), the only remaining pure-BEAM move,
~12% on the blend; and a **NIF backend**, which is the owner's decision and
not an optimization.

## New: the corpus gap this round exposed

`Degenerate.schema` (#203) has a standing property — every type in it is a
whole number of bytes. That is load bearing for what it catches and it is
also a ceiling: no clause boundary inside it ever lands mid-byte, so this
round's grouped write clause (52-bit budget) and grouped read clause (49-bit
window) **always pick the same element count** there. The disagreement was
untested by construction.

Two new corpus units close it, adopted from an adversarial review of this PR:

- **`Clauses.schema`** — element widths where the two sides choose different
  k. At 13 bits the write clause takes four elements (52, the whole budget)
  and the read clause three (39, inside the window); at 17 it is three
  against two; at 26, two against one. Counts run 0, below a clause, exactly
  a clause, one past it, and the bound. Plus a fixed mid-byte array, grouping
  across a nested struct boundary, a union of empty arms behind a tag, and
  `string`/`bytes` at zero, partial and full length behind a 5-bit lead so
  the align inside them is a real barrier.
- **`Joins.schema`** — the same aimed at the static-offset state machine:
  arms that agree and disagree on width, a branch with no `else`, a branch
  inside a branch, an align that regains staticness on one path only, an
  array that gives it up on one path only, unions of unequal arms at mid-byte
  offsets, and a long static run after an align.

Because these shapes are NOT byte-aligned, one shared stream would not equal
a concatenation of the shapes written alone — and the Elixir emitter returns
each message as its own binary from bit zero, so it cannot write a shared
stream at all. Each shape is therefore written to its own stream and flushed,
and the golden is those concatenated. **All nine legs byte-compare against
the C++ pin**, and each shape is held to its own declared bit width so a
doubled or dropped clause is named rather than smeared into one total.

## Constraints held

- **No existing wire golden moved.** `git diff origin/main -- testdata/wire/`
  shows only the two NEW corpus goldens, `clauses.bin` and `joins.bin`.
  `make update-goldens` rewrote every wire golden under
  `SCHEMA_UPDATE_WIRE_GOLDENS=1` and every pre-existing one came back
  byte-identical.
- corpus_id `6b213fbfa1a03a99` unchanged; `bench/` otherwise untouched.
- Adding units to `examples/` moves the `example` unit's protocol id in
  `testdata/golden/id.txt`, exactly as #198 did when it added
  `Degenerate.schema`.
- Rebased onto main. #198 pinned `testdata/golden/elixir/Degenerate.ex` with
  the pre-round emitter, so merging unrebased would have turned main red on
  `TestGoldenSourceElixir`; that source golden is re-pinned here.
- Behaviour and raise text identical; nine-backend `make test` green;
  regeneration zero-diff (`make generated-current`); shape-gate clean with no
  new ledger entries; `mix format`, `modernize`, `gofmt` and `go vet` clean.
- `serialize.elixir` is untouched and could not be otherwise — generated
  Elixir has no runtime dependency at all, so the whole seam the owner's
  grant opened is on the emitter side of it.


gafferongames:
## The profile, and the ranked lever list — posted before the work (#174's method)

### 1. Where the time actually goes

Instrument: wall-clock A/B on the canonical shape, `bench_mixed` (438 bytes;
8 entities x 135 bits, 80 stats x 18 bits, 4 loadout bytes), rotating the
same 64 pre-decoded instances the bench rotates, 100k ops per measurement,
M2 Air, OTP 29.0.5 / Elixir 1.20.4. Decomposition by shortening the lists in
the instance, which is honest because the writer's list length is data.

```
write        7.344 us/op
read         3.561 us/op        (round_trip 10.748 us/op)

write splits            us      per element     share
  stats, 80 elems      2.632    32.9 ns          36%
  entities, 8 elems    2.970   371.5 ns          40%
  header+footer+loadout 1.742                    24%
```

**The write path is twice the read path.** eprof agrees on the ordering
(stats 38.7%, entities 23.7% of write) and is not quoted for magnitudes —
162000 traced calls into the stats loop inflate the very function the wall
clock independently indicts. BEAM profiling distorts; the A/B is the ruling.

**The mechanism, named.** `mergeW` flushes every whole byte after EVERY
field, so **325 `bs_append` BIF calls ride one message** (131 static field
sites + 8x14 + 80x2 + 4x1 in the loops). The read side's `rd/3` opens a
fresh match context per FIELD — 14 of them per entity, 2 per stat.

**Not a routing question.** Generated Elixir has no runtime dependency;
serialize.elixir is not on this path at all. Every lever below is emitter-only.

### 2. The levers, measured in isolation before ranking

Micro-benchmarks reproducing the shipped loop shapes exactly (the shipped
shape reproduces to 32.8 vs the in-situ 32.9 ns/stat, so the model is faithful).
These live in scratch, not in the repo — the shape gate (#200) is right that
hand-coded shapes do not belong in `bench/`.

| write, per stat (18 bits, 2 fields) | ns |
|---|---|
| shipped: 1 append per field | 32.8 |
| grouped: 1 append per element | 28.0 |
| grouped + one map read per element | **20.2** |
| unrolled x4 (72 bits = 9 bytes), 2 appends / 4 elements | 14.3 |
| no append at all (arithmetic control) | 12.6 |

| write, per entity (135 bits, 14 fields) | ns |
|---|---|
| shipped shape | 348 |
| 4 grouped appends + one map read | **101** |

| read, per stat | ns |
|---|---|
| shipped: one `rd` per field | 27.6 |
| one `rd` per element + shift/mask | 14.8 |
| one 56-bit-window `rd` per 2 elements | **7.7** |
| one **72-bit**-window `rd` per 2 elements | 29.4 |
| progressive `<<..., rest::binary>>` match, 4-element period | 13.2 |
| one `rd` per element, no struct built (control) | 11.2 |

Two results that decide the design:

- **The 72-bit window is SLOWER than the shipped 40-bit one (29.4 vs 27.6).**
  A 72-bit window boxes: the BEAM's fixnum is 60-bit signed, so any
  intermediate at or above 2^59 costs a heap bignum. The whole design is
  fenced by that boundary — this is why the group budget below is 52 bits
  (7 leftover + 52 = 59) and not "a 64-bit word".
- **Progressive `rest::binary` matching (13.2) barely beats a grouped `rd`
  (14.8) and loses to a wider grouped `rd` (7.7)**, while costing 8-phase
  clause specialization to generate. The match-context-reuse idiom is not
  worth its generator complexity here. Recorded as a negative result.

### 3. Ranked by win / effort

| # | lever | est. win | effort | take |
|---|---|---|---|---|
| A | **Grouped write appends.** `mergeW` stops flushing per field: accumulate into `scratch` with the generator's statically-known offsets and flush once per group of <= 52 bits, with explicit flushes at every barrier (function tail, loop-helper call and loop tail, align, string bytes, branch and union arm joins). 325 appends/msg -> ~90. | write **~1.7x** | low — one choke point (`mergeW`) plus barrier calls | yes, first |
| B | **One map read per element.** Emit `%{f1: v1, ...} = e` once instead of N `e.field` accesses; each `.` is its own `get_map_elements` with its own raise branch, one destructure is one instruction. | write **~1.3x** on top of A | low-medium — field-path rewriting in the emitter | yes |
| C | **Grouped reads through a wider window.** One `rd` per group of <= 49 bits into a group register, fields extracted by static shift/mask; a second helper `rdw` opens a 56-bit window (49 = 56 - the 7-bit worst-case shift) while `rd` keeps the cheap 40-bit window for narrow groups. | read **~2x** | low-medium — one choke point (`readR`) plus group resets at the same barriers | yes |
| D | **Static-offset specialization.** Where the bit offset is statically known (everything before the first counted array), the flush size is a literal and `flush`/`scratch_bits` bookkeeping disappears entirely. | write ~1.1x | low, once A exists | yes if A lands clean |
| E | **`fr/1` halves its allocations.** The float32 rounding helper builds two 4-byte binaries where one suffices (match the same binary twice, once as int and once as float). 18 calls/msg. | write ~1.02x | trivial | yes |
| F | **Unroll loops to the byte-alignment period** (4 stats = 9 bytes; 8 entities = 135 bytes), so a whole period is emitted with byte-literal flushes. Measured 14.3 vs 20.2 ns/stat. | write ~1.15x | high — period clauses plus a remainder path per element type | no — named, not built |
| G | **Phase-specialized loop clauses** (8 clauses per element type, one per entry bit-phase), which would make every offset static inside a loop. | unknown, <= F | high — 8x generated code per array element type | no — named, not built |
| H | **iodata instead of binary append.** Measured 23.2 ns/stat against the shipped 28.0 for the same append count — but grouped appends (A) already reach 20.2, and iodata then owes a final `iodata_to_binary` pass. | negative once A lands | low | **no — measured and rejected** |
| I | **NIF backend** (#174's reserved second tier). Out of scope for this round by the issue's own sequencing: it goes on the table after the pure-BEAM floor is published, and it is the owner's decision, not an optimization. | large | large | no |

Expected combined: write ~2.2x, read ~2.1x, blended `round_trip` ~2.1x.
That is 3699% -> roughly 1700%. It does not reach the generated-JS tier (478%),
and the reason will be stated with its mechanism rather than smoothed.

gafferongames:
## The round's result: 3624% → 1451% of generated C/C++, and the floor named

Four more levers since the ranked list above — D and J from it, K and L
re-ranked onto it on evidence. Every one measured, one commit each, and the
two that did not pay are recorded as results rather than quietly kept.

### The certification numbers

`bench/run.sh --quick`, both legs in ONE sitting, same box (M2 Air, OTP
29.0.5 / Elixir 1.20.4), corpus_id `6b213fbfa1a03a99` on every row. The
published blend is the §2.9 `round_trip`. The eight non-Elixir rows are the
in-sitting control: none of them moves more than 2%, so the Elixir move is
the change and not the weather.

| language | before | after |
|---|---:|---:|
| c | 97% | 99% |
| cpp | **100%** | **100%** |
| java | 144% | 148% |
| rust | 151% | 153% |
| cs | 355% | 361% |
| dart | 359% | 368% |
| js | 465% | 469% |
| go | 572% | 589% |
| **elixir** | **3624%** | **1451%** |

Elixir's own rates: write 133,697 → 365,823 msg/s (**2.74x**), round_trip
93,764 → 234,112 msg/s (**2.50x**). The sweep's elixir round_trip row
tripped §2.3's noise gate at 15.3% spread, so the leg was re-run twice
alone: 1491% (spread 1.8%) and 1456% (spread 1.0%). 1451% is the sweep's
own figure and the one quoted; the re-runs bracket it.

Whole round, against the pre-round baseline, on the scratch A/B (the
committed variant corpus, 64 rotating instances, 150k ops, 4 runs x 4
interleaved passes, best per pass, median of passes):

| | before | after A/B/C/E | after D/J/K/L | round |
|---|---:|---:|---:|---:|
| write | 7.431 | 3.203 | **2.602** | **2.86x** |
| read | 3.529 | 2.155 | **1.779** | **1.98x** |
| round_trip | 12.696 | 5.374 | **4.354** | **2.92x** |

### The four levers

**D — the static offsets the generator already knew.** A write surface
starts at a known empty scratch and stays known until the message's own data
decides a length, so through that region every shift is a literal, every
flush is a literal-width binary segment, the `scratch_bits` variable is not
maintained at all, and the tail is `data` or `<<data::binary, scratch>>`
outright. Staticness is given up at a loop call and at a branch whose arms
disagree, and regained at every align.

**MEASURED: at or below the noise floor. ~2% on write, and the honest
reading is zero.** head 3.122 vs D 3.136 by median over five passes, 3.095
vs 3.038 by best. The isolated micro says why, and the two agree: the whole
pre-lever-D bookkeeping shape costs 113.8 ns per twelve 32-bit fields and
the post-lever-D shape 103.6 — 9% of a region that is 24% of the write. 2%
is exactly what 9% of 24% predicts. **The measurement did not disappoint the
mechanism; the mechanism was small.** The lever stays because it also made
K possible to reason about, and because the code it deletes is real, but it
is recorded as ~0, not as what it looked like it should be worth.

**J — a scalar array element sized its own read window.** The run fuser is
reached through `emitReadItems`, and a scalar array element is not: it went
straight to the scalar read with the run unknown, so a one-byte element
opened the 56-bit window to take eight bits — and took that window's longer
`:binary.decode_unsigned` tail fallback with it, two bytes sooner than the
40-bit window would. **MEASURED: no change on this shape** (read 2.054 vs
2.069, inside the noise). The loadout array is four elements of a 438-byte
message; the defect is real and the shape it appears in here is too small to
show it. Recorded as correct-and-unmeasurable, kept because a schema with a
large scalar array pays it and this one does not.

**K — one append for a whole clause of elements.** This is where the write
time was, and the ranked list had it at "no — named, not built" on a cost
estimate that turned out to be wrong. The micro that re-ranked it: twelve
appends of one 32-bit segment cost 103.6 ns; the same 48 bytes in three
appends of four segments cost 79.0, and in one append of twelve, 73.3. The
same twelve appends with a DYNAMIC segment width cost 103.7 — identical.
**The append is the expense and the arithmetic around it is not.**

The list clause now takes k elements, k being the most whole elements whose
widths fit the 52-bit group budget, capped at four; the single-element
clause behind it is the remainder, so no list length has to divide anything.
No byte-alignment period and no phase specialization are needed — which is
what made F look expensive and is why the estimate was wrong. Stats: 80
appends → 40. Loadout: 4 → 1.

**MEASURED: write 3.097 → 2.65 us/op, 1.17x. round_trip 1.11x.**

**L — one window decode for a whole clause of elements.** K's other half.
A read clause decodes k elements under one window, k bounded by the wide
window's 49 usable bits. Stats: 80 decodes → 40. Loadout: 4 → 1.

**MEASURED: read 2.089 → 1.838 us/op, 1.14x. round_trip 1.07x.**

Both needed the element emitters to take the variable they bind (so a clause
can hold `e1` and `e2`), and the raise TEXT is held across it: the display
map resolves every slot back to `e`, so a message still reads `e.delta is
above the wire maximum`.

### The floor, with its mechanism

At 1451% the port is **14.5x the fastest language** — inside the five-to-
fifteen band #174 asked to have published rather than smoothed, and still
3.1x short of the generated-JS tier (469%). Here is what holds it there,
measured rather than asserted. The shipped stats clause, decomposed by
removing one layer at a time from the real shape:

| | ns/stat | what the layer costs |
|---|---:|---|
| shipped: destructure + checks + merge + append | 13.19 | |
| no append | 5.23 | **the append: 7.96** |
| no merge either | 3.20 | the merge: 2.03 |
| destructure only | 2.44 | the contract checks: 0.76 |
| bare list traversal | 0.72 | the two field reads: 1.72 |

Two things follow, and they are the floor.

**1. The append still dominates, and the fixnum boundary caps how much of it
can be removed.** The BEAM's fixnum is 60-bit signed, so a scratch register
cannot carry more than 59 bits without heap-allocating a bignum — measured
back in lever A, a boxed window is slower than the flush it saves. That is
why the group budget is 52 and the read window 49, why a stats clause takes
two elements and not four, and why no amount of further grouping can put a
whole 438-byte message in one construction. **The cap is the BEAM's integer
representation, not the design.**

**2. Under the serialization there is a per-element cost no codec can
remove.** Walking a list of 80 maps and reading two fields from each costs
2.44 ns/stat before a single bit is packed — 19% of the shipped clause —
and the contract checks the SPEC requires add 0.76 more. C does the same
work in registers with branch-predicted compares.

### Ranked next, named and not built

**M — one binary construction per barrier, not one per flush.** The one
lever that can still move the append cost, and the closest thing to #174's
original hypothesis: a barrier emits ONE construction carrying several
scratch registers as several segments, instead of one construction per
flush. A register still cannot exceed 59 bits, but a *construction* can
carry as many registers as it likes, so four stats can ride one append as
two segments. **Prototyped and measured: 14.62 → 12.61 ns/stat, 1.14x on
the stats loop, and the bytes come out identical to the shipped shape** —
the mechanism is proven, only the emitter work is left. Its bigger prize is
the entity loop, which flushes three times per element for 40% of the write.
Estimated ~20% on write, ~12% on the blend; 1451% → roughly 1300%.

**N — a NIF backend.** #174's reserved second tier, and after this round it
is the only thing that reaches the JS tier. It is the owner's decision and
not an optimization: per-platform native compilation, and a NIF fault takes
the VM down.

Ruled out and measured, unchanged from the list above: F is superseded by K
(and was mis-ranked, recorded), G is not needed once K showed phase
specialization was never the requirement, H stays rejected.

### Constraints held

Wire bytes never moved: no file under `testdata/wire` is touched by any
commit on this branch, corpus_id `6b213fbfa1a03a99` on every bench row
before and after, the full per-variant decode/re-encode gate green in all
nine legs. Behaviour identical: the same inputs refused, the same raise
text. Every generated module compiles without a warning,
`mix format --check-formatted` is clean, and the nine-backend `make test`
chain is green. `serialize.elixir` is untouched and could not be otherwise:
generated Elixir has no runtime dependency at all, so the whole seam the
owner's grant opened is on this side of it.

The micro-benchmarks quoted here live in scratch, not in the repo — #200's
shape gate is right that hand-coded shapes do not belong in `bench/`.

gafferongames:
## The adversarial review's five, answered — and its corpora adopted

The review returned BLOCK on five items, none touching the wire, which it
verified clean by its own commands. All five are fixed; its two schemas are
now part of the corpus.

**1 — REBASE (hard blocker). Fixed.** The branch's merge-base was `aa6c62e`
(#200) but main is `52691a0` (#198), which added `examples/Degenerate.schema`
and pinned `testdata/golden/elixir/Degenerate.ex` with the OLD emitter.
Merging as-is turned main red on `TestGoldenSourceElixir`, and it did:
reproduced before the fix, exactly as described. Rebased, regenerated,
re-pinned. The re-pin is a SOURCE re-pin only, and here is the proof rather
than the assertion: `make update-goldens` rewrote every wire golden under
`SCHEMA_UPDATE_WIRE_GOLDENS=1` and **every one came back byte-identical**,
`git diff origin/main -- testdata/wire/` shows nothing but the two new corpus
goldens, and all nine legs byte-compare `degenerate.bin` against the C++ pin
green.

**2 — COMMIT THE CSVs. Fixed, by re-measuring.** The original sweep's CSVs
were never written anywhere that survived, so there was nothing to commit —
which is precisely the finding. Both halves were therefore re-run as a fresh
paired A/B in one sitting after the rebase and committed. The stamps resolve
on origin: BEFORE `52691a0` (main), AFTER `fef86ed` (the last code commit on
this branch — commits after it add only this data and prose, so the stamp
reconstructs the measured tree). No dead SHA, and no need for the
"measurement predates the rebase" caveat.

**3 — THE HEADLINE. Fixed, and the review was right about the mechanism.**
1451% was the row that tripped §2.3's noise gate at 15.3% spread against the
15.0% `spreadNoisy` in `bench/tools/relative.go:86` — the repo's own tool
would have excluded it — and because the headline uses the max statistic, a
noisier row carries a luckier max. It was also the most flattering of the
three figures then in hand. Rather than pick among them, the pair is
re-measured and committed: **3743% → 1457%**, elixir spreads 0.22% and 0.38%,
nowhere near the gate. That lands on the review's own clean re-run (1456%).
The substance is unaffected.

**4 — "none of them moves more than 2%". Corrected, and it was false in my
data too.** Five of seven non-cpp controls hold inside 1.3%; **two do not** —
js +3.05% and java −9.28%, the two JIT legs, both with the noisier half in
BEFORE. The claim is gone and the real per-language figures are in the body.

On direction: the five tight controls all moved the same way (+0.3% to
+1.3%), so cpp was marginally faster in the AFTER sitting. That inflates
every other language's percentage, which means it runs **against** this
headline — 1457% is if anything slightly high and the improvement slightly
understated. Conservative, as the review argued.

**5 — THE #174 FRAMING. Fixed.** The body now says what the correction on
#174 says: the issue's hypothesis is **not** retired. Its ARITHMETIC half is
(lever D, ~0, with the literal-vs-dynamic-width micro at constant append
count proving why). Its CONSTRUCTION half is lever M — live, prototyped,
byte-identical output, 1.14x on the stats loop, and its micro (1 append of 12
segments 73.3 ns vs 12 appends of 1 at 103.6, **1.41x**) is direct evidence
for exactly the mechanism #174 proposed.

## The corpora, promoted

The review's `Adv.schema` and `Adv2.schema` were still on disk. They are
adopted as **`examples/Clauses.schema`** and **`examples/Joins.schema`**,
byte-compared by all nine legs against C++-pinned goldens, in the same shape
as `Degenerate.schema`.

The gap is real and structural, and the generated code confirms it directly.
For `W13` (`[..12]uint16 | max = 8191`) the emitter produces a **write clause
of four elements** (`[e1, e2, e3, e4 | rest]`, 52 bits — the whole budget)
and a **read clause of three** (`rdw(data, bits_read, 39)`). The two sides
choose different k, and both boundaries land mid-byte — which
`Degenerate.schema` cannot reach, because its standing property is that every
type in it is a whole number of bytes.

One correction to the review's framing, found while adopting them: the
single-stream design it used for the reference dump is the only one that
works. These shapes are not byte-aligned, so a shared stream does not equal a
concatenation — and the Elixir emitter returns each message as its own binary
from bit zero, so it has no way to write a shared stream at all. Each shape
is therefore written alone and flushed, and the golden is those concatenated.
As a cross-check, **the pinned `clauses.bin` and `joins.bin` are byte-identical
to the review's own independent C++ dumps** (343 and 247 bytes), and all 90
shape widths in the harness were derived by hand and matched it.

## Verification

`make test` green, nine backends. `make generated-current` — regeneration
zero-diff. Nothing under `testdata/wire` moved except `clauses.bin` and
`joins.bin`. corpus_id `6b213fbfa1a03a99` unchanged and `bench/` otherwise
untouched. Shape gate clean, 19 exemptions, ledger unchanged from main.
`mix format --check-formatted`, `modernize`, `gofmt`, `go vet` all clean.
Adding units to `examples/` moves the unit's protocol id in
`testdata/golden/id.txt`, exactly as #198 did when it added
`Degenerate.schema`.

PR #207, opened 2026-08-31
elixir #174: lever M — one binary construction per barrier

gafferongames:
Issue #174. **Lever M** — the one lever #202 named, prototyped and did not
land. With the NIF backend ruled out, it is the last planned Elixir lever, so
the number it lands is **the BEAM's published floor for now**.

**Headline, and it is below the estimate.** M makes the write path
**1.17x** faster and moves the §2.9 blend somewhere between **0% and 6%**,
against the ~12% #202 estimated. The write half is unambiguous and measured
by two instruments; the blend half is small, and the certification pair could
not resolve it at all. Both are reported as measured, neither is smoothed.

## What M is

Lever A closed the merge group across the fields of a scope and K closed it
across the elements of a clause, but neither could close it across the
GROUP's own budget: a barrier-free run wider than 52 bits had to empty the
scratch to keep merging, and every emptying was its own `bs_append`. A
`bench_mixed` header is 82 static bytes and paid ten of them; an entity
element paid three.

The emptying and the append were never the same thing. A full group now
**spills** — the scratch's whole bytes bind to a fresh register and the
scratch keeps merging — and the append is deferred to the **barrier**, where
every register that spilled since the last one rides ONE binary construction
as its own segment. A spill is a bind and a shift where it was an append and
a shift, so the statement count is unchanged and only the BIF calls move.
The dynamic-offset path (a loop helper's body) defers identically, each spill
naming the width local it computed — #174's own micro says that costs
nothing: twelve appends of a DYNAMIC-width segment measured 103.7 ns against
103.6 for literal widths.

The write clause's `k` follows from it. What a clause shares is one APPEND,
and the append is now the barrier's rather than the group's, so the 52-bit
budget no longer bounds `k` — only the unroll cap of four does.

| | before | after |
|---|---|---|
| static append sites in `generated/bench/elixir` | 79 | **44** |
| `bench_mixed` header | 10 appends | **1** (ten segments) |
| entity element | 3 appends | **1** |
| stats clause | 2 elements, 40 appends | **4 elements, 20 appends** |

## Measured

The lever's own figure comes from the **interleaved** instrument levers K and
L used: both arms in the same window, order rotated per pass, so a spike hits
both. Two sittings, `bench/elixir` under the pinned BEAM toolchain, the same
runner and the same corpus over two generated `Bench.ex` texts, max statistic,
median of per-pass ratios.

**Ten passes, main vs M** — M won 9 of 10 on write:

| | main | lever M | ratio |
|---|---:|---:|---:|
| write | 361,140 | 424,619 | **1.17x** |
| round_trip | 235,207 | 249,436 | **1.05x** |

**Nine passes, three-way**, separating M's two halves — C is the deferred
appends alone, with the clause `k` left where K had it:

| | B/A (full M) | C/A (appends only) | B/C (the clause widening) |
|---|---:|---:|---:|
| write | **1.168** | 1.099 | **1.070** |
| round_trip | **1.072** | 1.054 | 1.016 |

Both halves earn their place: deferring the appends is 1.10x on write on its
own, and widening the clause adds a further 1.07x. Neither is scaffolding.

## The certification pair, and what it could not show

`bench/run.sh --quick`, both halves 2m39s apart, corpus_id `6b213fbfa1a03a99`
on every row, **both committed**:
`bench/results/2026-09-01-elixirleverm-{before,after}-quick-arm64-macbook.csv`.
BEFORE stamps main at `d398ec4`, AFTER stamps this branch at `f96d749`.

| | before | after | ratio |
|---|---:|---:|---:|
| elixir write | 383,329 | 434,318 | 1.133x |
| elixir round_trip | 248,616 | 254,142 | 1.022x |
| **blend (§2.9)** | **1413%** | **1409%** | — |

**The controls say to read that second row as nothing.** The AFTER half was
globally the faster of the two — but not uniformly, and the original text of
this paragraph overstated it (CORRECTED at review, and commit 894414a's
message carries the old claim): the sixteen non-elixir rows actually ranged
**-0.25% to +4.10%** (cs round_trip -0.12%, dart round_trip -0.25%, dart
write -0.02% came in DOWN; cpp round_trip +1.9%). The wider, mixed-sign
control band strengthens the conclusion rather than weakening it: elixir
write at +13.3% is well clear of that drift; elixir round_trip at +2.2% is
inside it, and net of cpp the blend moves 0.3%.

That is not evidence the lever does nothing. It is evidence **this instrument
cannot resolve this question.** One sweep is ONE sample per arm, and elixir's
round_trip rate varies far more between runs in a sitting than within one:
across the nine interleaved passes of the same main-branch code it ran
219,874 to 247,320, a **12.5% range**, against the 0.55% spread inside a
single run. A 6% effect does not survive that. The sitting is labelled SHARED
in both preambles, which it was — a sibling session's C# round was live on the
same box.

So what a sweep actually recorded is **1413% -> 1409%**. The interleaved
round_trip ratio would imply roughly 1330%, and I am **not** publishing that
number: no sweep measured it, and the §2.9 blend is a sweep statistic. The
honest position is that this lever's effect on the blend is smaller than the
sweep's own between-run variance, and a quiet box would be needed to put a
figure on it.

## Why it is below the ~12% estimate — the mechanism, not a smoothing

The estimate came from #174's micro: one append of twelve segments at 73.3 ns
against twelve appends of one at 103.6, **1.41x**. That is the cost of appends
*in isolation*. In the codec they are one term of write, and write is ~65% of
round_trip on this shape — so a 1.17x write is at most a 1.10x round_trip
before anything else is counted, and the blend statistic is round_trip.

**The "slower derived read" is an arithmetic artifact, not a measurement**
(REWRITTEN at review — the original paragraph here presented a 6-15% derived-
read slowdown as an unexplained reproducible effect). Derived read is
round_trip MINUS write, never a measured row (§2.9 says exactly this and why).
Any lever that improves write more than round_trip drives that derived number
backwards BY CONSTRUCTION, and its error band is enormous: read-time is ~35%
of round_trip here, so a 1% error in round_trip moves derived read ~2.9%, and
round_trip's own between-run range on unchanged code is 12.5% — a ±30%+ band
on the derived figure. The three "instruments" shared one round_trip number,
so they were never independent replicates. **M is neutral on real read
traffic by construction**: the read emission is byte-identical (all 1,232
lines diffed), and read-heavy workloads decode binaries that arrived from a
socket — already plain non-writable refc binaries, unreachable by anything a
writer-side lever changes. The one true residue: the round-trip loop pays an
emasculation copy when the reader touches the still-writable binary the same
process just built — a loop artifact, identical in size on both branches, and
the reason round_trip cannot capture the whole write gain.

## Constraints held

- **The Elixir wire is byte-identical to the C++ pin.** The gate that
  actually carries an Elixir-only change is `test/elixir/main.exs`
  byte-comparing generated output against `testdata/wire/*.bin` — green, and
  the review proved it discriminating (a reversed barrier segment order goes
  red on six checks). ("No wire golden moved" is true but near-vacuous for
  this diff: goldens are produced by the C++ leg, which an Elixir emitter
  change cannot touch.) corpus_id `6b213fbfa1a03a99`.
- **The fixnum envelope is untouched**, which is what this whole round rests
  on. `mergeW`'s guard is unchanged, so a group is still at most 52 bits and
  the scratch at most 7 + 52 = 59; a register is filled by exactly the spill
  that used to append it, so it holds what the scratch held and nothing is
  grown. Registers stay live to the barrier instead of dying at the append —
  the deepest run in the tree spends 13 of the BEAM's 1024 (write_rigid_body, sc0..sc12 plus scratch — the review counted; the argument is unchanged).
- **Clauses.schema and Joins.schema pass**, which is the point of them: their
  mid-byte clause boundaries are exactly where a grouped write can disagree
  with a grouped read, and this lever changes how many elements a write clause
  takes. All nine legs byte-compare against the C++ pin.
- **Raise text and refusal semantics identical.** The diff ADDS 34 raise sites
  — the new clause slots — and removes none; the display map resolves every
  slot back to the element, so a message still reads `"e.delta is above the
  wire maximum"`. Every range, count, mask and length check is where it was.
- The barrier's construction is emitted in `mix format`'s own shape for a long
  bitstring, so `--check-formatted` passes on the emitted text.
- Nine-backend `make test` green; regeneration zero-diff; every generated
  module compiles under `--warnings-as-errors`; shape-gate clean with no new
  ledger entries; `mix format`, modernize, gofmt and `go vet` clean. The seven
  SOURCE goldens the emitter moved are re-pinned in the lever's own commit.
- `serialize.elixir` untouched — generated Elixir has no runtime dependency.

### DIFF
--- a/bench/results/2026-09-01-elixirleverm-after-quick-arm64-macbook.csv
+++ b/bench/results/2026-09-01-elixirleverm-after-quick-arm64-macbook.csv
@@ -1,0 +1,49 @@
+# schema bench results
+# date: 2026-08-31T23:42:31Z
+# host: macbook  arch: arm64  os: Darwin 25.6.0
+# cpu: Apple M2
+# build: Release
+# cpp compiler: Apple clang version 21.0.0 (clang-2100.1.1.101)
+# cpp flags: -O3 -DNDEBUG -DSERIALIZE_RELEASE -DBENCH_OPT="O3" -std=c++17 -Wall -Wextra -Werror -ffp-contract=off -fno-rtti -Igenerated/bench/cpp -I../serialize
+# c compiler: Apple clang version 21.0.0 (clang-2100.1.1.101)
+# c flags: -O3 -DNDEBUG -DBENCH_OPT="O3" -std=c99 -Wall -Wextra -Werror -Igenerated/bench/c -I../serialize.c (serialize.c compiled in, no LTO)
+# go: go version go1.27.0 darwin/arm64 (go run, default optimized build)
+# rust: cargo 1.98.0 (797e8a9bc 2026-08-05) (Homebrew); rustc 1.98.0 (88d9e12ae 2026-08-18) (Homebrew) (cargo run --release at opt-level 3, no LTO)
+# dotnet: 10.0.400 (dotnet run -c Release, workstation GC)
+# node: v26.7.0 (NODE_ENV=production — serialize.js's caller-trust release mode; the runner records the mode that ran in its checks column)
+# java: openjdk 21.0.12.1 2026-08-18 LTS (generated codecs, zero runtime dependency; default JVM flags, no -ea)
+# dart: Dart SDK version: 3.13.2 (stable) (Tue Aug 25 01:01:12 2026 -0700) on "macos_arm64" (generated codecs, zero runtime dependency; AOT executable)
+# elixir: 1.20.4 (generated codecs, zero runtime dependency; pinned BEAM toolchain)
+# pinning: none
+# noise: SHARED fanless M2 Air, a sibling Claude session (the C# round) live in the estate; load avg recorded per half in the PR. --quick is the iteration instrument, never certification (BENCH-STANDARD §2.8). AFTER half of a paired same-sitting A/B for lever M on branch elixir-lever-m: this branch's Elixir emitter output in generated/bench/elixir; every other leg is the in-sitting control. The lever's own figure comes from the ten-pass INTERLEAVED elixir A/B recorded on the PR, not from this pair — a shared box moves an unpaired half.
+# schema commit: f96d749
+# serialize commit: cebaed2 (main) [build-verified: /Users/glenn/rowan-working/serialize]
+# serialize.c commit: 37db942 (main) [build-verified: /Users/glenn/rowan-working/serialize.c]
+# serialize.go commit: ae730d5 (main) [build-verified: /Users/glenn/rowan-working/serialize.go]
+# serialize.rs commit: 6b52aa6 (main) [build-verified: /Users/glenn/rowan-working/serialize.rs]
+# serialize.cs commit: f08327e (main) [build-verified: /Users/glenn/rowan-working/serialize.cs]
+# serialize.js commit: e391f1e (main) [build-verified: /Users/glenn/rowan-working/serialize.js]
+lang,bench,path,iters,bytes_per_op,runs,median_msgs_per_sec,min_msgs_per_sec,max_msgs_per_sec,median_mb_per_sec,spread_pct,corpus_id,family,linkage,checks,opt,inline
+cpp,bench_mixed,write,4000000,438,3,7502917,7500031,7511501,3134.04,0.15,6b213fbfa1a03a99,gen,hdr,removed,O3,unknown
+cpp,bench_mixed,round_trip,4000000,438,3,3576671,3569319,3580272,1494.01,0.31,6b213fbfa1a03a99,gen,hdr,removed,O3,unknown
+lang,bench,path,iters,bytes_per_op,runs,median_msgs_per_sec,min_msgs_per_sec,max_msgs_per_sec,median_mb_per_sec,spread_pct,corpus_id,family,linkage,checks,opt,inline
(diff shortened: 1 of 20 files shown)


gafferongames:
Review verdict: MERGE, all conditions now landed. The reviewer verified the spill machinery structurally (a Go flattener rendering every multi-segment construction back into the append sequence it replaced — every static byte-width sequence identical to main, every dynamic sequence identical in count and order), ran its OWN negative control (reversed segment order -> six checks red including the C++-pin byte compare), and worked the fixnum envelope by hand (the same 59-bit ceiling main already hit, not raised). It also DISSOLVED the derived-read anomaly: an arithmetic artifact by construction (round_trip minus write, ±30% band from round_trip's own 12.5% variance, three instruments sharing one number), with M provably read-neutral — the body's paragraph is rewritten accordingly. Conditions landed: the false control sentence corrected in place with the correction named (commit 894414a's message carries the old claim — noted rather than rewritten, history stays); the segW join-reset guard (one line, zero-diff regen, segN kept monotonic to avoid renumber churn); nine->13 registers; the wire-gate claim restated in the terms that actually carry an Elixir-only diff. Estate-wide finding routed to #194: §2.9 F1's datadriven filename rule is followed by no CSV in the repo including main's own — the ledger issue owns sub-era bookkeeping and should either enforce or amend it.

PR #208, opened 2026-08-31
cs: the codec pass — union batch cores, scoped batches, and the flat word codec (write 1.59x, round_trip 1.60x; 361% -> 219% of generated C++)

gafferongames:
The C# backend was the laggard on the current ledger — 361% of generated C/C++ on `bench/results/2026-09-01-arm64-macbook.csv` (main at d00cf49) — and the only backend never audited for generation quality (#170's matrix). This is that audit and the levers it found.

**Headline** (`--only cs`, both halves committed, two sittings back to back, `bench_mixed`, family `gen`, `checks=always`):

| path | before max | after max | before median | after median | |
|---|---|---|---|---|---|
| write | 2,104,563 | **3,342,814** | 2,103,454 | 3,334,630 | **1.59x** |
| round_trip | 992,155 | **1,594,245** | 990,865 | 1,586,844 | **1.60x** |

Against the same day's quiet-box reference at the same `corpus_id` (cpp round_trip median 3,467,551), on the round_trip statistic: **350% → 219% of generated C++**, and on write 347% → 219%. Zero allocations on the hot path throughout (the leg's own assert). Wire bytes never moved: full `make test` green across nine backends on every commit, degenerate corpus included.

---

## 1. MEASURE FIRST — where the time actually was

Instrument: `DOTNET_JitDisasm` + `DOTNET_JitDisasmSummary`, `DOTNET_TieredCompilation=0`, FullOpts, against the built `bench/cs` Release binary (BENCH-STANDARD.md §4.1's own recipe). Not dotnet-trace: every hot call inlines, so a sampled stack would attribute the whole codec to one frame and say nothing.

**Finding 1 — this is NOT Go's distribution.** #198 found ~86% of Go's time inside runtime calls the compiler would not inline. C# has no such wall: `ReadBenchMixed` and `WriteBenchMixed` contain ZERO direct `bl` into serialize.cs. Every entry point the generated code touches inlines. Anyone porting #198's conclusions here would have optimised the wrong thing.

**Finding 2 — the cost was the generated code's own composition.** `BenchMixed` reaches a union, and the batch plan excluded any type that reached one, so the whole message ran on the stream and reached its element types through their STREAM entries — each of which opens a batch and closes it again:

| site | elements | BeginBatch/End round trips | real calls |
|---|---|---|---|
| `entities` | 8 | 8 | 8 |
| `stats` | 80 | 80 | 80 |
| `game_event` | 1 | 1 | 1 + 1 arm |

89 capture/restore round trips per 438-byte message, on the two arrays carrying 72% of the wire bits.

**Finding 3 — what the inlined bodies do.** `BitWriter`'s packer step masks, shifts, ORs into scratch, tests `newScratchBits >= 64`, conditionally stores a qword and recovers the spill: a data-dependent branch and a conditional store PER FIELD. `MixedEntity` paid fourteen for 135 bits.

**Finding 4 — read-path code density.** `ReadBitsUnchecked`'s slack-free window-assembly fallback loop is inlined at EVERY field read site — a loop that never runs on a slack buffer. `ReadBenchMixed` was 8952 bytes of code for 869 bytes of IL.

## 2. THE LEVERS, ranked, and what each returned

| # | lever | where | status |
|---|---|---|---|
| A | **Union batch cores + a scoped batch per composition site** — 89 capture/restore round trips become 3 | emitter | **LANDED — write +34%, round_trip +11%** |
| A' | Whole-message batch, the obvious companion to A | emitter | **REFUSED BY MEASUREMENT** — write 2.10→2.04, round_trip 0.97→0.76 M msg/s. See §3. |
| B1 | **The flat word codec, write half** — values into locals, OR'd into 64-bit chunk words at literal shifts | emitter | **LANDED — write +17%, round_trip +9%** |
| B2 | **The flat word codec, read half** — one bounds check per chunk, one sticky-error test per run | emitter | **LANDED — round_trip +30%** |
| E | Public refusal-latch on `ReadStream`/`ReadBatch`, so B2 could fold the ranged reads it leaves on the runtime call | serialize.cs | **RANKED AND REFUSED — worth ~1%.** See §4. |
| C | Hoist the read slack-path out of line (`NoInlining` cold helper) so the fallback stops being pasted at every read site | serialize.cs | not taken — B2 cut the read sites from 14 to 3 per entity, which removes most of what made this attractive |
| D | Branch-free 64-bit chunk primitives on the batch (`value >> 1 >> (63 - sb)` handles `sb == 0` branchlessly) | serialize.cs | not taken — ranked behind C for the same reason |
| F | Group K array elements into one run (80 stats × 18 bits = 80 chunk touches today; K=8 would make it 30) | emitter | **the next real lever** — Go's `flatGroupRun`, not ported here (land-and-expand) |
| G | Unroll small fixed scalar arrays into runs (`loadout [4]uint8`) | emitter | small, not taken |
| H | The C# bench driver's write/read delegates (one indirect call per op cpp and rust do not pay) | bench/cs | **NOT TAKEN — harness, and cross-leg comparability is not mine to move** |

## 3. Lever A' — the measured refusal, stated

Once unions carry cores, nothing stops the whole message body running on one batch. It is slower, and materially. The JIT DID inline the 12.7 KB core (entry IL size 23), so this is not the address-exposure law from the original batch work firing. It is the delegated sites: `BenchMixed`'s body has six (`fixed`, two `bytes` runs, `ufixed`, `uint128`, `int128`), each a Sync down and a Recapture back, and between them a ref struct the register allocator must keep alive across 12.7 KB of code.

`batchEntry` now states it: density is necessary, `b == 0` is the rest.

## 4. Lever E — refused on evidence, not deferred

B2 folds a ranged read into a run only where the range check is VACUOUS (`max - min == 2^bits - 1`), or where this emitter's read path already refuses without latching (the full-range unsigned path). Everything else keeps the runtime call, because `SerializeInt` LATCHES `ValueOutOfRange` and a generated comparison would not — cs publishes `checks=always` and test/cs pins that latch.

The obvious follow-on was a public refusal-latch on `ReadStream`/`ReadBatch` so the rest could fold. **A hand-written prototype that folded every ranged read of the two hot cores, latch semantics and all, measured round_trip 1.55 M msg/s. The landed restriction measures 1.54.** The API addition is worth about 1%. Recorded as refused.

## 5. Discipline

- **Wire bytes never move.** Full `make test` green across nine backends on every commit — degenerate corpus, `test/cs` and `test/cs-ludicrous` (which pin the corrupt-wire refusals) included. The bench's own golden + per-variant round-trip gates pass at `corpus_id 6b213fbfa1a03a99`.
- **Negative control, run rather than assumed.** With `flatMaskLit` forced to emit no mask — the flat form's masking invariant deliberately broken — `make test` produces 65 failure lines. The gate that proves this form does see it.
- **Checks semantics preserved exactly.** Same inputs refused, same errors, on every path. The two consequences fusion always has (which error surfaces when a stream is both truncated and out of range; a failed run leaving the destination untouched) are named in `flat.go`, and are the ones Rust, Go, Java, Dart and JS-flat already carry.
- **Zero allocations on the hot path**, still 0 bytes/pass.
- **One commit per lever, one measurement per commit**, attribution never bundled.
- **Provenance**: `bench/run.sh`'s `schema commit` line now carries its branch, like every runtime line already did — a small fix for a failure mode this repo has had twice (a sweep that measured the wrong tree; a published result whose SHA was rebased away). Law-home in BENCH-STANDARD.md §3.5.
- The `bits` family rows in the committed pair are flagged in the CSVs themselves as NOT attributable to this change: `BitsBench.cs` touches no generated code, yet its rows move 11% across the halves at sub-3% spread. Process-level, an open question, not a result.

### DIFF
--- a/bench/BENCH-STANDARD.md
+++ b/bench/BENCH-STANDARD.md
@@ -987,6 +987,13 @@ benchmarks two different versions of the same C++ library and publishes both.
 > `msbuild -getItem:Compile`) before it is printed — **and a leg that cannot
 > prove its runtime path REFUSES (no rows) rather than reporting.**
 
+**The schema commit records its BRANCH too** (added 2026-09-01, alongside the
+runtime lines that already did). The failure it closes is the twin of the one
+below: a sweep that silently measured the wrong tree, and a CSV that could not
+have said so. A bare SHA also stops resolving the moment it is rebased away,
+which is how a published result came to carry a dead SHA; the branch name
+survives the rebase and says which tree was on the bench.
+
 This clause was earned on **2026-08-15**: `bench/rust/Cargo.toml`,
 `bench/go/go.mod` and `bench/cs/schemabench.csproj` hardcoded their runtime
 paths while `run.sh` recorded `$SERIALIZE_RS`/`GO`/`CS` in the preamble — so a
--- a/bench/results/2026-09-01-cscodec-after-arm64-macbook.csv
+++ b/bench/results/2026-09-01-cscodec-after-arm64-macbook.csv
@@ -1,0 +1,84 @@
+# ---------------------------------------------------------------------------
+# PR #208 (cs-word-codec) — the C# codec pass, AFTER (cs-word-codec, 28b771d) half.
+#
+# ONE LEG ONLY (--only cs): nothing else in the tree changed, so the eight
+# other legs would have re-measured main against main. The cpp/c reference
+# rows this pair is read against are the quiet-box sweep at
+# bench/results/2026-09-01-arm64-macbook.csv, same corpus_id
+# 6b213fbfa1a03a99, same box, same day.
+#
+# TWO SITTINGS, both committed, taken back to back in the same window with the
+# other half. The box was shared with another worker's timed elixir runs
+# (PR #207) for part of the session; the pair is reported as max AND median
+# per BENCH-STANDARD.md rather than as a single number.
+#
+# THE `bits` FAMILY ROWS ARE NOT ATTRIBUTABLE TO THIS CHANGE, said out loud
+# rather than left for a reader to trip over. bench/cs/src/BitsBench.cs
+# exercises serialize.cs's raw BitWriter/BitReader and NO generated code at
+# all; neither the emitter nor the runtime moved under it. Its rows still
+# differ across the halves (bitpacker write 1222 -> 1087 MB/s at 0.6-3% spread
+# in sitting 2), which can only be process-level: JIT method placement and
+# cache/branch state after a differently shaped bench_mixed pass in the same
+# process. Read the `gen` rows; the `bits` rows here are an open question, not
+# a result.
+# ---------------------------------------------------------------------------
+# ---- sitting 1 ----
+# schema bench results
+# date: 2026-08-31T23:47:21Z
+# host: macbook  arch: arm64  os: Darwin 25.6.0
+# cpu: Apple M2
+# build: Release
+# cpp compiler: Apple clang version 21.0.0 (clang-2100.1.1.101)
+# cpp flags: -O3 -DNDEBUG -DSERIALIZE_RELEASE -DBENCH_OPT="O3" -std=c++17 -Wall -Wextra -Werror -ffp-contract=off -fno-rtti -Igenerated/bench/cpp -I../serialize
+# c compiler: Apple clang version 21.0.0 (clang-2100.1.1.101)
+# c flags: -O3 -DNDEBUG -DBENCH_OPT="O3" -std=c99 -Wall -Wextra -Werror -Igenerated/bench/c -I../serialize.c (serialize.c compiled in, no LTO)
+# go: go version go1.27.0 darwin/arm64 (go run, default optimized build)
+# rust: cargo 1.98.0 (797e8a9bc 2026-08-05) (Homebrew); rustc 1.98.0 (88d9e12ae 2026-08-18) (Homebrew) (cargo run --release at opt-level 3, no LTO)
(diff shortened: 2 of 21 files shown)


gafferongames:
Adversarial review complete: **MERGE**.

The reviewer rebuilt main's compiler and ran a ~930,000-trial reflection differential against this branch's emitter (shipped corpus + a probe schema covering the 384-bit cap edges, chunk straddles, negative-min offsets, union arm shapes): zero wire-byte, verdict, bits-consumed, or error differences. The 65-failure mask negative control reproduced to the exact digit. Regeneration deterministic, no wire goldens moved, all three CSV stamps resolve, headline ratios recomputed from the committed CSVs and confirmed (1.59x/1.60x, 219% on the same-day cpp reference).

Four advisory findings, none wire-affecting, filed as follow-on issues after merge:
1. The #198 guard (1:1 piece↔item) is a property of code shape with no gate — a C# twin of `TestFlatFallbackReEmitsItemsNotPieces` is owed BEFORE lever F makes classification 1:N.
2. The fusion-consequence comment understates its class (wrong `const` / nonzero `reserved` also qualify, reproduced 35x on the shipped corpus with identical verdicts) + a dropped SPEC §4.3 citation in the flat read path.
3. `batchEntry`'s `b == 0` basis is false two ways (`TFixed`/128-bit score as scalars but delegate; `densityUnion` takes the heaviest arm's `b`, not max) — pre-existing miscount, now load-bearing for a ruling.
4. The zero-alloc "assert" prints to stderr and fails nothing — gate it or rename it.

Also noted: the batch-entry retraction from `BenchPacket`/`RegainAfterAlign`/`TestData` is a real scope change the body does not disclose; covered by land-and-expand but recorded here.

PR #223, opened 2026-09-01
SPEC-TABLES: the ruled model — tolerant wire, self-contained tables, reflection, relocatable and built wide

gafferongames:
The design half of today's tables feature. Rewrites SPEC-TABLES.md from the fenced/append-only/document draft to the model the owner ruled today (recorded verbatim in §11): versioning is wire tolerance (flatbuffers-class evolution by name-hash identity), `was =` rename aliasing + compile-time collision refusal close the two identity holes, tables are self-contained (root table = a user's format, no document concept), the wire is neutral (no schema packing opinions), reflection descriptors carry the tool/editor surface, and relocatability + measure/save parallel generation are load-bearing requirements. C++ backend today, others refused by name.

Draft until the implementation PR lands beside it — the two get reconciled at review (exact kind list, refusal wordings) and merge together.

### DIFF
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -50,11 +50,10 @@ table ShipConfig
 }
 ```
 
-A table body is a type body — the full field grammar of SPEC §4.2, hosted
-by `table`: bare and ranged integers, `bits(N)`, `bool`, floats and
-compressed floats, `fixed`/`ufixed`, enums, flags, strings, bytes, bounded
-arrays, unions, `if` branches, and declared types as field groups. Two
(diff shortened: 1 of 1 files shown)


gafferongames:
Scope note: the previous draft was "views and tables" — one document for two features. This rewrite is the TABLES spec only; the view (#157, generic reflection over type declarations) is no longer specified here. Tables carry their own reflection surface (§6), which covers the tool/editor need the owner named. If a general view over types resurfaces, it is its own design under #157, starting fresh against the current model.

PR #228, opened 2026-09-01
Tables: evolution-tolerant data declarations, C++ backend

gafferongames:
## The model

Tables are for data that outlives builds — tools-to-game pipelines, editors that walk properties by name, save/load to memory or disk — declared with `table`, carried on an evolution-tolerant TLV wire where field identity is a name hash, so any reader takes any data in both directions (unknown fields skip, absent fields default, changed kinds skip rather than misdecode, out-of-range values clamp, and every event is counted in a report). Packets stay exactly what they are: exact-match, bit-packed, protocol-id-gated — tables never enter the projection, never move the protocol id, and never change a generated packet byte.

Flatbuffers/protobuf ideas apply to tables, not to types: tables must go wide in generation (the encode surface is a measure/save split, so subtables can be measured in parallel, prefix-summed, and scatter-written from N workers) and must be relocatable (trivially copyable, standard-layout, pointer-free, arrays inline — enforced by generated `static_assert`s). **The exact-capacity guarantee**: for any value that saves at all, `TableSave(value, buf, TableMeasure(value))` succeeds and writes exactly that many bytes — elision (including all-default nested tables) is decided before any byte is emitted, so a scatter-writing worker can be handed exactly `sizes[i]` and nothing more. Tested corpus-wide, including the one-byte-short refusal. Contract ported from prior internal work; the `was` rename attribute and the measure/save split are new here.

## What is generated

For `--lang cpp`, a unit that declares tables gains one `<Base>Table.h` per schema file, holding for the whole table closure (every `table` plus everything one references, transitively):

- storage structs for the `table` declarations (closure `type`s reuse the structs from `<Base>.h`), with declared defaults as member initializers
- `TableMeasureX( const X & )` — exact encoded size, writes nothing
- `TableWriteX( TableWriter &, const X & )` / `TableSaveX( const X &, uint8_t * buffer, int64_t capacity )` — returns bytes written (== measure) or -1
- `TableReadX( const uint8_t *, int64_t, X &, TableReport & )` — prefill defaults in place, overlay, count events
- `TableTypeX()` reflection descriptors: name, wire id/kind, storage offset, array bounds and `_count`/`_length` companions, declared ranges, enum name functions, branch guards — enough to walk/print/diff/edit any table value with no RTTI and no schema files
- relocatability `static_assert`s per closure member
- zero allocation (tested), **no serialize dependency** — the test leg compiles without the serialize include path to prove it

Every other target refuses a unit that declares tables, by name ("tables are C++-only today; the <lang> table backend is a named follow-on").

## The wire, on one page

Neutral by ruling: plain little-endian scalars and length-prefixed regions — implementable by a third party from this description alone. None of the packet wire's density machinery (no bitpacking, no range compression) rides here; schema's opinionated declarations (ranged ints, compressed floats, bits(N), enums, flags) are declaration-side conveniences that validate/clamp on load and map to neutral kinds.

- A table body is a sequence of fields, terminated by `u16 0`.
- Field: `u16 id` (fold16(fnv1a32(name)), 0 rebounds to 1; the `was` alias's hash after a rename), `u8 kind`, payload.
- Kinds: 1 bool(u8) · 2–5 i8/i16/i32/i64 · 6–9 u8/u16/u32/u64 · 10/11 f32/f64 (bit pattern) · 12 string (`u16 len` + bytes) · 13 table (`u32 len` + nested body) · 14 array (`u32 body_len` + `u8 elem_kind` + `u16 count` + elements; table elements are each `u32`-length-prefixed; bytes(N) is an array of u8) · 15 union (`u8 tag`; tag 0 = None is the whole payload; else `u32 len` + selected arm as a nested body).
- Mapping: enums ride at their storage width (out-of-set → None, counted); flags ride u64; bits(N) rides the smallest u-kind with a width clamp; ranged ints ride at full storage width and clamp on load; compressed-float triples ride as plain f32 and clamp to [min, max]. No packing opinion leaks onto the wire (audited; none found in the ported contract either).
- Elision: a field at its declared default stays off the wire; an all-default table is 2 bytes; None unions, empty strings/arrays and all-default nested tables elide; guarded fields stay off the wire when the guard says so, and the reader's prefilled defaults stand in.

## Deviations from the ported contract

1. **`was = "old_name"`** (new): rename aliasing — the field's wire id stays the old name's hash. Refused: `was` naming the field's own current name, empty/non-string values, `was` outside table bodies (by name), and any two fields of one closure member whose effective ids collide (the prior collision refusal, extended over was-ids).
2. **Measure/save split** (owner-ruled): the prior encode surface was write-with-capacity only; `TableMeasureX` + `TableSaveX` are added so generation can go wide. Measure mirrors write's elision decisions exactly; the test asserts measure == save everywhere.
3. **Tables are invisible to the packet wire** (forced by this repo's structure): here a `type` cannot reference a `table`, a table is not a union payload, and tables live beside the decl stream (`Unit.Tables`/`File.Tables`), not in it — the packet backends, the projection and the protocol id never see them. The prior implementation let tables double as packet types; schema's packet emitters are byte-frozen, so tables stay a separate world.
4. **`const`/`reserved`/`align` refused in table bodies** (by name): they describe bit positions, and the table wire has no bit positions.
5. **string/bytes/array extents past 65535 refused in closures**: table-wire lengths and counts ride in u16 (schema's packet wire permits larger; the table wire does not).
6. **Table declarations take no qualification** (day one): no tags, no cpp_native mapping on tables themselves (closure `type`s keep theirs).
7. **Table.h renders folded literals** (no symbolic constant rendering): `string(MaxName)` declares fine; the header prints the folded bound. The packet headers keep their symbolic rendering.
8. **Codec prototypes precede definitions** (layout, not contract): declaration order in the schema is free.
9. **Cross-file include cycles among Table headers are refused by name** (the prior code would emit mutual includes).
10. Same-as-prior refusals, kept: `fixed(I,F)` and `int128`/`uint128` have no table-wire kind; arrays of unions are a named follow-on; self/cyclic table nesting is refused under the existing composition-cycle rule (tables join the cycle graph).
11. **Array elements are bounded by the field's body length** (deliberate deviation from the prior GENERATED reader, ruled by the tolerance model): the prior implementation's generated reader let a lying element count read past the field body into the following fields' bytes — fabricated values accepted silently, then the same bytes decoded again as fields. Its own hand-written generic decoder bounds elements to the body slice, and the ruled model ("changed kinds skip, NEVER misdecoded; malformed = stop + partial + flag") arbitrates the disagreement that way: elements decode through a sub-reader over the field body, a truncated element keeps the decoded prefix and flags `malformed`, and the parent continues at the next field. Probed by test with crafted body_len/count lies.
12. **Write and measure validate storage invariants** (restores prior-encoder behavior the first cut of this port dropped): a `_count`/`_length` outside its bound — negative included — fails the save (`TableMeasure` returns -1, `TableSave` returns -1) instead of reading out of the array onto the wire; an out-of-range union tag fails measure exactly as it fails write.

## Expressiveness gate: "Config.bin / Assets.bin in tables"

Surveyed the production config/asset declarations and loaders (read-only). Each structural need, verified in the compiled test leg:

| Structural need | How tables express it | Verified |
|---|---|---|
| Per-field defaults + ranges (`max_players int32 = MaxPlayers \| min = 1, max = MaxPlayers`) | Same spelling in table bodies; absent-field decode materializes the default, out-of-range clamps and counts | round-trip + evolution + clamp tests |
| Nested tables and bounded collections (`armor_config [..NumArmorLevels]ArmorConfig`, `colliders [..Max]Collider`) | Table-typed fields and `[..N]Table` arrays, const-named bounds; `string(MaxLevelNameLength)` const capacities | corpus `RootConfig`/`LoadoutConfig`, `TestTableRefusals` bounds cases |
| Enum/flags/float/bool/int fields (`ship_type ShipType`) | All native in table bodies; enums clamp out-of-set to None and surface name functions in descriptors | round-trip + reflection tests |
| Guarded optional groups (`has_gunner_settings bool` + `if … { gunner_settings GunnerSettings }`) | `if` guards in table bodies; untaken side stays off the wire, defaults stand in | guard test |
| The container: root envelope (magic + hash) holding per-collection table instances | **The worked answer:** declare a root table with bounded arrays of config tables — the DATA is schema's; the few-line magic/hash envelope stays user code by design (opinion-free, no document concept, no prescribed roots) | corpus `RootConfig` + USAGE's nesting section |
| The evolution doctrine (unknown skip, absent default, kind-change skip, clamp, warn loudly, only structural damage rejects) | This is the wire contract verbatim; the report carries the warn-loudly counts | both-directions evolution tests with exact counter asserts |

Named gap (follow-on, not implemented): the production declarations carry a `json = "name"` field attribute — a tool-side JSON authoring surface, not wire. It joins the follow-ons list.

## Test inventory

- `test/tables/main.cpp` (compiled + run in `make test`, **without** the serialize include path): corpus round trip over ALL 15 wire kinds — every scalar kind now rides and is pinned by descriptor (i8/i16/i32/i64, u8/u16/u32/u64, f32/f64, bool, string, table, array, union) — plus bits, enum, flags, fixed/counted arrays, nested/cross-file tables, guards; all-default = 2 bytes; measure == save everywhere; **exact-capacity suite** (save into a buffer of exactly measure's size succeeds and byte-matches, one-byte-short refuses — across all-default nested elision, guarded elision, deep nesting, fixed table arrays, both evolution shapes); **storage-invariant suite** (count above bound, negative count/length, deep nested violation, invalid union tag — measure and save both refuse); **bounded-element probes** (body_len/count lies yield empty or prefix-only arrays, malformed flagged, following fields decode intact — never fabricated elements); both-directions evolution (old reader × new data, new reader × old data) with exact unknown/kind_mismatch asserts; `was` rename identity in both directions; clamp (ranged int, bits width, over-long string) with counts; malformed framing (truncation) with partial-result keep; reflection walk (ids pinned against an independent C++ implementation of the hash, ranges, enum names, guards, nested descriptors, offset-based generic access); relocation by memcpy; the parallel measure/prefix-sum/scatter-write shape; too-small buffer refusal.
- `internal/check/tables_test.go`: 19 named refusals (see deviations above), table-runtime name claims, id function pins (frozen values), `was` id override, tables-invisible-to-packet-IR, and **adding/editing a table moves neither the projection nor the protocol id**.
- `compiler/tables_test.go`: all eight non-C++ targets (plus aliases) refuse by name; cpp emits Table headers only for table units; **non-table generated files are byte-identical with and without tables**; generated table code contains no allocation, no standard containers, no serialize include/symbols, and carries the codecs and relocatability asserts.
- `internal/check/diagnostics_test.go`: bare `was` (valued-attribute guard), table qualification refusal.

## Locked paths and protocol id, untouched

- No file under `internal/codegen/c/`, `internal/codegen/cpp/`, `generated/c/`, `generated/cpp/`, `generated/c-ludicrous/` is modified (`git diff --stat` on this branch shows none). The Table emitter is a new sibling package, `internal/codegen/cpptable/`, wired in `compiler/builtin.go`.
- `make test` regenerates every committed `generated/` tree; the tree is clean afterwards (the examples corpora declare no tables, and a table-free unit's output is byte-identical by test).
- Protocol-id independence is tested twice: projection/id equality in `internal/check/tables_test.go`, and non-table byte equality in `compiler/tables_test.go`.

## Follow-ons (named, not now)

- Per-language table backends (go, rust, c, cs, js, dart, java, elixir) — each currently a by-name refusal.
- `json = "name"` tool-side authoring attribute.
- Arrays of unions on table-closure paths.
- Symbolic constant rendering in Table headers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,7 @@ GO_SOURCES   := $(shell find cmd internal -name '*.go') go.mod
 SCHEMAS      := $(wildcard examples/*.schema)
 SCHEMAS128   := $(wildcard examples128/*.schema)
 SCHEMAS_BENCH := $(wildcard bench/corpus/*.schema)
+SCHEMAS_TABLES := $(wildcard tables/examples/*.schema)
 
 all: bin/schema
 
@@ -193,6 +194,24 @@ build/guard-generated/.stamp: bin/schema test/guard/Alpha.schema test/guard/Beta
 build/schema_test_guard: build/guard-generated/.stamp test/guard/main.cpp
 	@mkdir -p build
 	$(CXX) $(CXXFLAGS) -Ibuild/guard-generated test/guard/main.cpp -o $@
+
+# The tables corpus (SPEC-TABLES.md): the tabledemo unit plus the
+# two-generation evolution pair (tblv1/tblv2), generated at build time into
+# build/ — test-only, never part of the committed generated/ tree.
+build/tables-generated/.stamp: bin/schema $(SCHEMAS_TABLES) test/tables/V1.schema test/tables/V2.schema
+	@mkdir -p build/tables-generated
+	./bin/schema generate --lang cpp --out build/tables-generated/examples tables/examples
+	./bin/schema generate --lang cpp --out build/tables-generated/v1 test/tables/V1.schema
+	./bin/schema generate --lang cpp --out build/tables-generated/v2 test/tables/V2.schema
+	@touch $@
+
+# Deliberately compiled WITHOUT -I$(SERIALIZE): the generated Table headers
+# carry no serialize dependency, and this build proves it stays that way.
+build/schema_test_tables: build/tables-generated/.stamp test/tables/main.cpp
+	@mkdir -p build
+	$(CXX) -std=c++17 -Wall -Wextra -Werror -ffp-contract=off \
+		-Ibuild/tables-generated/examples -Ibuild/tables-generated/v1 -Ibuild/tables-generated/v2 \
+		test/tables/main.cpp -o $@
 
 build/schema_test_random: generated/cpp/.stamp test/random_main.cpp
 	@mkdir -p build
@@ -342,9 +361,10 @@ build/schema_test_c_ludicrous: generated/c-ludicrous/.stamp test/c-ludicrous/mai
 		-O2 -ffp-contract=off -Igenerated/c-ludicrous -I$(SERIALIZE_C) \
 		test/c-ludicrous/main.c $(SERIALIZE_C)/serialize.c -o $@ -lm
 
-test: build/schema_test build/schema_test_guard build/schema_test_random build/schema_test_ludicrous build/schema_test_c build/schema_test_c_ludicrous build/schema_test_bench build/schema_test_bench_c generated/go/.stamp generated/rust/.stamp generated/cs/.stamp generated/js/.stamp generated/dart/.stamp generated/java/.stamp generated/elixir/.stamp generated/go-ludicrous/.stamp generated/rust-ludicrous/.stamp generated/cs-ludicrous/.stamp generated/js-ludicrous/.stamp generated/dart-ludicrous/.stamp generated/java-ludicrous/.stamp generated/elixir-ludicrous/.stamp generated/bench/go/.stamp generated/bench/rust/.stamp generated/bench/cs/.stamp generated/bench/js/.stamp generated/bench/dart/.stamp generated/bench/java/.stamp generated/bench/elixir/.stamp build/java-test/.stamp build/java-test-ludicrous/.stamp build/java-bench/.stamp
+test: build/schema_test build/schema_test_guard build/schema_test_tables build/schema_test_random build/schema_test_ludicrous build/schema_test_c build/schema_test_c_ludicrous build/schema_test_bench build/schema_test_bench_c generated/go/.stamp generated/rust/.stamp generated/cs/.stamp generated/js/.stamp generated/dart/.stamp generated/java/.stamp generated/elixir/.stamp generated/go-ludicrous/.stamp generated/rust-ludicrous/.stamp generated/cs-ludicrous/.stamp generated/js-ludicrous/.stamp generated/dart-ludicrous/.stamp generated/java-ludicrous/.stamp generated/elixir-ludicrous/.stamp generated/bench/go/.stamp generated/bench/rust/.stamp generated/bench/cs/.stamp generated/bench/js/.stamp generated/bench/dart/.stamp generated/bench/java/.stamp generated/bench/elixir/.stamp build/java-test/.stamp build/java-test-ludicrous/.stamp build/java-bench/.stamp
 	./build/schema_test
 	./build/schema_test_guard
+	./build/schema_test_tables
 	./build/schema_test_random
 	./build/schema_test_ludicrous
 	cd test/c && ../../build/schema_test_c
@@ -430,6 +450,9 @@ generated-current: test
 check: bin/schema
 	./bin/schema check examples
 	./bin/schema check examples128
+	./bin/schema check tables/examples
+	./bin/schema check test/tables/V1.schema
+	./bin/schema check test/tables/V2.schema
 	./bin/schema check bench/corpus/Bench.schema
 	./bin/schema check bench/corpus/RealWorld.schema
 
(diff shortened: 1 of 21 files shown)


gafferongames:
Adversarial review complete across two passes: initial BLOCK's real findings (exact-capacity save with elided nested tables, unbounded array-element decode, unvalidated storage invariants, kind coverage) all repaired and re-verified by the reviewer's own probes — including lying lengths at every nesting level, element bodies overrunning their array body, and full truncation batteries. The element-bounding fix is a declared deviation from the prior generated reader, arbitrated by the owner's tolerance model with the prior generic decoder as precedent (upstream note filed). Wire byte-verified against hand-derived goldens; id derivation pinned; evolution proven both directions with count and value assertions; ProtocolId independence and locked paths verified untouched. Verdict: MERGE.

PR #243, opened 2026-09-01
Tables: pointer semantics — derived modes, the arena lifecycle, and the cooked form

gafferongames:
Tables gain pointer semantics. The founding line, in the owner's words: *"types
can remain value semantics. tables should ALLOW pointer semantics"* — *"we
can't be a generic system if we don't have pointers to tables."*

`next *Node` declares a pointer to a table. The compiler DERIVES what follows
and nobody declares it: a table with no pointer in its by-value closure stays a
plain struct and pays nothing; a table with one is built through a lock-free
arena, locked once into a single packed region, and read through a root
pointer. That region also cooks — written verbatim behind a build-locked header
so a later run points at its root without copying or parsing it — while the
tolerant wire stays the format of record.

---

## Every element, labelled

**OWNER-RULED**

- Pointer semantics for tables, and the C-like spelling `next *Node`.
- The mode is COMPILER-DERIVED, never declared: *"i wouldn't want to manually
  have to specify this… the compiler can work it out."*
- Const-vs-mutable is a LIFECYCLE, not a schema marking: *"maybe you 'lock' a
  table and it is constant from that point forward."*
- The two forms are genuinely different at runtime; neither is contorted to
  look like the other.
- The builder is multithreaded, and *"I prefer lockless if possible."*
- The generated surface, verbatim from the approved snippet: struct with a
  clean name, `<Name>Measure` / `<Name>Save` / `<Name>Load` free functions,
  `<Name>Builder` with `Alloc<T>()` and `Lock()` as methods.
- Tables and types share one symbol table — *"i vote yes."*
- The zero-cost gate: *"make sure we don't pay this cost when we have nice,
  small messages… when the table is inferred to be value types only."*
- No generated size constants — *"you don't need to recreate sizeof."*
- Pointer-bearing tables are never held by value; file-format-scale tables
  *"can't be a struct."*
- A pointed-at region form exists at all: load a big file, point at the root,
  no copy, no full parse.

**COORDINATOR-DESIGNED** (my calls; veto any of them)

- The **segmented slab arena** as the builder's storage, chosen under the
  lockless preference. SPEC §14 weighs all four models considered.
- **Cook/Open** as the region form's verbs — grounded in the game-industry
  vocabulary the owner used, but the naming choice is mine.
- **Pointer-to-`type` refused by name.** Inferred from "types remain value
  semantics" rather than quoted — the owner never ruled on the spelling
  `*SomeType`, and this is my reading of what the split implies.
- **Self-relative references inside a region**, so a deref is one add with no
  base pointer and a region relocates by pure `memcpy` with zero fix-up.
- **A high-water mark rather than a visited set** in `Open`'s walk.
- **Only pointer edges charge depth**; by-value nesting charges none.
- **Wire v1 is a tree**: aliasing is not preserved. Loud in the spec, in the
  generated comments, and asserted by a test rather than hidden.
- **A depth cap (128) instead of cycle detection**, on every walk.
- The **layout id** construction: a compile-time digest of the schema's
  packed-layout facts — fields keyed by WIRE ID, so a `was` rename holds —
  mixed with this build's own `sizeof` and per-field `offsetof`, so schema
  drift and ABI drift both refuse.
- Argument orders and the exact spellings of `SaveBody` / `LoadBody` /
  `LoadMeasure` / `OpenWalk`.

**FOLLOW-ONS, filed not built** (SPEC §15)

Graph and DAG identity; arrays of pointers; lifting the depth cap with a flat
indexed node encoding; cross-endian `Open` (v1 REFUSES a foreign byte order —
the descriptor-driven fix-up pass is filed with its mechanism sketched); a
hash-guarded fallback loader; and a generic JSON walk over the reflection
descriptors, which is one implementation over the descriptor surface rather
than per-table codecs.

---

## The zero-cost gate — stated at the grain it holds

The owner's criterion: a value-only table gets none of the new machinery. That
is true at two different grains, and the earlier draft of this section blurred
them. Precisely:

**Per UNIT.** A unit whose tables are all fixed-size emits nothing new at all —
no arena runtime, no reference type, no builder, no `<atomic>`, and not one
extra descriptor column. Proof: the entire pointer-free corpus
(`tables/examples`, `test/tables/V1.schema`, `test/tables/V2.schema`)
regenerates **byte-identical** to a rename-applied baseline built at commit
`974ff0f`. `diff -r` over both trees: **clean, zero differences.** Closing that
found three real leaks — the `is_pointer` column, the `variable` column, and a
reworded comment — all now gated on the unit having pointers.

**Per TABLE.** Inside a unit that DOES have pointers, a value-only table gets
**no Builder, no pointer-graph walkers, and codecs character-for-character
identical to the pointer-free ones** — no ctx parameter, no depth parameter, no
template. What it shares with the unit are the per-UNIT facts above: the arena
runtime and the two descriptor columns are emitted once per header and belong
to the unit, not to any table in it. (It does gain one thing: an `Open` bounds
walk, because `Open` must bound its count companions — see C2 below.)

**Standing gates**, so both stay true:

- `make tables-zero-cost` (inside `make test`) greps every pointer-free
  corpus header — now including `P1` — for every pointer-machinery symbol and
  fails the build on a hit.
- `TestZeroCostForValueOnlyTables` pins the per-UNIT property.
- `TestPointerSurfaceEmitted` pins the per-TABLE property, asserting exactly
  the list above rather than more than it can hold.

---

## The design

**Mutable: a segmented slab arena.** Equal-size 4 MiB segments whose count
saturates the u32 reference space exactly (1024 → 4 GiB), handed to workers in
64 KiB slabs with one compare-exchange each. No atomic on the per-node path, no
lock anywhere. Nodes are born at final offsets and a segment never moves, so a
`T*` from `Alloc` stays valid while other workers allocate and while the arena
grows.

The rejected model is named in the generated header and in SPEC §14: one buffer
under a lock, grown by `realloc`. A `realloc` moves the buffer under a worker
mid-write; offsets fix identity but not the references already resolved from
them, and the corruption surfaces much later. Segments never move, so that bug
cannot be written here. Slack is bounded and stated: one slab tail per worker
plus one per segment.

**`Lock()` is one way and it IS the compaction.** The arena packs into a single
exact region, root at base, every reference rewritten self-relative. Lock's
output and Load's output are the same representation, read through one view
API. No unlock: re-editing means loading into a fresh builder, which is a copy
and says so.

**Threading contract, stated in the header:** allocating on your own worker is
safe concurrently; writing fields of a node another worker allocated is your
own synchronization; `Lock`/`Save`/`Cook`/`Open` are single-threaded, after the
join.

**On the wire**, a pointer rides as its pointee's table body — framing
*identical* to a by-value nesting, which is why a field can change between the
two without moving a byte (there is a both-directions evolution test for
exactly that). Null elides. A non-null pointer rides even when its pointee is
all-default, or null and empty would be one value.

**The cooked form** is an accelerator, not an archive. `<Name>Open` validates
before it points: a bounds walk over the reference graph and the counts that
bound a traversal of it, never a field value.

That walk is **linear in the region**, and it takes a proof rather than an
assertion — the first draft said "O(references)" and was wrong. Forward,
in-range and aligned are not enough: a forged file whose references alias
forward is a legal-looking DAG that a stateless walk explores exponentially.
The walk carries a **high-water mark**. Packing lays nodes out in pre-order by
bump allocation and the walk follows that order, so a genuine region always
satisfies `at >= watermark`; requiring it and advancing past each node makes
the walk consume region bytes monotonically — it visits each byte at most once,
it terminates, and it rejects mid-node overlaps too. Not a visited set, which
would allocate in proportion to the graph. One integer buys all three. The pack
order and the walk order are now one invariant, stated at both sites. No
trust-mode bypass.

---

## Test inventory

Twelve new cases in the tables leg (`test/tables/main.cpp`), plus the corpus
and Go-side additions:

| Case | What it holds |
|---|---|
| `test_pointer_lifecycle` | build → measure → save → Lock → relocate by memcpy → LoadMeasure → Load → re-save byte-identical |
| **exact-capacity battery** | save into exactly `Measure`'s answer succeeds and byte-matches a roomy save; one byte short refuses — now across a **pointer graph**, extending the #228-F4 battery |
| `test_lock_layout_stable` | two independent builds `memcmp` equal; two cooks byte-identical |
| `test_pointer_alias` | two references to one node are two nodes — in the packed form AND across the wire |
| `test_pointer_cycle_refused` | a self-pointing node: Measure, Save, CookMeasure and Lock all refuse; nothing recurses |
| `test_pointer_depth_cap` | a chain at the cap saves; one link past it refuses everywhere |
| `test_builder_grow` | 200 000 allocations across many slabs and segments; a `T*` taken first is still valid last |
| `test_builder_workers` | four workers single-threaded (deterministic) **plus a real four-thread concurrent build**, joined before Lock |
| `test_cooked_form` | cook → open → point; cook-open-cook stable; **nine refusals**: magic, layout id, truncation, sub-header size, unaligned base, out-of-region reference, backward reference, misaligned reference, out-of-bound count |
| `test_pointer_evolution_old_reader_new_data` | P1 (by-value) reads P2 (pointered): first node decodes, the chain's tail is one unknown |
| `test_pointer_evolution_new_reader_old_data` | P2 reads P1: one node allocated from the by-value body, `next` null — plus load-into-builder, edit, re-Lock |
| `test_pointer_null_and_empty` | a null pointer elides; a non-null pointer at an all-default node does not — the distinction survives the round trip |
| `test_pointer_reflection` | derived mode surfaced, `is_pointer`, the target descriptor, a self-referential pointer resolving to its own type |

Seven more from the adversarial round, each the reviewer's own repro reduced to
an assertion:

| Regression | Finding it guards |
|---|---|
| `test_open_walk_is_linear` | B1 — a forged aliased-forward DAG at n = 16/26/40/64 is refused; n = 64 would not return unbounded |
| `test_open_refuses_overlap` | B1 — an 8-aligned reference landing mid-node |
| `test_lock_deterministic_on_dirty_heap` | B2 — heap dirtied 0xA5 then 0x5C between builds; identical regions AND cooked files, zero padding |
| `test_depth_agrees_through_by_value_nesting` | C1 — the 127-chain through a by-value nested variable table round-trips all four walks |
| `test_open_bounds_nested_fixed_companions` | C2 — a by-value nested fixed table's count, and one reached through a pointer |
| `test_open_refuses_dirty_reserved` | the header's reserved words, all 20 bytes swept |
| `test_descriptors_are_constant` | C3 — the surface read from four threads, no first-use state |
| `test_cross_file_pointer_unit` | R1 — a multi-file pointered unit: a native-mapped type, a fixed table and a variable table nested across file boundaries, round-tripped and forged |

Go-side: `TestTableModeDerivation` (six shapes, propagation up by-value edges
only), `TestPointerRecursionIsLegal`, eight new refusal cases,
`TestZeroCostForValueOnlyTables`, `TestPointerSurfaceEmitted`,
`TestPointerGenerationDeterministic`, `TestLayoutIdMovesWithTheSchema`.

Corpus: `tables/pointers/Graph.schema` (a value-only table, a value-only
pointer target, a self-recursive list, a tree, a variable table nested by
value, a bounded array of them) in its **own unit**, so `tables/examples` stays
pointer-free and keeps proving the gate; `test/tables/P1|P2.schema` for the
pointer evolution pair.

**Negative control run:** inverting the aliasing assertion turns the tables leg
red at the expected line and exits 1. The battery is wired in, not decorative.

---

## Adversarial round: disposition per finding

All BLOCKING and SHOULD-FIX findings are fixed in this branch; the reviewer's
batteries live at `scratchpad/adv/`.

| # | Finding | Disposition |
|---|---|---|
| B1 | `Open`'s walk exponential on a forged aliased DAG (26 nodes = 312 ms) | FIXED — high-water mark; invariant stated at both Pack and OpenWalk with "neither may change alone" |
| B2 | Lock/Cook copied uninitialised arena padding into regions and cooked files | FIXED — segments `calloc`'d; the determinism test now dirties the heap first |
| B3a | `table Root` / `table Const` emitted non-compiling headers | FIXED — methods moved to `GetRoot`/`AsConst`; every other builder member name refused as a table name |
| B3b | six generated spellings missing from the claimed-name registry | FIXED — plus `TableInfo`/`TableFields` from the C3 fix; a refusal test each |
| C1 | the four walks spent depth differently | FIXED — only pointer edges charge depth, one rule in one place |
| C2 | `Open` never bounded nested fixed tables' companions | FIXED — every closure member gets an Open walk; union tags bounded too |
| C3 | TSAN race in the descriptors' lazy link | FIXED — constant-initialised namespace data, no first-use state; TSAN clean |
| C4 | layout id cried wolf on `was`, missed a moved member | FIXED — keyed by wire id, mixes per-field `offsetof` |
| R1 | *(introduced by C2)* a multi-file pointered unit emitted a non-compiling header | FIXED — a member's Open walk is emitted by the file that DECLARES it, keyed on the UNIT having pointers rather than the file |

**Negative controls run on the two subtle ones.** Restoring `malloc` turns the
determinism regression red on both assertions. Removing the watermark makes the
forged DAG **ACCEPTED** again at n = 16 and 24 — the defect returns exactly
where the test looks for it.

**The alloc path was already TSAN-clean and stays so**: the race was in the
reflection surface, not in allocation.

**R1, and the corpus gap that hid it.** Widening `Open`'s descend-set (C2) was
right; emitting the walks per FILE was not — a pointered unit whose second file
declares neither a variable table nor a pointer target named walks that existed
nowhere. `tables/pointers` is now a genuinely multi-file unit: `Parts.schema`
holds a native-mapped type and a fixed table and is deliberately both
pointer-free AND pointer-target free (the first attempt at this corpus missed
the bug because its fixed table *was* a pointer target, which put it in the
walker set anyway); `Marks.schema` holds the cross-file variable member;
`Album` nests all three across file boundaries. Since Album's `Colour` storage
spells `::ColourMath`, the cross-file native mapping is covered too — the walk
is reached through a derived-to-base conversion.
`test_cross_file_pointer_unit` round-trips it and forges three cross-file
corruptions. The byte-identity proof was **re-run after this fix and is still
clean**.

Re-running the reviewer's own battery: **240 126 checks, one remaining
"failure"** — their `CHECK` that garbage in the header's reserved words still
opens. That was minor 3; it is now refused, so their assertion of the old
behaviour is what fails.

---

## Gates

All run locally, all green:

- `make test` — **exit 0**, all nine backend legs plus the tables leg, **zero
  warnings** from the generated modules. `make tables-zero-cost` runs inside it.
- `gofmt -l` clean · `go vet ./...` clean · `golangci-lint run` **0 issues** ·
  `modernize` clean · `make shape-gate` clean (19 ledgered files, no new shape
  knowledge).
- Locked paths untouched: `internal/codegen/{c,cpp}/`,
  `generated/{c,cpp,c-ludicrous}/`.

`golangci-lint` and `modernize` each found real things across the two rounds (an
unused helper, a range-over-int, an if-else chain, a `slices.Contains` loop);
all fixed.

---

Draft on purpose. `ROUND-LOG.md` on the branch carries the unit-by-unit record.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,7 +217,8 @@ build/tables-generated/.stamp: bin/schema $(SCHEMAS_TABLES) $(SCHEMAS_TABLES_POI
 # is recorded in the round log; this is the standing gate.)
 .PHONY: tables-zero-cost
 tables-zero-cost: build/tables-generated/.stamp
-	@for f in build/tables-generated/examples/*Table.h build/tables-generated/v1/*Table.h build/tables-generated/v2/*Table.h; do \
+	@for f in build/tables-generated/examples/*Table.h build/tables-generated/v1/*Table.h \
+	          build/tables-generated/v2/*Table.h build/tables-generated/p1/*Table.h; do \
 		if grep -nE "TableArena|TableSlot|TableWorker|TableRef|TableRegion|kTableSegment|kTableSlab|kTableMaxDepth|is_pointer|Builder|LayoutId|OpenWalk|PackMeasure|LoadMeasure|Cook|Open\\(" $$f; then \
 			echo "ZERO-COST GATE FAILED: pointer machinery leaked into $$f"; exit 1; \
 		fi; \
--- a/Makefile
+++ b/Makefile
@@ -230,7 +230,7 @@ tables-zero-cost: build/tables-generated/.stamp
 build/schema_test_tables: build/tables-generated/.stamp test/tables/main.cpp
 	@mkdir -p build
 	$(CXX) -std=c++17 -Wall -Wextra -Werror -ffp-contract=off -pthread \
-		-Ibuild/tables-generated/examples -Ibuild/tables-generated/pointers \
+		-Ibuild/tables-generated/examples -Ibuild/tables-generated/pointers -Itest/tables \
 		-Ibuild/tables-generated/v1 -Ibuild/tables-generated/v2 \
 		-Ibuild/tables-generated/p1 -Ibuild/tables-generated/p2 \
 		test/tables/main.cpp -o $@
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,7 @@ SCHEMAS      := $(wildcard examples/*.schema)
 SCHEMAS128   := $(wildcard examples128/*.schema)
 SCHEMAS_BENCH := $(wildcard bench/corpus/*.schema)
 SCHEMAS_TABLES := $(wildcard tables/examples/*.schema)
+SCHEMAS_TABLES_POINTERS := $(wildcard tables/pointers/*.schema)
 
 all: bin/schema
 
@@ -198,19 +199,39 @@ build/schema_test_guard: build/guard-generated/.stamp test/guard/main.cpp
 # The tables corpus (SPEC-TABLES.md): the tabledemo unit plus the
 # two-generation evolution pair (tblv1/tblv2), generated at build time into
 # build/ — test-only, never part of the committed generated/ tree.
-build/tables-generated/.stamp: bin/schema $(SCHEMAS_TABLES) test/tables/V1.schema test/tables/V2.schema
+build/tables-generated/.stamp: bin/schema $(SCHEMAS_TABLES) $(SCHEMAS_TABLES_POINTERS) test/tables/V1.schema test/tables/V2.schema test/tables/P1.schema test/tables/P2.schema
 	@mkdir -p build/tables-generated
 	./bin/schema generate --lang cpp --out build/tables-generated/examples tables/examples
+	./bin/schema generate --lang cpp --out build/tables-generated/pointers tables/pointers
 	./bin/schema generate --lang cpp --out build/tables-generated/v1 test/tables/V1.schema
 	./bin/schema generate --lang cpp --out build/tables-generated/v2 test/tables/V2.schema
+	./bin/schema generate --lang cpp --out build/tables-generated/p1 test/tables/P1.schema
+	./bin/schema generate --lang cpp --out build/tables-generated/p2 test/tables/P2.schema
 	@touch $@
+
+# The ZERO-COST GATE (SPEC-TABLES.md): a table with no pointer in its by-value
+# closure must pay NOTHING for the pointer machinery — no builder, no arena, no
+# handles, no lifecycle surface, no extra descriptor columns. The pointer-free
+# corpus's generated headers must not contain one symbol of it. (The stronger
+# one-time proof — byte-identical emission against the pre-pointer baseline —
+# is recorded in the round log; this is the standing gate.)
+.PHONY: tables-zero-cost
+tables-zero-cost: build/tables-generated/.stamp
+	@for f in build/tables-generated/examples/*Table.h build/tables-generated/v1/*Table.h build/tables-generated/v2/*Table.h; do \
+		if grep -nE "TableArena|TableSlot|TableWorker|TableRef|TableRegion|kTableSegment|kTableSlab|kTableMaxDepth|is_pointer|Builder|LayoutId|OpenWalk|PackMeasure|LoadMeasure|Cook|Open\\(" $$f; then \
+			echo "ZERO-COST GATE FAILED: pointer machinery leaked into $$f"; exit 1; \
+		fi; \
+	done
+	@echo "tables zero-cost gate: value-only tables carry no pointer machinery"
 
 # Deliberately compiled WITHOUT -I$(SERIALIZE): the generated Table headers
 # carry no serialize dependency, and this build proves it stays that way.
 build/schema_test_tables: build/tables-generated/.stamp test/tables/main.cpp
 	@mkdir -p build
-	$(CXX) -std=c++17 -Wall -Wextra -Werror -ffp-contract=off \
-		-Ibuild/tables-generated/examples -Ibuild/tables-generated/v1 -Ibuild/tables-generated/v2 \
+	$(CXX) -std=c++17 -Wall -Wextra -Werror -ffp-contract=off -pthread \
+		-Ibuild/tables-generated/examples -Ibuild/tables-generated/pointers \
+		-Ibuild/tables-generated/v1 -Ibuild/tables-generated/v2 \
+		-Ibuild/tables-generated/p1 -Ibuild/tables-generated/p2 \
 		test/tables/main.cpp -o $@
 
 build/schema_test_random: generated/cpp/.stamp test/random_main.cpp
@@ -365,6 +386,7 @@ test: build/schema_test build/schema_test_guard build/schema_test_tables build/s
 	./build/schema_test
 	./build/schema_test_guard
 	./build/schema_test_tables
+	$(MAKE) tables-zero-cost
 	./build/schema_test_random
 	./build/schema_test_ludicrous
 	cd test/c && ../../build/schema_test_c
@@ -451,8 +473,11 @@ check: bin/schema
 	./bin/schema check examples
 	./bin/schema check examples128
 	./bin/schema check tables/examples
+	./bin/schema check tables/pointers
 	./bin/schema check test/tables/V1.schema
 	./bin/schema check test/tables/V2.schema
+	./bin/schema check test/tables/P1.schema
+	./bin/schema check test/tables/P2.schema
 	./bin/schema check bench/corpus/Bench.schema
 	./bin/schema check bench/corpus/RealWorld.schema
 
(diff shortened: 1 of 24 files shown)


gafferongames:
Adversarially reviewed across three passes. Initial verdict found three real blockers — an exponential Open walk on forged aliased-forward cooked files (killed by a high-water mark whose correctness is the stated Pack/OpenWalk pre-order invariant, negative-control-proven), uninitialised arena padding reaching cooked files on disk (calloc + a determinism test that dirties the heap first), and legal schemas emitting non-compiling headers (builder-member and generated-verb name collisions, registry completed) — plus depth-accounting unification across all four walks, Open bounding nested fixed tables' companions, descriptors moved to constant-initialised data (TSAN-clean, no lazy link exists to race), and the layout id rekeyed by wire id + offsetof so a was-rename holds while any moved byte refuses. The repair pass introduced one regression (cross-file by-value members in multi-file pointered units), caught by the reviewer's own repro, fixed unit-keyed with the double-definition case verified ODR-clean, and the corpus gained the exact multi-file shape that had been missing. Final pass: both reviewer repros compile and run end-to-end, zero-cost byte-identity re-verified against the reviewer's own baseline, 240k-check battery at zero failures, all gates green. Every design element labeled owner-ruled or coordinator-designed, audited for accuracy. Verdict: MERGE.

PR #254, opened 2026-09-02
Table wire: name-hashed enums and union arms, u32 lengths and counts, the docs pass (#249 #250 #252)

### LINKED ISSUE #249: Table wire: enum values and union arms ride by NAME HASH; flags stay positional under an append-at-end rule
Decided 2026-09-02 on the owner's "or what you want, if you have a better idea" after his append-only ruling. The decision and the reasons are on the record here; the owner may overrule.

**Enums and union arms: identity by name, on the TABLE wire only.** An enum value rides as the u16 fnv1a32 xor-folded hash of the variant NAME (the same fold fields use, §5, 0 mapped to 1). A union body opens with the u16 arm id (0 = empty) then the arm body. On load, an id the reader cannot name → None for an enum, empty for a union, counted in the report as `unknown`. The compiler refuses id collisions among one enum's variants and among one union's arms, naming both. Reflection descriptors carry the id per variant. Types keep ordinals: same-build wire, nothing moves there.

**Why hash and not append-at-end.** (1) It is the wire's own principle applied to its last two positional places: §4's "add anywhere, remove freely, reorder freely" becomes true of variants. (2) It is enforced by construction; append-at-end is a rule the compiler cannot check because it retains no history, and a violation is silent. (3) Cost is one byte per enum value on a byte wire, paid only for non-default values (elision, §3).

**Flags stay positional.** A flags value rides as its unsigned storage (a u64 of bits); there is no cheap name-identified form. Rule on the page (§4, §5, USAGE): flags variants are appended at the END only; inserting or reordering remaps stored files silently. This is the one ordinal that keeps the owner's append-only rule.

**Scope.** SPEC-TABLES.md §3 (wire), §4 (events), §5 (identity: the variant collision refusal), §11; cpptable codecs and descriptors; checker: variant/arm id collisions; corpus: enum insert-in-the-middle and union insert-in-the-middle, both directions, plus the flags rule as a documented case; negative control: revert to ordinal and show the remap goes undetected. The table wire moves; v2.3.0/v2.4.0 table files do not survive it, stated in the PR body and the v2.5.0 notes (schema#248).

One PR with schema#250 and schema#252.

### LINKED ISSUE #250: Table wire: u32 string lengths and array counts — no 16-bit ceilings anywhere on the wire
Owner ruling, 2026-09-02 (Glenn, live, on the tables language review item 5): "we should not limit at 16 bits."

**Today.** §3: strings carry a u16 byte length; an array body opens with kind (u8) and count (u16); extents past 65535 are refused by name (§2.2, §11). Body lengths are already u32.

**The change.** String byte length → u32. Array count → u32. The extent refusal at 65535 goes; the new ceiling is the u32 range and the checker refuses only extents that cannot be counted or measured in int64. Storage-side `_length`/`_count` companions are already int32; verify nothing in the C++ storage or the reflection descriptors (array bound, count companion) truncates. Measure/save byte-exactness battery must still hold (§9). Fuzz/hostile: a u32 count the body cannot cover yields the bounded prefix and malformed, never a fabricated value (§4 already says this; the test must cover the wide case).

**Scope.** SPEC-TABLES.md §2.2, §3, §11; cpptable codecs; corpus gains an extent past 65535; the wire moves (see schema#A).

One wire edit with schema#A if one worker takes both.

### LINKED ISSUE #252: Tables docs: the default is wire contract, teach if for optional and * for identity, and close the SPEC.md grammar gap
From the tables language review, 2026-09-02. Owner accepted ("Rest seems fine").

- **The default is part of the wire contract.** A field at its declared default is elided (§3), so changing a default rewrites history: a v1 writer elided `damage = 21.0`; a v2 reader with `damage = 25.0` loads 25.0 from that file, and no report event fires. Keep elision; add the rule to §4 and §5 in one sentence each: a default change is a semantic edit to every stored file, and `was` does not cover it. USAGE.md gets the same sentence beside the elision paragraph.
- **Optional is `if`, identity is `*`.** USAGE.md and SPEC-TABLES.md §2.1 teach `settings *Settings // an optional subtable`; one pointer flips the parent to variable-length and the whole builder lifecycle. Teach the guarded field (`has_x bool / if has_x { x X }`) as the spelling for an optional section, and `*` for recursion, sharing (after schema#C) and size. Reword the two examples.
- **SPEC.md §4.2 grammar** has no `TableDecl` production, no `"*" ident` in `Scalar` (table bodies only), and no `was` under Attributes. Add them; `table` is already in the reserved list.
- **§11 names the collision rule but not the claimed suffixes.** List the 24 name-first suffixes `tableGeneratedVerbs` claims (internal/check/check.go) on the page, so a user meets the rule before the refusal.
- **The cross-file cycle refusal** is justified by C++ headers including each other. Keep the rule; give it a language-neutral statement (a unit's table files form a DAG by reference) so the port does not inherit a C++-only reason.

Docs-only; can ride with schema#A's PR or land alone. Gate item for schema#248.

gafferongames:
Closes #249. Closes #250. Closes #252.

One wire edit and its docs pass. **The TABLE wire moves. The types wire does not** — no `ProtocolId`, no projection token and no generated `type` byte changes here, which is §10's independence held by `TestTablesMoveNoProtocolId` (green) and, mechanically, by the committed `generated/` tree coming out of `make test` byte-unchanged.

## What moved on the wire

| | before | after |
|---|---|---|
| enum value | raw ordinal, kind = the enum's storage width (u8/u16/u32/u64) | **u16 hash of the VARIANT NAME**, kind `7` (u16) always; `None` = `0` |
| union body | `u8 tag` (ordinal + 1, 0 = empty), then `u32 len` + arm | **`u16 arm id`** (hash of the arm's NAME, 0 = empty), then `u32 len` + arm |
| string payload | `u16` byte length + bytes | **`u32`** byte length + bytes |
| array body header | `elem kind (u8)`, `count (u16)` | `elem kind (u8)`, **`count (u32)`** |
| `bytes(N)` | array of u8 with the u16 count | array of u8 with the **u32** count |
| extents | refused past 65535 by name | **no wire ceiling** — the language's own int32 storage cap is the only limit |

Unchanged: field ids, the u16 zero terminator, the kind numbers, the u32 body lengths on table/array/union, flags (raw u64, positional bits), and pointer framing.

**v2.3.0 / v2.4.0 table files do not survive this.** Every enum value, every union arm, every string and every array on a stored table file is framed differently; there is no compatibility path and none is offered. The types wire is untouched, so no peer is forced into a redeploy by this.

## Identity, and the refusals that keep it sound

One fold now serves three vocabularies — field names, enum variants, union arms — `fold16(fnv1a32(name))` with 0 rebounding to 1, so `0` stays reserved for the terminator, `None`, and the empty arm. New refusals, scoped to the TABLE CLOSURE (the packet wire keeps its ordinals untouched):

- two variants of one enum, or two arms of one union, whose ids collide — both named;
- `| max = K` headroom on an enum in a table closure: a headroom value is reserved by number and named by nothing, so it has no identity to ride under, and the table wire needs no headroom because a variant may be added anywhere.

**Flags stay positional**, and the append-at-the-end-only rule is now on the page in SPEC-TABLES §4 and §5 and in USAGE, with the worked wrong/right pair from #249. A test pins it: kind `9`, no variant id, the mask itself on the wire.

## Docs (#252)

- **The default is wire contract** — one paragraph in SPEC-TABLES §4, cited from §3 and §5, and beside USAGE's elision paragraph: an elided field means the reader's declared default, so a default change is a semantic edit to every stored file and `was` does not cover it.
- **`if has_x { x X }` is the spelling for an optional section; `*` is for recursion, sharing and size.** Both `settings *Settings // an optional subtable` examples reworded (SPEC-TABLES §2.1, USAGE).
- **SPEC §4.2 grammar**: `TableDecl` added to `Declaration`, `"*" ident` added to `Scalar` (noted table-bodies-only), `was` documented under Attributes.
- **SPEC-TABLES §11** lists the generated suffixes `tableGeneratedVerbs` claims, so a user meets the collision rule before the refusal.
- **The cross-file cycle refusal** is restated language-neutrally — a unit's table files form a DAG by reference — with the C++ include consequence kept as the C++ note.
- **§3 is now implementable from itself**: kind numbers by value, one payload row per kind, the skip rule per kind, and field order — proved by a from-the-page encoder, byte-identical to the generated `Save` (below).
- **An array BOUND is not wire identity** (§4, USAGE): resize a bounded array — a literal, a constant, or an `E.Max + 1` expression that moved when its enum grew — and stored files still load. Past the reader's bound: bounded prefix, `clamped`. Short of it: the tail keeps its declared defaults. `malformed` stays what it was — a count the body cannot cover.
- **§4 states the kind vocabulary's one coarse edge**: an enum field and a plain `uint16` field are both kind `7`, so an edit between the two is not a kind mismatch.

## Negative controls

Both were applied to the emitter, run, and reverted before committing; the branch carries neither.

**(i) enum identity reverted to the ordinal** (`TableEnumId`/`TableEnumValue` emit `ordinal + 1` instead of the name hash; everything else untouched, so the framing is identical and only identity changes):

```
FAIL test/tables/main.cpp:438: grade->variant_id != NULL && grade->variant_id( 2 ) == field_id( "Gold" )
FAIL test/tables/main.cpp:444: out.grade == tblv2::Grade::Gold
FAIL test/tables/main.cpp:460: report.unknown == 1
FAIL test/tables/main.cpp:461: out.grade == tblv1::Grade::None
FAIL test/tables/main.cpp:470: !report2.malformed && report2.unknown == 0
FAIL test/tables/main.cpp:471: out2.grade == tblv1::Grade::Gold
FAIL test/tables/main.cpp:747: grade->variant_id( 1 ) == field_id( "Bronze" )
FAIL test/tables/main.cpp:748: grade->variant_id( 2 ) == field_id( "Silver" )
FAIL test/tables/main.cpp:749: grade->variant_id( 3 ) == field_id( "Gold" )
tables test: 9 failure(s)
```

Line 444 is the one that matters: V1 wrote `Gold`, V2 inserted `Silver` in the middle, and under ordinals V2 reads the stored `Gold` back as `Silver` — silently, with an empty report. Line 461 is the other direction: a variant the reader has no name for is read as its neighbour instead of `None`.

**(ii) array counts restored to u16** (write, read, measure and the pointer-side sizing pre-pass; string lengths left at u32 so the control isolates the count):

```
FAIL test/tables/main.cpp:321: report2.malformed
FAIL test/tables/main.cpp:322: out2.items_count == 1 && out2.items[0] == 10
FAIL test/tables/main.cpp:596: out.payload_length == 70000 && out.payload[69999] == (uint8_t) ( 69999 & 0xff )
FAIL test/tables/main.cpp:597: out.samples_count == 70000 && out.samples[69999] == (uint16_t) 69999
FAIL test/tables/main.cpp:623: out2.samples_count == 2
FAIL test/tables/main.cpp:624: out2.samples[0] == 11 && out2.samples[1] == 22
```

70000 truncates to 4464 in a u16 count, so the wide extents test loses 65536 elements and 65536 bytes without a word of complaint. (The full revert — leaving the reader's clamp at `uint16_t` too — does not even compile: `-Werror` refuses `keep > 70000` and `keep = 70000` in the generated `WideTable.h`, because a u16 count cannot hold the corpus's own bound.)

## Gates

- `make test` — the full nine-language battery, green (31s).
- `make tables-zero-cost` — green. It is a **grep gate**, not a stored baseline: it refuses pointer machinery in a pointer-free unit's headers. No baseline to regenerate, so none was regenerated. The value-only corpus's generated text changes only where the wire legitimately changed (enum kind and codec, union arm id, u32 lengths and counts) plus the new `variant_id` descriptor column.
- Committed `generated/` tree byte-unchanged — the packet wire did not move.
- `TestTablesMoveNoProtocolId` (SPEC-TABLES §10) — green.
- Evolution corpus, both directions: `test_evolution_enum_insert_old_data` / `_new_data` and `test_evolution_union_insert_old_data` / `_new_data` — a variant inserted in the MIDDLE, old Gold still loads as Gold, an unnameable variant lands on `None`/empty and counts.
- measure == save at exact capacity — the existing battery, plus the wide case (`WideBlobSave` at exactly `Measure`'s answer, and one byte short refusing).
- `go vet`, `gofmt`, `go mod tidy -diff`, `modernize`, `shapegate` — all clean.

## Calls I made that the issues did not decide

Each is documented where it lands, and each is reversible.

1. **`None` rides as `0` for an enum, not as `hash("None")`.** It is the fold's reserved id, it mirrors the union's `0 = empty`, and no declared name can reach it. (SPEC-TABLES §3, §5.)
2. **The unknown-variant counter is `unknown`**, reused rather than added. #249 named `unknown`; the semantics agree — the writer named something this reader cannot name — and `TableReport`'s shape stays put. §4 now says `unknown` counts field ids, enum variant ids and arm ids. An unnameable arm was previously `kind_mismatch`; it is `unknown` now.
3. **A stored value no variant names is REFUSED on the way out** — measure and save return -1, exactly as they already do for an out-of-range union tag — rather than being silently written as `None`. Writing `None` over a value the program is holding is silent data loss, which this wire does not do anywhere else.
4. **Enum `| max = K` headroom is refused in a table closure** (new refusal, SPEC-TABLES §5, §11). It is the direct consequence of (3): headroom values have no name, so under name hashing they cannot ride at all, and refusing the declaration is better than refusing every value at save time. Nothing in the corpus used it.
5. **Union fields carry their arm vocabulary in the descriptor**: `enum_max` = arm count and `enum_name` = tag → arm name, which were `-1`/`NULL` for a union before. Without them the new `variant_id` column has nothing to be indexed by, and #249 asks for descriptors to carry the id per variant.
6. **§11 lists 23 suffixes, not the 24 #252 estimated** — that is what `tableGeneratedVerbs` actually holds today; the list on the page is the array, verbatim.
7. **The enum id functions are emitted per header behind a macro guard** (`<PKG>_SCHEMA_TABLE_ENUM_<NAME>`) rather than only in the file that declares the enum: two files of a unit may both reach one enum, and the guard makes one definition survive per translation unit whatever the include order.

---

# Review repairs (adversarial review, MERGE WITH CONDITIONS)

All three conditions and the nit are addressed; the two observations are recorded as prose. Six further commits, cadence-pushed.

## Condition 1 — §3's union framing, stated once, exactly as the codec writes it

The reviewer was right: §3 listed `union` under the "u32 byte length then body" bullet AND described the arm-id form separately, so read together the page specified one length the codec never writes. `§3`'s payload section is now **one table, one row per kind**:

| kind | payload |
|---|---|
| `12` string | `L`, then `L` bytes |
| `13` table | `L`, then `L` bytes of table body |
| `14` array | `L`, then `element kind (u8)`, `N` (u32), then the elements |
| `15` union | `arm id (u16)`, and when it is not 0, `L` then `L` bytes of table body |

with the sentence the reviewer asked for made explicit: **the union carries no outer length.** The "any reader can skip any field" promise is now a stated rule rather than a claim — kinds `1`–`11` skip their fixed width, `12`/`13`/`14` read `L` and skip it, and `15` reads the u16 arm id and stops there if it is 0, else reads `L` and skips it — with kind 15 named as the one payload a skipper must know the shape of. §3 also now states **field order**: this implementation writes declaration order, a reader may not rely on it, a repeated id is legal input and the last occurrence wins, and byte-identical output against this implementation therefore requires matching its declaration order as well as its framing.

**Proof.** A Go encoder written from the rewritten §3 (and §5's fold) alone — no schema package imported, no generated header read — encoding three corpus instances spanning every kind: `tblv1::Cfg` (i32, f32, enum, string, nested table, counted i32 array, counted and fixed enum arrays, union), `tabledemo::LoadoutConfig` (flags mask, fixed and counted arrays of TABLES), `tabledemo::ProfileConfig` (bytes, every integer width, f64, fixed f32 array, a guarded nested table). Scratch dir, `go run`, deleted after:

```
$ go run .          # encoder written from the rewritten SPEC-TABLES.md §3 alone
cfg_from_spec3     20cd0409000000e9ca0a000008417c0c070371df300c0500000068656c6c6ff9b20d0900000029af0a000018410000f6090e0d00000004020000000a0000001400000072d20727da1a300e0b00000007030000000734000027da8a080e0b000000070300000027da07340000d8f40e0d00000004020000006400000065000000997d0e1100000004030000000700000008000000090000003ae30fc92109000000bbe30a0000003f00000000
loadout_from_spec3 72d20727da1a300e090000000702000000073427da8a080e0b000000070300000027dae7a30734c52f090600000000000000f7050d0d000000a9150a0000204240ab0101000047160e1c0000000d020000000200000000000d000000a9150a0000204240ab01010000d5440e120000000d0100000009000000e437040300000000000000
profile_from_spec3 df300c0a000000706c61796572206f6e65b0f30e0800000006030000000102fad86108090300003c0202f45d8803d08aa06705000efad5feffffff1cb706c8426907409cbdf4090000000000010000e0960b000000000000c03fdd970e150000000a040000000000803f000000400000404000008040dbb40101789f0d8400000072d20727da1a300e090000000702000000073427da8a080e0b000000070300000027dae7a30734c52f090600000000000000f7050d0d000000a9150a0000204240ab0101000047160e1c0000000d020000000200000000000d000000a9150a0000204240ab01010000d5440e120000000d0100000009000000e4370403000000000000000000

$ ./emit            # the generated C++ Save, same three instances
cfg_generated      20cd0409000000e9ca0a000008417c0c070371df300c0500000068656c6c6ff9b20d0900000029af0a000018410000f6090e0d00000004020000000a0000001400000072d20727da1a300e0b00000007030000000734000027da8a080e0b000000070300000027da07340000d8f40e0d00000004020000006400000065000000997d0e1100000004030000000700000008000000090000003ae30fc92109000000bbe30a0000003f00000000
loadout_generated  72d20727da1a300e090000000702000000073427da8a080e0b000000070300000027dae7a30734c52f090600000000000000f7050d0d000000a9150a0000204240ab0101000047160e1c0000000d020000000200000000000d000000a9150a0000204240ab01010000d5440e120000000d0100000009000000e437040300000000000000
profile_generated  df300c0a000000706c61796572206f6e65b0f30e0800000006030000000102fad86108090300003c0202f45d8803d08aa06705000efad5feffffff1cb706c8426907409cbdf4090000000000010000e0960b000000000000c03fdd970e150000000a040000000000803f000000400000404000008040dbb40101789f0d8400000072d20727da1a300e090000000702000000073427da8a080e0b000000070300000027dae7a30734c52f090600000000000000f7050d0d000000a9150a0000204240ab0101000047160e1c0000000d020000000200000000000d000000a9150a0000204240ab01010000d5440e120000000d0100000009000000e4370403000000000000000000

$ diff <(go run . | sed s/_from_spec3//) <(./emit | sed s/_generated//)
(no output — byte-identical on all three)
```

The union field is visible in `cfg`: `3ae3` = `id("effect")`, `0f` = kind 15, `c921` = `id("ward")`, `09000000` = the arm's length, then the arm body `bbe3 0a 0000003f 0000`. No outer length, exactly as the page now says.

## Condition 2 — an unnameable union arm RESETS to empty

Fixed in `emitTableReadField`'s `default:` arm: `value.<f>.type = <U>Type::None;` before the `unknown++`. §4 and §5 said an unnameable arm "reads as empty"; the codec counted and skipped but left the type alone, so the sentence held only by the placement-new prefill. The enum path already reset explicitly — the two vocabularies now agree.

The battery gains the reviewer's repro as `test_repeated_id_unnameable_variant`: one buffer carrying `effect` and `grade` twice, the first occurrences real (a `ward` arm, `Gold`), the second naming an arm id and a variant id no build has (`0xBEEF`). Both must land on the reader's empty value, and `unknown == 2`.

**Negative control (iii)** — the reset line removed, everything else untouched:

```
FAIL test/tables/main.cpp:572: out.effect.type == tblv1::EffectType::None
tables test: 1 failure(s)
```

Exactly the reviewer's `expected type=0, got type=2`. Restored; `tables test passed`.

## Condition 3 — enum arrays in the corpus

The reviewer's greps were right: not one `[..N]Enum` or `[N]Enum` existed, so `emitEnumElementCheck`, the enum branch of `emitTableWriteElement` and the enum branch of `emitTableReadScalarFrom` under an array all shipped dead — `grep -l element_id build/tables-generated/*/*.h` found nothing.

`LoadoutConfig` gains `grades [..4]Grade` and `podium [3]Grade`; the V1/V2 pair gains the same two, so a **middle insert inside an enum array** is covered both directions. All three files now emit `element_id`:

```
$ grep -c element_id build/tables-generated/examples/TablesTable.h build/tables-generated/v1/V1Table.h build/tables-generated/v2/V2Table.h
build/tables-generated/examples/TablesTable.h:10
build/tables-generated/v1/V1Table.h:10
build/tables-generated/v2/V2Table.h:10
```

New assertions: round trip at exact capacity over both shapes (with `None` as an element, riding as 0); a middle-inserted variant leaving every stored element on its own variant, both directions; an element id the reader cannot name landing on `None`, counted, with the element AFTER it decoding normally; and the per-element refusal on both shapes — counted (only elements below the count are examined, proven by lowering the count and watching measure go positive) and fixed (every element rides, so every element is checked).

**Negative control (iv)** — `emitEnumElementCheck` emitting nothing:

```
FAIL test/tables/main.cpp:640: tblv1::CfgMeasure( counted ) == -1
FAIL test/tables/main.cpp:649: tblv1::CfgMeasure( fixed ) == -1
tables test: 2 failure(s)
```

Restored; `tables test passed`.

## Nit — the refusals name the edge that reached them

All three variant refusals (headroom, enum collision, union collision) now name the first field that pulled the vocabulary into the closure, including through a union arm's payload type:

```
$ ./bin/schema check H1.schema     # enum E | max = 8 { A, B }; table Filler; table Tab { e E }
enum E: | max = 8 reserves values above the declared variants, and table Tab's field e reaches it,
putting E in a table closure — a headroom value has no NAME, … (SPEC-TABLES.md §5)

$ ./bin/schema check H3.schema     # the same enum reached only THROUGH a union arm's payload
enum E: … and type Payload's field grade reaches it, putting E in a table closure — …

$ ./bin/schema check R1.schema     # enum E { agL8, arFY } — the rebound-to-1 collision
enum E: variants agL8 and arFY collide on table-wire id 0x0001, and table Tab's field tier
reaches it, putting E in a table closure — rename one (SPEC-TABLES.md §5)
```

## Owner-raised, folded in: an array's BOUND is not wire identity

A field is its name hash and its kind; neither carries an extent. SPEC-TABLES §4 and USAGE now state what the reader already does, and the V1/V2 pair holds it both directions in one edit — `MaxSlots` shrinks 6 → 3 while `tally`, sized `[Grade.Max + 1]`, GROWS 3 → 4 because `Grade` gained `Silver`, so one instance exercises the shrink and the grow at once each way round. `clamped` for a count past the reader's bound with the bounded prefix kept; declared defaults in the tail for a count short of it; `malformed` reserved for a count the body cannot cover (asserted not set in either direction).

## Observations recorded as prose, not fixed

- §4 now says an enum field and a plain `uint16` field are both kind `7`, so an edit between the two is not a kind mismatch — the raw value is read as a variant hash and lands on `None` unless it happens to name one.
- Field order is stated in §3 (above), including that a reader may not rely on it.

## Gates after the repairs

`make test` green (full nine-language battery, tables test, zero-cost gate), committed `generated/` tree byte-unchanged, `go vet` / `gofmt` / `go mod tidy -diff` / `modernize` / `shapegate` clean.


gafferongames:
## Adversarial review — verdict: **MERGE WITH CONDITIONS**

Reviewed in a fresh clone (`~/rowan-working/schema-review254`, `gh pr checkout 254`), no part in writing it. Every claim below has the command that produced it. Findings first.

---

# Findings

## 1. §3 is not implementable from itself: the union framing over-reads by 4 bytes  *(condition — docs fix, one line)*

The PR's own headline docs claim is **"§3 is now implementable from itself: kind numbers by value, every payload's framing, both container headers."** I implemented an encoder in Go from §3 alone — no schema code imported, only the page — and byte-compared it against the generated C++ `Save` for the same instance. Seven field kinds matched exactly. The union did not.

§3 says both of these:

> - **table**, **array** and **union** carry a **u32 byte length**, then the body …
> - A **union body** opens with its **u16 arm id** … and, when the id is not 0, a **u32 byte length** then the arm's value as a table body.

Read together that is `id, kind=15, u32 len, [u16 arm id, u32 len, arm]`. The shipped codec writes `id, kind=15, u16 arm id, u32 len, arm` — no outer length. `emitTableWriteField`'s union case goes straight from `w.put8(15)` to the arm-id switch, `emitTableMeasureField` charges `3 + 2 + 4 + arm`, and `TableReader::skip` case 15 consumes exactly that.

Repro — a table with `grade` (enum), `grades` (`[..4]Grade`), `trio` (`[3]Grade`), `effect` (union), `label` (string), `payload` (bytes), `samples` (`[..70000]uint16`); a Go encoder written from §3 alone, printed twice, and the generated C++ `ProbeSave` for the same value:

```
$ go run .                       # hand encoder, from SPEC-TABLES.md §3 only
as_written_in_spec_3: 72d20727da1a300e090000000702000000073427da374b0e0b000000070300000027da000000003ae30f0f000000c92109000000bbe30a0000003f00006ae10c020000006f6baa440e0800000006030000000102039aaf0e07000000070100000009000000
without_outer_length: 72d20727da1a300e090000000702000000073427da374b0e0b000000070300000027da000000003ae30f        c92109000000bbe30a0000003f00006ae10c020000006f6baa440e0800000006030000000102039aaf0e07000000070100000009000000

$ ./cpp/emit                     # generated ProbeSave, same instance
generated_cpp_save:   72d20727da1a300e090000000702000000073427da374b0e0b000000070300000027da000000003ae30f        c92109000000bbe30a0000003f00006ae10c020000006f6baa440e0800000006030000000102039aaf0e07000000070100000009000000
```

`without_outer_length` is byte-identical to the generated save; `as_written_in_spec_3` carries `0f000000` — one extra u32 — after `…3ae30f` (the `effect` field header). The same divergence, and only that divergence, showed up on the `tblv1::Cfg` instance too (82 bytes generated vs 86 hand-written).

The rationale clause is wrong for union as well: *"so any reader can skip any field without understanding it"* is false for kind 15 under the real wire — a generic skipper must know the union's internal shape, which is exactly why `skip` special-cases it. `main` carried the same over-claim (`Arrays, unions and nested tables carry a u32 byte length then the body`), so this is not regressed by the PR — but the PR rewrote this bullet and staked a claim on it, so it is the PR's to close. Either drop `union` from that bullet, or restate the union payload in one place as `u16 arm id [, u32 byte length, arm body]`.

## 2. §4's "the union reads as empty" is not what the reader does  *(condition — one line of emitter)*

§4 and §5 both say an arm the reader cannot name means the union **reads as empty**. The generated `default:` arm counts `unknown` and skips the body, but never assigns `type = …Type::None` — where the enum path explicitly does (`value.grade = Grade::None;`). On a fresh `Load` the placement-new prefill hides it. Repeat the field id and it shows:

```
$ ./cpp/dup
union: load=1 type=2 (0=None) unknown=1 malformed=0
  SPEC-TABLES §4 says the union reads as EMPTY -> expected type=0, got type=2
enum:  grade=0 (0=None) unknown=1  -> the enum path DOES reset to None
```

(Buffer: `effect` = `ward{charge=0.5}`, then `effect` again with arm id `0xBEEF`.) Hostile-input-only, memory-safe, no fabrication — but the two vocabularies disagree, and the union side contradicts the sentence this PR wrote. Fix is `value.%s.type = %sType::None;` in the `default:` case of `emitTableReadField`.

## 3. The enum-ARRAY paths this PR adds have zero corpus coverage  *(condition — corpus)*

`emitEnumElementCheck` (measure), the enum branch of `emitTableWriteElement` (save) and the enum branch of `emitTableReadScalarFrom` under an array (load) are new in this PR and are never instantiated:

```
$ grep -nE '\[[^]]*\](Grade|Mode)' tables/examples/*.schema tables/pointers/*.schema test/tables/*.schema
(no matches)
$ grep -l 'element_id' build/tables-generated/*/*.h
(none)
```

Not one `[..N]Enum` or `[N]Enum` field exists anywhere in the corpus, so all three paths are dead in the built output. I exercised them by hand against a scratch schema and they are **correct** — the unnameable-element refusal fires on both the counted and the fixed array, and the read path resets to `None` and counts `unknown` — but nothing in the repo would catch a regression. One field on `LoadoutConfig` closes it.

## 4. The headroom refusal names the enum, not the path that pulled it into the closure  *(nit)*

```
$ ./bin/schema check H1.schema
H1.schema:2:1: enum E: | max = 8 reserves values above the declared variants, and E is in a table's closure — a headroom value has no NAME, … (SPEC-TABLES.md §5)
```

The error exists *because of* closure membership, and the diagnostic does not say which table or field created it. The collision diagnostics name both variants; this one leaves the user to find the edge. In a unit with many tables that is a search.

## 5. Two observations, no action asked

- **§3 never states field ORDER within a table body.** Declaration order in practice. It does not affect decodability (ids are keys), but it does mean a third-party encoder written from the page cannot be expected to produce byte-identical output, which is the standard §3 sets for itself.
- **An enum field and a plain `uint16` field are both kind `7`.** So a `uint16` → enum edit is not caught by the kind-mismatch rule — the raw value is read as a variant hash. Uniform-u16 makes this crisper than the old storage-width scheme rather than worse, and §4's list of tolerated events could say so in a clause.

---

# The two worker calls, assessed

**(a) Enum `| max = K` headroom refused in a table closure — verified, and I agree it is the only consistent answer.**

```
$ ./bin/schema check H1.schema      # enum E | max = 8 { A, B }; table Tab { e E }
enum E: | max = 8 reserves values above the declared variants … (SPEC-TABLES.md §5)   exit=1
$ ./bin/schema check H3.schema      # the same enum reached only THROUGH a union arm
enum E: | max = 8 reserves values above the declared variants … exit=1
$ ./bin/schema check H2b.schema     # the same enum, no table in the unit
exit=0
$ ./bin/schema check H5.schema      # headroom enum beside a table, outside its closure
exit=0
$ ./bin/schema check H4.schema      # | max = exactly the variant count (no headroom)
exit=0
```

Fires inside the closure including through a union arm, silent outside it, silent when `max` equals the variant count.

Could headroom ride by ordinal-with-a-flag instead? It is implementable — reserve a bit of the u16 (fold to 15 bits, use bit 15 to mean "ordinal + 1"), or reserve a high id band and refuse any name that folds into it. Both cost a permanently narrower name space (a 15-bit fold doubles the accidental-collision rate on every field, variant and arm in the language, to buy a feature nothing in the corpus uses), and both put a **second identity mechanism** on a wire whose entire thesis is one. And they buy nothing: headroom exists on the packet wire to reserve *bit width* so a later variant does not widen the field and move the protocol id. On the table wire an enum is always kind 7 / two bytes and a variant may be added anywhere for free — the problem headroom solves does not exist here. Refusing is right, and §5 states the consequence plainly (an enum with headroom cannot be shared between packet code and a table closure without splitting it).

**(b) A stored value no variant names is refused at measure and save — verified, all three shapes.**

```
$ ./cpp/hostile
== 3(b) a stored value no variant names is refused on the way out ==
  ok:   ProbeMeasure( p ) == -1        # scalar enum field
  ok:   ProbeSave( p, buf, sizeof(buf) ) == -1
  ok:   ProbeMeasure( p ) == -1        # counted enum array element
  ok:   ProbeSave( p, buf, sizeof(buf) ) == -1
  ok:   ProbeMeasure( p ) == -1        # fixed enum array element
  ok:   ProbeSave( p, buf, sizeof(buf) ) == -1
== 3(b) Load of a hostile UNKNOWN enum id: None + unknown, no crash ==
  ok:   ProbeLoad(...)   ok: out.grade == Grade::None   ok: rep.unknown == 1   ok: !rep.malformed
== 3(b) Load of a hostile UNKNOWN union arm id ==
  ok:   ProbeLoad(...)   ok: out2.effect.type == EffectType::None   ok: rep2.unknown == 1   ok: !rep2.malformed
```

(built `-fsanitize=address,undefined`). Measure and save agree with each other and with the union-tag precedent; load lands on `None`/empty and counts, no crash. Finding 2 above is the one seam.

---

# Verified good

**Identity (§5).** `ir.VariantId` *is* `ir.FieldId` — same function, so the folds cannot diverge. The rebound to 1 is the interesting case: I searched the ≤4-char identifier space and found 186 names folding raw to 0 (rebound to 1) and 187 folding naturally to 1, so the rebound really can collide — and it is caught:

```
$ ./bin/schema check R1.schema    # enum E { agL8, arFY }; table Tab { e E }
enum E: variants agL8 and arFY collide on table-wire id 0x0001, and E is in a table's closure — rename one (SPEC-TABLES.md §5)
$ ./bin/schema check R2.schema    # the same enum, type T instead of table Tab
exit=0                            # scoped to the closure: the packet wire is untouched
$ ./bin/schema check R3.schema    # a variant literally named None
variant None is a compile error — every enum has None = 0 implicitly (SPEC §4.2)
```

Nothing declared can reach `0`; `id("None") = 0xc7b4`, and `None` as a spelling is refused by the language before the table rules see it. Scoping is real (`checkTableVariantIdentity` walks only the closure, and the closure walk goes through union arm types).

**u32, hostile.** All under ASan+UBSan:

```
== 4 a u32 count of 0xFFFFFFFF with a tiny body ==
  ok: out3.samples_count == 2   ok: samples[0]==111 && samples[1]==222   ok: rep3.malformed   ok: rep3.clamped == 1
== 4 a u32 count of 0xFFFFFFFF on an ENUM array ==
  ok: out4.grades_count == 1    ok: rep4.malformed
== 4 a string length larger than the remaining buffer ==
  ok: !okd   ok: rep5.malformed   ok: out5.label_length == 0
== 4 a bytes body length larger than the remaining buffer ==
  ok: !okd6  ok: rep6.malformed
== 4 measure == save at exact capacity, and one byte short refuses ==
  measure = 280094
  ok: n == m   ok: n2 == m   ok: memcmp(exact, roomy, m) == 0   ok: ProbeSave(p, exact, m-1) == -1
  ok: round-trip clean, label/payload/samples all 70000, grades and trio and the union intact
```

The bounded prefix holds, the clamp fires, nothing is fabricated from a neighbour's bytes, and 70000-element extents survive at exact capacity. The repo has **no table-wire fuzz corpus** (`internal/fuzz` fuzzes schema *source*, not wire bytes), so I wrote a smoke fuzz: 200 000 iterations of random and mutated buffers into `ProbeLoad` / `RootConfigLoad` / `LoadoutConfigLoad` / `WideBlobLoad` under `-fsanitize=address,undefined -fno-sanitize-recover=all` — **clean**. Worth landing something like it eventually; not this PR's debt.

**Zero cost and independence.**

```
$ make test    # (dist/ symlinked in from a populated tree; the fresh clone has none)
tables test passed
… nine-language battery, go test ./... all ok      EXIT=0
$ git status --porcelain
(clean — the committed generated/ tree is byte-unchanged after every generate step re-ran)
$ go test ./internal/check/ -run TestTablesMoveNoProtocolId -v
--- PASS: TestTablesMoveNoProtocolId (0.00s)
```

`gofmt -l .` empty, `go vet ./...` silent, `make check` clean, `make fmt` leaves no drift.

**Both negative controls reproduce, exactly as the PR body reports them.** Applied to my own checkout, run, reverted.

(i) `TableEnumId`/`TableEnumValue` emitting `ordinal + 1`:

```
FAIL test/tables/main.cpp:438 … 444 … 460 … 461 … 470 … 471 … 747 … 748 … 749
tables test: 9 failure(s)
```

(ii) array counts back to u16 (write, read, measure, and `pointers.go`'s sizing pre-pass; string lengths left at u32):

```
FAIL test/tables/main.cpp:321 … 322 … 596 … 597 … 623 … 624
tables test: 6 failure(s)
```

Same lines, same order, same counts as the PR body claims — including the third claim: taking the reader's clamp to `uint16_t` as well does not compile.

```
build/tables-generated/examples/WideTable.h:281:31: error: result of comparison of constant 70000 with expression of type 'uint16_t' is always false [-Werror,-Wtautological-constant-out-of-range-compare]
build/tables-generated/examples/WideTable.h:281:50: error: implicit conversion from 'int' to 'uint16_t' changes value from 70000 to 4464 [-Werror,-Wconstant-conversion]
```

Restored; `git diff --stat` empty and `./build/schema_test_tables` back to `tables test passed`.

**Call #7 (per-header macro-guarded enum identity) holds — and is not covered by the corpus.** Only `TablesTable.h` emits a guard today. I built the two-file case the call describes (`A.schema`: `enum Grade` + `table Alpha`; `B.schema`: `table Beta { g Grade }`), and both headers emit the pair:

```
$ grep -c 'SCHEMA_TABLE_ENUM_GRADE' gen/*Table.h
gen/ATable.h:3
gen/BTable.h:3
$ c++ -Wall -Werror -I gen both.cpp  && ./both      # include A then B
alpha=7 beta=7
$ c++ -Wall -Werror -I gen both2.cpp && ./both2     # include B then A
alpha=7 beta=7
```

The guard-name collapse I went looking for (`ToUpper` folding two enum names together) is unreachable — `enum Grade` beside `enum GRADE` is already refused by the C-form symbol rules.

**Docs (#252).** `TableDecl`, `"*" ident` and `was` are all in SPEC §4.2. The §11 suffix list is the array verbatim:

```
$ bash /tmp/verbs.sh
code count: 23   doc count: 23
--- diff (code vs doc) ---
IDENTICAL SETS
```

The default-is-wire-contract paragraph is in SPEC-TABLES §4 and §5 and beside USAGE's elision paragraph; the optional-is-`if` / `*`-is-identity rewording is in SPEC-TABLES §2.1 and USAGE, with both `settings *Settings` examples replaced by `palette *Palette`; the cross-file cycle rule is restated language-neutrally with the C++ consequence demoted to a note. A scan of the added lines for historical register (`previously|used to|no longer|formerly|changed from|deprecat|renamed from|moved from`) returns nothing — the two hits are present-tense prose. The USAGE reflection loop compiles and runs as written:

```
grade    Grade    kind= 7  None=0x0000  Bronze=0x3407  Gold=0xda27
effect   Effect   kind=15  None=0x0000  boost=0xdcc5  ward=0x21c9
```

**Provenance.** Six commits, every one authored *and* committed `Rowan Claude <rowan@mas-bandwidth.com>`. Thirteen files, all of them in scope. No `git add -A` artefacts; `ROUND-LOG.md`, `notes/` and `working/` untouched.

---

# Verdict: MERGE WITH CONDITIONS

The wire this PR ships is right. I attacked it from the page, from the sanitizer and from the checker and could not make it fabricate, misdecode or crash, and both negative controls reproduce to the line. The conditions are all *around* the wire, not in it:

1. **Fix §3's union framing** (Finding 1) — one line, and it is the PR's own claim. Repro: the two hex strings above.
2. **Reset the union to `None` on an unnameable arm** (Finding 2) — one line in `emitTableReadField`'s `default:`. Repro: `./cpp/dup`.
3. **Put one enum array in the corpus** (Finding 3) — three new emitter paths currently ship untested. Repro: the two greps above.

Finding 4 (diagnostic doesn't name the closure path) and the two observations in Finding 5 are ruled-not-now material if you'd rather.

*Reviewed in a throwaway clone; the branch was never edited, and every local patch made for the negative controls was reverted and re-verified green.*

gafferongames:
## Delta review of the five repair commits — verdict: **MERGE**

Fresh clone, `gh pr checkout 254`, deltas only (`bdac88a..HEAD`). I reproduced each condition myself rather than reading the worker's pastes; the branch was never edited and every local sabotage was reverted and re-verified green.

---

## Condition 1 — §3's union framing — **closed**

I threw away my old encoder and wrote a new one from the **rewritten** §3 alone: the per-kind payload table, the array-element rules, the elision list, and §3's own statement that this implementation orders fields by declaration. Twelve fields across nine shapes — `int32`, an elided `float32`, an enum scalar, a `string`, a nested table, a counted `int32` array, a counted enum array (with `None` as an element), a **fixed** enum array, a counted array under a shrinkable constant, a fixed array sized `[Grade.Max + 1]`, and the union:

```
$ diff <(cd d1 && go run .) <(./emit 2>/dev/null)
IDENTICAL — the rewritten §3 encodes the union exactly as the codec writes it

measure=160 save=160
20cd04070000007c0c070371df300c020000006869f9b20d0900000029af0a000060400000f6090e0d0000000402
0000000a0000001400000072d20727da1a300e090000000702000000073400008a080e0b000000070300000027da
00000734d8f40e0d00000004020000006400000065000000997d0e11000000040300000000000000050000000000
00003ae30fc92109000000bbe30a0000003f00000000
```

The union tail reads `3ae3` (`id("effect")`) `0f` (kind 15) `c921` (`id("ward")`) `09000000` (the arm's `L`) then the arm body — no outer length, exactly as the table row and the "**The union carries NO outer length**" sentence now say. The new skip paragraph also matches `TableReader::skip` rule for rule (1–11 fixed width; 12/13/14 read `L` and skip `L`; 15 reads the arm id, stops on 0, else `L`). Nothing left in the document implies an outer length:

```
$ grep -n 'union' SPEC-TABLES.md | grep -iE 'length|u32|frame|framing|skip'
215:  **The union carries NO outer length** — its `arm id` sits where the other
```

The field-order paragraph also makes a new normative claim — "a body carrying an id more than once is legal input, the last occurrence wins" — so I tested that too, with two nameable occurrences of a scalar and of an array:

```
last-occurrence-wins: grade=2 items_count=1 items[0]=7 -> PASS
```

## Condition 2 — the unnameable union arm resets to empty — **closed**

My original repro, re-derived and re-run against the repaired build:

```
union repeated-id: load=1 type=0 (0=None) unknown=1 malformed=0 -> PASS
enum  repeated-id: grade=0 (0=None) unknown=1 -> PASS
0 failure(s)
```

`type=2` before, `type=0` now, with `unknown` still counted and nothing flagged malformed. The two vocabularies agree.

## Condition 3 — enum arrays in the corpus — **closed**

The paths are live in all three generated units, and the schemas that make them so are real fields, not test-only scaffolding:

```
$ grep -c element_id build/tables-generated/{examples/TablesTable.h,v1/V1Table.h,v2/V2Table.h}
…/TablesTable.h:10   …/V1Table.h:10   …/V2Table.h:10
$ grep -nE '\[[^]]*\](Grade|Mode)' tables/examples/*.schema test/tables/*.schema
tables/examples/Tables.schema:60,61   test/tables/V1.schema:53,54   test/tables/V2.schema:52,53
```

I ran the worker's stated sabotage and two of my own, each on my own checkout, each reverted:

**(a) `emitEnumElementCheck` emits nothing** — the worker's control, reproduced:

```
FAIL test/tables/main.cpp:691: tblv1::CfgMeasure( counted ) == -1
FAIL test/tables/main.cpp:700: tblv1::CfgMeasure( fixed ) == -1
tables test: 2 failure(s)
```

Same two assertions and same count as the PR body; the line numbers there (640/649) are from before `ca380e5` landed ~50 lines above them. Cosmetic staleness in the paste, not in the tree.

**(b) enum ARRAY elements ride as the ordinal instead of the name hash** — mine, the array analogue of negative control (i):

```
FAIL main.cpp:473: !report2.malformed && report2.unknown == 0 && report2.clamped == 0
FAIL main.cpp:475: out2.grades[0] == Grade::Gold && out2.grades[1] == Grade::Bronze
FAIL main.cpp:476: out2.podium[0] == Grade::Bronze
FAIL main.cpp:477: out2.podium[1] == Grade::Gold
FAIL main.cpp:516: !report3.malformed && report3.unknown == 1
FAIL main.cpp:518: out3.grades[0] == Grade::Gold
tables test: 6 failure(s)
```

Both array shapes, both directions. The corpus addition genuinely bites.

**(c) the enum read's reset to `None` removed** — mine:

```
FAIL test/tables/main.cpp:621: out.grade == tblv1::Grade::None
tables test: 1 failure(s)
```

One failure, from the *scalar* repeated-id case. See the note below.

## Nit — the refusals name the reaching edge — **closed**

All four, including through a union arm's payload type, and the closure scoping is intact:

```
$ ./bin/schema check N1.schema
enum E: | max = 8 … and table Tab's field e reaches it, putting E in a table closure — …
$ ./bin/schema check N2.schema        # reached only through union U's arm payload
enum E: | max = 8 … and type Payload's field grade reaches it, putting E in a table closure — …
$ ./bin/schema check N3.schema
enum E: variants agL8 and arFY collide on table-wire id 0x0001, and table Tab's field tier reaches it, …
$ ./bin/schema check N4.schema
union U: arms costarring and liquid collide on table-wire id 0xf4d0, and table Tab's field pick reaches it, …
$ ./bin/schema check N5.schema        # both vocabularies present, no table reaches them
exit=0
```

## Owner addition — an array's BOUND is not wire identity — **verified independently**

I did not read the new test and agree with it; I wrote my own against the V1/V2 pair. `MaxSlots` 6→3 and `tally` `[Grade.Max + 1]` 3→4 in one instance, so each direction exercises a shrink and a grow at once:

```
== V1 (slots bound 6, tally [3]) -> V2 (slots bound 3, tally [4]) ==
  ok: !rep.malformed        ok: rep.clamped == 1        ok: r.slots_count == 3
  ok: slots[0..2] == 100,101,102     ok: tally[0..2] == 10,11,12     ok: tally[3] == 0
  (report: unknown=0 kind_mismatch=0 clamped=1 malformed=0)
== V2 -> V1, the other direction ==
  ok: !rep.malformed        ok: rep.clamped == 1        ok: r.slots_count == 3
  ok: slots[3] == 0 && slots[5] == 0    ok: tally[0..2] == 20,21,22
  (report: unknown=0 kind_mismatch=0 clamped=1 malformed=0)
== and the line §4 draws: a count the BODY cannot cover IS malformed ==
  ok: rep.malformed         ok: rep.clamped == 0        ok: r.slots_count == 2
0 failure(s)
```

Clamped and not malformed in both directions, the grown tail holds its declared defaults, and the `clamped`/`malformed` line §4 draws holds at the boundary — a count inside the reader's bound but past the body is framing damage with `clamped == 0`.

## Observations — recorded as §3/§4 prose, and both are accurate

Field order is now stated, including that a reader may not rely on it and that byte-identical output needs matching declaration order (which is precisely what my encoder had to do). The kind-`7` overlap between an enum field and a plain `uint16` is stated in §4's kind-mismatch row in the terms the codec actually behaves in.

## Gates and provenance

```
$ make test                        EXIT=0
tables test passed
… nine-language battery, go test ./... all ok
$ git status --porcelain           (clean — committed generated/ tree byte-unchanged)
$ gofmt -l . ; go vet ./... ; go mod tidy -diff      (all silent)
```

Five commits, every one authored **and** committed `Rowan Claude <rowan@mas-bandwidth.com>`. Eight files, all in scope; no `ROUND-LOG`, `notes/`, `working/` or stray artefacts. As regression insurance I re-ran my wire-level smoke fuzz over the *repaired* corpus — 200 000 random and mutated buffers into five `Load` entry points (both generations, `LoadoutConfig` with its new enum arrays, `RootConfig`, `WideBlob`) under `-fsanitize=address,undefined -fno-sanitize-recover=all`: **clean**.

---

## One note, not a condition

**The enum ARRAY element's reset to `None` is correct but unguarded.** Sabotage (c) produced only one failure, from the scalar repeated-id case — `test_unnameable_enum_element_read` stayed green, because a fresh `Load`'s prefill masks a missing reset when no earlier occurrence wrote the slot. The gap is reachable: a repeated `grades` id whose second occurrence carries an unnameable element id.

```
--- against the SABOTAGED (c) headers ---
grades[0] = 2 (0 = None, 2 = Gold), unknown = 1  -> LEAKED the earlier element
--- against the SHIPPED headers ---
grades[0] = 0 (0 = None, 2 = Gold), unknown = 1  -> None, correct
```

**The shipped behaviour is right** — this is a missing assertion, not a defect, and it is the array twin of the case condition 2 just closed for the scalar. One more buffer in `test_repeated_id_unnameable_variant` closes it. Land as is; take it whenever.

---

# Verdict: **MERGE**

All three conditions closed and independently reproduced, the nit closed, both observations folded into the page in accurate terms, and the owner addition verified from my own harness in both directions including the `clamped`/`malformed` boundary. Battery, zero-cost gate and packet-wire independence green; provenance clean.

Issue #253, opened 2026-09-02
Tables: JSON in and out of ONE table instance, driven by the reflection descriptors — the un-opinionated half of packing


gafferongames:
Owner ruling, 2026-09-02 (Glenn, live): "the general idea of reading JSON file into a table can move into schema" — "that's not opinionated. but only one table at a time? So we separate packing from table reading JSON, the actual linking up by enums has to be replicant only since that is opinionated."

**What schema takes.** For every table in a unit's closure, C++ first: `<Name>FromJson( value, text, size, &report )` fills ONE table instance from one JSON text, and `<Name>ToJson( value, buffer, capacity )` writes one out. Both are ONE generic walk over the reflection descriptors (SPEC-TABLES.md §8: name, kind, offset, bounds, array bound + count companion, union, nested table, branch guards) — no per-table codec, which is what makes it schema's rather than pack's. Absent keys take defaults; unknown keys are counted in the report; out-of-range values clamp and count; enum and flags values are spelled by variant NAME; a union is an object with one key (the arm name); a guarded (`if has_x`) group's presence in the text sets its guard. Nested tables and arrays of tables recurse through the same walk. The JSON dialect accepts trailing commas (the authoring files in the wild carry them). No allocation on the read path beyond the caller's instance; the writer targets a caller buffer with measure/write symmetry like the wire.

**The one attribute that rides with it.** A field-level JSON key mapping (`| json = "type"`) — the name-mapping follow-on §15 already lists — lives on the table declaration and is therefore schema's. Decide and document the guard-presence spelling (today's opinionated grammar has `json_presence = "key"`; the un-opinionated form may be: the guard bool is set by the presence of the guarded group's key, no attribute).

**What schema does NOT take.** Directories, one file per instance, linking instances into a root table's collection by an enum key, the envelope (content hash, protocol id). That is PACK and it stays in replicant by ruling — it calls these functions once per file.

**Corpus.** Round-trip every table in tables/examples through ToJson → FromJson → Save and byte-match the wire; a hostile battery (wrong kinds, overflow, deep nesting, unknown keys); negative control: sabotage the walk's offset arithmetic and show the round-trip red.

Gate item for schema#248; the first consumer is space#441 (the dogfood). Queued after schema#249/#250/#252.

gafferongames:
## 16. The text form: JSON in and out of one table

**Ruled 2026-09-02.** Reading a JSON text into a table, and writing one out,
[...]

PR #263, opened 2026-09-02
The tables baseline: an optional committed projection, and the check that refuses silent edits until it moves with a reason (#256)

### LINKED ISSUE #256: The tables baseline: an OPTIONAL committed projection of every table's defaults, flag order and kinds, and a check that refuses a silent edit until the baseline moves deliberately
Owner, 2026-09-02 (Glenn, live): "Yeah I really like this idea, and it's optional, so that's good."

**The problem.** After #249 and #255, the table wire's silent class is two edits the compiler cannot see because it keeps no history: changing a specified DEFAULT (old files that elided the old default now load the new one, no report can fire) and reordering FLAGS (positional bits). A third, changing a field's KIND, is reported at load but is still a lossy edit nobody was warned about at compile time. Rules on the page are not gates.

**The mechanism, opt-in.** `schema tables-baseline <unit> > tables.baseline` writes a canonical text projection of the unit's TABLE closure: per table, per field — name, wire id, kind, array shape (fixed/bounded/enum-keyed, and the bound's VALUE), specified default (exact text), enum and flags vocabularies in order, union arm vocabulary, `was` aliases. Committed beside the schemas, like a golden. When the file is present, `schema check` (and therefore every generate) diffs the live projection against it and REFUSES, naming the field and the edit, on:

- a specified default changed (or added/removed, which changes elision)
- a flags variant inserted, removed or reordered (append at end is allowed)
- a field's kind changed
- an enum-keyed array's enum swapped for another

and WARNS (not refuse) on the reported-lossy class: a bound shrunk, a variant removed. Everything the wire absorbs safely (add/remove/reorder fields, append variants, grow bounds, `was` renames) passes silently. Updating the baseline is the deliberate act: `schema tables-baseline --update`, in the same commit as the edit, which is the code-review moment.

**Optional by construction:** no baseline file, no check — the language does not require it, the way the projection golden is the repo's discipline and not the user's. A team that wants the gate commits the file.

**Independence:** the baseline never touches the type wire, the projection or the protocol id (§10).

**Held by test:** each refusal class with a fixture pair and its negative control (remove the check → the edit passes); a full round trip on tables/examples where the baseline is regenerated and byte-identical; the reported-lossy class warns and does not refuse.

Gate item for schema#248; queued after #254, #253, #255.

gafferongames:
Closes #256. **Rebased onto `85becbf`, docs from main** — `SPEC.md`,
`SPEC-TABLES.md` and `USAGE.md` are byte-identical to main's, and this branch
touches no other doc but `VERSIONING.md`'s export note. The page's §18 carries
the policy; this is the implementation of it.

Two edits to a table schema change what data ALREADY WRITTEN means, and
neither one can be reported by any reader:

- **A specified default.** A field holding its default is elided
  (SPEC-TABLES.md §3), so the reader's declared default is what the absence
  MEANS. Change it and every stored file changes meaning — no report event
  fires, because nothing was lost or skipped.
- **A flags variant's position.** A mask carries bits, not names, so
  inserting or reordering a variant silently remaps every stored file.

The compiler retains no history and cannot see either edit on its own. This
PR gives it one, on the owner's terms: optional, diffable, and never moved
without a recorded reason.

## What lands

**`schema tables-baseline <unit>`** prints a canonical text projection of the
unit's table closure — members sorted, fields in declaration order, one fact
per line: name, wire id, kind, array shape with the bound's EVALUATED value,
string/bytes capacity, the specified default as exact canonical text
(evaluated, so a constant that moved through an expression into a default
shows up as the value it now produces), and the `was` alias; then each enum's
variants in order with their ids, each flags' variants positionally, and each
union's arms in order with their ids and payloads.

**`tables.baseline` in the unit directory turns the check on**, and nothing
else does — no file, no check. `schema check` (and so `schema generate`) then
diffs live against committed:

- **REFUSE, naming table.field and the edit** — a specified default changed,
  added or removed; a flags variant inserted, removed, reordered or RENAMED IN
  PLACE (a rename moves no byte, a new meaning on a spent bit remaps every
  stored file, and nothing distinguishes them — so the author says which); a
  field's wire kind or an array's element kind changed; a field's referent
  dropped or swapped for one whose identities do not ride; an enum-keyed
  array's key enum swapped, the day that construct exists.
- **WARN** — a bound or a capacity shrunk; an enum variant or union arm
  removed; a DECLARATION renamed, or otherwise no longer in the closure under
  its baseline name (a paired rename names what carries it on; an unpaired one
  names the closest candidate and its score). The runtime read report already
  counts what is lost (`clamped`, `unknown`), so this reports rather than
  stops.
- **PASS, in silence** — everything the wire absorbs: fields added, removed,
  reordered or renamed under `was`; variants added anywhere; flags variants
  APPENDED; bounds and capacities grown; bounded made fixed or the reverse.

**`schema tables-baseline --update --reason "<text>" <unit>`** rewrites the
file and appends a dated entry to the `## history` section inside it, naming
every edit it recorded, old value to new, beside the reason. `--update`
without `--reason` is refused. That history is the one record the wire lacks —
the log of every intentional break — and it is what a person consults years
later when an old save or an old tool file reads back wrong. It is idempotent:
a unit that has not moved rewrites nothing.

The type wire, the wire-shape projection and the protocol id are untouched
(§10); `TestTablesMoveNoProtocolId` is green.

## Design notes

- **One representation** serves rendering, parsing and diffing: a field is its
  name, its wire id and an ordered list of `key=value` tokens. The whole
  compatibility policy is then one table, `DefaultTokenPolicy`, mapping a
  token key to what a change of it means.
- **A declaration rename is followed, not lost.** A declaration name is not on
  the wire, so renaming one is absorbed — but it must not take everything
  inside the declaration out of coverage for that commit. Members match by
  name first, then across a rename by identity overlap: at least half of the
  vanished member's identities, uniquely. The vanished name warns either way,
  so a coverage hole is never invisible.
- **A field's referent must be able to STAND IN for the one it replaces**, and
  each vocabulary's standard is its own. A table body — and a union arm's
  payload, the same fact one level in — is read by field id, so it stands in
  when every id still rides AND THE FACTS UNDER THOSE IDS ARE UNCHANGED: a
  twin declaration carrying the same id under a different specified default
  rewrites the meaning of every stored body, and id membership alone would let
  it through. Those facts are judged by `diffTokens` itself, so the referent
  path and the in-place path cannot disagree about what a field fact means. An
  enum value is its variant name hash and a union body opens with its arm name
  hash, so those stand in when the names survive; a flags mask carries no names
  at all, so it stands in only when the old variants sit at the same bits.
  Dropping the referent — an enum-typed field respelled as its raw `uint16` —
  always refuses: SPEC-TABLES §4 says both ride as kind 7, so no reader can
  report it.
- **Two kinds of edit, and they never judge each other.** A declaration edited
  IN PLACE is judged by its own walk, where verdicts follow what the runtime
  can report; a field REPOINTED at a different declaration is judged by the
  referent rule, which asks about substitutability. A rename is the first
  wearing the second's clothes, so a PAIRED rename is left to the walk — the
  same wire loss draws the same verdict whether or not the author also
  renamed, and a test holds the two equal.
- **`--update` repairs a baseline the checker cannot read**, salvaging the
  `## history` section verbatim. Every parse refusal names that command as the
  remedy, so the remedy has to run — and the history is the one part of the
  file nothing can regenerate.
- **The constructs #265 landed are facts here too.** `?T` renders
  `optional=true` and is judged on NOTHING — T, ?T and *T are one framing
  (§3.1), so §18.1's "recorded and judged on nothing" is mechanical. `[E]T`
  renders `array=keyed bound=N key=E` under its own wire kind 16, so a
  keyed/positional spelling change refuses through the `kind` row, and swapping
  the key enum is judged by the referent rule under the ENUM standard, because a
  keyed array's slots ride under their variants' name ids (§3.2).
- **A key enum reaches the closure only through the array**, never through a
  field's own type, so the projection had to learn to walk `KeyEnumRef` — without
  it `Team`, `Hull` and `Weapon` were absent from the corpus baseline and every
  keyed collection's vocabulary was uncovered. Found by the fixtures, held by
  `TestKeyEnumIsCovered`.
- **A keyed array's bound is not an extent.** It IS the key enum's size, and the
  slots ride under variant name ids (§3.2): an unknown key is skipped and counted
  `unknown`, with no bounded prefix and nothing clamped. The extent row stands
  aside for a keyed shape — the enum walk already reports each variant that went,
  by name, once — while a positional array's shrunk bound still warns.
- **Fields match by WIRE ID, not by name.** The id is the identity a reader
  keys on, which is what makes a `was` rename pass and a bare rename read as
  the remove-plus-add that it is.
- **The table-wire kind mapping moved to `ir`.** The C++ backend held the only
  copy of SPEC-TABLES §3's kind vocabulary and the baseline records the same
  fact — two copies of a wire mapping are two wires. `cpptable` now names the
  one definition.

## Gates

- One fixture pair **per refusal class and per warn class**, each with two
  negative controls: the DISCRIMINATION control (the same field edited in the
  direction the wire absorbs — a bound grown, a flags variant appended, a field
  added) and the ATTRIBUTION control (the same edit re-judged with that one
  policy row removed, which must no longer produce that finding — what proves
  it came from the check under test and not from something else in the walk).
  Every class has both now: the four vocabulary walks are gated on their own
  policy rows (`member`, `flags-position`, `enum-variant`, `union-arm`), which
  is the seam the earlier revision's flags cases had no way to ablate through.
- Twelve absorbed edits that must pass in **silence** — a baseline that cries
  wolf is a baseline someone deletes.
- The committed corpus baselines (`tables/examples`, `tables/pointers`) pinned
  as goldens: regeneration must reproduce them byte for byte.
- `--update` without `--reason` refused; a foreign, unreadable or
  future-version baseline refused by name.
- `tables-baseline` added to the public-API command matrix, so the external
  module that rebuilds the CLI proves the new command too.
- Full battery, the zero-cost gate and the independence test green.

## Judgment calls, labelled

- **String/bytes capacity is in the warn class** beside the array bound. The
  issue named "a bound shrunk"; a capacity is a bound, and shrinking one
  truncates stored values exactly as a shrunk count clamps them.
- **Value ranges (`min`/`max`) are deliberately NOT recorded.** Tightening one
  clamps and the read report counts it, and the fact set the issue enumerated
  is the spec — the docs say so rather than leaving the omission silent.
- **A `type` in a table closure renders as `table`.** On the table wire the
  keyword that declared a closure member changes no byte.
- **A union arm's payload swap and a nested-table swap are one rule, not two.**
  The review filed the payload comparison separately; both are a table body
  read by field id, so both go through the same substitutability check rather
  than two rules that could drift apart.
- **The rendering `Version` is 3, and the rule for bumping it is written down.**
  ANY NEW JUDGED TOKEN BUMPS IT: a token this rendering emits and an older one
  did not reads as an ADDITION to every older file, and the added-token branch
  refuses an addition on every judged row — so an unbumped rendering greets an
  untouched schema with a refusal per field. Recording a fact nothing judges
  (an `optional`) does not need a bump; adding a rule (`key`, and earlier the
  `enum=`/`flags=`/`union=` split) does. A file at an older rendering gets one
  clean "regenerate it" refusal and keeps its history through `--update`. No
  released build has ever written a baseline, so the only files this has ever
  exposed are this branch's own intermediates.
- **The hook is on the driver, not in `check`.** `Compiler.TablesBaseline` +
  `Compiler.OnWarn` are the two new policy fields; `internal/check` is
  untouched, which keeps this rebasing cleanly under the other branches in
  flight.

## Open for the owner

**A removed FIELD passes in silence; a removed enum variant or union arm
WARNS** — and both produce the same runtime event, one `unknown` count
(SPEC-TABLES §4). Both readings are faithful to #256's own text, and the
rationale in §16 does not currently discriminate between them. Left as it
stands, flagged for a ruling rather than decided here.

## What the page states, and what the implementation also does

Two behaviours exist in the code that §18 does not describe. Both were added in
response to review and neither is in the spec text; flagging rather than
editing, since the docs come from main wholesale:

1. **The tie and the second question.** §18.3 says a vanished name is matched to
   the candidate carrying the most of its identities "when that is at least half
   of them **and unique**". The implementation goes further: when more than one
   candidate reaches that bar it decides on a second question — whose own facts
   under the shared ids are closest — and takes a candidate only when it wins
   both, strictly. Without that, a brand-new declaration reusing four field names
   outscores the real rename that dropped one, and collects four confident
   refusals for edits nobody made.
2. **A third unpaired message.** §18.3 lists two ("a paired one naming the
   declaration that carries it on, an unpaired one naming the closest candidate
   and its score"). There is a third: when two candidates cannot be separated,
   the warning names both. Under the two-message text a tie falls into the
   below-half wording and prints "carries 4 of its 4 identities, below the half
   needed" — a reason the numbers in the same sentence disprove.

Neither changes a verdict on data; both change what a person is told. A sentence
each in §18.3 would close the gap.

## Section numbering

Numbering follows the spec-of-record: this is `SPEC-TABLES.md §18` (§16 is the
JSON walk, §17 the packer). Every citation in the code names the subsection it
means — the check §18.2, the file §18.1, the override and its history §18.4, the
gate §18.6 — and `grep -rn "§16" --exclude='SPEC*.md' --exclude=USAGE.md` over the
tree comes back empty.

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -99,6 +99,17 @@ emitters — carries no promise and may change in any
 release. Building on the compiler means `compiler.Generator` and `ir`, which is
 the same door the built-in backends come through.
 
+The tables baseline (SPEC-TABLES.md §16) adds to that covered surface:
+`compiler.TablesBaselineText` and `compiler.UpdateTablesBaseline`, the
+`TablesBaseline` and `OnWarn` policy fields on `compiler.Compiler`, and in `ir`
+the table-wire kind vocabulary — the `TableKind*` constants and
+`TableScalarKind` / `TableFieldKind` / `TableElemKind`, which are wire law and
+therefore frozen (SPEC-TABLES.md §3). The baseline FILE's own rendering carries
+its own version on its first line, independent of these: a bump there makes
+every committed baseline stale at once and is repaired with
+`schema tables-baseline --update --reason "..."`, which preserves each file's
+history section.
+
 **The SPEC is versioned with the compiler.** [SPEC.md](SPEC.md) is normative;
 where the compiler and the SPEC disagree, one of them is a bug.
 
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -99,7 +99,7 @@ emitters — carries no promise and may change in any
 release. Building on the compiler means `compiler.Generator` and `ir`, which is
 the same door the built-in backends come through.
 
-The tables baseline (SPEC-TABLES.md §16) adds to that covered surface:
+The tables baseline (SPEC-TABLES.md §18) adds to that covered surface:
 `compiler.TablesBaselineText` and `compiler.UpdateTablesBaseline`, the
 `TablesBaseline` and `OnWarn` policy fields on `compiler.Compiler`, and in `ir`
 the table-wire kind vocabulary — the `TableKind*` constants and
--- a/bench/SHAPE-GATE.allow
+++ b/bench/SHAPE-GATE.allow
@@ -99,3 +99,14 @@ consts bench/tools/variantgen/main.go 3 three comments stating the 438-byte gold
 
 paths test/bench/c_main.c 1 the §1.5 golden pinner: correctness code, no timing
 paths test/bench/main.cpp 1 the §1.5 golden pinner: correctness code, no timing
+
+# ---------------------------------------------------------------------------
+# A CALENDAR DATE IS NOT A CLOCK.
+#
+# The tables baseline (SPEC-TABLES.md §16) stamps each history entry with the
+# day it was written, so a person reading the intentional-break log years
+# later knows when the break happened. It measures no interval and times
+# nothing; the timing check is lexical and cannot tell a date from a clock.
+# ---------------------------------------------------------------------------
+
+timing internal/baseline/file.go 1 the history entry's calendar date — a day written into a text log, no interval measured (prose)
--- a/bench/SHAPE-GATE.allow
+++ b/bench/SHAPE-GATE.allow
@@ -103,7 +103,7 @@ paths test/bench/main.cpp 1 the §1.5 golden pinner: correctness code, no timing
 # ---------------------------------------------------------------------------
 # A CALENDAR DATE IS NOT A CLOCK.
 #
-# The tables baseline (SPEC-TABLES.md §16) stamps each history entry with the
+# The tables baseline (SPEC-TABLES.md §18.4) stamps each history entry with the
 # day it was written, so a person reading the intentional-break log years
 # later knows when the break happened. It measures no interval and times
 # nothing; the timing check is lexical and cannot tell a date from a clock.
--- a/cmd/schema/main.go
+++ b/cmd/schema/main.go
@@ -46,7 +46,7 @@ func main() {
 	}
 	// Warnings are never quiet and never gated on --verbose: a warning nobody
 	// reads is a warning that does not exist. Today they come from the tables
-	// baseline's warn class (SPEC-TABLES.md §16).
+	// baseline's warn class (SPEC-TABLES.md §18.2).
 	c.OnWarn = func(msg string) { fmt.Fprintln(os.Stderr, "warning: "+msg) }
 
 	switch os.Args[1] {
@@ -75,7 +75,7 @@ func main() {
 		unit := loadUnit(c, os.Args[2:])
 		fmt.Print(ir.WireProjection(unit))
 	case "tables-baseline":
-		// The TABLES BASELINE (SPEC-TABLES.md §16). Printing is the default:
+		// The TABLES BASELINE (SPEC-TABLES.md §18). Printing is the default:
 		// the same canonical projection the committed file holds, so a
 		// pipeline can diff without writing anything. --update moves the file
 		// and records WHY in its history — and never without --reason,
--- a/cmd/schema/main.go
+++ b/cmd/schema/main.go
@@ -4,6 +4,7 @@
 //	schema generate   --lang cpp --out <dir> [dir]   emit generated code
 //	schema id         [dir|files...]                 print the protocol id
 //	schema projection [dir|files...]                 print the wire shape the id hashes
+//	schema tables-baseline [--update --reason "..."]  print or move the tables baseline
 //	schema fmt        [dir|files...]                 canonicalize schema files in place
 //	schema version                                   print the build identity
 //
@@ -43,6 +44,10 @@ func main() {
 			fmt.Printf("formatted %s\n", path)
 		}
 	}
+	// Warnings are never quiet and never gated on --verbose: a warning nobody
+	// reads is a warning that does not exist. Today they come from the tables
+	// baseline's warn class (SPEC-TABLES.md §16).
+	c.OnWarn = func(msg string) { fmt.Fprintln(os.Stderr, "warning: "+msg) }
 
 	switch os.Args[1] {
 	// Accept every spelling a newcomer will try. `--version` costs nothing to
@@ -55,6 +60,7 @@ func main() {
 		fs := flag.NewFlagSet("check", flag.ExitOnError)
 		fs.BoolVar(&verbose, "verbose", false, "report the package and protocol id on success")
 		_ = fs.Parse(os.Args[2:]) // ExitOnError: Parse never returns an error
+		c.TablesBaseline = true
 		unit := loadUnit(c, fs.Args())
 		if verbose {
 			fmt.Printf("ok: package %s, protocol id 0x%016x\n", unit.Package, unit.ProtocolId)
@@ -68,6 +74,48 @@ func main() {
 		// and that is a review question, not an implementation detail.
 		unit := loadUnit(c, os.Args[2:])
 		fmt.Print(ir.WireProjection(unit))
+	case "tables-baseline":
+		// The TABLES BASELINE (SPEC-TABLES.md §16). Printing is the default:
+		// the same canonical projection the committed file holds, so a
+		// pipeline can diff without writing anything. --update moves the file
+		// and records WHY in its history — and never without --reason,
+		// because an intentional break that nobody declared is the failure
+		// this file exists to prevent.
+		//
+		// Neither mode runs the baseline check on load: --update is the tool
+		// for moving a baseline the check is refusing, and it cannot be
+		// blocked by the refusal it exists to resolve.
+		fs := flag.NewFlagSet("tables-baseline", flag.ExitOnError)
+		update := fs.Bool("update", false, "rewrite the unit's tables.baseline and record the move in its history")
+		reason := fs.String("reason", "", "why the baseline moves — mandatory with --update, and written into the file's history")
+		fs.BoolVar(&verbose, "verbose", false, "name the file written")
+		_ = fs.Parse(os.Args[2:]) // ExitOnError: Parse never returns an error
+		paths, err := compiler.GatherPaths(fs.Args())
+		if err != nil {
+			fail(err)
+		}
+		u, err := c.Load(paths)
+		if err != nil {
+			fail(err)
+		}
+		if !*update {
+			if *reason != "" {
+				fatalf("--reason belongs with --update: it records why a baseline MOVED")
+			}
+			fmt.Print(compiler.TablesBaselineText(u))
+			break
+		}
+		path, rewrote, err := compiler.UpdateTablesBaseline(u, paths, *reason)
+		if err != nil {
+			fail(err)
+		}
+		if verbose {
+			if rewrote {
+				fmt.Printf("wrote %s\n", path)
+			} else {
+				fmt.Printf("%s is already current\n", path)
+			}
+		}
 	case "fmt":
 		// standalone formatting; check/generate/id already format before
 		// processing, so this exists for editors and pre-commit hooks
@@ -99,6 +147,7 @@ func main() {
 		out := fs.String("out", "generated", "output directory")
 		fs.BoolVar(&verbose, "verbose", false, "list the files emitted")
 		_ = fs.Parse(os.Args[2:]) // ExitOnError: Parse never returns an error
+		c.TablesBaseline = true
 		unit := loadUnit(c, fs.Args())
 		files, err := c.Generate(unit, *lang, compiler.Options{})
 		if err != nil {
@@ -133,6 +182,7 @@ func usage() {
   schema generate   [--lang c|cpp|cs|dart|elixir|go|java|js|rust] [--out generated] [--verbose] [dir|files...]
   schema id         [dir|files...]
   schema projection [dir|files...]
+  schema tables-baseline [--update --reason "..."] [--verbose] [dir|files...]
   schema fmt        [--verbose] [dir|files...]
   schema version
 
--- a/compiler/baseline.go
+++ b/compiler/baseline.go
@@ -5,7 +5,7 @@ import (
 	"github.com/mas-bandwidth/schema/v2/ir"
 )
 
-// TablesBaselineText renders the unit's TABLES BASELINE (SPEC-TABLES.md §16):
+// TablesBaselineText renders the unit's TABLES BASELINE (SPEC-TABLES.md §18.1):
 // a canonical text projection of its table closure — every closure member's
 // fields with their wire ids, kinds, array shapes, evaluated defaults and
 // `was` aliases, and every enum, flags and union it reaches. One fact per
--- a/compiler/baseline.go
+++ b/compiler/baseline.go
@@ -1,0 +1,34 @@
+package compiler
+
+import (
+	"github.com/mas-bandwidth/schema/v2/internal/baseline"
+	"github.com/mas-bandwidth/schema/v2/ir"
+)
+
+// TablesBaselineText renders the unit's TABLES BASELINE (SPEC-TABLES.md §16):
+// a canonical text projection of its table closure — every closure member's
+// fields with their wire ids, kinds, array shapes, evaluated defaults and
+// `was` aliases, and every enum, flags and union it reaches. One fact per
+// line, stable and diffable, exactly as `schema tables-baseline` prints it.
+//
+// It carries no history: the history section belongs to the committed file,
+// and [UpdateTablesBaseline] is what writes it.
+func TablesBaselineText(u *ir.Unit) string {
+	return baseline.Render(u).Text()
+}
+
+// UpdateTablesBaseline rewrites the unit's committed baseline — the
+// tables.baseline beside its schema files — and appends a dated entry to the
+// history section inside it naming every edit that changed what already-written
+// data means, and the reason.
+//
+// The reason is mandatory: moving the baseline is the declaration that a break
+// is intentional, and an undeclared move would defeat the whole point of
+// keeping the file. It is idempotent — when the projection has not moved, the
+// file is untouched and rewrote is false.
+//
+// paths are the unit's *.schema files, as [GatherPaths] returns them; the
+// baseline lives in their directory.
+func UpdateTablesBaseline(u *ir.Unit, paths []string, reason string) (path string, rewrote bool, err error) {
+	return baseline.Update(u, paths, reason)
+}
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -50,7 +50,7 @@ type Compiler struct {
 	// stage the file, or fail the build.
 	OnFormat func(path string)
 
-	// TablesBaseline turns on the TABLES BASELINE check (SPEC-TABLES.md §16):
+	// TablesBaseline turns on the TABLES BASELINE check (SPEC-TABLES.md §18.2):
 	// when a tables.baseline sits in the unit's directory, Load diffs the
 	// unit's table closure against it and REFUSES the edits the table wire
 	// cannot report — a changed specified default, a moved flags bit, a
--- a/compiler/compiler.go
+++ b/compiler/compiler.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/mas-bandwidth/schema/v2/internal/baseline"
 	"github.com/mas-bandwidth/schema/v2/internal/check"
 	"github.com/mas-bandwidth/schema/v2/internal/format"
 	"github.com/mas-bandwidth/schema/v2/internal/parser"
@@ -48,6 +49,20 @@ type Compiler struct {
 	// actually rewrote. The CLI prints "formatted <path>"; a build tool might
 	// stage the file, or fail the build.
 	OnFormat func(path string)
+
+	// TablesBaseline turns on the TABLES BASELINE check (SPEC-TABLES.md §16):
+	// when a tables.baseline sits in the unit's directory, Load diffs the
+	// unit's table closure against it and REFUSES the edits the table wire
+	// cannot report — a changed specified default, a moved flags bit, a
+	// changed field kind. No file means no check. The CLI sets it for `check`
+	// and `generate`.
+	TablesBaseline bool
+
+	// OnWarn, when set, receives each non-fatal report a load produced — the
+	// baseline's warn class (a shrunk bound, a removed enum variant or union
+	// arm): data survives, something is lost, and the runtime read report
+	// counts it. The CLI writes them to stderr.
+	OnWarn func(msg string)
 
 	gens      map[string]Generator // every accepted --lang spelling
 	canonical []string             // one per registered generator, its first name
@@ -127,6 +142,17 @@ func (c *Compiler) Load(paths []string) (*ir.Unit, error) {
 	u, cerrs := check.Unit(files)
 	if len(cerrs) > 0 {
 		return nil, Diagnostics(cerrs)
+	}
+	if c.TablesBaseline {
+		warns, berrs := baseline.Check(u, paths)
+		for _, w := range warns {
+			if c.OnWarn != nil {
+				c.OnWarn(w)
+			}
+		}
+		if len(berrs) > 0 {
+			return nil, Diagnostics(berrs)
+		}
 	}
 	return u, nil
 }
--- a/internal/baseline/baseline.go
+++ b/internal/baseline/baseline.go
@@ -171,6 +171,15 @@ func Render(u *ir.Unit) *Unit {
 		// the branch structure is deliberately not a fact here.
 		for _, f := range st.Fields {
 			t.Fields = append(t.Fields, renderField(f))
+			// an ENUM-KEYED array's key enum is a vocabulary on the wire, not
+			// just a spelling: its slots ride under their variants' NAME ids
+			// (SPEC-TABLES.md §3.2), so its variants are covered exactly as an
+			// enum-typed field's are. It reaches the closure through KeyEnum,
+			// never through the field's own type.
+			if f.KeyEnumRef != nil && !seenEnum[f.KeyEnumRef.Name] {
+				seenEnum[f.KeyEnumRef.Name] = true
+				out.Enums = append(out.Enums, renderEnum(f.KeyEnumRef))
+			}
 			switch ref := f.Type.Ref.(type) {
 			case *ir.Enum:
 				if !seenEnum[ref.Name] {
@@ -241,11 +250,25 @@ func renderField(f *ir.Field) Field {
 			add("type", f.Type.Name)
 		}
 	}
-	switch f.Array {
-	case ir.ArrayFixed:
+	// PRESENCE IS RECORDED AND JUDGED ON NOTHING (SPEC-TABLES.md §18.1): a
+	// field moving between T, ?T and *T moves no byte (§3.1), so the fact is
+	// here to be read in a diff and nowhere in the policy.
+	if f.Type.Optional {
+		add("optional", "true")
+	}
+	switch {
+	case f.KeyEnum != "":
+		// an enum-keyed array is its own wire kind (§3.2) and its slots ride
+		// under their variants' NAME ids, so the key enum is a referent like
+		// any other — and the keyed and positional bodies can never be
+		// decoded as one another, which the kind fact already says.
+		add("array", "keyed")
+		add("bound", strconv.FormatInt(f.ArrayBound, 10))
+		add("key", f.KeyEnum)
+	case f.Array == ir.ArrayFixed:
 		add("array", "fixed")
 		add("bound", strconv.FormatInt(f.ArrayBound, 10))
-	case ir.ArrayCounted:
+	case f.Array == ir.ArrayCounted:
 		add("array", "bounded")
 		add("bound", strconv.FormatInt(f.ArrayBound, 10))
 	}
--- a/internal/baseline/baseline.go
+++ b/internal/baseline/baseline.go
@@ -44,8 +44,11 @@ import (
 
 // Version is the baseline RENDERING's own version, on the file's first line —
 // the same discipline ir.ProjectionVersion keeps for the protocol id. Bumping
-// it makes every committed baseline stale at once, deliberately and visibly.
-const Version = 1
+// it makes every committed baseline stale at once, deliberately and visibly:
+// an older file gets the "regenerate it" refusal, and `--update` regenerates
+// it with its history intact, instead of a flood of diffs against facts this
+// rendering spells differently.
+const Version = 2
 
 // FileName is the baseline's name in the unit directory. Its presence is what
 // turns the check on: no file, no check.
--- a/internal/baseline/baseline.go
+++ b/internal/baseline/baseline.go
@@ -1,4 +1,4 @@
-// Package baseline is the TABLES BASELINE (SPEC-TABLES.md §16): an optional
+// Package baseline is the TABLES BASELINE (SPEC-TABLES.md §18): an optional
 // committed projection of a unit's table closure, and the check that refuses
 // the edits the table wire cannot report.
 //
@@ -48,7 +48,14 @@ import (
 // an older file gets the "regenerate it" refusal, and `--update` regenerates
 // it with its history intact, instead of a flood of diffs against facts this
 // rendering spells differently.
-const Version = 2
+//
+// THE RULE FOR BUMPING IT: any NEW JUDGED TOKEN bumps the version. A token
+// this rendering emits and an older one did not reads as an ADDITION to every
+// older file, and the added-token branch of the diff refuses an addition on
+// every judged row — so an unbumped rendering greets an untouched schema with
+// a refusal per field. Recording a fact nothing judges (an `optional`) does
+// not need a bump; adding a rule does.
+const Version = 3
 
 // FileName is the baseline's name in the unit directory. Its presence is what
 // turns the check on: no file, no check.
--- a/internal/baseline/baseline.go
+++ b/internal/baseline/baseline.go
@@ -221,8 +221,22 @@ func renderField(f *ir.Field) Field {
 	if elem := ir.TableElemKind(f); elem != 0 {
 		add("elem", strconv.Itoa(elem))
 	}
+	// the referent rides under a key that says WHAT it names, because the two
+	// are judged differently: a table's fields ride by id inside their own
+	// body, and a vocabulary's do not ride at all (SPEC-TABLES.md §4 — an
+	// enum field and a plain uint16 field are both kind 7, so the runtime
+	// cannot report an edit between them). See DefaultTokenPolicy.
 	if f.Type.Kind == ir.TNamed {
-		add("type", f.Type.Name)
+		switch f.Type.Ref.(type) {
+		case *ir.Enum:
+			add("enum", f.Type.Name)
+		case *ir.Flags:
+			add("flags", f.Type.Name)
+		case *ir.Union:
+			add("union", f.Type.Name)
+		default:
+			add("type", f.Type.Name)
+		}
 	}
 	switch f.Array {
 	case ir.ArrayFixed:
@@ -231,10 +245,6 @@ func renderField(f *ir.Field) Field {
 	case ir.ArrayCounted:
 		add("array", "bounded")
 		add("bound", strconv.FormatInt(f.ArrayBound, 10))
-		// ENUM-KEYED ARRAYS, when the construct lands: one line here —
-		//     add("key", f.ArrayKeyEnum)
-		// — and the `key` row already sitting in DefaultTokenPolicy makes
-		// swapping the key enum a refusal.
 	}
 	if f.Type.Size != 0 {
 		add("size", strconv.FormatInt(f.Type.Size, 10))
@@ -331,14 +341,14 @@ func Parse(path string, data []byte) (*Unit, error) {
 
 	head := strings.Fields(lines[0])
 	if len(head) != 2 || head[0] != magic {
-		return nil, fmt.Errorf("%s: not a schema tables baseline (its first line must read %q) — regenerate it with: schema tables-baseline --update --reason \"...\"", path, magic+" "+strconv.Itoa(Version))
+		return nil, fmt.Errorf("%s: not a schema tables baseline (its first line must read %q)", path, magic+" "+strconv.Itoa(Version))
 	}
 	v, err := strconv.Atoi(head[1])
 	if err != nil {
 		return nil, fmt.Errorf("%s: unreadable baseline version %q", path, head[1])
 	}
 	if v != Version {
-		return nil, fmt.Errorf("%s: baseline version %d, this compiler writes %d — regenerate it with: schema tables-baseline --update --reason \"...\"", path, v, Version)
+		return nil, fmt.Errorf("%s: baseline version %d, this compiler writes %d", path, v, Version)
 	}
 	u.Version = v
 
@@ -396,6 +406,7 @@ func (u *Unit) parseMemberLine(path string, lineno int, section string, fields [
 			return bad()
 		}
 		f := Field{Name: fields[1]}
+		var haveId bool
 		for _, tok := range fields[2:] {
 			k, val, ok := strings.Cut(tok, "=")
 			if !ok {
@@ -406,10 +417,16 @@ func (u *Unit) parseMemberLine(path string, lineno int, section string, fields [
 				if err != nil {
 					return bad()
 				}
-				f.Id = uint16(id)
+				f.Id, haveId = uint16(id), true
 				continue
 			}
 			f.Tokens = append(f.Tokens, Token{Key: k, Value: val})
+		}
+		// the wire id is the field's IDENTITY here — a line without one names
+		// nothing the diff can match, and a file that carries such a line is
+		// reported rather than half-read
+		if !haveId {
+			return fmt.Errorf("%s:%d: field %s carries no id= — the wire id is a field's identity in this file", path, lineno, fields[1])
 		}
 		t := &u.Tables[len(u.Tables)-1]
 		t.Fields = append(t.Fields, f)
@@ -438,10 +455,24 @@ func (u *Unit) parseMemberLine(path string, lineno int, section string, fields [
 		un := &u.Unions[len(u.Unions)-1]
 		un.Arms = append(un.Arms, v)
 	case "flags":
-		if fields[0] != "variant" || len(u.Flags) == 0 {
+		if fields[0] != "variant" || len(u.Flags) == 0 || len(fields) != 3 {
 			return bad()
 		}
 		fl := &u.Flags[len(u.Flags)-1]
+		// LINE ORDER IS THE FACT, and `bit=` states it in the file so a human
+		// reading a diff does not have to count. The parser holds the two to
+		// each other: a hand-edit that moves a line without moving its bit is
+		// a file this parser cannot read, and it says so rather than guessing
+		// which half the author meant.
+		k, val, ok := strings.Cut(fields[2], "=")
+		if !ok || k != "bit" {
+			return bad()
+		}
+		bit, err := strconv.Atoi(val)
+		if err != nil || bit != len(fl.Variants) {
+			return fmt.Errorf("%s:%d: flags variant %s is written at bit=%s but stands at position %d — bit position is line order here, and the two disagree",
+				path, lineno, fields[1], val, len(fl.Variants))
+		}
 		fl.Variants = append(fl.Variants, fields[1])
 	default:
 		return bad()
--- a/internal/baseline/baseline.go
+++ b/internal/baseline/baseline.go
@@ -1,0 +1,460 @@
+// Package baseline is the TABLES BASELINE (SPEC-TABLES.md §16): an optional
+// committed projection of a unit's table closure, and the check that refuses
+// the edits the table wire cannot report.
+//
+// # WHY IT EXISTS
+//
+// The table wire is evolution-tolerant by construction (SPEC-TABLES.md §4):
+// fields may be added, removed and reordered, variants inserted anywhere,
+// bounds grown, names changed under `was`. Every one of those edits is
+// invisible to a reader in the good sense — nothing is lost and nothing is
+// misread. Two edits are invisible in the bad sense, and they are the reason
+// this file exists:
+//
+//   - A SPECIFIED DEFAULT is part of the wire contract. A field holding its
+//     default is elided, so the reader's declared default is what the absence
+//     MEANS. Change it and every file already written changes meaning, with no
+//     report event, because nothing was lost or skipped.
+//   - A FLAGS variant is positional. A mask carries bits, not names, so
+//     inserting or reordering a variant silently remaps every stored file.
+//
+// The compiler retains no history and cannot see either edit on its own. The
+// baseline is that history, kept where a human can read it: a text file in the
+// unit directory, diffed on every check.
+//
+// # THE SHAPE OF THE ANSWER
+//
+// One representation serves rendering, parsing and diffing: a member is a list
+// of fields, and a field is its name, its wire id and an ORDERED LIST OF
+// TOKENS — `kind=4`, `default=21`, `bound=8`. The whole compatibility policy
+// is then one table, [DefaultTokenPolicy], mapping a token key to what a
+// change of it means. Adding a fact to the projection is a line in the
+// renderer plus a row in that table, which is what keeps the enum-keyed array
+// hook (`key=`) a one-line addition the day the construct lands.
+package baseline
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/mas-bandwidth/schema/v2/ir"
+)
+
+// Version is the baseline RENDERING's own version, on the file's first line —
+// the same discipline ir.ProjectionVersion keeps for the protocol id. Bumping
+// it makes every committed baseline stale at once, deliberately and visibly.
+const Version = 1
+
+// FileName is the baseline's name in the unit directory. Its presence is what
+// turns the check on: no file, no check.
+const FileName = "tables.baseline"
+
+// HistoryHeading opens the intentional-break log inside the file. Everything
+// from this line to the end of the file is history, preserved verbatim by
+// every update.
+const HistoryHeading = "## history"
+
+const magic = "schema-tables-baseline"
+
+// A Unit is one baseline: the projection of a unit's table closure, plus the
+// history the file carries.
+type Unit struct {
+	Version int
+	Package string
+	Tables  []Table
+	Enums   []Enum
+	Flags   []Flags
+	Unions  []Union
+
+	// History is the `## history` section's lines, verbatim and in file
+	// order. Rendering writes them back untouched; an update appends.
+	History []string
+}
+
+// A Table is one member of the table closure: a `table` declaration, or a
+// `type` reached from one. The two are one thing on the table wire — the
+// keyword that declared it changes no byte — so the baseline spells both
+// `table`.
+type Table struct {
+	Name   string
+	Fields []Field
+}
+
+// A Field is one field of a closure member: its declared name, its EFFECTIVE
+// table-wire id (the `was` alias applied), and the wire facts as ordered
+// tokens.
+type Field struct {
+	Name   string
+	Id     uint16
+	Tokens []Token
+}
+
+// A Token is one `key=value` wire fact on a field's line. The key is what
+// [DefaultTokenPolicy] judges.
+type Token struct {
+	Key   string
+	Value string
+}
+
+// Get returns the value of one token and whether the field carries it.
+func (f Field) Get(key string) (string, bool) {
+	for _, t := range f.Tokens {
+		if t.Key == key {
+			return t.Value, true
+		}
+	}
+	return "", false
+}
+
+// An Enum is one enum in the closure: variants in declaration order, each with
+// the wire id it rides under.
+type Enum struct {
+	Name     string
+	Variants []Variant
+}
+
+// A Variant is one enum variant or union arm, with its name-hash id.
+type Variant struct {
+	Name    string
+	Id      uint16
+	Payload string // union arms only
+}
+
+// A Flags is one flags declaration in the closure. The ORDER IS THE FACT:
+// bit i is variant i, and no name rides on the wire at all.
+type Flags struct {
+	Name     string
+	Variants []string
+}
+
+// A Union is one union in the closure: arms in declaration order, each with
+// its name-hash id and its payload type.
+type Union struct {
+	Name string
+	Arms []Variant
+}
+
+// Render projects a checked unit's table closure. The result is exactly what
+// [Unit.Text] writes and what the check diffs against; it carries no history
+// (the file's own history is merged in by the update path).
+func Render(u *ir.Unit) *Unit {
+	out := &Unit{Version: Version, Package: u.Package}
+
+	closure := ir.TableClosure(u)
+	names := make([]string, 0, len(closure))
+	for name := range closure {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+
+	seenEnum := map[string]bool{}
+	seenFlags := map[string]bool{}
+	seenUnion := map[string]bool{}
+
+	for _, name := range names {
+		st := u.Tables[name]
+		if st == nil {
+			st = u.Structs[name]
+		}
+		if st == nil {
+			continue
+		}
+		t := Table{Name: name}
+		// DECLARATION ORDER: st.Fields is the flattened body, branch fields
+		// included. A guard removes a field from the bytes exactly as a
+		// default does, and both are absorbed by the reader's default — so
+		// the branch structure is deliberately not a fact here.
+		for _, f := range st.Fields {
+			t.Fields = append(t.Fields, renderField(f))
+			switch ref := f.Type.Ref.(type) {
+			case *ir.Enum:
+				if !seenEnum[ref.Name] {
+					seenEnum[ref.Name] = true
+					out.Enums = append(out.Enums, renderEnum(ref))
+				}
+			case *ir.Flags:
+				if !seenFlags[ref.Name] {
+					seenFlags[ref.Name] = true
+					out.Flags = append(out.Flags, Flags{Name: ref.Name, Variants: append([]string(nil), ref.Variants...)})
+				}
+			case *ir.Union:
+				if !seenUnion[ref.Name] {
+					seenUnion[ref.Name] = true
+					out.Unions = append(out.Unions, renderUnion(ref))
+				}
+			}
+		}
+		out.Tables = append(out.Tables, t)
+	}
+
+	sort.Slice(out.Enums, func(i, j int) bool { return out.Enums[i].Name < out.Enums[j].Name })
+	sort.Slice(out.Flags, func(i, j int) bool { return out.Flags[i].Name < out.Flags[j].Name })
+	sort.Slice(out.Unions, func(i, j int) bool { return out.Unions[i].Name < out.Unions[j].Name })
+	return out
+}
+
+func renderEnum(e *ir.Enum) Enum {
+	out := Enum{Name: e.Name}
+	for _, v := range e.Variants {
+		out.Variants = append(out.Variants, Variant{Name: v, Id: ir.VariantId(v)})
+	}
+	return out
+}
+
+func renderUnion(un *ir.Union) Union {
+	out := Union{Name: un.Name}
+	for _, v := range un.Variants {
+		out.Arms = append(out.Arms, Variant{Name: v.Name, Id: ir.VariantId(v.Name), Payload: v.Type})
+	}
+	return out
+}
+
+// renderField lists a field's wire facts. The token ORDER here is the file's
+// column order; every key it can emit has a row in [DefaultTokenPolicy].
+func renderField(f *ir.Field) Field {
+	out := Field{Name: f.Name, Id: ir.TableFieldId(f)}
+	add := func(k, v string) { out.Tokens = append(out.Tokens, Token{Key: k, Value: v}) }
+
+	add("kind", strconv.Itoa(ir.TableFieldKind(f)))
+	if elem := ir.TableElemKind(f); elem != 0 {
+		add("elem", strconv.Itoa(elem))
+	}
+	if f.Type.Kind == ir.TNamed {
+		add("type", f.Type.Name)
+	}
+	switch f.Array {
+	case ir.ArrayFixed:
+		add("array", "fixed")
+		add("bound", strconv.FormatInt(f.ArrayBound, 10))
+	case ir.ArrayCounted:
+		add("array", "bounded")
+		add("bound", strconv.FormatInt(f.ArrayBound, 10))
+		// ENUM-KEYED ARRAYS, when the construct lands: one line here —
+		//     add("key", f.ArrayKeyEnum)
+		// — and the `key` row already sitting in DefaultTokenPolicy makes
+		// swapping the key enum a refusal.
+	}
+	if f.Type.Size != 0 {
+		add("size", strconv.FormatInt(f.Type.Size, 10))
+	}
+	if f.HasDefault {
+		add("default", DefaultText(f))
+	}
+	if f.WasName != "" {
+		add("was", f.WasName)
+	}
+	return out
+}
+
+// DefaultText renders a specified default as exact canonical text — the
+// EVALUATED value, never the author's spelling, so a constant that moved
+// through an expression into a default shows up as the value it now produces.
+func DefaultText(f *ir.Field) string {
+	switch {
+	case f.DefVariant != "":
+		return f.DefVariant
+	case f.DefInt != nil:
+		return f.DefInt.String()
+	case f.Type.Kind == ir.TBool:
+		return strconv.FormatBool(f.DefBool)
+	default:
+		// shortest round-tripping form: exact, and stable across builds. The
+		// trailing ".0" is put back on a whole number so a float default
+		// never reads as an integer one.
+		s := strconv.FormatFloat(f.DefFloat, 'g', -1, 64)
+		if !strings.ContainsAny(s, ".eEnN") {
+			s += ".0"
+		}
+		return s
+	}
+}
+
+// Text renders the baseline as the committed file: the projection, then the
+// history section verbatim.
+func (u *Unit) Text() string {
+	var b strings.Builder
+	fmt.Fprintf(&b, "%s %d\n", magic, u.Version)
+	fmt.Fprintf(&b, "package %s\n", u.Package)
+
+	for _, t := range u.Tables {
+		fmt.Fprintf(&b, "\ntable %s\n", t.Name)
+		for _, f := range t.Fields {
+			fmt.Fprintf(&b, "    field %s id=0x%04x", f.Name, f.Id)
+			for _, tok := range f.Tokens {
+				fmt.Fprintf(&b, " %s=%s", tok.Key, tok.Value)
+			}
+			b.WriteString("\n")
+		}
+	}
+	for _, e := range u.Enums {
+		fmt.Fprintf(&b, "\nenum %s\n", e.Name)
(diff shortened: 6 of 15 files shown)


gafferongames:
## Adversarial review — verdict: **MERGE-WITH-CONDITIONS**

Fresh clone, `gh pr checkout 263`, every claim re-run from scratch. The machinery
is sound and every class the PR claims to catch, it catches. What follows is what
it does **not** catch, proven on `tables/examples` against the committed baseline.

---

# Findings

## 1. BLOCKING — renaming a declaration disables the check for everything inside it, in the same commit

`Diff` matches closure members by NAME and `continue`s when a baseline member is
not in the live projection (`internal/baseline/diff.go`, four sites: tables,
`diffEnums`, `diffUnions`, `diffFlags`). A declaration rename moves no byte — the
table name is not on the wire (SPEC-TABLES §3) — so it is correctly absorbed. But
it also silently drops **every field and every variant of that declaration out of
coverage for that commit**, and a rename plus a semantic edit is one ordinary
refactor.

The flagship refusal, defeated by renaming the table it lives in:

```
$ perl -pi -e 's/damage      float32 = 21\.0/damage      float32 = 999.0/' Tables.schema
$ /tmp/schemabin check .
tables.baseline: WeaponConfig.damage: specified default 21.0 -> 999.0 — this edit changes
what data already written MEANS, and no reader can report it; ... (SPEC-TABLES.md §16)
schema: 1 error(s)
[exit 1]

$ # the same edit, plus `WeaponConfig` -> `Weapon`:
$ /tmp/schemabin check .
[exit 0]
```

and the projection shows the default really did move, unreported:

```
$ diff <committed baseline> <(schema tables-baseline .)
47,48c47,48
< table WeaponConfig
<     field damage id=0x15a9 kind=10 default=21.0
---
> table Weapon
>     field damage id=0x15a9 kind=10 default=999.0
```

The same hole in all four vocabularies, each a single commit:

| edit | verdict |
|---|---|
| table DECL renamed + a specified default changed | **PASS** |
| flags DECL renamed + bits reordered | **PASS** |
| enum DECL renamed + a variant dropped | **PASS** |
| union DECL renamed + an arm dropped | **PASS** |

`TestAbsorbedEdits` pins "a whole table removed from the closure" as silent — that
decision is right, but it was only tested for *removed*, never for *renamed*, and
the code cannot tell them apart.

**Fix, small:** when a baseline member or vocabulary name is not present in the
live projection, emit a **Warn** ("`WeaponConfig` is no longer in the closure — its
fields are no longer covered by this baseline"). Removals stay legal, the coverage
hole stops being invisible, and a rename is one line of output away from being
obvious in review.

## 2. BLOCKING — `"type": RulePass` lets a field's flags/enum/union type be swapped silently

`DefaultTokenPolicy` justifies `type` as RulePass with *"the nested member has its
own section; its fields ride by id"*. That is true of a nested **table**. It is
false of a flags, an enum or a union: swapping the named type is exactly the remap
§16 exists to refuse.

```
$ # add `flags Traits { Turbo, Shielded, Cloaked }` and point the field at it
$ diff <committed baseline> <(schema tables-baseline .)
22c22
<     field perks id=0x2fc5 kind=9 type=Perks
---
>     field perks id=0x2fc5 kind=9 type=Traits
65,68c65,68
< flags Perks                    > flags Traits
<     variant Shielded bit=0     >     variant Turbo bit=0
<     variant Cloaked bit=1      >     variant Shielded bit=1
<     variant Turbo bit=2        >     variant Cloaked bit=2

$ /tmp/schemabin check . ; echo $?
0
```

The file records the entire bit remap and the check says nothing. Same for enums:

| edit | verdict |
|---|---|
| field's flags TYPE swapped for a reordered flags decl | **PASS** — every stored mask remapped |
| field's enum TYPE swapped for a different vocabulary | **PASS** — cf. "variant removed", which WARNS |
| enum-typed array element -> raw `uint16` | **PASS** |

The last one deserves its own line: SPEC-TABLES §4 states explicitly that an enum
field and a `uint16` field are both kind `7`, *"so an edit between the two is not a
kind mismatch"* — i.e. the runtime cannot report it. That is the definition of this
baseline's job, and it is the one such edit the baseline also cannot report.

```
$ perl -pi -e 's/podium      \[3\]Grade/podium      [3]uint16/' Tables.schema
$ /tmp/schemabin check . ; echo $?
0
```

**Fix:** judge `type` when the referent is a flags/enum/union (RuleFixed, or Warn
at minimum), and keep RulePass for a table referent, where the by-id argument holds.

## 3. The `--update` remedy the diagnostics advertise does not work

Every parse failure ends with *"regenerate it with: schema tables-baseline --update
--reason"*. `Update` calls `Parse` first and returns its error, so the advertised
remedy reproduces the identical message:

```
$ printf 'hello world\n' > tables.baseline
$ schema check .
tables.baseline: not a schema tables baseline (its first line must read
"schema-tables-baseline 1") — regenerate it with: schema tables-baseline --update --reason "..."
$ schema tables-baseline --update --reason "regenerating as instructed" .
schema: tables.baseline: not a schema tables baseline (its first line must read
"schema-tables-baseline 1") — regenerate it with: schema tables-baseline --update --reason "..."
[exit 1]
```

Identical loop on the future-version path — which is the one that matters, because
it is the path a `Version = 2` bump puts **every committed baseline in the estate**
on:

```
$ schema check .
tables.baseline: baseline version 2, this compiler writes 1 — regenerate it with: ...
$ schema tables-baseline --update --reason "regenerating as instructed" .
schema: tables.baseline: baseline version 2, this compiler writes 1 — regenerate it with: ...
[exit 1]
```

The only escape is `rm tables.baseline`, which destroys the `## history` section —
the artifact this PR calls *"the one record the wire lacks"*. **Fix:** on a version
or magic mismatch, `--update` should salvage the `## history` lines verbatim and
rewrite the projection (it already knows where the heading is), or the message
should stop naming a command that cannot run.

## 4. `payload=` is rendered and parsed, then never compared

`Render` writes `arm buff id=0xeae6 payload=Buff`, `parseMemberLine` reads it into
`Variant.Payload` — and `diffUnions` compares arm names only. Swapping an arm's
payload type is a kind change one level in, which is precisely the reasoning
`"elem": RuleFixed` is built on ("*an array's element kind is that same fact, one
level in*"):

```
$ perl -pi -e 's/    buff   Buff$/    buff   Debuff/' Tables.schema
$ /tmp/schemabin check . ; echo $?
0
```

## 5. The PR body's negative-control claim is not what the tests do

> *"One fixture pair per refusal class and per warn class, **each with two negative
> controls**"*

11 refusal cases; 6 carry `token:` (attribution). The 5 that do not are **all five
flags cases** — the flagship refusal, and the one walk whose logic is hand-written
outside `DefaultTokenPolicy`. 2 of 4 warn cases likewise (enum variant removed,
union arm removed). `Diff` gives `diffFlags`/`diffEnums`/`diffUnions` no policy
seam, so `without()` cannot reach them.

I built the missing attribution control by ablating `diffFlags` to `return nil` and
rebuilding — **it holds**:

```
flags inserted                exit=0 (silent)
flags reordered               exit=0 (silent)
flags renamed                 exit=0 (silent)
default changed               exit=1 tables.baseline: WeaponConfig.damage: specified default 21…
kind changed                  exit=1 tables.baseline: ProfileConfig.experience: wire kind 8 -> …
enum variant removed (warn)   exit=0 warning: ... variant Gold removed…
```

So the check is right and the claim is wrong. Either add a seam and the control, or
say "discrimination control on every class, attribution control on the token-policy
classes".

## 6. Consistency question for the owner, not a defect

`unknown` is one runtime counter over field ids, enum variant ids and union arm ids
(SPEC-TABLES §4). This PR **warns** on a removed enum variant / union arm, on the
stated ground that the read report counts what is lost — and passes a **removed
field in silence**, which produces the same `unknown` count. Both are faithful to
#256's own text. Worth one ruling so the rationale in §16 discriminates.

Related, and answering the question posed: an enum variant **renamed** warns
(remove+add on a name-hash identity), and a union arm renamed warns. That reads
correct under the invariant — the value becomes unnameable, reads `None`, counts
`unknown`. Dropping a `was` alias, by contrast, is silent:

```
$ perl -pi -e 's/speed       float32 = 500\.0 \| was = "velocity"/speed       float32 = 500.0/' Tables.schema
$ /tmp/schemabin check . ; echo $?
0
```

The baseline holds the exact datum (`was=velocity`, `id=0x2e46`) needed to say so,
and matching by id turns it into an ordinary field removal. Given #6 above, the
same ruling covers both.

## 7. Minor

- `bit=N` is written into the file and **never read** by `Parse` — line order is
  the fact. Hand-reordering two `variant` lines while keeping their `bit=` values
  correct produces a refusal for a reorder that never happened. Safe direction
  (false refusal, not false pass), but it contradicts the file comment *"deliberately
  strict: a file this parser cannot read is reported rather than guessed at"*.
- `"key": RuleFixed` is a live policy row for a token no renderer emits, plus a
  forward-looking `// ENUM-KEYED ARRAYS, when the construct lands` note in source.
  Reasonable given #265 is in flight, but it is scaffolding and a non-present-state
  comment under the documentation doctrine.
- 20 new identifiers land on the semver-covered surface (15 `ir.TableKind*` + 3 `ir`
  funcs, 2 `compiler` funcs, 2 `Compiler` fields) with no VERSIONING note. Minor per
  policy, but worth stating.

## 8. §16 collision — flagged, not a defect of this PR

PR #267 (`tables-json-walk`, #253, draft) also numbers its new section `## 16.` in
SPEC-TABLES.md. Whichever lands second must renumber, and **this PR cites §16 from
14 places in 11 files** — including runtime error text a user sees. Sequence
deliberately.

---

# Verified good

Every one of these re-run from scratch, not taken from the PR body.

**All nine refusal classes REFUSE, naming `table.field` and the edit** — default
changed / added / removed; flags inserted / removed / reordered / renamed-in-place;
field kind changed; array element kind changed. Messages are precise and name the
consequence.

**All four warn classes WARN and do not stop the build** — bound shrunk, capacity
shrunk, enum variant removed, union arm removed. Discrimination controls (bound
grown, flags appended, bounded->fixed, field added) are silent.

**The owner's explicit scope item — evaluated values — is met.** A default written
as `BaseDamage + Bonus` records as `default=21`, and moving the constant is caught:

```
$ perl -pi -e 's/const Bonus      = 11/const Bonus      = 15/' Save.schema
$ schema check .
tables.baseline: Weapon.damage: specified default 21 -> 25 — this edit changes what data
already written MEANS ... [exit 1]
```

Same for a capacity declared through a constant expression (warns, 20 -> 16).

**`--update` mechanics.** Bare `--update` refused; whitespace-only `--reason`
refused; `--reason` without `--update` refused; history entry records
`old -> new [refuse]` / `[warn]` per edit beside the reason and date; re-running is
byte-identical with no new entry ("already current"); both committed corpus
baselines regenerate byte-for-byte through the CLI; `tables-baseline` deliberately
does not run the check, so it can move a baseline the check is refusing.

**Foreign/corrupt/future baselines refused by name**, each with its own message:
future version, non-baseline file, empty file, wrong package, unknown section,
unreadable member line.

**Robustness:** 800 corrupted baselines (400 random truncations + 400 random byte
flips of the committed file) through `schema check` — **0 panics**.

**The kind-vocabulary move is provably behaviour-preserving.** `cpptable`'s
`tableScalarKind` body moved to `ir.TableScalarKind` verbatim, constants identical;
generated C++ from `tables/examples` and `tables/pointers` is **byte-identical**
between `main` and this branch (`diff -r`, both corpora).

**Public API boundary intact.** `go list -f '{{.Imports}}' ./cmd/schema` →
`compiler` and `ir` only. `internal/publicapi` (external-module CLI rebuild, now
covering `tables-baseline`), `TestTablesMoveNoProtocolId`, and
`TestZeroCostForValueOnlyTables` all green; full `go test ./...` green; shape gate
clean (the `time.Now()` ledger entry is one call, a `2006-01-02` format string, no
interval measured — the justification is correct); all 8 CI checks green.

**My own ablation control:** stubbing `baseline.Check` to `return nil, nil` turns
every fixture above silent (exit 0) and turns the suite red in exactly the file-side
tests (`TestUpdateRecordsTheBreak`, `TestCheckReportsWarningsWithoutRefusing`,
`TestCheckRefusesAForeignBaseline`). Every finding is attributable to the check.

**Docs.** SPEC-TABLES §16 and the §11 row are present-state reference register.
USAGE's worked example is not illustrative — it reproduces **verbatim**:

```
$ schema tables-baseline .
schema-tables-baseline 1
package shipdemo

table ShipConfig
    field damage id=0x15a9 kind=10 default=21.0
    field speed id=0x2e46 kind=10 default=500.0 was=velocity
    field name id=0x30df kind=12 size=32
```

**Provenance clean.** 8 commits, author and committer `rowan@mas-bandwidth.com` on
every one. No strays.

---

# Verdict

**MERGE-WITH-CONDITIONS.** The design is right, the fact set is right, the gate is
honest, and it is opt-in so nothing in the estate regresses. But the owner's
invariant is *"the save games would have to not ever break compatibility with stuff
already written"*, and findings **1** and **2** let a one-token edit inside an
ordinary refactor walk straight past the two refusals the whole file exists for —
a renamed declaration, or a swapped flags type. Both fixes are small and local to
`diff.go`.

Conditions before merge:
1. A **Warn** when a baseline member or vocabulary name is no longer in the live
   projection (finding 1).
2. `type` judged when the referent is a flags/enum/union (finding 2).
3. Either make `--update` able to run on a baseline the parser rejects, salvaging
   `## history`, or stop advertising it in that message (finding 3).

Follow-ons, not blocking: compare `payload=` (4); the negative-control claim (5);
the field-removal vs variant-removal ruling (6); the `bit=` nit (7).
And sequence the §16 numbering against #267 (8).

gafferongames:
All three conditions are met, plus the follow-ons. Every scenario in the review
re-run against the branch.

## 1. A declaration rename no longer disables the check

Members match by NAME first, then across a rename by identity overlap: the
unmatched live declaration of the same kind carrying the most of the vanished
member's identities, when that is at least half of them and unique. A vanished
name WARNS either way — a paired one saying where it went, an unpaired one
saying nothing carries its identities any more — so the coverage hole is never
silent, and where it pairs the contents are judged under the new name.

The flagship case from the review, on `tables/examples`:

```
$ perl -pi -e 's/damage      float32 = 21\.0/damage      float32 = 999.0/;
               s/\bWeaponConfig\b/Weapon/g' Tables.schema
$ schema check .
warning: tables.baseline: table WeaponConfig: no longer in the closure under that
name; Weapon carries 6 of its 6 identities, so it is judged as the rename it looks like
tables.baseline: Weapon.damage: specified default 21.0 -> 999.0 — this edit changes
what data already written MEANS ... (SPEC-TABLES.md §16)
schema: 1 error(s)   [exit 1]
```

All four rows of your table are now caught, each a fixture pair with a
discrimination control and an attribution control (`TestVanishedMembers`), plus
`TestVanishedMemberWithNoSuccessor` for the removal that really is a removal.

## 2. Referents are judged, per vocabulary

A field's referent now rides under a key that says WHAT it names — `type=`,
`enum=`, `flags=`, `union=` — and each is judged on whether that vocabulary's
own identities ride (SPEC-TABLES §3):

| referent | rides when |
|---|---|
| table (`type=`, and a union arm's `payload=`) | every field id the old one carried is still carried — a body decodes by id |
| enum | every variant NAME survives — a value is its name hash |
| union | every arm NAME survives — a body opens with its arm's name hash |
| flags | the old variants sit at the SAME BITS — a mask carries no names at all |

Dropping the referent outright refuses on its own line, because that is the one
§4 calls out as unreportable:

```
$ perl -pi -e 's/podium      \[3\]Grade/podium      [3]uint16/' Tables.schema
$ schema check .
tables.baseline: LoadoutConfig.podium: enum Grade removed — the field keeps its wire
kind, so no reader can report the change (SPEC-TABLES.md §4) ...   [exit 1]
```

and the flags swap now reports the remap it found:

```
tables.baseline: LoadoutConfig.perks: flags Perks -> Traits, and bit 0 was Shielded
and is Turbo under Traits — a mask carries bits, not names, so every stored file is
remapped silently ...   [exit 1]
```

This subsumes **finding 4**: an arm's payload is a table body one level in, so
it is the same rule rather than a second one that could drift from it.

```
$ perl -pi -e 's/^    buff   Buff$/    buff   Debuff/' Tables.schema
$ schema check .
tables.baseline: union Effect.buff: arm payload Buff -> Debuff, and 1 of 1 field ids
do not ride — that much of every stored body reads back as declared defaults [exit 1]
```

## 3. `--update` is now the remedy it advertises

`Update` parses LENIENTLY: it salvages the `## history` lines verbatim,
regenerates the projection from the unit, and records that the previous
projection could not be diffed. The version-bump path — the one a `Version = 2`
bump puts the whole estate on:

```
$ schema check .
tables.baseline: baseline version 2, this compiler writes 1 — regenerate it with:
schema tables-baseline --update --reason "...", which preserves the ## history section
$ schema tables-baseline --update --reason "the rendering version moved; regenerated" .
$ sed -n '/## history/,$p' tables.baseline
## history
### 2024-01-01 — the break nobody may forget
- Weapon.damage: 10.0 -> 21.0 [refuse]

### 2026-09-02 — the rendering version moved; regenerated
- baseline REGENERATED over an unreadable one (baseline version 2, this compiler
  writes 1); the projection is this unit as it stands and the history above is
  preserved, but the previous projection could not be diffed, so no per-edit lines follow
$ schema check .   [exit 0]
```

The remedy is also named in ONE place now, so every parse refusal carries it.

## 5. The negative-control claim is now true

The four vocabulary walks are gated on their own policy rows — `member`,
`flags-position`, `enum-variant`, `union-arm` — so `without()` reaches them and
every class has an attribution control, the five flags cases included. The
control asserts the NAMED finding is gone rather than total silence, because an
edit can legitimately trip two rows at once (respelling an enum-typed field as
`uint16` also drops its default, and both refusals are correct).

## 7. Minor

- `bit=` is now READ and held to line order: two variant lines swapped without
  their `bit=` values is a file the parser refuses, saying so. A field line with
  no `id=` is refused too — the wire id is a field's identity here.
- The forward-looking `// ENUM-KEYED ARRAYS, when the construct lands` comment
  is gone. The `key` policy row stays with its test, under a present-state
  comment: the table describes the fact set, not the grammar.
- VERSIONING.md now records the new semver-covered surface.

## 6 and 8

**Finding 6** is in the PR body under *Open for the owner*, unchanged in
behaviour — a removed field passes, a removed variant warns, and both produce
the same `unknown` count. Worth one ruling.

**Finding 8**: numbering follows the spec-of-record PR; nothing renumbered here.
Stated in the PR body.

One thing I fixed that the review did not raise: the salvage history entry
embedded the absolute path of the file it repaired. That file is read on another
machine, years later.

All 8 CI checks green.

gafferongames:
## Delta review — three repair commits re-verified from a fresh clone

Fresh `git clone` + `gh pr checkout 263`, everything below reproduced rather than
read. **Findings 1, 3, 4, 5 and the `bit=` nit are closed.** Finding 2 is closed
for enum, flags and union referents and **one case of it survives on the table
referent** — repro below. Verdict at the end.

---

# Remaining blocker — a table referent rides on ID MEMBERSHIP, not on what those ids MEAN

`identitiesRide` answers `"type"`/`"payload"` with `ridesBySet(tableIdents(base).sets[was], tableIdents(live).sets[now], …)`
— set membership of field ids and nothing more. SPEC-TABLES §16 now states the same
rule: *"a nested table … rides when every field id the old one carried is still
carried."* That condition is **not sufficient**. A field id riding is not the same
as the field meaning what it meant.

Point a nested field at a twin declaration that carries the same ids under
different facts, and keep the original referenced elsewhere so nothing vanishes and
no rename pairing fires:

```
package twindemo

type Buff   { multiplier float32 = 1.0 }
table Holder { keep Buff
               swap Buff }
```
```
$ schema tables-baseline --update --reason "first baseline" .
$ sed -n '/^table Holder/,/^$/p' tables.baseline
table Holder
    field keep id=0x7a33 kind=13 type=Buff
    field swap id=0xe3a3 kind=13 type=Buff
```

Now `swap` alone is repointed at a twin whose one shared id carries a different
**specified default**:

```
type Buff { multiplier float32 = 1.0 }
type Boon { multiplier float32 = 7.0 }     // same id 0x32e0, different default
table Holder { keep Buff
               swap Boon }

$ schema check . ; echo $?
0
```

Silent. No warning, no refusal. Every stored `Holder` body's elided `multiplier`
inside `swap` now means 7.0 instead of 1.0 — the flagship class of #256, intact.
The same shape with a changed **kind** is equally silent:

```
type Boon { multiplier float64 = 1.0 }     // kind 10 -> 11 inside the body
$ schema check . ; echo $?
0
```

**The two halves of this PR's own idea already disagree about it.** When the old
declaration *does* vanish, `pairMembers` pairs it with the twin and judges it in
full — a changed default on the shared field refuses:

| edit | verdict |
|---|---|
| `buff Buff` -> twin carrying every old id, identical facts | pass (+ the vanished warn) |
| `buff Buff` -> twin carrying every old id, **different default** — *twin vanishes* | **REFUSE** |
| `swap Buff` -> twin carrying every old id, **different default** — *original stays referenced* | **PASS, silent** |

So the strict answer already exists in the codebase; the referent path just doesn't
reach it.

**Fix, local:** `identitiesRide` for `"type"` and `"payload"` should compare the two
referents' field TOKEN SETS under the shared ids — `diffTokens` is already written
and already carries the whole policy — rather than comparing id sets. Answering
"the ids ride" is the right question asked one level too shallow.

To be explicit on the question put to me: **a table-typed field swapped to another
table carrying every old id is safe only when the facts under those ids are also
unchanged.** With that added, the pass is correct and I would sign it off — it is
genuinely the same table under another name. As it stands the pass is unguarded.

---

# Closed, with the repro

## Finding 1 — the rename hole. Closed, with defence in depth.

All four rows now WARN on the vanished name and REFUSE the edit under the paired
name:

```
--- table DECL renamed + specified default changed
warning: table WeaponConfig: no longer in the closure under that name; Weapon carries
6 of its 6 identities, so it is judged as the rename it looks like
tables.baseline: Weapon.damage: specified default 21.0 -> 999.0 — ... [exit 1]

--- flags DECL renamed + bits reordered
warning: flags Perks: no longer in the closure under that name; Traits carries 3 of 3 …
tables.baseline: LoadoutConfig.perks: flags Perks -> Traits, and bit 0 was Shielded and
is Cloaked under Traits — a mask carries bits, not names ...
tables.baseline: flags Traits: variant Shielded moved from bit 0 to bit 1 ... [exit 1]

--- enum DECL renamed + a variant dropped        -> warn + 3 refusals [exit 1]
--- union DECL renamed + an arm dropped (1 of 2) -> warn + refusal    [exit 1]
```

**The half edge holds.** `bestScore*2 >= len(want)` is measured against the
BASELINE's identity count, so adding fields never dilutes and only dropping them
does. Exactly half pairs and is judged in full:

| overlap | paired? | the semantic edit inside |
|---|---|---|
| 6 of 6 | yes | caught |
| 3 of 6 (exactly half) | yes | **caught** — `Weapon.damage: specified default 21.0 -> 999.0` |
| 2 of 6 (below half) | no | **still caught**, by the referent rule at each referencing field |
| 0 of 6 | no | **still caught**, same |

**On "warn only" at zero / below-half overlap — acceptable, with one message fix.**
A REFERENCED declaration is protected regardless of pairing: the referent rule fires
at every field that names it, so 2-of-6 and 0-of-6 both refuse (`nested table
WeaponConfig -> Weapon, and 4 of 6 field ids do not ride`). The warn-only path is
therefore confined to an UNREFERENCED root-level declaration. I built that case
deliberately — a root table renamed, 3 of 4 fields dropped, and the survivor's
default changed in the same commit:

```
warning: tables.baseline: table SaveFile: no longer in the closure — nothing left
carries its identities, so this baseline no longer covers it
[exit 0]                      # the default change on the surviving field is silent
```

I judge warn-only **correct** here. Pairing below half would mis-pair unrelated
declarations and produce false refusals, which is what gets a baseline deleted;
refusing every vanished name would refuse the removals #256 rules absorbed. The
warning names the table and says coverage is gone, which is the honest answer.

But the message is wrong in exactly this case: `Archive` demonstrably carries 1 of
the 4 identities and the text says *"nothing left carries its identities"*. That
sentence is only true at zero overlap, and on this path the message is the **only**
thing standing between the user and the hole. It should say what it actually found —
the closest candidate and its score, below the half needed to call it a rename.

## Finding 2 — referents. Closed for the three vocabularies.

| edit | verdict |
|---|---|
| `[3]Grade` -> `[3]uint16` (referent dropped; §4: kind 7 both) | **REFUSE** |
| `grade Grade` -> `uint16` | **REFUSE** |
| `perks Perks` -> `uint64` | **REFUSE** |
| field's flags type -> a REORDERED flags decl | **REFUSE** |
| field's flags type -> a PREFIX-IDENTICAL decl (bits ride) | pass (+ vanished warn) |
| field's enum type -> a DIFFERENT vocabulary | **REFUSE** |
| field's enum type -> a SUPERSET (all names ride) | **PASS** |
| union arm `payload` swapped (finding 4) | **REFUSE** |

The prefix rule for flags and the name-set rule for enum/union are each the right
identity rule for that vocabulary, and the discriminations are clean.

## Finding 3 — `--update` is the remedy it advertises. Closed.

```
$ schema check .
tables.baseline: baseline version 2, this compiler writes 1 — regenerate it with:
schema tables-baseline --update --reason "...", which preserves the ## history section

$ schema tables-baseline --update --reason "regenerating after a rendering-version bump" /abs/path/sal
wrote /abs/path/sal/tables.baseline
```
```
## history
### 2024-01-01 — the record the wire lacks
- an intentional break nobody wants to lose

### 2026-09-02 — regenerating after a rendering-version bump
- baseline REGENERATED over an unreadable one (baseline version 2, this compiler
  writes 1); the projection is this unit as it stands and the history above is
  preserved, but the previous projection could not be diffed, so no per-edit lines follow
```

History salvaged verbatim, the loss of diff fidelity recorded honestly, and the
check then passes. Same on a foreign non-baseline file with a history section.
**No machine paths in the file** even when `--update` is invoked with an absolute
path (`grep` for the tmp prefix, `/Users/`, `/private/`: none).

## Finding 5 — attribution seams. Closed; I re-ran them myself.

Every walk is now gated on its own policy row, so I wrote my own controls against
the exported map rather than reusing the PR's. All eleven classes are singly
ablatable:

```
--- PASS: TestReviewerAttribution/flags_inserted
--- PASS: TestReviewerAttribution/flags_reordered
--- PASS: TestReviewerAttribution/flags_renamed_in_place
--- PASS: TestReviewerAttribution/flags_variant_removed_mid
--- PASS: TestReviewerAttribution/last_flags_variant_removed
--- PASS: TestReviewerAttributionWalks/enum_variant_removed
--- PASS: TestReviewerAttributionWalks/table_renamed_->_member_warn
--- PASS: TestReviewerAttributionReferents/flags_referent_dropped
```

Three cases needed a second row deleted, and in each the residue was a *different
legitimate row* co-firing on the same edit, not an attribution leak — union-arm
removal also drops `Debuff` from the closure (`member`), `grade Grade = Silver` ->
`uint16` also drops the default (`default`), `effect Effect` -> `uint16` also moves
the kind (`kind`). Deleting the pair leaves nothing. The flags class — the flagship,
and the one I had to ablate by hand last round — now has a real seam.

## Finding 7 (`bit=`). Closed, and better than I asked for.

```
tables.baseline:66: flags variant Cloaked is written at bit=1 but stands at position 0
— bit position is line order here, and the two disagree — regenerate it with: ...
```
The parser now holds line order and `bit=` to each other instead of ignoring one.
VERSIONING.md also gained the export note.

---

# Conditions and nits

1. **Renaming raises the verdict.** The same wire loss gets two verdicts depending
   on whether you also renamed the declaration:

   | | alone | + declaration renamed |
   |---|---|---|
   | enum variant `Gold` dropped | WARN | **REFUSE** |
   | union arm `debuff` dropped | WARN | **REFUSE** |

   It errs safe and the message is accurate, but §16 says a rename means the
   contents *"keep being judged under its new name"* — and they are judged more
   harshly instead. Either state the asymmetry or let the referent rule warn where
   the vocabulary walk warns.

2. **The unpaired warning's text.** See finding 1 above — "nothing left carries its
   identities" is asserted when a candidate carries some, just below half.

3. **`Version` was not bumped though the rendering moved** (`type=` ->
   `enum=`/`flags=`/`union=`; the corpus baseline changed with it). That constant
   exists for exactly this. A file written by the previous build of this branch does
   not get the designed "regenerate it" message — it gets a flood of false refusals
   with the wrong noun:

   ```
   tables.baseline: LoadoutConfig.grade: nested table Grade removed — the field keeps
   its wire kind, so no reader can report the change (SPEC-TABLES.md §4)
   tables.baseline: LoadoutConfig.grade: enum Grade added
   … 6 more
   ```
   No released build has ever written a baseline, so the blast radius is this branch
   only — but it is a one-character fix that turns a confusing failure into the
   designed one.

4. **Both PASS lists are now wrong on renames.** SPEC-TABLES §16 lists, under
   **"PASS, in silence"**, *"a DECLARATION renamed, whose contents keep being judged
   under its new name"*, and USAGE repeats it (*"bounds grown; a declaration
   renamed"*). Measured, a bare rename warns:

   ```
   warning: tables.baseline: table WeaponConfig: no longer in the closure under that
   name; Weapon carries 6 of its 6 identities, so it is judged as the rename it looks like
   ```
   §16 self-corrects two paragraphs later ("Either way the vanished name WARNS");
   USAGE's summary does not. A rename belongs in the WARN list in both.

5. **§16 still collides with #267** (`tables-json-walk`, #253), and this PR now
   cites §16 from more places than before. Unchanged from my last comment.

---

# Otherwise green, all re-run

No regressions: the original 9 refusal classes still refuse, the 4 warn classes
still warn, the discrimination controls (bound grown, flags appended, variant
inserted mid, `was` rename, bounded->fixed) are still silent. Full `go test ./...`
green; `TestTablesMoveNoProtocolId` and `TestZeroCostForValueOnlyTables` green;
shape gate clean; `gofmt -l` and `go vet` clean; all 8 CI checks green. Generated
C++ still **byte-identical to `main`** on both tables corpora — the independence
holds through the delta. Both committed corpus baselines regenerate byte-for-byte.
600 corrupted baselines (300 truncations + 300 byte flips) through `schema check`:
**0 panics**. Provenance: the three new commits are all
`rowan@mas-bandwidth.com`, author and committer, no strays.

---

# Verdict

**One blocker left, and it is one function.** The rename hole is properly closed —
better than I proposed, because the pairing rule and the referent rule cover each
other, and I could not get past both. `--update`, the seams and the `bit=` nit are
clean. The residue is that a table referent is judged on which ids exist rather
than on what they mean, which lets a twin declaration carry a changed default or a
changed kind through in silence — the flagship class, still reachable.

Make `identitiesRide` compare the referents' field facts under the shared ids
(`diffTokens` already does it, and `pairMembers` already reaches the strict answer
by the other path) and this is a **MERGE** from me. Nits 1–4 are small and worth
taking in the same pass; nit 5 is a sequencing call, not a code change.

Noting for the record that this stays HELD behind #269 regardless.

gafferongames:
Blocker closed, and the four nits with it. Everything below re-run against the
branch.

## The blocker — a referent must STAND IN, not merely carry the same ids

You were right that the question was asked one level too shallow. `identitiesRide`
is gone; the table/payload path now compares the two declarations' FIELD FACTS
under the shared ids, and it does it through **`diffTokens` itself**, so the
referent path and the in-place path cannot drift apart about what a fact means.
Your repro, verbatim:

```
$ sed -n '/^table Holder/,/^$/p' tables.baseline
table Holder
    field keep id=0x7a33 kind=13 type=Buff
    field swap id=0xe3a3 kind=13 type=Buff

$ # swap -> a twin carrying id 0x32e0 with default 7.0
$ schema check .
tables.baseline: Holder.swap: nested table Buff -> Boon, and multiplier's specified
default 1.0 -> 7.0 — this edit changes what data already written MEANS ...  [exit 1]

$ # the same twin with kind float64 instead of float32
tables.baseline: Holder.swap: nested table Buff -> Boon, and multiplier's wire kind
10 -> 11 — ...  [exit 1]

$ # the control: a superset twin, every id AND every fact riding
$ schema check . ; echo $?
0
```

A missing id still refuses on its own line (`1 of 1 field ids do not ride`), so
finding 4's payload case is unchanged. Fixtures: both twins, on a nested field and
on a union arm's payload, each with the substitutable superset as its
discrimination control and its own attribution control.

## Nit 1 — renaming no longer raises the verdict

The two rules were double-judging renames. They are now separated on the page and
in the code: **a declaration edited IN PLACE is judged by its walk** (verdicts
follow what the runtime can report); **a field REPOINTED at a different
declaration is judged by the referent rule** (which asks about substitutability).
A rename is the first wearing the second's clothes, so a **paired** rename is left
to the walk.

| | alone | + declaration renamed |
|---|---|---|
| enum variant `Gold` dropped | WARN | **WARN** |
| union arm `debuff` dropped | WARN | **WARN** |

`TestRenamingDoesNotRaiseTheVerdict` holds the two equal by asserting zero
refusals on both sides. Nothing you signed off downgraded: the flags-type swap,
the enum-type swap to a different vocabulary and the dropped-referent cases all
still refuse — they are unpaired repoints, which is what the referent rule is for.
The fixtures for those now keep the original declaration referenced by a second
field, so they exercise the repoint rather than an accidental rename.

## Nit 2 — the unpaired message says what it found

```
warning: tables.baseline: table Config: no longer in the closure — the closest
declaration, Archive, carries 1 of its 13 identities, below the half needed to call
it a rename, so this baseline no longer covers it
```

The zero-overlap wording is now only used at zero overlap ("no declaration carries
any of its identities"), and both branches have a test.

## Nit 3 — `Version` bumped to 2

```
$ schema check .          # a file written by the previous build of this branch
tables.baseline: baseline version 1, this compiler writes 2 — regenerate it with:
schema tables-baseline --update --reason "...", which preserves the ## history section

$ schema tables-baseline --update --reason "the rendering moved to 2" .
$ sed -n '/## history/,$p' tables.baseline
### 2026-09-02 — the first baseline over the tables corpus
- baseline created over 9 tables — data written BEFORE this point is not covered by it

### 2026-09-02 — the rendering moved to 2
- baseline REGENERATED over an unreadable one (baseline version 1, this compiler
  writes 2); ... the history above is preserved ...
$ schema check .   [exit 0]
```

The designed message instead of the flood, exactly as you described. Both committed
corpus baselines regenerated at version 2.

## Nit 4 — both PASS lists

A rename warns, so it is out of **PASS** and into **WARN** in SPEC-TABLES §16 and
in USAGE's summary. §16 also now states the two-kinds-of-edit rule explicitly and
gives the substitutability standard per vocabulary; USAGE says what "stand in"
means and why a same-id twin is refused.

## Nit 5

Unchanged — a sequencing call. The PR body says numbering follows the
spec-of-record PR.

---

Full `go test ./...`, `TestTablesMoveNoProtocolId`, the tables C++ battery and the
zero-cost gate green locally; golangci-lint v2.12.2, modernize, shape gate,
`gofmt -l`, `go vet` clean; all 8 CI checks green. PR body updated.

gafferongames:
## Final delta review — **MERGE**

Fresh clone, `gh pr checkout 263`, 13 commits, everything below reproduced. **The
blocker is closed** — properly, and further than I asked. Four small things below,
none of which change a verdict on real data in the unsafe direction.

---

# The blocker: closed

`identitiesRide` is gone; `substitutable` / `bodyRides` now reuses `diffTokens`, so
the referent path and the in-place path cannot disagree about what a field fact
means. My exact repro, both twins, with the original still referenced so nothing
vanishes and no rename pairing fires:

```
--- twin with a different DEFAULT under the same id
tables.baseline: Holder.swap: nested table Buff -> Boon, and multiplier's specified
default 1.0 -> 7.0 — this edit changes what data already written MEANS ...  [exit 1]

--- twin with a different KIND under the same id
tables.baseline: Holder.swap: nested table Buff -> Boon, and multiplier's wire kind
10 -> 11 — ...                                                              [exit 1]

--- SUBSTITUTABLE superset (every id, every fact, plus a new field)
[exit 0]                                                                    # silent

--- twin missing one of the old ids
tables.baseline: Holder.swap: nested table Buff -> Boon, and 1 of 1 field ids do not
ride — that much of every stored body reads back as declared defaults        [exit 1]
```

Naming `Holder.swap` and the fact, as asked. The same on a union arm's payload:

```
tables.baseline: union Effect.buff: arm payload Buff -> Boon, and multiplier's
specified default 1.0 -> 7.0 — ...                                          [exit 1]
```
and the substitutable arm payload passes in silence.

**Better than the fix I proposed:** it recurses, and the message carries the whole
path. A twin identical except that its OWN nested referent points at a different
declaration:

```
tables.baseline: Holder.swap: nested table Buff -> Boon, and child's nested table
Inner -> InnerB, and v's specified default 1.0 -> 9.0 — ...                  [exit 1]
```
with a `visiting` guard so a chain of simultaneous renames cannot walk in a circle.

I accept the substitutable pass as safe: every id rides and every fact under those
ids is unchanged, so an old body decodes to the same values — it genuinely is the
same table under another name.

# The other four, re-verified

**(2) Repoints still refuse, and the fixtures are honest repoints.** The PR's new
fixtures keep the original referenced (`swap_perks`/`swap_grade`/`swap` beside the
originals, with the comment saying why), so they exercise the referent rule and not
the rename pairing. Measured on the corpus with a `legacy` field added to keep the
original referenced:

| repoint | verdict |
|---|---|
| flags -> a differently-ordered vocabulary | **REFUSE** |
| flags -> a prefix-identical vocabulary | PASS |
| enum -> a different vocabulary | **REFUSE** |
| enum -> a superset (all names ride) | PASS |
| `[3]Grade` -> `[3]uint16` (dropped) | **REFUSE** |
| `grade Grade` -> `uint16` | **REFUSE** |
| `perks Perks` -> `uint64` | **REFUSE** |

**(3) Renaming no longer raises the verdict.** The asymmetry I flagged is gone:

| | alone | + declaration renamed |
|---|---|---|
| enum variant `Gold` dropped | WARN | **WARN** |
| union arm `debuff` dropped | WARN | **WARN** |

and the rows that must still refuse still do — table renamed + default changed,
flags renamed + bits reordered, both REFUSE. `TestRenamingDoesNotRaiseTheVerdict`
green.

**(4) Both messages reachable and correct.**
```
below half:  table SaveFile: no longer in the closure — the closest declaration,
             Archive, carries 1 of its 4 identities, below the half needed to call
             it a rename, so this baseline no longer covers it
zero:        table SaveFile: no longer in the closure — no declaration carries any
             of its identities, so this baseline no longer covers it
```

**(5) `Version = 2` does exactly what it is for.** A v1 file now gets one clean
refusal instead of last round's flood of wrong-noun diffs, and the remedy keeps the
history — including entries that predate the bump:

```
$ schema check .
tables.baseline: baseline version 1, this compiler writes 2 — regenerate it with:
schema tables-baseline --update --reason "...", which preserves the ## history section
$ schema tables-baseline --update --reason "regenerating after the rendering-version bump" .
wrote tables.baseline
```
```
## history
### 2026-09-02 — the first baseline over the tables corpus
- baseline created over 9 tables — data written BEFORE this point is not covered by it

### 2025-01-01 — an earlier intentional break
- the record the wire lacks

### 2026-09-02 — regenerating after the rendering-version bump
- baseline REGENERATED over an unreadable one (baseline version 1, this compiler
  writes 2); the projection is this unit as it stands and the history above is
  preserved, but the previous projection could not be diffed, so no per-edit lines follow
```
Check then passes. Both corpus baselines carry `schema-tables-baseline 2` and
regenerate **byte-identically**.

**(6) Battery.** All 17 `go test` packages green; `TestTablesMoveNoProtocolId`,
`TestZeroCostForValueOnlyTables`, `TestRefusals`, `TestVanishedMembers`,
`TestRenamingDoesNotRaiseTheVerdict`, `TestUnpairedVanishedMemberNamesWhatItFound`,
`TestVanishedMemberWithNoSuccessor`, `TestAbsorbedEdits` all PASS. `gofmt -l` and
`go vet` clean, shape gate clean, all CI checks green. Full class regression re-run
from scratch: 9 refusal classes refuse, 4 warn classes warn, 8 discrimination
controls silent. **Independence holds** — generated C++ byte-identical to `main` on
all four corpora (`tables/examples`, `tables/pointers`, `examples`, `examples128`).
`cmd/schema` still imports only `compiler` and `ir`; `publicapi` green. 600
corrupted baselines: **0 panics**. Provenance: 13 commits, one author and committer
throughout, `rowan@mas-bandwidth.com`.

**Docs.** The REFUSE row and the prose now agree, and USAGE matches: the row states
STAND IN with the condition spelled out ("every identity survives AND, for a table
or a union arm's payload, the facts under the shared field ids are unchanged"), a
declaration rename sits in WARN in both, and PASS-in-silence no longer claims it.

---

# Four small things, none blocking

**1. The tie message states something the same line disproves.** When two candidates
carry the same number of identities, `pairMembers` correctly declines to pair — but
`vanishedFinding` has no branch for a tie, so it falls into the `carried > 0` case:

```
warning: table A: no longer in the closure — the closest declaration, A2, carries
4 of its 4 identities, below the half needed to call it a rename, ...
```

4 of 4 is not below half. The verdict is right (warn, no pairing); the reason given
is false, and it is false about the numbers printed in the same sentence. This is
the function just rewritten so that *"the message is the only thing standing between
a reader and the coverage hole"* — worth the third branch it is missing.

**2. A mis-pair attributes refusals to an unrelated declaration.** `A{x,y,z,w}`
renamed to `A2` keeping 3 of 4, while an unrelated new `B` carries all 4:

```
warning: table A: ... B carries 4 of its 4 identities, so it is judged as the rename
it looks like
tables.baseline: B.x: specified default 1 -> 77 — ...
tables.baseline: B.y: specified default 2 -> 88 — ...
tables.baseline: B.z: specified default 3 -> 99 — ...
tables.baseline: B.w: specified default 4 -> 66 — ...
tables.baseline: Holder.a: nested table A -> A2, and 1 of 4 field ids do not ride — ...
```

`B` is brand new; none of those four "edits" happened. Over-refusal, so the
direction is safe, and the warning does name `B` so a reader can see the mis-pair —
but four confident refusals describing edits that were never made is the kind of
noise that gets a baseline deleted. A sentence in §16 saying the pairing is a
heuristic and naming the tell would cover it.

**3. `TestRenamingDoesNotRaiseTheVerdict` holds above the pairing threshold only.**
Below it the referent rule fires and the same wire loss refuses:

```
drop 3 of 4 enum variants, no rename   -> 3 warnings, exit 0
the same drop + the enum renamed       -> Config.g: enum Grade -> Rank, and 3 of 4
                                          variant names do not ride ...  exit 1
```
That is the documented rule working as designed (below half is a repoint, not a
rename), and the warning says so. But the test's name asserts the principle without
the qualifier, and both its cases sit above the threshold. Scope the name or add the
below-threshold case as the stated exception.

**4. Observed and correct, recording it so nobody re-derives it:** the referent rule
is stricter than the in-place walk on field removal — dropping a field from a
renamed declaration is absorbed (`Buff` -> `Boon` minus a field: warn only), while
repointing at a declaration missing that field refuses. That is exactly the
rename-vs-repoint split the new doc block in `diff.go` sets out, applied
consistently. No action.

**5. §16 still collides with #267** (`tables-json-walk`, #253) — unchanged, a
sequencing call.

Cosmetic: three lines in USAGE.md now run to 95–98 columns against the file's wrap.

---

# Verdict

**MERGE.** Three rounds, and the check now holds against everything I could throw at
it: the rename hole, the referent hole, the twin, the twin one level down, the
repoints, the dropped referents, the mis-pair, the threshold edges, 600 corrupted
files. The two paths that judge a field fact are now the same code, which is why I
stopped finding new instances of the same bug. Item 1 is worth a follow-up line and
items 2–3 a sentence each; none of them is a reason to hold this.

Lands after #269 and #265.

gafferongames:
All four taken. Commit `9b6c351`, all 8 CI checks green.

**1. The tie branch.** `vanishedFinding` has the third branch it was missing, and
the tie no longer borrows the below-half reason:

```
warning: tables.baseline: table A: no longer in the closure — A2 and A3 each carry
enough of its 4 identities to be the rename and the evidence does not separate them,
so nothing is paired and this baseline no longer covers it
```

**2. The mis-pair.** Your shape, before and after:

```
--- before
warning: table A: ... B carries 4 of its 4 identities, so it is judged as the rename
tables.baseline: B.x: specified default 1 -> 77 — ...        # never happened
tables.baseline: B.y: specified default 2 -> 88 — ...        # never happened
tables.baseline: B.z: specified default 3 -> 99 — ...        # never happened
tables.baseline: B.w: specified default 4 -> 66 — ...        # never happened
tables.baseline: Holder.a: nested table A -> A2, and 1 of 4 field ids do not ride

--- after
warning: tables.baseline: table A: no longer in the closure — A2 and B each carry
enough of its 4 identities to be the rename and the evidence does not separate them,
so nothing is paired and this baseline no longer covers it
tables.baseline: Holder.a: nested table A -> A2, and 1 of 4 field ids do not ride —
that much of every stored body reads back as declared defaults
schema: 1 error(s)
```

Identity overlap alone cannot separate a rename from a stranger reusing its field
names, so when more than one candidate reaches the threshold the match asks a
second question — whose own facts under the shared ids are closest (`factDistance`,
which skips referent tokens so it cannot depend on the pairing it is deciding) —
and takes a candidate only when it wins **both, strictly**. Otherwise nothing is
paired and the warning names the contenders. The one real loss is still refused,
under the declaration it actually happened in.

Three fixtures: your mis-pair shape, the 4-of-4 tie, and
`TestPairingTakesTheRenameWhenItIsTheOnlyCandidate` — the ordinary rename the
tiebreak must not cost, which still pairs and still catches the default change
inside it.

**3. Scope stated, exception tested.** `TestRenamingDoesNotRaiseTheVerdict` is now
`TestPairedRenameDoesNotRaiseTheVerdict`, and `TestBelowThresholdRenameIsARepoint`
holds the case you measured: two of three variants dropped (the defaulted one
surviving, so the loss is the only edit) warns in place and refuses when the
declaration is also renamed, with the warning saying the pairing did not happen.

**4. Recorded in `diff.go`**, so nobody re-derives it:

> The split is deliberately ASYMMETRIC ON FIELD REMOVAL, and the asymmetry is the
> point rather than an oversight: dropping a field from a declaration — renamed or
> not — is absorbed, because a reader defaults what a writer no longer sends, while
> repointing a field at a declaration that never had that field refuses, because
> nothing says the author knew what the old bodies carried. The first is the schema
> evolving; the second is a substitution claiming to be equivalent, and only the
> second makes that claim.

**Docs.** SPEC-TABLES §16 now says the match is a heuristic and names the tell — a
refusal under an unexpected declaration is a mis-pair, and the warning above it
names the declaration the check believed in. USAGE says it too, and the three lines
past the file's wrap are rewrapped.

§16 numbering unchanged — sequencing, per the PR body. Standing by for the rebase
onto #269.

gafferongames:
Rebased onto `85becbf`. `SPEC.md`, `SPEC-TABLES.md` and `USAGE.md` are taken from
main wholesale and are byte-identical to it (`git diff origin/main -- <the three>`
is empty); the branch's only other doc change is VERSIONING.md's export note. All
8 CI checks green.

## The two constructs, as facts

**`?T`** renders `optional=true` and is judged on **nothing** — `T`, `?T` and `*T`
are one framing (§3.1), so §18.1's "recorded and judged on nothing" is now
mechanical rather than a promise. Both directions are in `TestAbsorbedEdits`.

**`[E]T`** renders `array=keyed bound=N key=E` under its own wire kind 16, which
falls out of the kind mapping in `ir` that the C++ emitter now names rather than
duplicating:

```
field turrets id=0x48ad kind=16 elem=13 type=TurretConfig array=keyed bound=4 key=Weapon
field gunner  id=0x2bc9 kind=13 type=GunnerConfig optional=true
```

A keyed/positional spelling change is a kind change and the `kind` row refuses it;
the key enum is judged by the referent rule under the **enum** standard, because a
keyed array's slots ride under their variants' name ids (§3.2):

```
$ # turrets [Weapon]TurretConfig -> [4]TurretConfig
tables.baseline: HullConfig.turrets: wire kind 16 -> 14
tables.baseline: HullConfig.turrets: array key enum Weapon removed — the field keeps
its wire kind, so no reader can report the change   [exit 1]

$ # teams [Team]TeamConfig -> [Hull]TeamConfig
tables.baseline: KeyedConfig.teams: array key enum Team -> Hull, and 3 of 3 variant
names do not ride — stored values naming them read as None   [exit 1]

$ # gunner ?GunnerConfig -> GunnerConfig, and enum Weapon grown by one
[exit 0]   # silent: presence moves no byte, and a new slot is one nothing has written
```

## A hole the fixtures found

A key enum reaches the closure **only** through the array, never through a field's
own type — so `Render` was not collecting it. `Team`, `Hull` and `Weapon` were
absent from the corpus baseline and every keyed collection's vocabulary was
uncovered: dropping a variant of a key enum, which loses those slots, passed in
silence. `Render` now walks `KeyEnumRef`, and `TestKeyEnumIsCovered` holds it —
the drop warns with its own attribution control, and growing the enum is absorbed.
Both corpus baselines regenerated; `tables/examples` now carries the keyed and
optional shapes.

The `key` policy row stops being a rule with no construct behind it, and the
synthetic hook test that stood in for one is deleted.

## Version stays at 2

No v2 file can contain a keyed array — no compiler that wrote one could parse the
spelling — and an added `optional` fact is judged on nothing, so v2 files stay
readable. §18.1's example says 2 and it is right.

## Two sentences §18 does not have that the implementation needs

Flagging, not editing, since the docs come from main wholesale. Neither changes a
verdict on data; both change what a person is told.

1. **The second question.** §18.3 says the match is taken "when that is at least
   half of them **and unique**". The implementation goes further: when more than
   one candidate reaches the bar it decides on whose own facts under the shared ids
   are closest, and takes a candidate only when it wins both, strictly. Without
   that, a brand-new declaration reusing four field names outscores the real rename
   that dropped one and collects four confident refusals for edits nobody made.
2. **A third unpaired message.** §18.3 lists two. There is a third: when two
   candidates cannot be separated the warning names both. Under the two-message
   text a tie falls into the below-half wording and prints "carries 4 of its 4
   identities, below the half needed" — a reason the numbers in the same sentence
   disprove.

Both are in the PR body too. Ready to merge on your word.

gafferongames:
## Last delta — **MERGE**, after one sed

Fresh clone, rebased branch (13 commits on `85becbf`), everything reproduced. The
`?T` / `[E]T` work is sound and the hole it found is real and closed. Two items
below: one is a rebase artifact that ships wrong text to users, the other is a
recorded justification that is provably false. Neither risks data.

---

# The hole, reproduced and closed

`Render` was not collecting a key enum, because it reaches the closure only through
`KeyEnumRef` and never through a field's own type. Built the parent (`cf8824b`) and
had it write **its own** baseline, so nothing is stale:

```
$ /parent/schema tables-baseline --update --reason "parent's own baseline" .
$ grep '^enum ' tables.baseline
enum Grade                         # Team, Hull and Weapon: absent

$ # drop Hull.Freighter and Weapon.Mine
$ /parent/schema check .
warning: HullConfig.turrets: array bound 4 -> 3 (…clamped)
warning: KeyedConfig.hulls: array bound 4 -> 3 (…clamped)
[exit 0]                           # the lost VARIANTS: unreported
```
```
$ # same experiment on 6322d02
enum Grade / enum Hull / enum Team / enum Weapon
warning: enum Hull: variant Freighter removed — stored values naming it read as None and count unknown
warning: enum Weapon: variant Mine removed — stored values naming it read as None and count unknown
```

A keyed collection's whole vocabulary was uncovered, and the fixtures found it. That
is the gate working the way a gate is supposed to.

# The two constructs

| edit | verdict |
|---|---|
| keyed -> positional (`[Weapon]TurretConfig` -> `[3]TurretConfig`) | **REFUSE** — `wire kind 16 -> 14`, + `array key enum Weapon removed` |
| positional -> keyed (`[3]Grade` -> `[Grade]Grade`) | **REFUSE** — `wire kind 14 -> 16`, + `array key enum Grade added` |
| key enum SWAPPED (`[Hull]` -> `[Team]`) | **REFUSE** — `array key enum Hull -> Team, and 3 of 3 variant names do not ride` |
| key enum RENAMED (`Hull` -> `Chassis`) | absorbed, warn names the rename |
| key enum variant INSERTED mid | **PASS** — the whole point of keyed |
| key enum variant DROPPED | WARN, by name |
| `?T` -> `T` | **PASS**, silent |
| `T` -> `?T` | **PASS**, silent |

The kind-16 split does the work the spec says it does (§3.2: "the keyed and the
positional spellings cannot be mistaken for one another"), and judging the key enum
under the ENUM standard is right — §3.2 has slots riding under their variants' name
ids, so name survival is exactly the substitutability question.

`optional` recorded and judged on nothing is spec-backed (§18.1, §3.1: `T`, `?T` and
`*T` are one framing). I checked the row is genuinely inert, not accidentally inert:
deleting `"optional"` from the policy changes nothing in either direction.

**My own attribution controls on the two new rows pass**: the key-enum swap is
attributable to `"key"` alone; keyed->positional to `"kind"`+`"key"`; `"optional"`
to nothing. The PR's own fixtures carry discrimination controls (a second keyed
field added; a variant inserted mid), and the scaffolding `TestEnumKeyedArrayHook`
is gone, replaced by fixtures on the real spelling — the `key` row is no longer a
rule with no construct behind it.

# Last round's three items: closed

`cf8824b` fixed both message defects, and better than I framed them. The tie now has
its own branch, and the mis-pair no longer invents edits:

```
TIE (A2 and B both carry 4 of 4):
warning: table A: no longer in the closure — A2 and B each carry enough of its 4
identities to be the rename and the evidence does not separate them, so nothing is
paired and this baseline no longer covers it

MIS-PAIR (A renamed to A2 keeping 3 of 4, unrelated B reusing all four names):
warning: table A: … the evidence does not separate them, so nothing is paired …
tables.baseline: Holder.a: nested table A -> A2, and 1 of 4 field ids do not ride — …
```

Where last round `B` collected four confident refusals for edits nobody made, the
pairing now declines, names both contenders, and the real edit is caught by the
referent path. Refusing to guess is the right answer.

# Green

Docs equal main — `git diff origin/main -- SPEC-TABLES.md SPEC.md USAGE.md` is
**empty**; only `VERSIONING.md` remains as a doc delta. That also retires my §16
collision note: main now numbers the baseline **§18**, JSON §16, packing §17.

All 17 test packages green; `TestKeyEnumIsCovered`, `TestRefusals`, `TestWarnings`,
`TestVanishedMembers`, `TestAbsorbedEdits`, `TestRenamingDoesNotRaiseTheVerdict`,
`TestCorpusBaselinesAreCurrent`, `TestCorpusPassesItsOwnCheck`,
`TestZeroCostForValueOnlyTables`, `TestTablesMoveNoProtocolId` all PASS. `gofmt -l`
and `go vet` clean, shape gate clean, all CI checks green. Both corpus baselines
regenerate **byte-identically**. Independence holds: generated C++ byte-identical to
`origin/main` on all four corpora. Provenance: 13 commits, `rowan@mas-bandwidth.com`
as author and committer on every one.

---

# Two things, and one of them ships to users

## 1. Twelve stale `§16` citations — three of them in user-facing error text

The rebase moved the baseline's home from §16 to §18, the docs followed, the code
did not. `internal/baseline/baseline.go` has one `§18.1`, which is the tell that the
sweep was started and not finished:

```
$ grep -rn "SPEC-TABLES.md §16" --include='*.go' .
cmd/schema/main.go:49          cmd/schema/main.go:78
internal/baseline/baseline.go:1    internal/baseline/diff.go:2
internal/baseline/file.go:2        internal/baseline/file.go:82   ← user-facing
internal/baseline/file.go:93   ← user-facing (EVERY refusal)
internal/baseline/file.go:112  ← user-facing
internal/baseline/baseline_test.go:1   internal/publicapi/publicapi_test.go:143
compiler/baseline.go:8         compiler/compiler.go:53

$ sed -n '1648p' SPEC-TABLES.md
## 16. The text form: JSON in and out of one table
```

So every refusal this feature emits — its primary diagnostic — now sends the reader
to the JSON walk:

> `… if you mean it, record it: schema tables-baseline --update --reason "..." (SPEC-TABLES.md §16)`

Same for the parse-failure remedy and the `--update needs --reason` message. A sed,
but it is wrong text in the binary and this feature's whole value is that a person
reads its messages years later.

## 2. The `Version` stays at 2 reasoning is false, and I can show it

The commit message records:

> *"The rendering version stays at 2: no v2 file can contain a keyed array, because
> no compiler that wrote one could parse the spelling."*

The branch is rebased onto `85becbf`, which already carries #265's `[E]T` grammar —
so the parent build **does** parse the spelling and **does** write a v2 file
containing keyed arrays, rendered without `key=`:

```
$ /parent/schema tables-baseline .
schema-tables-baseline 2
    field turrets id=0x48ad kind=16 elem=13 type=TurretConfig array=fixed bound=4
    field teams   id=0x9ae1 kind=16 elem=13 type=TeamConfig   array=fixed bound=4
```

Feed that v2 file to the new build with **the schemas untouched**:

```
tables.baseline: HullConfig.turrets: array key enum Weapon added — this edit changes
what data already written MEANS …
tables.baseline: KeyedConfig.teams: array key enum Team added — …
tables.baseline: KeyedConfig.hulls: array key enum Hull added — …
tables.baseline: ScoreBoard.per_team: array key enum Team added — …
schema: 4 error(s)
```

Four spurious refusals on an unedited schema — precisely what `Version` exists to
prevent, and precisely the failure the 1 -> 2 bump fixed last round on the identical
argument ("a file written by the previous build of this branch").

**Impact is nil for anyone outside this PR** — no released compiler has ever written
a baseline — so this is a judgment call, not a blocker. But the general claim
underneath it ("an added fact needs no bump") is what the repro disproves: `key` is
a `RuleRefs` row, and the added-token branch refuses `RuleFixed`/`RuleRefs`
additions, so *any* future added referent fact will refuse on every older file.
Either bump to 3, or correct the recorded reason to say what is actually true —
that the only exposed files are this branch's own intermediates.

## 3. Nit: the `bound` shrink note is wrong for a keyed array

```
warning: HullConfig.turrets: array bound 4 -> 3 (a stored count past the new bound
keeps the bounded prefix and counts clamped)
```

For `array=keyed` the bound is the key enum's size, and §3.2 says slots ride under
variant name ids: an unknown key is **skipped and counted `unknown`**, there is no
prefix, and nothing is clamped. It also double-reports — the enum walk emits the
correct explanation (`variant Freighter removed — … read as None and count unknown`)
in the same run. `shrinkNote` could take the `array` value, or `bound` could be
skipped when the shape is keyed since the enum walk already covers it.

---

# Verdict

**MERGE**, once item 1 lands — a sed over twelve `§16` citations, three of which are
the messages this feature exists to be read through. Item 2 is Rowan's call, and I
would take the bump for the same reason the last one was taken; at minimum the
recorded reason should stop asserting something the parent build disproves. Item 3
is a follow-on.

Nothing here is a hole. Four rounds in, the check has held against the rename hole,
the referent hole, the twin, the twin one level down, repoints, dropped referents,
mis-pairs, threshold edges, keyed/positional spellings, key-enum swaps and 600
corrupted files — and the one gap this round was found by the PR's own fixtures
before I got to it.

Lands after #269.

gafferongames:
All three taken. Commit `151ddc2`, all 8 CI checks green.

## 1. The citations — twelve, and the grep is empty

```
$ grep -rn "§16" --exclude='SPEC*.md' --exclude='USAGE.md' --exclude-dir=.git .
$ echo $?
1
```

Each now names the subsection it actually means rather than the section as a
whole: the check `§18.2`, the file `§18.1`, the override and its history `§18.4`,
the gate `§18.6`. The three user-facing ones read right now:

```
… if you mean it, record it: schema tables-baseline --update --reason "..." (SPEC-TABLES.md §18)
… regenerate it with: schema tables-baseline --update --reason "...", which preserves
  the ## history section (SPEC-TABLES.md §18.4)
--update needs --reason: … what a person reads years later when an old file refuses
  (SPEC-TABLES.md §18.4)
```

I also swept every other `SPEC-TABLES.md §N` citation in the Go tree against main's
headings — §2.x, §3.x, §4, §5, §7, §8, §11, §15 — and all of them still resolve.
`VERSIONING.md` and `bench/SHAPE-GATE.allow` carried one each; both fixed.

## 2. `Version = 3`

You were right and the repro is decisive: the branch sits on a main that already
carries #265's grammar, so the parent build **does** write a v2 file with
`kind=16 … array=fixed` and no `key=`. I rebuilt your scenario from the committed
baseline — keyed rows stripped back to `array=fixed`, `optional` removed, the
Hull/Team/Weapon enum sections deleted, header set to 2 — and fed it to the new
build on an untouched schema:

```
--- before
tables.baseline: HullConfig.turrets: array key enum Weapon added — …
tables.baseline: KeyedConfig.teams: array key enum Team added — …
tables.baseline: KeyedConfig.hulls: array key enum Hull added — …
tables.baseline: ScoreBoard.per_team: array key enum Team added — …
schema: 4 error(s)

--- after
tables.baseline: baseline version 2, this compiler writes 3 — regenerate it with:
schema tables-baseline --update --reason "...", which preserves the ## history
section (SPEC-TABLES.md §18.4)
schema: 1 error(s)

$ schema tables-baseline --update --reason "the rendering moved to 3" .
$ # history intact, then: schema check .   [exit 0]
```

The general rule is now written at the constant rather than argued case by case:
**any new judged token bumps the version**, because a token this rendering emits
and an older one did not reads as an addition, and the added-token branch refuses
an addition on every judged row. Recording a fact nothing judges (an `optional`)
does not need a bump; adding a rule does. The disproved sentence is gone from the
commit message and the PR body, and neither now claims more than "the only files
this has ever exposed are this branch's own intermediates".

Corpus baselines regenerated at 3.

## 3. The keyed bound

The extent row now stands aside for `array=keyed` — the bound IS the key enum's
size, slots ride under variant name ids (§3.2), and the enum walk already says what
went, correctly and once:

```
--- before
warning: HullConfig.turrets: array bound 4 -> 3 (a stored count past the new bound
keeps the bounded prefix and counts clamped)     # wrong, and a second time
warning: enum Hull: variant Freighter removed — …

--- after
warning: enum Hull: variant Freighter removed — stored values naming it read as None
and count unknown
warning: enum Weapon: variant Mine removed — …
```

`TestKeyedBoundIsNotJudgedAsAnExtent` holds it with its discrimination control: a
positional array's shrunk bound in the same unit still warns, so this is a
discrimination and not a blanket exemption.

Docs still byte-identical to main. Ready to merge.

Issue #251, opened 2026-09-02
The cook: load trusted data from tools — Go cooker, C++ Open, C# Open (was 'wire v2')


gafferongames:
From the tables language review, 2026-09-02. Owner accepted the recommendation to decide this BEFORE the nine-language port ("Rest seems fine").

**Three follow-ons in §15 are one problem.** Wire v1 nests a pointee as its parent's body (§3.1), so: a chain of N nodes is N levels deep and a list of 129 nodes fails to SAVE; two pointers to one node write two bodies and load as two nodes (aliasing lost); `[..N]*Node` is refused, workaround being a by-value array behind a pointer.

**The design to write first, then build.** A pointered root writes a NODE TABLE: every reachable node body once, in pre-order, each length-prefixed; a pointer field rides as a u32 node index (0 = null, so absence and null agree with v1's elision rule). Depth is no longer nesting, so the cap is gone from the wire (the pack/Open walks keep their linear high-water-mark proof, and the LOAD walk needs its own bound: a visited-index guard or an index-monotonic rule, decided in the spec). Identity is preserved: one index, one node. An array of pointers is an array of u32. Value-only tables and by-value nesting are UNCHANGED (zero-cost gate holds byte-for-byte). Tolerance: an old-layout reader meeting a node table it cannot name skips by length and counts unknown; state what evolution between a by-value field and a pointer field means under v2 (v1 promised wire-identical; v2 must either keep that or say plainly it does not).

**Order of work.** 1) SPEC-TABLES.md §3.1 rewritten as the v2 wire, implementable from the page, with the load-walk bound proven on the page; 2) a cold adversarial read of the spec text BEFORE code; 3) cpptable emitter; 4) corpus: 200+ node chain, a DAG that round-trips as a DAG, `[..N]*Node`; 5) negative controls for each of the three properties. The §15 list loses three entries.

Blocked behind schema#A/#B landing (same wire section). Gate item for schema#248.

gafferongames:
Spec-first PR is up as a draft: #264. Five owner decisions in its body (region node directory + back-references; 2-byte type id per node; cycles writer-side only; by-value↔pointer no longer wire-identical; table-typed union arm form). Awaiting Glenn.

gafferongames:
[...]

Issue #259, opened 2026-09-02
The byte buffer primitive: `data *bytes` — an exact-size blob node in the variable class, so images and assets live inside a table at their own size


gafferongames:
Owner ruling, 2026-09-02 (Glenn, live): "yes, I like byte buffer as primitive", on: "So we can include say, images or assets inside them."

**Why.** `bytes(N)` is fixed inline storage of N per instance; in a gigabyte variable table a blob field that repeats per node costs N per node whether used or not, because the region packs storage verbatim. Owner's framing: "the key with variable tables is imagine they could be very large, like a gigabyte. you'd not in that case want to blow out memory with extra union tables you don't need and so on."

**The primitive.** `data *bytes` — a pointer to an UNBOUNDED byte buffer, legal only in a table body (the variable class, §2.1's rules: no default, no arrays of them in v1). Storage: the four-byte relocatable reference plus a u32 length beside it. Arena: `builder.AllocBytes( n )` returns the node at exactly n bytes. Region: packed at its used length, self-relative like every reference (§6.3). Wire: a bytes payload with a u32 length — the framing §3 already has — so any reader skips it by length; null is absent. Cooked: `Open`'s walk bounds the blob by its length as it bounds every node; a pointer into a mapped file IS the asset, no copy, no parse. `*string` is the sibling with the same shape (Rowan's addition, labelled): a string node at its used length.

**What schema does not do.** Interpret the blob: a format tag beside it is the user's field. **Tolerant wire loads copy the blob** (a gigabyte on the wire path is a gigabyte read); the cooked path is the zero-copy one — both sentences in the docs.

**Consequence for the message/lazy case (owner's "OOOOO"):** a sub-document or a rarely needed union arm can ride as its own wire bytes in a blob and be decoded only when asked, keeping a gigabyte file's memory proportional to what is touched. Documented as a pattern, not a construct.

**Held by test:** an Assets-shaped corpus with per-node blobs of varying size, exact-capacity measure==save, region byte-exact after Lock, Open pointing at a blob in a mapped file, hostile length past the region refused, zero-cost gate unchanged for fixed units.

Gate item for schema#248, variable-class group, after the fixed-class items the dogfood needs (#253, #255, #257, #256) and beside #251 (wire v2).

gafferongames:
[...]

Issue #258, opened 2026-09-02
Tables as MESSAGES: a union whose arms are tables, so a tool message set evolves by name — the case the spec's opening promises and the corpus never proves


gafferongames:
Owner, 2026-09-02 (Glenn, live): "if you had a set of messages for tooling, you'd want to safely evolve those tables / messages."

**Finding, by command (main at d09a10d):** a union arm cannot be a `table`:

```
Tool.schema:16:10: OpenDocument is a table, not a union payload — a payload is a declared `type`; tables nest in tables directly (SPEC-TABLES.md)
```

So the message-set pattern — `table ToolMessage { sequence uint32, body ToolBody }` with `union ToolBody { open OpenDocument, save SaveDocument }` — is expressible today ONLY when every message is a `type`. A message therefore cannot carry a nested table, a pointer, or anything that is itself a document. SPEC-TABLES.md's opening claims tables serve messages "just as much" as files; the corpus holds no message case.

**The iteration.** Inside a TABLE closure, a union may take table arms (a union declared for the type wire keeps refusing them — types stay value semantics). After #249 arms ride by u16 name hash, so a union of tables is an evolvable message set by construction: add a message = add an arm; a reader that lacks the arm reads empty and counts `unknown`; remove or reorder freely. Mode derivation (§2.2) runs through union arms: a table arm with a pointer makes the union's holder variable-length. JSON form (#253): the one-key object already specified. Reflection (#254 gives unions their arm vocabulary) names table arms' descriptors.

**Corpus:** `tables/examples/Messages.schema` — a tool message set with three arms (a type, a fixed table, a table nesting a table), round-trip, plus a V1/V2 pair adding an arm in the middle and removing one, both directions. **Docs:** SPEC-TABLES.md §2 (union arms), §11 (the refusal narrows to the type wire), and a short "messages" example in USAGE beside the config one — the two uses the opening line names, both shown.

Gate item for schema#248; queued with the language iterations after #254.

gafferongames:
Owner ruling, 2026-09-02 (Glenn, live): "I have no problem for the message and table case (not types) if you do some allocations while you read or write for unions, this way you can have a pseudo-union in golang see."

[...]

PR #264, opened 2026-09-02
SPEC: table wire v2 — the flat indexed node encoding (#251, spec-first)

gafferongames:
Spec-first step of #251. **Docs only** — one file, `SPEC-TABLES.md`, rebased onto `69bc4ec`: main's page with the v2 sections re-applied, main winning wherever the two disagreed outside v2's scope. No compiler, emitter, corpus or test change.

Draft → [cold read](https://github.com/mas-bandwidth/schema/pull/264#issuecomment-5503430395) → repair → owner rulings → [delta read](https://github.com/mas-bandwidth/schema/pull/264#issuecomment-5504283466) (LAND AFTER FIXES) → **this pass**. F1, F2 and F3 are in, and decision 4 is taken.

## The design

```
a pointer to a table:   id (u16), kind = 17, index (u32)

the node table, in one or more fields of the root body:
    id = 0xFFFF, kind = 12, L (u32), then L bytes
  the first field opens with node_count (u64); every field carries
  WHOLE RECORDS;  record = type id (u64), length (u32), body
```

`0` null, `1` the root, `2…` the records, in first-visit depth-first pre-order. No pointer edge is a nesting level, identity survives, `[..8]*Node` is legal. **Load is a scan** — the record's type id is what makes it one. Buffers stay inline as §2.5 frames them and table union arms stay inline as §2.6 does; v2 touches table pointer edges and nothing else. §7 is rewritten from the owner's governing sentence down — cooking is fundamentally an optimization; `Open` checks the header and points; two parts, data and attribution, and a shipped build carries only data; `OpenValidated` is the named check.

## The three fixes

**F1 — a record never straddles a field.** A writer opens the next field when the record would not fit, so every multi-byte read a reader makes lies inside one contiguous payload. No segmented cursor, no copy to hand a body to the generated decoder — **§6.5's "the load path allocates nothing" stays literally true**, and the generated body decoder never learns that chunking exists. Still ceiling-free, still deterministic, and the naive per-chunk reader becomes the correct reader. The reader's two smaller silences are answered on the page too: a short field is legal input (the fill rule is the writer's, and byte-identity requires matching it), and a node table that cannot be read whole is **malformed entire**, not salvaged in part — numbering is positional, so dropping a field would renumber every record after it.

**F2 — a node body past `0xFFFFFFFF` is a named save-time refusal.** The old sentence bounded each payload, not their sum; two 2 GiB `*bytes` in one table are four gigabytes of body under §2.5. Refusal, not a `u64` length: the repair is more nodes, which is the shape the flat encoding wants, and widening would cost four bytes on every node in every save to frame a structure nobody should build. §11 carries the row.

**F3 — the three header facts that are semantic.** Both part lengths and the layout id are **64 bits** — a 32-bit length would reimpose at cook time the ceiling §3.1 just removed, on exactly the mesh catalogue a cook exists for, and the layout id is now the sole guard between a runtime and a foreign region. The magic is read **bytewise first**, since it establishes the byte order every other header field is in. And `Open`'s check list is now **complete** rather than closed at three: magic, byte order, layout id, both part lengths against the caller's `size`, base alignment. Everything else about the header stays the implementation's business, as the reader allows.

## Decision 4 — kind `17`, taken

A node index and a plain `uint32` are the same four bytes. Under a shared kind the edit between them reports nothing in **both** directions, and the direction that matters — an index read back as a number — is silent *always*, not merely often. The kind byte already rides, so a distinct number costs **zero bytes** and one row in the fixed-width skip rule.

§3's "an unknown kind is not skippable" is the real objection, and the page now states its scope exactly, because the same section refuses a kind for the node table: **it bites for a kind added after readers exist.** The node table must stay skippable by a reader that never heard of it; a pointer field need not, because the set is closed before any reader ships and a kind spent later would not be available at all. That is the ruling-5 exception — a decision that becomes unavailable, not an optimization deferred.

§4.1 returns to **two** silent edits, and the third construction-closed case joins `[E]T`'s kind `16` in the list of classes a kind number closed.

**The cost that is not zero, and the page says it**: a reader built before kind `17` existed meets it as framing damage and stops that body. Bounded by the same fact that makes the kind affordable — no such reader ships — and a pointered save was already unreadable by a build of the tree encoding, since the node table carries what the bodies used to.

## Verified, not asserted

Fresh scratch Go encoder + independent decoder from the final text. The **naive per-chunk reader** — no segmented cursor, no copy, a contiguous sub-slice per body — succeeds at every chunk cap, which is F1's whole point:

```
cap=4096  wire=129 bytes  1 field   naive per-chunk reader: OK, identity held
cap=  60  wire=136 bytes  2 fields  naive per-chunk reader: OK, identity held
cap=  40  wire=150 bytes  4 fields  naive per-chunk reader: OK, identity held
cap=  20  wire=150 bytes  4 fields  naive per-chunk reader: OK, identity held
```

Sharing preserved, `owner = 1` (an index below the node carrying it) surviving, the elided null reading back as null, `node_count` agreeing with the scan at every cap. A reader that declares the field as `uint32` gets exactly one `kind_mismatch` on it — the edit kind `17` exists to report. A reader without kind `17` cannot frame the body, which is the cost stated above, reproduced rather than assumed. **Scratch deleted.**

## Costs, unchanged by this pass

12 B/node of wire record header (u64 type id, u32 length) and 16 B/node of attribution, which a shipped build need not carry: 18.75% / 25% at 64-byte nodes, 37.5% / 50% at 32-byte, 1.17% / 1.56% at 1 KiB. Table is in §6.3 with the lever named — node size, not encoding. Kind `17` moved no byte. Record-aligned chunking costs under one record of slack per field plus the field's seven bytes.

## Open for the owner

Nothing. The delta read's three musts are in, decision 4 is taken on its numbers, and every should is on the page: `unknown` counts per **transport field** rather than per schema difference; the distributed cook farm is named as `OpenValidated`'s home; type id `0` keeps its rebound with the four vocabularies uniform.

Negative controls for #251 step 5: a forged region whose reference names a node no traversal enters; a round trip through a record whose type id this build cannot name, then a `Cook` of the result; a `node_count` that disagrees with the scan; a node table chunked across several fields, read by a per-chunk reader; a node body over `0xFFFFFFFF`; a pointer field read by a schema declaring `uint32`; and a corpus DAG shaped `Scene { a *X, b *Y }` with both naming one `P`, the shape that produces a lower-index reference.

### DIFF
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -173,10 +173,10 @@ table Scene
 No pointer, no allocation and no change of mode. Reach for `*` when
 the structure genuinely needs it: a table that refers to ITSELF (a chain, a
 tree), a large subtree you would rather not carry by value, or one node
-several parents name (sharing is a named follow-on, §15 — v1 writes a
-tree). One pointer anywhere in a table's by-value closure flips it to
-VARIABLE-LENGTH (§2.2) and with it the whole builder lifecycle, so the
-choice is a real one.
+several parents name — sharing is preserved end to end, on the wire and
+in both runtime forms (§3.1). One pointer anywhere in a table's by-value
+closure flips it to VARIABLE-LENGTH (§2.2) and with it the whole builder
+lifecycle, so the choice is a real one.
 
 The spelling is C's, deliberately: it reads as what it is. The rules,
 each refused by name (§11):
@@ -189,10 +189,15 @@ each refused by name (§11):
 - **A pointer is declared inside a table body, and nowhere else.** A
   `type` body refuses one: types remain value semantics, and that is the
   founding line of the split.
-- **An array of pointers is a named follow-on** (§15). Declare a bounded
-  array of tables by value, or a pointer to a table that holds the array.
 - **A pointer field takes no specified default.** A fresh pointer is
   null, and null is the only value a default could name.
+
+**An array of pointers is a bounded array like any other.** `[..8]*Node`
+declares eight reference slots and rides as an array of node indices
+(§3.1); a null element is index `0`. It is the spelling for a node with a
+fixed fan-out, and it costs four bytes a slot instead of a whole table by
+value. An array of `*bytes` or `*string` stays refused (§2.5): a buffer
+takes no index, so there is nothing to put in the slots.
 
 A pointer's STORAGE is a four-byte relocatable reference — never a
 machine address — which is what keeps §9's relocatability true with
@@ -443,8 +448,9 @@ something asks for it — which keeps a very large file's resident memory
 proportional to what is touched rather than to what exists.
 
 Refused by name: `*bytes` or `*string` outside a table body; a specified
-default on one (a fresh reference is null); an array of them, as for any
-pointer (§15); `?` on one, because null already IS absence (§2.3).
+default on one (a fresh reference is null); an array of them — a buffer
+takes no node index, so there is nothing to put in the slots (§3.1); `?`
+on one, because null already IS absence (§2.3).
 
 ### 2.6 Union arms may be tables
 
@@ -507,7 +513,7 @@ schema's codebase:
 - **The kinds are a closed set**, and these are their numbers: `1` bool,
   `2` i8, `3` i16, `4` i32, `5` i64, `6` u8, `7` u16, `8` u32, `9` u64,
   `10` f32, `11` f64, `12` string, `13` table, `14` array, `15` union,
-  `16` enum-keyed array.
+  `16` enum-keyed array, `17` pointer index.
 - **Payloads**, one row per kind. `L` is a u32 byte length; `N` is a u32
   element count. Nothing is aligned and nothing is padded.
 
@@ -522,6 +528,7 @@ schema's codebase:
   | `14` array | `L`, then the array body: `element kind (u8)`, `N`, then the elements |
   | `15` union | `arm id (u16)`, and when it is not 0, `L` then `L` bytes of arm body |
   | `16` enum-keyed array | `L`, then the body: `element kind (u8)`, `N` = the number of SLOTS PRESENT, then N pairs of `variant id (u16)`, `L (u32)`, `L` bytes of element (§3.2) |
+  | `17` pointer index | 4 bytes, the u32 node index (§3.1) |
 
   **An array's ELEMENT KIND is part of its identity, not only its framing.**
   For kinds `14` and `16` a reader compares the element kind it declares
@@ -539,10 +546,16 @@ schema's codebase:
   kind `13` payload carries.
 
   **Spellings that add no row, and the one way they differ.** A `?T`
-  optional field is framed exactly as the non-optional `T` (§2.3); `*T` is
-  framed exactly as a by-value `T` (§3.1); and `*bytes` and `*string` are
-  framed exactly as `bytes(N)` and `string(N)` (§2.5). Each family is ONE
-  FRAMING under several declaration spellings.
+  optional field is framed exactly as the non-optional `T` (§2.3), and
+  `*bytes` and `*string` are framed exactly as `bytes(N)` and
+  `string(N)` (§2.5). Each family is ONE FRAMING under several
+  declaration spellings. **`*T` naming a TABLE is the exception**: it
+  rides as a node index under its own kind `17` (§3.1), because a body
+  that may be named twice cannot also sit inline at one of its names.
+  The distinct kind is what makes moving a field to or from `*T` a
+  REPORTED edit rather than a quiet one, for the same reason kind `16`
+  exists (§3.2): a node index and a plain `uint32` are the same four
+  bytes, and only the kind can tell a reader which it is holding.
 
   **What differs inside a family is ELISION, and only at the empty end.**
   Content decides for the by-value spellings and presence decides for the
@@ -554,7 +567,9 @@ schema's codebase:
   the claim: a schema may move a field among them and no byte moves for
   such a value. At the empty end the bytes differ and no reader
   misdecodes — an elided field reads as absent (`?T`), null (`*T`) or the
-  declared default (`T`), which is correct in every direction.
+  declared default (`T`), which is correct in every direction. Moving a
+  field ACROSS families — between `*T` and `T` or `?T` — is not a free
+  edit: it changes kind, and §4 counts it (§3.1).
   - **Array elements.** For a scalar element kind the elements sit back to
     back at that kind's fixed width. For element kind `13` (table) each
     element is preceded by its own `L`. `bytes(N)` rides as an array of
@@ -566,9 +581,10 @@ schema's codebase:
     reader accepts it.
 - **Skipping a field you cannot name** needs the kind byte and nothing
   else, which is what makes an unknown field survivable (§4). Three rules
-  cover the set: kinds `1`–`11` skip their fixed width; kinds `12`, `13`,
-  `14` and `16` read `L` and skip `L` bytes; kind `15` reads the `u16` arm
-  id and stops there if it is 0, else reads `L` and skips `L` bytes.
+  cover the set: kinds `1`–`11` and `17` skip their fixed width; kinds
+  `12`, `13`, `14` and `16` read `L` and skip `L` bytes; kind `15` reads
+  the `u16` arm id and stops there if it is 0, else reads `L` and skips
+  `L` bytes.
   **A kind a reader does not know at all is not skippable** and is framing
   damage, which is why the set is closed and why a new kind is a wire
   change rather than an addition.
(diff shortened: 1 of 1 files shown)


gafferongames:
Converted to DRAFT: spec-first by design, and it carries five decisions that are the owner's (listed in the body). It does not land until Glenn rules; the emitter step follows the ruling. Parked on schema#248 for his return.

gafferongames:
## Cold adversarial read — #251 step 2

Read against `origin/main`'s §3.1 first, then the diff, then #251 / #259 / #258 / the #248 doctrine comments. Every framing claim below was checked with a scratch Go encoder + an **independent** decoder written from §3/§3.1 alone, plus a model of §7's `Open`. The scratch program has been deleted; nothing from it is on this branch, and the branch was not edited.

**Headline:** the wire itself is sound and implementable. The framing round-trips, the `L == 4` discriminator is provably unambiguous, and load really is a scan that terminates on every hostile node table I could build. The problems are all on the **region** side (decision 1) and in **contradictions between sections** — including one that lets a forged cooked file through `Open`.

---

# DEFECTS

## D1 — §7's back-reference rule is unsound. The stated justification is false.

> "A hit whose type id is the declared target's is accepted and NOT descended into, **because a node below the mark has already been entered and its own slots already validated**"

The premise does not hold. The mark advances past *region bytes*, not past *entered nodes*. A forward reference may jump over a node the walk never enters; that node is then "below the mark" with its slots unvalidated, and a back-reference to it is accepted.

Modelled `Open` exactly as §7 states it (16-byte nodes, one self-relative slot each, directory validated first). Forged region, 48 bytes:

```
directory:  offsets [0, 16, 32]   strictly increasing, inside the region, storage fits  -> passes
root @0  .head = +32  -> node @32   FORWARD  (>= mark)  -> entered, mark = 48
node @32 .next = -16  -> node @16   BACKWARD -> directory hit, right type -> ACCEPTED, not descended
node @16 .next = +0x7FFFFFF0        NEVER VALIDATED
```

Result: `Open` returns the root. Nodes entered: `{0, 32}`. Node `@16` was never entered, and the first `NodeAt( y->next )` the runtime does hands out a pointer 2 GB outside a 48-byte region. This is the exact failure class §7 exists to prevent, and there is no trust-mode bypass to blame.

**Two fixes, smallest first.**

1. *One integer.* In a genuine region the walk enters nodes in directory order, so: require a forward reference to be **the next directory entry**, and a back-reference to have **directory index < entered count**. Then "below the mark" and "already entered" become the same statement and the proof is true as written.
2. *Better — drop the traversal.* The directory now names **every** node start **and** its type. That is the same fact the type id gives `LoadMeasure`, and the page already draws the right conclusion there: *"Two bytes a record removes the walk instead of bounding it."* Apply it here. `Open` becomes a linear scan over the directory: for each entry, take its type, validate that node's slots against the directory. This
   - fixes D1 by construction,
   - validates **unreachable** nodes too (the walk never does),
   - deletes the high-water mark and §14's "why the cooked walk still needs order state",
   - **breaks the pack-order == walk-order coupling** that §7 currently pins at two sites,
   - is the same asymptotics: `O(R + P log N)`.

   My recommendation for decision 1 is this one. It is strictly less machinery than what the page has.

**Related, on the same bullet:** the header still says *"The walk is LINEAR IN THE REGION"* while the body correctly says *"linear in the region plus one search per back-reference"*. It is `O(R + B log N)`, `B ≤ R/4`. Say the real bound in the header.

## D2 — The node directory's `0` sentinel contradicts "strictly increasing", and collides with the root's own offset.

Three statements, three different rules:

- §6.3: "a trailer of `N` `u32` offsets from the region's base, **strictly increasing** (and **`0`** for a node that was not materialized)" — self-contradictory as written.
- §7 proof: "its **non-zero** offsets strictly increase" — the right rule.
- §7 refusal list and §11: "a node directory that **does not increase**" — the wrong rule again.

This is not cosmetic. It fires on the ordinary evolution case:

```
[4] rec1 type 0xBEEF (a table this build has no name for), rec2 Node
    -> rec1 keeps its index, counted unknown, referrers read null   (correct, per §3.1)
    -> its directory entry must therefore be the 0 sentinel
    -> Lock/Cook that region and Open REFUSES it: "does not increase"
```

A file from a newer schema loads fine, then cannot be cooked. Worse: **`0` is also the root's own offset** (§6.3: "position `0` is the root, at offset `0`"), so a reader resolving a wire index through a 0 entry lands on the root rather than null — fabrication, silently, with no report event.

Fix: use a distinct sentinel (`0xFFFFFFFF`), and make all three sites say "offsets of materialized nodes strictly increase".

## D3 — §3.1 specifies the wire of `*bytes`; §2.1 and §11 still refuse it.

§3.1: "**`0` is the RAW BYTE NODE** … A `*bytes` field's node takes this form" and "a `*bytes` field requires `0`".
§2.1: "**A pointer targets a `table`, and only a `table`.**"
§11: "`*T` where T is a `type`, enum, flags or union — value-semantics data has no identity to point at".

The page specifies the encoding of a construct the same page refuses. A third party implementing from §3.1 cannot tell whether `data *bytes` is legal input. #259 is unlanded, and — worth flagging — **its own ruled wire is a different shape**: "Wire: a bytes payload with a u32 length — the framing §3 already has — so any reader skips it by length". §3.1 makes it a node table record reached through a kind-8 index instead. That is a design change to an unlanded, unruled item, made on a page that cannot cite a ruling for it.

Recommend: **keep** the `0` type-id reservation (reserving is present-state law and costs nothing), **delete** the `*bytes` sentences and the length-companion sentence, and let #259's own PR add §2.1 permission and the wire form together. Same for `*string` (PR body question 5 — correctly not on the page; keep it that way).

## D4 — "exactly one node table" rests on a false premise.

> "There is exactly one node table in a save. **A variable-length table is never held by value (§6.2)**, so a save has one root and one table"

§2.2's mode rule exists *precisely because* a variable-length table can be nested by value — that is the whole propagation rule: "A table is VARIABLE-LENGTH if it declares a pointer, **or if anything it nests by value — directly, or as a bounded array — is variable-length**." §6.2's "never held by value" is about the API surface (not copied as a struct), not about nesting.

The *rule* is fine and well-defined. State it directly: **only the root body carries a node table; no nested body ever does, by-value or record.** Do not derive it from §6.2.

## D5 — The numbering rules and the union-arm rule disagree.

§3.1's numbering walk lists "**a union's set arm where one is set**" as a place the walk *descends through*, alongside by-value nestings — never as a node that *takes an index*. But the later bullet says a table-typed arm in a variable-length table "rides as a node reference: `arm id (u16), L = 4, index (u32)`", which requires the numbering pass to have assigned it one.

An implementer cannot write the numbering pass from this. Say explicitly that a table-typed union arm in a variable-length table is a **node edge** (takes an index at first visit), not a by-value descent. See also R5 — I think this rule should come out entirely for now.

## D6 — v2 introduces an aggregate 4 GiB ceiling; §2.2 says there is none.

The node table is **one** field, `kind 12`, framed by **one `u32` L**. Every node body in a save lives inside it. So a pointered save is capped at 4 GiB − 1 of node bodies, aggregate.

§2.2, unchanged: "**Extents have no wire ceiling**: lengths and counts ride as u32 (§3), so the only limit is the language's own".

Under v1 there was no aggregate cap — the root body is framed by the buffer, and each nesting carried its own `u32`. This is a real regression against the owner's own framing on #259 ("imagine they could be very large, like a gigabyte") — 1 GiB is fine, 4 GiB is a wall, and #259's blobs live inside that same field. Either state the cap plainly with a named save-time refusal, or widen the field. Silence is the defect.

## D7 — §8 breaks §2.2's zero-cost gate as written.

The paragraph opens "A unit that has pointers carries **two more facts**, and a unit that has none carries neither" — then lists **three**, the new one being "A table also carries its own **node type id** (§3.1)", unqualified by unit class. §2.2's gate: "for a unit whose tables are all fixed-size … **not one extra descriptor column**". Scope it, or the gate's own test fails.

(The compile-time refusals in §11 — table name id collisions, `0xFFFF` — *do* apply to pointer-free units, and that is right for the same reason the 23 suffixes are claimed for every member. Say so, as §11 already does for the suffixes.)

## D8 — `node_count` is attacker-controlled and never cross-checked.

`node_count` is a `u32` read straight from the wire, and §6.3 sizes the directory as `node_count + 1` entries.

```
[5] node_count = 0xFFFFFFFF, zero records, 13-byte file
    a reader that sizes the directory from node_count asks for 25.8 GB
```

The page says the reader must bound record parsing by `L` ("a record whose length runs past `L`: malformed") but never says `node_count` must agree with the records actually scanned. Say it: **the record scan is authoritative; a `node_count` that disagrees is malformed**, and the directory is sized from the scan.

## D9 — Present-state register (CLAUDE.md: no history, no decision narration).

- **`v1`/`v2` now mean two different things on one page.** §4's own examples use them as *schema* versions — `flags Perks { … } // v1: bits 0, 1, 2`, "A v1 writer elided `damage = 21.0`; a v2 reader…" — while §3.1 uses them as *wire* versions. §7 and §15 say "REFUSES in v1" / "v1 refuses a foreign byte order" meaning *this version*, which now reads as *wire v1*. Pick one vocabulary; rename the §4 examples to `before`/`after`.
- **History and decision narration on the page.** "v1 promised the two were interchangeable … v2 cannot keep that promise"; "The break is taken now, before the table wire ships and before the per-language backends are written"; "The corpus keeps the both-directions evolution test and **changes what it asserts**: not 'no byte moves', but …". That is the PR body's material, not the reference's. State the law: a pointer field is kind 8, a by-value nesting is kind 13, the edit between them is §4's kind mismatch. The corpus's assertion belongs in the corpus.
- §3.1's "Two ways of bounding a load-time TRAVERSAL were weighed instead, and both are rejected (§14)" duplicates §14 verbatim inside a normative section. §14 is the licensed home; §3.1 should carry the rule, not the weighing.

## Ambiguities I had to guess to write the decoder — each one a defect

| # | The question §3.1 does not answer | What I guessed |
|---|---|---|
| G1 | `node_count` disagrees with the records the scan finds | scan wins, count `malformed` (see D8) |
| G2 | type-checking a pointer that names **index 1** — the root has no record and therefore no wire type id | accept; a `*T` naming the root where `T` is not the root's type has nothing to test against and I let it through. Say what this does. |
| G3 | does the numbering walk enter pointer fields inside a **false `if` guard**? | no — otherwise the save writes orphan records for nodes no field names |
| G4 | who fills the directory? §6.5 says `LoadMeasure` reads framing only (it has no region yet); §3.1 says `Load` "walks it again" (once) — but `Load` must resolve a **forward** index while decoding record 1 | `Load` is two passes: fill the directory, then decode |
| G5 | per-node **alignment padding** in the region. A type-0 raw byte node of odd length forces it, and it interacts with both "strictly increasing" and "a misaligned reference" | pad to the region's alignment; page silent |
| G6 | a body carrying id `0xFFFF` **twice** — §3's "last occurrence wins" vs §3.1's "exactly one node table in a save" | last wins (§3 is the reader's rule; §3.1's is the writer's) |
| G7 | does a **fixed-extent** array of pointers elide when all-null? §3's carve-out ("fixed arrays of tables keep their elements") does not cover kind 14 / element kind 8 | elides |

---

# RISKS — design calls, with a recommendation and a number

## R1 — Decision 1: is the region change forced? Yes. Is the *mark* forced? No.

**The "keep the region a tree" alternative is not available**, and the PR body is right about why, but the reason is sharper than the exponent: load-time dedup **is** sharing, sharing in a region of self-relative offsets **is** a back-reference. "Tree region + DAG load" is a contradiction, not a trade. The only real alternative is a region that **duplicates** — and that (a) is exponential (a chain of K nodes each named twice = 2^K bodies; **K = 20 is 1,048,576 copies of the leaf**) and (b) breaks the identity §2.1 now promises in words ("sharing is preserved end to end … in both runtime forms"): `a->palette == b->palette` would be false in the region and true on the wire.

So: **keep the directory.** But the mark is not forced — see D1 fix 2. Recommendation for decision 1: **take the region change, and replace the mark-walk with a directory scan.** It is less code than what is on the page and it closes D1.

## R2 — Decision 1's price, in numbers. 6 bytes/node is 9.4% of a gigabyte at 64-byte nodes.

| node size | nodes / GiB | directory (6 B/node) | % of region | type ids on the wire (2 B/node) |
|---|---|---|---|---|
| 32 B | 33,554,432 | 192.0 MiB | **18.75 %** | 64.0 MiB (6.25 %) |
| **64 B** | **16,777,216** | **96.0 MiB** | **9.38 %** | **32.0 MiB (3.12 %)** |
| 128 B | 8,388,608 | 48.0 MiB | 4.69 % | 16.0 MiB (1.56 %) |
| 256 B | 4,194,304 | 24.0 MiB | 2.34 % | 8.0 MiB (0.78 %) |
| 1 KiB | 1,048,576 | 6.0 MiB | 0.59 % | 2.0 MiB (0.20 %) |

The cooked file grows by the same amount, and `Open` touches all of it.

I looked for a cheaper directory and there isn't one that keeps §6.5's promise. A **shared-nodes-only** directory would be near-free (shared nodes are a small minority of a real asset graph) and would still decide every back-reference — but `Load` needs index→offset for **every** node to resolve forward indices, and §6.5 forbids it a scratch allocation, so the full directory has to be resident anyway. The 4 offset bytes are load-bearing for `Load`; the 2 type bytes are load-bearing for `Open`. **Recommendation: accept 6 bytes/node, and put this table on the page** beside the owner's gigabyte doctrine — 9.4% at 64-byte nodes is exactly the kind of number "memory proportional to data present" wants stated, and the lever is node size, not encoding.

## R3 — Decision 4: the old-reader claim holds. The *new* silent edit does not get named.

Verified, on real bytes: a v1-layout reader over the v2 body lands **exactly 1 unknown + 2 kind_mismatch**, nothing misdecoded. Kind 12 skips by `L`; kind 8 skips 4. Decision 4 is correct as claimed, and choosing kind 12 over a new kind is not merely tidy — **a new kind number would be unskippable by every existing reader**, since skipping is defined only over the closed set. That is the strongest argument in §14-5 and the page under-sells it.

But: **a pointer now rides as kind 8, which is also a plain `u32` field.** The `uint32 ↔ *Node` edit is therefore *not* a kind mismatch — in one direction the reader gets a raw number bounds-checked as an index (it may even land on a node of the right type by luck), and in the other a node index reads back as a plausible small integer **with no report event at all**. §4 already names this hazard for the enum/`uint16` pair; it must now name it for the pointer/`u32` pair. **Recommendation: keep kind 8, add the sentence.**

## R4 — Decision 3: a 16-bit table-name id is scoped to the whole unit closure.

Field ids are scoped per table and variant ids per vocabulary; **table name ids are the first 16-bit namespace scoped to an entire closure**, which is a much larger population.

```
 100 tables -> P(some pair collides) =  7.3 %
 200 tables -> 26.2 %
 300 tables -> 49.6 %
 500 tables -> 85.1 %
```

A real Assets.bin schema will get there. The refusal is loud and the fix is a rename, so it is survivable — but it is a *rename of a public type* forced by a hash accident, and it will happen. **Recommendation: keep u16 and put the number on the page** so nobody is surprised; widening the node type id to u32 costs +2 B/node (+3.1% at 64-byte nodes) and is the fallback if a real closure hits it. (The `0xFFFF` field refusal is also reachable by ordinary names — `atwf`, `cupf`, `f8v` all fold there — so it is real, not theoretical, and correctly listed in §11.)

## R5 — Decision 7: a table-typed union arm as a node reference gives one `union` two layouts.

The `L == 4` discriminator is sound (verified: legal table bodies are 2 or ≥ 6 bytes; 3, 4 and 5 are unreachable, so `L == 4` names the index form and nothing else). That is not my concern.

My concern is the storage consequence. A named `union` used in a **fixed** table stores its table arm **inline**; the same named `union` used in a **variable** table stores a **4-byte index**. One symbol, two layouts — and §13.1's ruling is that tables and types share one symbol table. The generated `ToolBody` cannot be both.

I can see the motive, and it is the owner's own (#259): *"you'd not in that case want to blow out memory with extra union tables you don't need"* — inline arm storage costs `max(arm)` per instance, and making the arm a node makes memory proportional to the arm actually present. That is a good reason. It is **not on the page**, and it is a reason to design #258 around, not a reason to specify it here.

**Recommendation: cut the node-reference arm form from this PR.** Table arms are not a declared construct yet (#258 is unlanded and the compiler refuses them by name today), the wire rule cannot be implemented without D5's numbering fix, and the two-layouts problem needs a real answer. Let #258 carry the construct, the motive and the wire together. That also removes the only place the fixed and variable classes disagree about a *value*'s encoding, which is worth keeping clean.

## R6 — `LoadMeasure` amplification is unbounded on the page.

Smallest legal record = 8 wire bytes (`type id` 2 + `length` 4 + bare terminator 2) and it commands `sizeof(T)` region bytes:

| `sizeof(T)` | amplification | 1 MiB wire demands |
|---|---|---|
| 64 B | ×8 | 8 MiB |
| 344 B (a ShipConfig-ish record) | ×43 | 43 MiB |
| 4 KiB | ×512 | 0.50 GiB |
| 32 KiB (a table with a big `[..N]` array) | **×4096** | **4.00 GiB** |

This is not a hole — the caller owns the allocation and can refuse the number, which is exactly the C++ caller-owned form the owner ruled for. But the page presents `LoadMeasure`'s answer as a convenience and never says it is also **the defense**. One sentence: the ratio is bounded by `max sizeof(T) / 8`, and the caller is expected to bound it.

## R7 — §9's "going wide" now has a serial prologue, and §9 does not mention it.

The pre-order numbering is a **single-threaded DFS over every reachable node**, and it must run before any body can be written. The scatter-write still works *after* it, but a gigabyte structure now pays a serial pass that v1 did not need. §9 is unchanged.

Also: "measure assigns the numbering, save emits in it" reads as if the map is handed from one to the other, but the API (`SceneMeasure` / `SceneSave`) passes no map. Say plainly that the numbering is **deterministic and re-derived**, which is what makes `measure == save` hold — otherwise an implementer wires a hidden parameter.

## R8 — "A data cycle is refused at SAVE" is not true of every save path.

§11: "measure, save, cook and `Lock` all return failure with the cycle named."
§3.1: "a save from a **LOCKED region** needs no map at all, because the region's node directory already is the numbering."
§3.1: "a **cyclic node table loads into a cyclic structure**".

Load a hostile wire into a region (the game path), and you hold a cyclic region indistinguishable from a locked one (§6.2: "`Load` produces the same representation `Lock` does"). Save from it and there is no map, so no colouring, so no refusal — a schema-generated writer emits a cyclic wire. "Acyclicity is the WRITER's guarantee" then has a hole in it.

Verified the read side is fine: a 2-node cycle loads, terminates, and fabricates nothing (report all zero, which is correct per the page). This is only about the claim's scope.

**Recommendation: narrow the claim on the page** — the refusal covers saves from a **builder**; a region loaded from an untrusted wire is not re-proved, and a save from it reproduces what it was given. That is the honest present-state law and it costs nothing. (The tool path is already safe: load-into-builder then `Lock` refuses the cycle by name.)

Answering PR-body question 3 directly: **no, do not add a reader-side acyclicity check.** §14-6 prices it correctly and the price is right — 4 bytes a node plus a nesting-proof pass, for a property no schema-generated walk depends on. Keep it priced and unspent.

---

# VERIFIED GOOD

- **The framing round-trips, and §3.1's worked example is arithmetically correct.** Independent encoder + decoder on the section's own graph: **93 bytes**, `node_count = 3`, root = 1, A = 2, B = 3, P = 4, `P` written once and named three times, `B.next` elided and reading back null, report all zero. The record lengths (23 / 16 / 9) and the node table `L` (70) check out byte for byte. Same total as the PR body claims, reached independently.
- **The `L == 4` union-arm discriminator is provably unambiguous.** Legal table body lengths are 2 (bare terminator) or ≥ 6 (`2 + 1 + 1 + 2`); 3, 4 and 5 are unreachable. The claim is exactly right.
- **Load is a scan, and it terminates and does not fabricate on every hostile table I built.** Cycle `2↔3` (terminates, cyclic structure, no event — as the page says); index `99` with `node_count = 1` (malformed, pointer null); a `*Node` field naming a `Palette` record (kind mismatch, null); an unnameable type id (keeps its index, counted unknown, referrers null, following records still in step); a record length past `L` (malformed, the root body reads on past the field). A second root cannot be forged — index 1 has no record.
- **"Never followed" does stop a cycle from making `Open` loop.** Forward references strictly advance the mark; back-references are not descended; `Open` terminates on any input. (The cost of not descending is D1 — it also never validates a node reached *only* by a back-reference.)
- **The rejection of the index-monotonic rule (§14-3) is correct.** Pre-order is not topological: `Scene{ a *X, b *Y }` with `X.p` and `Y.p` both naming one `P` numbers `X=2, P=3, Y=4`, and `Y`'s reference names 3 < 4. Worth noting the page's own worked example does **not** exhibit this — every reference in it names a larger index — so the corpus DAG should be shaped like the above, not like §3.1's figure.
- **The closed kind set does not move, and that is load-bearing**, not tidiness: skipping is defined only over the closed set, so any *new* kind would be unskippable by every existing reader and would break the whole tolerance story. Kind 12 is the right call and §14-5's five-bytes-vs-two arithmetic is right.
- **§15 loses exactly the three entries and gains none.** No stale depth-cap text survives: §2.1's array-of-pointers refusal, §11's array-of-pointers row and §11's read-time-depth row are all gone, and the 128 cap appears nowhere outside the unrelated 128-bit type family.
- **The zero-cost gate holds on the wire.** Every rule v2 adds is conditioned on a pointer existing; a fixed-size table has none, therefore no node table, therefore byte-identical bytes under both wires. Traced through §3 rule by rule and nothing in the value-only path is touched. (D7 is about the *descriptor* half of the gate, not this one.)
- **Decision 5 is right and stated plainly.** By-value and pointer cannot stay wire-identical while identity is preserved — a body that may be named twice cannot also sit inline at one of its names. Taking the break now rather than after the nine-language port is the correct call, and the hybrid rejected in §14-4 is rejected for the right reason (a 200-node chain is singly referenced at every link).

---

## What the page cannot tell the owner

1. **Whether the region change is worth it is not decidable from §3.1** — §3.1 only shows the wire, where the change is cheap and clearly good. The whole price is in §6.3/§7, and the number that decides it is R2's table, which is not on the page.
2. **The cooked path's cost is unmeasured.** `Open` on a mapped gigabyte now reads a 96 MiB directory linearly and then binary-searches it randomly, once per back-reference. Nobody has measured that against v1's walk. Before the emitter step, that is one benchmark, not a discussion.
3. **Nothing here is held by test yet** — this is spec-only by design, but D1 and D2 are exactly the shapes that pass a happy-path corpus and fail a negative control. The negative controls #251 step 5 names should include: a forged region whose back-reference names an unentered node (D1), and a round trip through a record whose type id this build cannot name (D2).

gafferongames:
Owner rulings, 2026-09-02 (Glenn, live), on the five residual decisions:

1. "Can we keep the overhead and tracking down for the non-cooked version only, and then when cooking, split it into two parts, the cooked data, and the cooked attribution data so we can separate if needed. We may not need at runtime in a shipped build for example (it is read only and not mutable)"
2. "u64 now" (the node table's length)
3. "u64 now, why fuck around" (node type ids)
4. "refuse" (Cook on a region loaded from a newer schema)
5. "we can worry about optimization later, unless there are specific decisions we need to make now that will knowingly make things slower." — none known: the u64s cost bytes not time; the directory scan is linear.

Reading of (1) for the page, Rowan's, labelled: the per-node bookkeeping (type id, directory) lives in the REGION form; `Cook` writes TWO parts — the data and the attribution — with the header locating both; `Open` validates against the attribution once and the attribution may be dropped/unmapped after, since reads never touch it; a cooked file with no attribution cannot be validated and is REFUSED unless the caller explicitly opts into trusting it (the shipped-build case). The owner may strike the trust opt-in.

gafferongames:
Owner, same minute, correcting Rowan's reading of (1): "cooked data at a specific version should just load. It should probably only be loadable at that specific version." / "or a subset of versions AT MOST" / "consider an asset format that rarely changes over time." → The layout id is the gate: on a match, Open points without validating (the runtime default); the attribution is for an explicit validated open or tools; a mismatch refuses and the wire is the fallback; exact match now, a declared compatible set is the only widening contemplated. My trust opt-in is withdrawn.

gafferongames:
Owner, framing for §7 (verbatim): "the cook would be LOCAL to the version your game needs" / "or on some distributed system." / "your game will only ever load the locally cooked asset specific to the versions that build knows." / "the generic table (tooling) needs to handle all versions."

gafferongames:
Owner (verbatim): "or, the game, if cooking is not required for speed, would just use the generic table" — cooking is the game's choice per asset, never the format's requirement; the wire is what a game loads by default.

gafferongames:
Owner (verbatim), §7's governing sentence: "cooking is fundamentally an optimization".

gafferongames:
Owner's DEFINITION (verbatim): "convert the table data structure into a specific version that can be memory mapped, endian fixed up and just loaded quickly by a build at the specific version it is cooked to." / "that's cooking." / "so for example, if you want to be smarty pants, Assets.bin is a cook, Config.bin is a cook".

gafferongames:
Owner, clarifying the space example: "from some more generic format. OR if we don't need this optimization (we don't) Assets.bin / Config.bin is the more generic versioned table data structure format, so we have some flexibility." — space's bins stay the generic wire; the cook is the optimization it does not take.

gafferongames:
Owner's worked example for §7 (verbatim): "imagine a huge data file that specifies all the meshes used in the game for example, or all the texture files." / "that would be a cook" / "Config.bin and Assets.bin are small enough that the overhead of loading doesn't require a cook."

gafferongames:
# Delta read — rebased #264 (+694/−184 vs 85becbf)

## VERDICT: **LAND AFTER FIXES** — three, all small

The design is right and the rulings are applied faithfully. Every one of the nine defects from the first read is closed, and §14-2 now records *why* the marked traversal fails rather than merely what it costs. Fresh clone, reproduced with a new scratch encoder + independent decoder written from the rebased §3/§3.1/§2.5/§2.6 (deleted; branch never edited).

Three things the rebase introduced or left behind must move before this is law.

---

# MUST FIX

## F1 — The straddling rule contradicts §6.5's "the load path allocates nothing". Chunk at RECORD boundaries instead.

Reproduced at 1, 2 and 4 chunks. The v2 decode is identical every time — sharing, elision, `owner = 1`, report all zero — so the concatenation semantics are right. But:

```
chunk=106 -> 1 field   v2 decode OK   naive per-chunk reader: 3 records, ok
chunk= 53 -> 2 fields  v2 decode OK   naive per-chunk reader: FAILS "a record header does not fit in this chunk"
chunk= 27 -> 4 fields  v2 decode OK   naive per-chunk reader: FAILS "a record body runs past this chunk"
```

A conforming reader must be **segment-aware for every multi-byte read**: `node_count`, every `u64` type id, every `u32` length — and every whole record **body**. A straddling body has no contiguous view, so the segmented cursor propagates *into the generated per-table body decoder*, not just the node-table scan. My decoder had to `make([]byte, n)` to give a straddling body a contiguous view — **an allocation on the reading path, which §6.5 forbids.** The alternative, materializing the concatenated stream, is worse and equally forbidden. The page states neither, so an implementer guesses, and the cheap guess is wrong.

**Fix: close a chunk at a record boundary** — open a new reserved field when the next record would not fit. Then:

- still ceiling-free (unbounded chunk count, `node_count` is `u64`),
- still deterministic, so `measure == save` is untouched,
- the naive reader becomes the correct reader; no segmented cursor anywhere,
- §6.5 stays literally true and the generated body decoder is unchanged,
- the exception to §3's last-occurrence-wins keeps its entire justification.

The only thing record-aligned chunking cannot frame is a single record larger than one chunk — which the `u32` record length already forbids (F2). So this costs nothing at all.

Two smaller silences in the same bullet, whichever way it goes: is a **short non-final chunk** legal input, or refused? And what does a reader do when one chunk of the stream arrives with a **kind other than 12**? I guessed "accept" and "count one kind mismatch, drop that chunk".

## F2 — "no single node body can reach 4 GiB" is false, and nothing refuses the case.

> "The **length** is a `u32` and needs no more: every payload inside a body is itself `u32`-framed (§3), so no single node body can reach 4 GiB."

The argument bounds each payload, not their **sum**. Counter-example, entirely legal under §2.5 (which this PR does not touch):

```
table Asset { a *bytes   b *bytes }        a = b = 2 GiB
body = 2 × (2 + 1 + 4 + 2 GiB) = 4.00 GiB  >  u32 record length (4294967295)
```

That is precisely the construct #259 exists for — "imagine they could be very large, like a gigabyte". The aggregate ceiling went away and a per-node one stayed, defended by a non-sequitur.

**Fix, pick one:**

- **(a) A named save-time refusal** — "a node body past 4 GiB − 1", naming the node — and delete the sentence. Zero bytes; the user's fix is more nodes, which is the shape the design wants anyway. **Recommended.**
- **(b) `u64` record length.** Record header 12 → 16 bytes. At 64-byte nodes that moves wire overhead 18.75% → 25% for a case that is pathological. Not worth it.

## F3 — §7's header: most of it may stay implementation-defined; three facts may not.

I wrote the header from §7 and had to guess: magic width; layout-id width; both part lengths' widths; the number and width of the reserved words; field order; the byte order the header's own multi-byte fields are written in (the magic is the order check, so it must be read bytewise first); and the attribution part's alignment (its entries are `u64` pairs).

Most of that is legitimately unspecified — §7 never claims third-party implementability the way §3 does, and a cooked file never crosses builds. **Three are semantic, not cosmetic:**

1. **The two part lengths must be at least 64 bits.** `CookMeasure` hands back `int64_t`s. A 32-bit header field silently reimposes, at cook time, the ceiling §3.1 just removed — on exactly "a huge data file that specifies all the meshes used in the game". This is "u64 now, why fuck around" applied one section later.
2. **The layout id's width is now safety-critical.** Under the new rule, a match means `Open` validates *nothing*. The id is the sole guard between a runtime and a foreign region. State the width.
3. **"`Open` checks the header and points. Magic, byte order, layout id."** is now a **closed list**, and it is short two entries its own refusal list carries: truncation checked against `size`, and base alignment. Since nothing else is checked at all, that sentence has to be exactly the check.

---

# SHOULD — worth a clause, not a block

- **`unknown` now scales with data size.** Verified: N chunks → exactly N `unknown`, 0 `malformed`, plus the 2 `kind_mismatch` the pointer fields owe. The claim holds precisely. But one absent field in a 12 GiB save reports 3 unknowns. §3.1 says "one unknown per field", which is true and easy to misread; one clause saying the count is per **transport field**, not per schema difference, saves a tools author.
- **The cook farm crosses a machine boundary.** §7's own pipeline sentence contemplates "a distributed cook farm", and its output arrives with a matching layout id and is pointed at unvalidated. The gate is the owner's ruling and I am not relitigating it — one sentence naming that as `OpenValidated`'s home would close the loop the page itself opens.
- **§13.4's first bullet has no posted provenance.** "the only time we need the optimization is to reduce load times for assets. in short, it's the flatbuffer style optimization, don't parse, just make something that points at the mmap'd data structure loaded as is and *works*" appears in no comment on this PR, #251 or #248 — every other verbatim in §13.4 does, and I checked each. Post it, so the record matches the page.
- Type id `0` still rebounds but nothing reserves it now that the raw byte node is gone. Harmless — it keeps the four vocabularies uniform — but it is a reservation with no referent.

---

# DECISION 4 — I recommend the pointer index take its OWN KIND, and the window closes at ship

Numbers first. A plain `u32` read as an index is silent only when it is in range **and** names a node of the declared type; the in-range factor alone:

| node_count | P(in range) for a uniform `u32` |
|---|---|
| 100 | 0.000002 % |
| 10,000 | 0.000233 % |
| 1,000,000 | 0.0233 % |
| 16,777,216 | 0.391 % |

So that direction is rarely silent. **The other direction — an index read as a plain number — is always silent**, and that is the half that matters.

The page's objection to a new kind is §3's real rule: *"a kind a reader does not know is not skippable"*. **That objection does not apply here**, and this is the whole of the argument: it bites only for a kind added *after* a reader ships. Nothing has shipped — #248 holds v2.5.0, and the nine-language port is queued behind this very PR. A kind spent now is simply part of the closed set every reader will ever have.

Cost: **zero bytes** (the kind byte already rides) and one row in the fixed-width skip rule. Kind `16` is taken by §3.2's keyed body, so `17` is free and skips 4 bytes — the same class as kinds 1–11.

Buys: §4.1's third silent edit becomes an ordinary counted kind mismatch in **both** directions, leaving the silent class at **two** — and the other two, defaults and flags, cannot be closed at all. §4.1's answer for this one is discipline ("change the field's name when you change what it holds"), which is the answer flags get, and flags is the case the page itself calls unfixable.

This is not an optimization deferred under ruling 5 — it is a decision that becomes **unavailable at ship**, which is the exception that ruling names. Recommendation: **spend kind `17`, keep §4.1 at two, record the trade in §14.**

If the owner prefers kind 8, the page as written is honest — §4.1 names the edit, bounds it, and gives it an answer — so this is a should, not a must.

---

# VERIFIED — reproduced, not read

- **The rebased §3.1 encodes and decodes cleanly at `u64` type ids.** §3.1's own 4-node graph (root + A + B + P, P shared three times, `B.next` null, both nodes' `owner` naming the root): **129 bytes**, `node_count = 3`, root = 1, A = 2, B = 3, P = 4, `Node = 0x66bd1cc6d2f6b68d`, `Palette = 0xc8af536e2084ade0`. Sharing preserved, the null elided and reading back 0, and `owner = 1` — an index **below** the node carrying it — resolving correctly, which is the case the old index-monotonic rejection turned on and which the previous worked example did not exhibit. Byte-unambiguous throughout; the 12-byte record header (`u64` id + `u32` length) parses without lookahead.
- **The repeating field decodes identically at 1, 2 and 4 chunks** — the concatenation semantics are unambiguous *as semantics*. F1 is about what a reader must be, not about what the bytes mean.
- **An old-layout reader lands exactly what the page claims**: `unknown` = one per chunk, `kind_mismatch` = 2 (the two pointer fields), `malformed` = 0, at every chunk count.
- **§6.3's cost table is correct.** I recomputed it independently at the new widths and it matches to the digit: 12 B/node wire headers and 16 B/node attribution — 18.75% / 25% at 64-byte nodes, 37.5% / 50% at 32-byte, 1.17% / 1.56% at 1 KiB. The table is on the page, with the lever named (node size, not encoding).
- **D1 is closed, and closed the right way.** §7's check is a linear scan over the directory: every node visited once, no reference ever followed, unreachable nodes checked too, `O(R + P log N)`, no allocation, pack order and check independent. §14-2 records the marked traversal's *failure* — the 48-byte forgery — not just its cost.
- **D2 is closed.** The sentinel is `0xFFFFFFFFFFFFFFFF`, explicitly distinct from the root's `0`, so an index through it yields NULL and cannot fabricate the root; and §11 turns cooking a region that carries one into the owner's "refuse", naming the index.
- **D3 and D5 are closed by deferral, correctly.** §3.1 now says buffers are leaf, framed inline, charging no node and no depth; and a table union arm is a by-value nesting framed as §2.6 says. I diffed §2.5 and §2.6 against 85becbf: **byte-identical**. This PR specifies nothing about the landed constructs, the `L == 4` discriminator is gone, and with it the one-union-two-layouts problem.
- **D4, D6, D7, D8 closed**; G1–G7 all answered on the page — the false-guard edge, the index-`1` type check, the two-pass `Load`, per-type alignment of directory offsets, the all-null array elision, and "the record scan is authoritative" in as many words.
- **D9 closed.** The wire-version `v1`/`v2` vocabulary is gone from the page entirely, so §4's `// v1:` now unambiguously means a schema version; no "the break is taken now", no corpus narration, no changed-from language anywhere I could grep.
- **R7 closed on the page**: §9 states the serial numbering prologue plainly, and correctly notes the scatter-write itself is untouched and the prefix-sum simpler over top-level records.
- **§15 gains exactly one entry** — the declared compatible set — loses the wire-v2 entry it is landing, and rewords cross-endian from `Open` to COOKING, which follows the endianness ruling rather than adding to the list.
- **§13.4's verbatims match the posted comments**, one exception noted above; the pipeline, the definition, the mesh/texture pair and Config/Assets-stay-on-the-wire all read as ruled.

---

**Land it once F1, F2 and F3 are in.** F1 is the only one with any shape to it, and the fix makes the section simpler than it is now.

gafferongames:
Owner (verbatim, posted for the record so §13.4 matches): "the only time we need the optimization is to reduce load times for assets. in short, it's the flatbuffer style optimization, don't parse, just make something that points at the mmap'd data structure loaded as is and *works*". Also, Rowan's decision on the delta read's recommendation: a pointer index rides under its own kind, 17 — zero bytes, one skip row, closes the third silent edit while nothing has shipped (the owner's ruling-5 exception).

gafferongames:
# Final delta read — #264 on 69bc4ec (+814/−191)

## VERDICT: **LAND**

F1, F2 and F3 are fixed, kind `17` is taken with its cost stated honestly, and the scope is clean. Fresh clone, new scratch encoder + a deliberately **naive, zero-copy** decoder written from the rebased §3/§3.1/§7 alone (deleted; branch never edited). Everything below is reproduced, not read.

Two wording clarifications at the end — neither blocks.

---

## F1 — record-aligned chunking: the naive reader is now the correct reader

Wrote the decoder the obvious way: each `0xFFFF` field's payload is a self-contained unit, scanned in order, **subslices only, never a copy**. §3.1's worked graph, records sized 42 / 35 / 21, stream 106 bytes:

| fields | records per field | wire | naive zero-copy decode | copies |
|---|---|---|---|---|
| 1 | `[42 35 21]` | 129 B | correct, report all zero | **0** |
| 2 | `[42 35] [21]` | 136 B | correct, report all zero | **0** |
| 4 | `[] [42] [35] [21]` | 150 B | correct, report all zero | **0** |

Identical decode at every field count — sharing preserved, `owner = 1` resolving, `B.next` elided and reading 0, `Palette.id = 7`. **No segmented cursor, no copy to make a body contiguous, and the body decoder never learns chunking exists**, exactly as the page claims. §6.5's "the load path allocates nothing" is literally true.

The straddle rule is unambiguous: "the FIRST field's payload opens with the count; every field's payload then carries WHOLE RECORDS" plus "bytes left over inside a field make the whole table malformed" pins the boundary from both sides. The 4-field case exercises the edge the rule permits — a **first field carrying only the count and no records** — and "a reader accepts any chunking a writer chose, a short field is legal input" covers it without a guess.

**Whole-or-nothing, tested hostile.** Corrupted one record's length in the second field:

```
record length past its field -> 0 nodes materialized, malformed=1, nothing salvaged
```

No partial numbering, one event. The rule also caught a bug in my own first decoder, which is the good kind of evidence: I left the root's already-decoded `head = 2` in place, and "every pointer in the save reads null" says plainly that it must not. See N2 below for the one consequence that follows from it.

## F2 — the node-body ceiling is a named refusal

§11: "**A node body past `0xFFFFFFFF` bytes** (§3.1)". The false justification is gone; the `u32` record length is now a stated, refused limit rather than an unreachable one. Correct fix, zero bytes.

## F3 — the header: three widths fixed, the rest properly left alone

I wrote the header from §7 **with no guesses**:

```
magic        read BYTEWISE first -- it establishes the byte order of every field after it
layout id    64 bits            -- the sole guard; sized like a digest, not a version counter
data length  64 bits            -- CookMeasure answers in int64_t
attrib len   64 bits            -- 0 is the shipped-build file: data only
reserved     non-zero -> Open refuses
layout       data part first, padded to the region's alignment; attribution follows
```

and `Open`'s **whole** check, in the page's own order: magic → the byte order it establishes → layout id → both part lengths against the caller's `size` → base alignment. That is now a closed list and it reads as one.

Left to the implementation, and the page says so in as many words ("A cooked file never crosses builds, so most of the header's shape is the implementation's business"): magic width and value, reserved-word count, field order, attribution alignment. That is the right line to draw — §7 makes no third-party-implementability claim, unlike §3 — and the three that decide something semantic are exactly the three that got pinned.

## Kind `17` — taken, and the cost is stated where it belongs

The reader checks the coordinator asked for, both directions:

| reader | 1 field | 2 fields | 4 fields |
|---|---|---|---|
| **holds the field by value, has kind 17** | 1 unknown, 2 kind_mismatch, 0 malformed | 2 unknown, 2 kind_mismatch, 0 malformed | 4 unknown, 2 kind_mismatch, 0 malformed |
| **pre-17** | 1 kind_mismatch, 1 malformed | same | same |

Exactly N `unknown` (one per transport field) + one `kind_mismatch` per pointer field — `Scene` declares two — and nothing else. The by-value reader is the one that will exist; it is clean at every chunking.

The pre-17 row is §14-6's stated price, reproduced: it meets kind `17` at `head`, the root's first declared field, and stops the body — 1 kind_mismatch, 1 malformed, and it never reaches the node table at all. §14-6 says this ("meets kind `17` as framing damage and stops that body"), says why the node table must NOT take a kind for the same reason, and bounds it correctly ("no such reader ships"; a pointered save was already unreadable by a tree-encoding build). The trade is recorded where a future reader will look for it.

**§4.1 is back to exactly two** — a changed default, a moved flags bit — with the `*T`/`uint32` pair moved into the "made REPORTABLE by construction" list beside `[E]T`'s kind `16`. The parallel between the two closures is drawn explicitly, which is the right way to leave it.

## Scope — the rebase kept main's page

Mapped every hunk to its section: §2.1, §2.5 (one clause), §3, §3.1, §4, §4.1, §5, §6.2/6.3/6.5, §7, §8, §9, §11, §13, §14, §15, §16.2. **Nothing in §10, §12, §17 or §18.**

- **§15** against 69bc4ec: loses the wire-v2 entry it is landing, gains **exactly** the declared-compatible-set entry, and rewords cross-endian from `Open` to COOKING — which follows the endianness ruling rather than adding to the list.
- **§16.2** against 69bc4ec: gains **one paragraph** and changes no mapping row. It says the text form is a tree, so a shared node round-trips as several and a walker over untrusted data carries its own visit bound. That is a v2 consequence landing in the section that owns it, not a text-form redesign — in scope, and it would have been a hole if it were missing.
- **§2.5 and §2.6** are otherwise untouched; the one-clause change to §2.5 narrows "an array of them (§15)" to "an array of them — a buffer takes no node index, so there is nothing to put in the slots", which is the same refusal with the v2 reason attached.

---

## Two clarifications — not blockers

**N1 — "so a reader that gives up partway still has them" has two antecedents, and the near one is false.**

> "This implementation writes the fields LAST in the root body, after the root's own declared fields, so a reader that gives up partway still has them."

If "them" is the node-table fields, the sentence is backwards, and kind `17` now makes that demonstrable rather than theoretical: the pre-17 run above gives up at `head` — the root's *first* declared field — and never reaches fields written last. If "them" is the root's own declared fields, it is true and a good reason (a reader that dies inside a gigabyte node table has already read the root's real fields). Name the antecedent.

**N2 — whole-or-nothing plus fields-written-last means pointer resolution cannot be inline.**

The node table is discovered *after* the root's pointer fields have been read, so a reader has already stored `head = 2` when it learns the table is malformed. The page's rule is unambiguous — "every pointer in the save reads null" — but the ordering consequence is not stated: a conforming reader must defer pointer resolution until the node table is known good, or null what it already stored. One clause in the whole-or-nothing bullet saves an implementer the bug I wrote and then caught by re-reading it.

---

## VERIFIED — the ledger closes

Every defect from the first read (D1–D9, G1–G7) and every must-fix from the second (F1–F3) is closed on the page, and the three decisions I recommended against the draft — the directory scan over the marked traversal, record-aligned chunking, and spending a kind on the pointer index — are all taken with their reasoning recorded in §14 rather than in a PR body. §14-2 states the marked traversal's *failure* (the 48-byte forgery), not merely its cost; §14-6 states kind `17`'s non-zero cost and why item 5 refuses a kind for the node table. The page argues against itself in the places where that is what a reference should do.

**Land it.**

PR #265, opened 2026-09-02
Tables: optional fields `?T` and enum-keyed arrays `[E]T` (#260 #255)

### LINKED ISSUE #255: Enum-keyed arrays: `slots [ShipType]ShipConfig` — one slot per variant, keyed by variant id on the table wire
Owner, 2026-09-02 (Glenn, live), on the proposal that closes the last positional leak in tables: "I like the enum keyed arrays. That is cool."

**The problem it closes.** An array indexed by an enum ordinal (`[ShipType.Max + 1]ShipConfig`) is positional on the wire: insert a variant in the middle and every later slot lands in the wrong place, silently. Values ride by name after #249; the SLOTS did not.

**The construct.** `ships [ShipType]ShipConfig` — an array with exactly one slot per variant, indexed by the variant. Design of record, Rowan's, for the owner's check:

- **Storage:** `T ships[ShipType.Max + 1]`, indexed directly by the enum value (slot 0 is None's, as the space slot model already does); no count companion (every slot exists). A generated `ships[ShipType::Bomber]` just works.
- **Table wire (§3):** the array body carries `(variant id u16, element body)` pairs; a slot whose element is all-default is elided like any default; None's slot rides only if non-default. On load, each pair is placed by its variant id; an id this reader cannot name is skipped by length and counted `unknown`; a slot the writer did not send takes its default. Insert, remove, reorder variants: every slot still finds its home.
- **Type wire:** the same spelling is accepted and is exactly `[E.Max + 1]T` — positional, bitpacked, same-build, the protocol id moving as it does today. One construct, two wires, the same rule as every other kind.
- **JSON (§16 / #253):** the natural form falls out — an object keyed by variant name, `{ "Fighter": {...}, "Bomber": {...} }`.
- **Reflection:** the field descriptor marks the array enum-keyed and names the enum, so a walker prints slots by name.
- **Refusals:** the enum must be a declared enum (not flags); `[..E]` (bounded) is refused — a keyed array is complete by construction; an element that is a pointer is refused as for any array.
- **Zero-cost:** fixed-size when T is; the gate holds.

**Consumer:** space#441's root tables become `teams [Team]TeamConfig`, `ships [ShipType]ShipConfig`, etc. — the enum-keyed collection Config.bin always was, now as language.

**Held by test:** V1/V2 pair with a middle insert AND a removal, both directions, asserting every surviving slot lands by name; measure==save at exact capacity; negative control: encode slots positionally and show the middle-insert test red. Gate item for schema#248; queued after #254 and #253.

### LINKED ISSUE #260: Optional fields without pointers: `settings ?GunnerSettings` — present-or-absent by value, fixed-size, wire-identical to a pointer
Owner ruling, 2026-09-02 (Glenn, live), on the JSON guard-presence question: "The 'has_gunner_settings' is a hack because previously the gunner settings were a pointer to table. Perhaps the real fix here is to just have a pointer to the gunner settings table, which is originally how it was done in flatbuffers. If you want to have optional fields without needing pointers too, and that's elegant, then go for it."

**The construct.** `settings ?GunnerSettings` — an OPTIONAL by-value field, legal in a table body for any field kind that can nest (a table, a type, a scalar too if wanted). Storage: the value plus a generated presence bool (`settings_present`), so the holder stays FIXED-SIZE — Config.bin keeps the owner's fixed-down-to-root ruling. Wire (§3.1's framing, unchanged): absent → not written; present → ALWAYS written, even all-default, because elision is by presence not content. It is therefore wire-identical to `*GunnerSettings` and to a plain `GunnerSettings` nesting: a schema may move a field among the three and no byte moves (the corpus holds it both directions, as it does for pointer↔value today). Defaults: an absent optional reads as absent with the value at its defaults. JSON (#253 §16): the presence of the key IS the presence; an absent key is absent; the walk writes present ones only. Reflection: an `optional` flag and the presence-companion offset, like `counted`/`count_offset`. Refusals: `?` outside a table body; `?` on a pointer (a pointer is already optional); a default on an optional (presence is the only default).

**What it retires.** The `has_x bool | json_presence` + `if has_x { x X }` idiom for tables. `if` guards remain for what they are: a branch on a real field. Space's ShipConfig/TurretConfig move to `gunner_settings ?GunnerSettings` in the dogfood.

**Held by test:** round trip present/absent/all-default-present; the three-way evolution (optional ↔ pointer ↔ by-value) both directions; JSON presence round trip; zero-cost gate unchanged; negative control: elide an all-default present optional and show the presence test red.

Gate item for schema#248, fixed-class group (the dogfood needs it), with #253 and #255.

### LINKED ISSUE #261: Test: repeated enum-ARRAY id with an unnameable element must reset the slot to None (the array twin of the repaired union-arm case)
From the #254 delta review (non-blocking, take-it-whenever): removing the enum read's reset to None made only the scalar test red; `test_unnameable_enum_element_read` stayed green because a fresh Load's prefill masks a missing reset unless an earlier occurrence wrote the slot. The shipped behaviour is CORRECT (proved: the sabotaged build leaks Gold, the shipped build gives None); what is missing is the assertion — a repeated `grades` id whose second occurrence carries an unnameable element, asserting the slot reads None. Add it beside test_repeated_id_unnameable_variant. Negative control: remove the reset, show it red.

gafferongames:
**Rebased onto 9b13893, docs taken from main.**

SPEC-TABLES.md, SPEC.md and USAGE.md on this branch are byte-identical to main's — #269 is the design of record and already carries every construct here. The branch is code, corpus, tests and generated output only. One behaviour was corrected to match the page: see "Conformance to the page" below.

Closes #260, closes #255, closes #261.

Two TABLE-BODY constructs in the fixed class, and the assertion #261 asked for.
Rebased onto the merged page (#269).

## `?T` — optional fields (#260)

```
table ShipConfig
{
    health   float32 = 100.0
    settings ?GunnerSettings
    tier     ?int32
}
```

Storage is the value plus a generated `<field>_present` bool, so the holder
stays FIXED-SIZE. **Presence decides whether the field rides, never content**:
absent is not written, present is always written, all-default included.

That makes `?T`, `*T` and a plain `T` nesting one field on the wire — **with
one forced asymmetry, at all-default**: a by-value `T` has no presence bit, so
it cannot tell "absent" from "present with nothing to say" and it ELIDES, while
`?T` and `*T` both ride (12 vs 21 bytes on the corpus shape). Stated in §2.3,
§3.1 and §4, and held by the corpus in both the ordinary and the all-default
case.

## `[E]T` — enum-keyed arrays (#255)

```
enum ShipType { Fighter, Bomber, Scout }

table Fleet
{
    ships [ShipType]ShipConfig
}
```

```cpp
fleet.ships[ShipType::Bomber].health = 400.0f;
```

**Owner ruling, 2026-09-02:** *"the 0th entry in the array is not valid, and is
'null'"* / *"memory is cheap so you could just leave the array as-is, and just
put an error/assert on index by zero"* / *"accessing the 0th entry is an
error/assert"*.

So storage is `E.Max + 1` slots — a slot's number IS its variant's value, no
bias anywhere — and **slot 0 is None's and is never valid**. The generated
`TableKeyed<T, E, N>` carries that rule in one accessor rather than leaving it
a convention every caller re-derives:

- `operator[]( E )` asserts on `None` — the earliest a runtime key can be
  caught;
- `Slot<Key>()` static_asserts, so a compile-time `None` is a compile error:
  ```
  error: static assertion failed due to requirement '(ShipType)'\x00' != ShipType::None':
  slot 0 is None's and is never valid: None is the null key of an enum-keyed array
  ```
- its layout is exactly `T slots[N]` — trivially copyable, standard-layout, so
  the holder stays relocatable — and it is emitted only into a unit that
  declares a keyed array.

On the wire the slots ride keyed, so a middle insert, a removal or a reorder
leaves every surviving slot in its own home. A None key never rides; a stored
key of `0` is framing damage, flagged `malformed`. On the TYPE wire the
spelling is exactly `[E.Max + 1]T` — positional, one protocol id either way,
held by `TestKeyedSpellingIsTheSameTypeWire`.

## Review conditions — all five addressed

**F1 — the key enum bypassed both table-closure refusals.** An enum reaching a
closure ONLY as a keyed-array key was never inspected, so `| max = K` headroom
compiled (and then failed to save, permanently and silently) and a genuine
variant-id collision was caught only by clang's duplicate-case error. The key
now joins the closure vocabulary walk, and the diagnostic names the keying
field as the reaching edge:

```
$ ./bin/schema check headroom_key.schema
enum E: | max = 15 reserves values above the declared variants, and table T's field s
keys an array by it reaches it, putting E in a table closure — a headroom value has no
NAME ... (SPEC-TABLES.md §5)

$ ./bin/schema check coll_as_key.schema
enum E: variants Agj and Atj collide on table-wire id 0x798a, and table T's field s
keys an array by it reaches it, putting E in a table closure — rename one
```

Three fixtures in `TestTableRefusals`. Decision 5's "unreachable" claim was
wrong and is withdrawn.

**F2 — the keyed body has its OWN WIRE KIND, `16`.** A keyed body and a
positional one were both kind 14 with incompatible layouts, so the
`[E]T` ↔ `[E.Max + 1]T` edit was silent garbage — a new instance of exactly the
class this feature exists to close. Under two kinds it is an ordinary kind
mismatch. §3's closed set, its payload table and its skip rule ("kinds 12, 13,
14 and 16 read `L` and skip `L` bytes") all carry it, and V1/V2 now spell one
field `ledger` both ways so the fixture runs both directions:

```
positional writer -> keyed reader: kind_mismatch == 1, every slot still default
keyed writer -> positional reader: kind_mismatch == 1, every element still default
```

**F3 — the all-default asymmetry** is stated in §2.3, §3.1 and §4, and
`test_optional_three_way_evolution` now carries the case: `?T` and `*T` stay
byte-identical there, the by-value writer's bytes read back absent with a clean
report, and the nine-byte difference is asserted rather than implied.

**F4 — §2.2's gate sentence** is rescoped to the pointer machinery, which is
what it always meant and what the mechanical gate checks. The measured price of
the feature columns: `sizeof(TableFieldInfo)` **104 → 136**, static rodata only,
no codec branch and no include in a pointer-free unit.

**F5 — slot order** is stated in §3.2: writers emit ascending by variant
ordinal, readers must not rely on it (every slot is found by its key), and
byte-identity against this implementation requires matching that order — the
same rule §3 states for fields.

**F6 — filed as #271**, referenced here: cooked-region scalars are unvalidated,
so a corrupted `bool` byte is a trap representation `Open` accepts. Verified on
`origin/main` with no `?T` in the language, so it predates this work; `?T`
widens it because a `_present` byte is compiler-generated.

**The reviewer's randomised oracle, ported.** `test_keyed_variable_oracle`
sweeps 30,000 `Depot` shapes — which slots ride, chain length 0–3 per slot,
string lengths, `?Meta` on or off — asserting `measure == save` at exact
capacity, that `LoadMeasure` sizes a region `Load` fits, that every field
survives the trip, that `Cook`→`Open` agrees, and that a re-save is
byte-stable. 1.5s in the tables leg; against the pre-fix region pre-pass it
goes red in four more places than the pinned shape did.

## Conformance to the page

Re-read against main after the rebase, the implementation matched §2.3, §2.4, §3.2, §4, §8 and §11 everywhere except one behaviour, now fixed:

§3.2 on a stored key of `0` — *"The reader stops that body, keeps what it decoded, and flags malformed"*. This branch skipped the damaged pair and read on inside the same body, which is more permissive than §3.2 and than §4's framing-damage rule. The reader now breaks out of the body and the field's length carries the parent past it. The fixture puts a GOOD pair behind the null key and asserts it does NOT land, so the test measures the difference between stopping and skipping; reverting to skip-and-continue turns it red:

```
FAIL test/tables/main.cpp:2458: out.tokens[tblv1::Slot::Delta] == 0
tables test: 1 failure(s)
```

## A real defect the variable class surfaced

`<X>LoadMeasureBody` — the region sizing pre-pass, which reads framing and no
field value — read each element's u32 length where a keyed slot's variant id
sits, so every length after the first came out of a key's bytes and the caller
sized its region off garbage. Fixed, and held by `table Depot` in
tables/pointers (a keyed array of variable Layers, an optional fixed table and
a pointer) plus the oracle above. Pack and OpenWalk were already correct.

## The type wire did not move

`TestTablesMoveNoProtocolId` stays green and gained a case: a table carrying an
optional field AND an enum-keyed array moves neither the wire projection nor the
protocol id. `TestKeyedSpellingIsTheSameTypeWire` proves `[E]T` and
`[E.Max + 1]T` project identically and share one id.

## Negative controls

**1. Elide an all-default present optional** (`bytes += 3 + 4 + body` becomes
`if ( body > 2 )`, in measure and save):

```
FAIL test/tables/main.cpp:2115: empty_back.extra_present
FAIL test/tables/main.cpp:2207: w_bare_optional > w_bare_value
FAIL test/tables/main.cpp:2214: w_bare_pointer == w_bare_optional
FAIL test/tables/main.cpp:2215: memcmp( other, third, (size_t) w_bare_optional ) == 0
FAIL test/tables/main.cpp:2245: tblp2::LinkAt( empty_pointer->link ) != NULL
FAIL test/tables/main.cpp:2635: loaded->spare_present == root->spare_present
tables test: 6 failure(s)
```

**2. Encode keyed slots positionally** (writer writes `i`, reader places by the
raw key) — the middle-insert test goes red, and Beta's slot lands in Omega's:

```
FAIL test/tables/main.cpp:2330: report.unknown == 1
FAIL test/tables/main.cpp:2333: out.bank[tblv2::Slot::Beta].power == 20
FAIL test/tables/main.cpp:2334: strcmp( out.bank[tblv2::Slot::Beta].label, "b" ) == 0
FAIL test/tables/main.cpp:2337: out.bank[tblv2::Slot::Omega].power == 0
FAIL test/tables/main.cpp:2341: out.tokens[tblv2::Slot::Beta] == 7
FAIL test/tables/main.cpp:2342: out.tokens[tblv2::Slot::Omega] == 0
FAIL test/tables/main.cpp:2343: out.ranks[tblv2::Slot::Beta] == tblv2::Grade::Gold
FAIL test/tables/main.cpp:2344: out.ranks[tblv2::Slot::Omega] == tblv2::Grade::None
FAIL test/tables/main.cpp:2365: report.unknown == 3
FAIL test/tables/main.cpp:2368: out.bank[tblv1::Slot::Beta].power == 3
FAIL test/tables/main.cpp:2369: out.bank[tblv1::Slot::Gamma].power == 0
tables test: 11 failure(s)
```

**3. #261 — remove the enum read's reset to None.** Before this PR only the
scalar case went red; the array twin now bites too:

```
FAIL test/tables/main.cpp:623: out.grade == tblv1::Grade::None
FAIL test/tables/main.cpp:2493: out.grades[0] == tblv1::Grade::None
FAIL test/tables/main.cpp:2494: out.grades[1] == tblv1::Grade::None
tables test: 3 failure(s)
```

**4. Drop the keyed key skip in the region pre-pass** (the defect above) — the
pinned shape and the oracle both go red:

```
FAIL test/tables/main.cpp:2533: !report.malformed && report.unknown == 0 && report.kind_mismatch == 0
FAIL test/tables/main.cpp:2541: node != NULL && node->value == 100 + tier
FAIL test/tables/main.cpp:2541: node != NULL && node->value == 100 + tier
FAIL test/tables/main.cpp:2633: !report.malformed && report.unknown == 0 && report.kind_mismatch == 0 && report.clamped == 0
FAIL test/tables/main.cpp:2642: node != NULL && node->value == slot * 1000 + link
FAIL test/tables/main.cpp:2650: graphdemo::DepotSave( loaded, again, sizeof( again ) ) == need
FAIL test/tables/main.cpp:2651: memcmp( wire, again, (size_t) need ) == 0
tables test: 7 failure(s)
```

## Decisions the issues did not make — mine, labelled

1. **`?` applies to scalars: YES**, as recommended, and to enums, flags,
   nested types and nested tables. REFUSED on a union (`None` IS the absence)
   and on arrays, strings and `bytes` (each already carries a count or length
   whose zero is emptiness) — the latter a named follow-on in §15.
2. **Every keyed element is u32 length-prefixed, scalars included** — one skip
   rule for an unknown key whatever the element is, at 4 bytes per present slot.
3. **`[E]T` means keyed-on-table-wire, positional-on-type-wire wherever it is
   declared** — including in a `type` inside a table closure. A `type` keeps its
   raw-struct storage, so a keyed array declared there is a plain array with no
   keyed accessor and only its table-wire encoding is keyed; stated in §2.4 and
   exercised by `type ScoreBoard` in the corpus.
4. **A keyed array elides PER SLOT** (unlike a fixed array of tables, which
   writes every element): the key is identity there, so a default slot drops
   safely. An array with no present slot elides whole.
5. **A slot whose key no variant names is a SAVE failure**, not a silent drop.
   No longer justified as unreachable — F1 made it reachable, and F1's refusals
   are what close it at compile time.
6. **The five new descriptor columns are unconditional**; §2.2's gate is about
   pointer machinery, now said so on the page (F4).
7. **`key_name` and `key_id` take the SLOT INDEX**, which IS the variant's
   value, so a walker steps `[0, array_bound)` with no arithmetic of its own.
   Slot 0 is marked invalid by the one id no declared name can hold:
   `key_id(0) == 0`, `key_name(0) == "None"`.
8. **`?` binds to the whole field type and precedes an array bound**, so
   `?[..4]T` parses and is refused by name.

## Battery

`make test` green end to end (all nine backends, the fixed-point/128-bit unit,
the diagnostics suite, the goldens and the fmt-drift gate), `make check`,
`make fmt` clean, `make tables-zero-cost` green, `go test ./...` green,
`golangci-lint` and `modernize` clean.


gafferongames:
# Adversarial review — #265 (`?T`, `[E]T`)

Cold read, fresh clone at `3e84a11`, no part in writing it. Everything below was
re-derived or re-run here; the branch was never edited (negative controls ran in a
throwaway detached worktree, reverted).

**This PR is HELD from merge regardless of this verdict** — owner ruling 2026-09-02,
*"spec first always"* (#248): the spec-of-record PR lands first and every
implementation PR rebases onto it. Where the findings below change the page, they
should land there, not be bolted on here.

---

## VERDICT: **MERGE-WITH-CONDITIONS**

The construct is sound, the wire is implementable from the page, and the walk fix
generalises. Two conditions are real defects with proof, not taste: **F1** (a
compiler hole that lets a legal schema reach an unrecoverable save failure with no
diagnostic, and lets a genuine id collision past the one checker built to catch it)
and **F2** (a new silent misdecode in the exact class this feature exists to close).
Both are small and local. F3–F5 are page corrections.

---

# FINDINGS

## F1 — `[E]T` bypasses BOTH table-closure enum refusals. Decision 5's "unreachable" is false.

`checkTableVariantIdentity` (`internal/check/check.go:1808`–`1833`) collects a
closure's vocabularies by walking `f.Type.Ref` for `ir.TNamed` fields only. An
enum-keyed array's **key** enum is not `f.Type.Ref`. An enum that reaches a table
closure *only* as a key is therefore never inspected, and both §5 refusals are
skipped.

**(a) Headroom — a legitimate schema, an unrecoverable runtime failure, no diagnostic.**

```
package t
enum E | max = 15 { A, B }
table T { s [E]int32 }
```
```
$ ./bin/schema check headroom_key.schema
$                                          # accepted, silently
$ grep 'int32_t s\[' headroom_keyTable.h
    int32_t s[16] = {}; // [E]: one slot per variant, indexed by the value (E.Max + 1)
```
13 of those 16 slots are named by nothing. Writing one:
```
A only     : measure=24 save=24
A + slot 7 : measure=-1 save=-1
```
Permanent save failure, no message, no report event. Move the same enum to a scalar
field and the refusal fires correctly:
```
enum E: | max = 15 reserves values above the declared variants, and table T's field g
reaches it, putting E in a table closure — a headroom value has no NAME ...
```
So the hole is exactly the key path. This is realistic: `| max` exists for the packet
wire, and a shared enum keyed into a table need not appear as a scalar anywhere.

**(b) Id collision — the refusal that exists for this never fires.**

`Agj` and `Atj` both fold to `0x798a`. As a scalar field or as an array **element**,
schema refuses and names both. As a **key**:
```
package t
enum E { Agj, Atj }
table T { s [E]int32 }
```
```
$ ./bin/schema check coll_as_key.schema
$                                          # accepted
```
It is caught only downstream, by accident:
```
coll_as_keyTable.h:197:14: error: duplicate case value '31114'
  197 |         case 0x798a: out = E::Atj; return true;
```
That is clang, not schema — against §5's own promise (*"the compiler sees every id in
the closure and refuses once, forever"*) and against CLAUDE.md's rule that the
compiler owns this. The eight pending ports will not all reproduce that accident.

**Fix:** feed the key enum into `checkTableVariantIdentity`'s `enums` map alongside
`f.Type.Ref`. Both refusals then fire with their existing messages. Decision 5's
justification needs rewriting either way — it is currently reachable.

## F2 — `[E]T` and `[E.Max + 1]T` are one kind with incompatible bodies, and the edit between them is silent garbage.

§2.4: *"On the TYPE wire the spelling IS `[E.Max + 1]T` ... One construct, two wires."*
§4: *"A changed array BOUND (a literal, a constant, or an `E.Max + 1` expression that
moved): the array still loads, and the bound is not part of identity."*

A reader who believes both — and the page gives no reason not to — gets:

```
[Grade]int32 writer -> [Grade.Max + 1]int32 reader
  wrote  : [0 10 20 30]
  read   : [275463 655360 -1545142272 0]
  report : unknown=0 kind_mismatch=0 clamped=0 malformed=0
```

Both spellings are kind `14`. Nothing on the wire distinguishes a keyed body
(`elem kind, N, (key,L,elem)*`) from a positional one (`elem kind, N, elem*`), so no
report event can fire. This is a **new instance of the exact class §3.2's contrast
paragraph condemns** — *"insert a variant in the middle and every later slot lands one
place off, with nothing on the wire to say so and no report event that could fire"* —
introduced by the change that closes the old one.

The reverse direction (`[4]int32` writer -> `[Grade]int32` reader) happens to set
`malformed=1` on this shape, by luck of the framing rather than by design.

Minimum condition: **name the hazard on the page** — §4's kind-mismatch paragraph
already calls out its one known coarse edge (enum vs `uint16`, both kind `7`); this is
a second and much worse one, and §2.4's "same spelling" sentence needs the table-wire
caveat beside it. Whether the wire should distinguish the two bodies is the owner's
call and belongs on the spec PR.

## F3 — the three-way `?T`/`*T`/`T` identity is overclaimed in three places; it breaks at all-default.

```
P1  by value, all-default          12 bytes: df300c030000006f6e650000
P3  ?T present, all-default        21 bytes: df300c030000006f6e65b20b0d0200000000000000
P2  *T non-null, all-default       21 bytes: df300c030000006f6e65b20b0d0200000000000000
P1 == P3 ? NO
P2 == P3 ? yes
P1 all-default -> ?T reader: link_present = FALSE (report clean = yes)
P1 all-default -> *T reader: link = NULL
```

`?T` and `*T` agree (presence rides). A by-value `T` at all-default **elides** — v1's
own rule, §3 — and reads back as absent/null. Nine bytes apart, clean report.

The page states the promise unconditionally three times:
- §2.3 *"a schema may move a field among the three and no byte moves, in either direction"*
- §3.1 *"A schema may change any of them into any other and no byte moves, in either direction"*
- §4 *"not an evolution event at all — the bytes do not move"*

All three are false at all-default. The corpus's byte-identity check
(`test/tables/main.cpp:2135`–`2158`) uses `value = 66` only, so it cannot see it.

The asymmetry is **forced and correct** — a by-value `T` has no presence bit, so it
cannot distinguish "absent" from "present with nothing to say", and elision is what
makes old-reader/new-writer work. It is the one real asymmetry of the three-way
promise and it belongs on the page and in the test, not left as a surprise. Condition:
state it, and add the all-default case to `test_optional_three_way_evolution`.

## F4 — §2.2's zero-cost gate text says "not one extra descriptor column". Five just landed.

`SPEC-TABLES.md:162`–`166`:
> *no builder, no arena, no reference type, no lifecycle surface, **not one extra
> descriptor column**, not one extra branch in a codec, not one extra `#include`*

Untouched by this PR, and false. Measured — main's emitter vs the branch's over
**identical** value-only input (`origin/main`'s `tables/examples`), the only changes
are the five columns and their doc comment: no codec branch, no include, no pointer
machinery. So **decision 6 is substantively honest** and the mechanical
`make tables-zero-cost` grep gate is green. But the sentence is wrong, and has been
since #254's `variant_id`. The measurable price:

```
sizeof(TableFieldInfo) = 136        (branch)
sizeof(TableFieldInfo) = 104        (main)
```
+32 bytes per field descriptor in every unit, pointer-free ones included. Static
rodata only. Given "spec first always" and present-state-only, this PR was editing
that section's neighbourhood and should have fixed the sentence.

## F5 — §3.2 does not state slot ORDER within a keyed array body.

§3 tells a third-party implementer exactly what byte-identity costs for FIELDS
(*"matching its declaration order as well as its framing"*). §3.2 says nothing for
SLOTS. Writing an encoder from the page, ascending variant ordinal was the one thing I
had to guess. One sentence.

## F6 — NOT this PR, recorded because `?T` widens it: cooked-region bool bytes are unvalidated.

Verified on `origin/main` `393ff35`, with no `?T` in the language at all:
```
MAIN (origin/main 393ff35, no ?T in the language): declared bool corrupted -> Open ACCEPTED
runtime error: load of value 129, which is not a valid value for type 'bool'
```
`OpenWalk` validates pointer deltas and count/length companions but no scalar, so any
`bool` in an untrusted cooked file is a trap representation. Every `?T` adds one more
such byte, and unlike a declared `bool` a `_present` byte is compiler-generated, so a
user cannot avoid it. **Inherited, not a blocker here** — worth its own issue.

---

# VERIFIED GOOD

1. **The wire is implementable from the page — byte-identical on the first try.** A Go
   encoder written from §3/§3.2 alone (a `?int32` present, a `?Nested` absent, a
   `?Nested` present-all-default, and a `[Grade]Slot` with one elided slot, one
   present, one unknown to the reader) matched the generated C++ `Save` exactly:
   ```
   cpp: 58e9040700000049a60d020000000000d8f40e330000000d02000000e7a31100000009d604...
   go : 58e9040700000049a60d020000000000d8f40e330000000d02000000e7a31100000009d604...
   BYTE-IDENTICAL: from-the-page Go encoder == generated C++ Save (76 bytes)
   ```
   Slot order (F5) was the only guess. `measure == save == 76`.

2. **Skip-by-length holds for an unknown key at every element kind.** Reading those
   same bytes with a build whose `Grade` has no `Gold`:
   ```
   load ok=1  report: unknown=1 kind_mismatch=0 clamped=0 malformed=0
   tier_present=1 tier=7 | gone_present=0 | empty_present=1
   slot 1: power=0 tag=''      slot 2: power=20 tag='s'
   ```
   And for a keyed **scalar** array (decision 2):
   ```
   reader without Gold: Bronze=10 Silver=20 | unknown=1 malformed=0 kind_mismatch=0
   ```
   Cost, measured: 44 bytes for 3 present `int32` slots — the u32 length is 4 of the
   10 bytes per slot, 40% of the body (57% for `i8`). §3.2 already states the
   `2 + 4` price, which is the right call: one skip rule, one reader.

3. **All four negative controls reproduce.** NC1 (elide an all-default present
   optional) and NC3 (#261, drop the enum read's reset) reproduce the PR body's output
   verbatim. NC4 (drop the keyed key skip) likewise:
   ```
   FAIL test/tables/main.cpp:2418: !report.malformed && report.unknown == 0 && report.kind_mismatch == 0
   FAIL test/tables/main.cpp:2426: node != NULL && node->value == 100 + tier
   ```
   NC2 (positional slots) goes red on the same middle-insert assertions the body lists
   (2298, 2299, 2306, 2308, 2330, 2333); my sabotage differs slightly from the
   worker's, so the count is 25 rather than 11.

4. **The `LoadMeasure` fix generalises past its pinned instance — proved with a
   stronger instrument than the PR ships.** A memory-safety sweep is the *wrong*
   oracle here: I built a 400k-iteration hostile-wire fuzzer over `DepotLoadMeasure`,
   `DepotLoad`, `Cook`, `Open`/`OpenWalk` and the fixed readers under ASan+UBSan, and
   **negative-controlled the fuzzer against the pre-fix emitter — it stayed clean.**
   The defect is wrong data, not a bad access. So I wrote a correctness oracle instead:
   30,000 randomised `Depot` shapes (which slots ride, chain length 0–3 per slot,
   string lengths, `?Meta` on/off), asserting `measure == save`, that `LoadMeasure`
   sizes a region `Load` fits, that every field survives the trip, that `Cook`→`Open`
   agrees, and that a re-save is byte-stable:
   ```
   ORACLE GREEN  iters=30000  failures=0            (branch)
   ORACLE RED    iters=30000  failures=105849       (pre-fix emitter)
   ```
   Recommend landing something like it — the shipped test pins one shape.

5. **Every refusal fires by name, with messages that teach.** `?` in a `type` body / on
   a pointer / on a union / on an array, string or `bytes` / with a specified default;
   a field colliding with `<field>_present`; `[..E]`; `[F]` naming a `flags`; a pointer
   element. Decision 8 holds: `?[..4]T` parses and is refused by name, not by
   "expected a field type".

6. **Per-slot elision (decision 4) is lossless.** A keyed array's slots all exist by
   construction, so "present but default" and "absent" denote the same value and the
   distinction is correctly not made. A slot whose element is otherwise default but
   carries a PRESENT all-default `?T` still rides — elision is measure-based, so
   presence propagates through the nesting.

7. **Gates.** `make tables-zero-cost` green; committed `generated/` byte-unchanged;
   `TestTablesMoveNoProtocolId` and `TestKeyedSpellingIsTheSameTypeWire` green;
   `go test ./...` green; `make check` and `make fmt` clean (tree unchanged);
   `schema_test`, `_guard`, `_tables`, `_random`, `_ludicrous`, `_c`, `_c_ludicrous`
   all green locally; CI green on all eight checks. The Dart/Java/Elixir pins are not
   installed in this fresh clone, so those legs rest on CI's `test (ubuntu-latest)`.

8. **Docs.** §2.1's bool-guard idiom is properly retired for `?T`; SPEC §4.2 carries
   `?` in `Type` and the enum-bound note; §11 lists the new refusals; §15 banks the
   `?`-on-array follow-on; §8 documents the five columns. No history or decision
   narration in the added prose — present-state clean.

9. **Provenance clean.** Six commits, all `Rowan Claude <rowan@mas-bandwidth.com>`,
   author and committer, no strays.

---

## Conditions, in order of weight

1. **F1** — refuse a headroom or collision-bearing enum used as a keyed-array key;
   correct decision 5's reachability claim.
2. **F2** — at minimum, name the `[E]T` ↔ `[E.Max + 1]T` table-wire hazard on the page
   beside §4's existing coarse-kind edge and §2.4's "same spelling"; whether the wire
   should separate the two bodies is an owner call for the spec PR.
3. **F3** — state the all-default asymmetry in §2.3/§3.1/§4 and add the case to
   `test_optional_three_way_evolution`.
4. **F4** — fix §2.2's "not one extra descriptor column"; **F5** — one sentence on slot
   order in §3.2.
5. **F6** — file separately; it predates this work.

gafferongames:
# Delta review — #265 at `a8a48d4`

Second cold pass, fresh clone, `gh pr checkout 265`. Everything below was
re-derived or re-run here rather than read off the commit message; the branch was
never edited (controls ran in a throwaway detached worktree, since removed).

## VERDICT: **MERGE** — lands right after #269.

All five findings are closed, and closed at the root rather than papered over. One
item is **not** in this PR and belongs on **#269**: the spec-of-record's §2.4 is
missing two rules this branch's §2.4 now carries (below). Two cosmetic nits, neither
blocking.

---

# The five, re-verified

## F1 — closed, and the fix generalises further than asked

```
enum E | max = 15 { A, B }        table T { s [E]int32 }
  -> enum E: | max = 15 reserves values above the declared variants, and table T's
     field s keys an array by it reaches it, putting E in a table closure — ...

enum E { Agj, Atj }               table T { s [E]int32 }
  -> enum E: variants Agj and Atj collide on table-wire id 0x798a, and table T's
     field s keys an array by it reaches it, putting E in a table closure — rename one
```
Refused **by schema**, not by clang's duplicate-case accident. I also checked the
paths I had not: the same enums keyed inside a **`type` that sits in a table closure**
are refused too, naming the type's field as the edge — and an enum keyed in a `type`
that never reaches a table closure is correctly **not** refused, since the packet wire
identifies variants by ordinal. Control: the same enums as a scalar field still refuse.

## F2 — closed by giving the keyed body its own kind

My own repro of the exact edit that was silent garbage last pass:
```
keyed writer wrote 44 bytes, field kind byte = 16
  -> positional reader: [0 0 0 0]   report: unknown=0 kind_mismatch=1 clamped=0 malformed=0
positional writer wrote 30 bytes, field kind byte = 14
  -> keyed reader:      [0 0 0 0]   report: unknown=0 kind_mismatch=1 clamped=0 malformed=0
```
Was `[275463 655360 -1545142272 0]` with an all-zero report. Now an ordinary kind
mismatch, counted, nothing misdecoded, both directions. The `ledger` fixture
(`V1` positional / `V2` keyed, `test/tables/main.cpp:2375`) holds the same thing in
the corpus.

**The wire is still implementable from the page, and this time with no guess at all.**
I rewrote the from-the-page Go encoder against the branch's §3/§3.2 — kind 16,
ascending ordinal, no `None` key — for a table with a `?int32` present, a `?Nested`
absent, a `?Nested` present-all-default and a `[Grade]Slot` with one elided slot and
two present:
```
cpp: 58e9040700000049a60d020000000000d8f410330000000d02000000e7a31100000009d6...
go : 58e9040700000049a60d020000000000d8f410330000000d02000000e7a31100000009d6...
BYTE-IDENTICAL (76 bytes)
```
F5's slot-order sentence is what removed the one guess I had last time.

**Hostile keyed bodies, hand-built from §3.2 and fed to the generated reader:**
```
key 0 alone                                 unknown=0 kind_mismatch=0 malformed=1   slots=[0 0 0 0]
Silver, key 0, Gold  (neighbours must land) unknown=0 kind_mismatch=0 malformed=1   slots=[0 0 20 30]
Silver, UNKNOWN key, Gold                   unknown=1 kind_mismatch=0 malformed=0   slots=[0 0 20 30]
Silver, Gold  (clean)                       unknown=0 kind_mismatch=0 malformed=0   slots=[0 0 20 30]
```
A stored `0` key is framing damage (`malformed`), *not* counted as an unknown name,
and the decode continues with both neighbours landing — exactly as §3.2 says.

**And the case the new kind creates:** an old, keyed-free reader meeting a kind-16
field it has never heard of.
```
keyed-free reader vs a kind-16 field: load=1 keep=99 (want 99)
  report: unknown=1 kind_mismatch=0 clamped=0 malformed=0
```
Skipped by length, counted, the following field intact. That is what licenses the one
new codec arm noted below.

## F3 — stated, accurate, and held

§2.3 now carries *"One asymmetry, and it is forced: at ALL-DEFAULT the three
diverge"*, §3.1 the *"Except at all-default"* clause, §4 the *"one exception"* clause.
The behaviour matches the text exactly:
```
P1  by value, all-default          12 bytes: df300c030000006f6e650000
P3  ?T present, all-default        21 bytes: df300c030000006f6e65b20b0d0200000000000000
P2  *T non-null, all-default       21 bytes: df300c030000006f6e65b20b0d0200000000000000
P1 == P3 ? NO      P2 == P3 ? yes
P1 all-default -> ?T reader: link_present = FALSE (report clean = yes)
P1 all-default -> *T reader: link = NULL
```
The corpus holds the all-default case beside the ordinary one (`main.cpp:2195`+).

## F4 — §2.2 rescoped to what the mechanical gate actually checks

*"A fixed-size table pays nothing for the POINTER MACHINERY ... not one
pointer-shaped descriptor column"*. That is what it always meant and what the grep
gate enforces. `make tables-zero-cost` green; committed `generated/` byte-unchanged.

## F5 — slot order stated

*"writers emit ascending by variant ordinal, and readers must not rely on it"*, with
the byte-identity consequence spelled out as §3 does for fields.

---

# The owner's storage ruling, checked with my own assertions

Not the emitter's `static_assert`s — mine, in a separate TU:
```cpp
static_assert( std::is_trivially_copyable<K>::value );
static_assert( std::is_standard_layout<K>::value );
static_assert( sizeof( K ) == 4 * sizeof( Slot ) );
static_assert( offsetof( K, slots ) == 0 );
```
```
sizeof(Slot)=20  sizeof(TableKeyed<Slot,Grade,4>)=80  4*sizeof(Slot)=80
operator[](Grade::Silver) and Slot<Grade::Gold>() both compile
memcpy round trip: Silver=1 Gold=3
```
`ships[ShipType::Bomber]` reads as itself, and the holder is relocatable in fact, not
just by claim. Both guards controlled:
```
Slot<Grade::None>()   -> error: static assertion failed ... slot 0 is None's and is never valid
operator[](None)      -> Assertion failed: (key != E::None), function operator[]  exit=134
```
**Emission scope holds:** `TableKeyed` appears in zero of the keyed-free units
(`p1`, `p2`, `p3`), and in a keyed-free *and* optional-free unit regenerated from
`origin/main`'s own corpus by both compilers, the entire delta is the five descriptor
columns, their doc comment, and one codec line:
```
<             case 12: case 13: case 14:
>             case 12: case 13: case 14: case 16:
```
That is the unknown-field **skip** vocabulary, not feature machinery — §3 requires
every reader to skip kind 16 by length, and the old-reader test above proves it earns
its place. Correct, and consistent with the rescoped §2.2.

# The region pre-pass, and the four controls

My 30,000-shape randomised `Depot` oracle, rebuilt against this branch and corrected
to honour the None-slot ruling (slot 0 never populated, asserted to stay default):
```
ORACLE GREEN  iters=30000  failures=0
```
Against the pre-pass with only the key skip removed: `ORACLE RED  failures=82370`.
The shipped `test_keyed_variable_oracle` goes red on the same sabotage in 7 places.

400,000-iteration hostile mutation sweep over `DepotLoadMeasure`, `DepotLoad`,
`Cook`, `Open`/`OpenWalk` and the fixed readers under ASan+UBSan: **no sanitizer
report**. Stated honestly, as last pass: this instrument is memory-safety only — I
negative-controlled it before and it is blind to the wrong-data class, so the oracle
above is what carries the correctness claim, not this.

**All four negative controls red:**
- NC1 (elide an all-default present optional) — **6 failures now, was 2**: the new
  all-default three-way assertions (2207, 2214, 2215) and the ported oracle (2635)
  all bite. The battery got materially stronger.
- NC2 (positional slots) — 29 failures, including `banks.slots[0]` staying default.
- NC3 (#261, drop the enum reset) — 3, unchanged.
- NC4 (drop the pre-pass key skip) — 7, plus my oracle's 82,370.

# Gates and provenance

`go test ./...`, `make check`, `make fmt` (tree clean), `make tables-zero-cost`,
`TestTablesMoveNoProtocolId`, `TestKeyedSpellingIsTheSameTypeWire`, and
`schema_test{,_guard,_tables,_random,_ludicrous,_c,_c_ludicrous}` all green locally;
committed `generated/` byte-unchanged; CI green on all seven checks. Seven commits,
every one `Rowan Claude <rowan@mas-bandwidth.com>` as author and committer, no strays.

---

# One thing to carry to #269, and two nits

**FOR #269 — `origin/spec-design-of-record` §2.4 is missing both rules this branch
just added.** #265 rebases onto it, so they need reconciling there or they are lost:

1. **The `type`-body carve-out.** The branch's §2.4 says *"a keyed array declared in a
   `type` body is a plain array of `E.Max + 1` elements with no keyed accessor"* —
   which I confirmed in the generated code: `type ScoreBoard` gets a bare
   `int32_t per_team[4]`, while a table field gets `TableKeyed<TeamConfig, Team, 4>`.
   The spec-of-record's §2.4 has no such sentence, and states the None guard
   unconditionally — *"reaching it is an ERROR ... through the generated keyed type's
   `operator[]( E )`"* — which is not true in a `type` body, where slot 0 is freely
   indexable and there is no keyed type at all.
2. **The key-is-a-closure-vocabulary rule (F1).** The branch's §2.4 closes with *"A key
   is a closure vocabulary, so §5's two refusals are owed to it"*. Nothing in the
   spec-of-record says this — its §5 still carries the two refusals in value-only
   wording. That is the page half of the F1 fix.

**Nit 1 — a mangled diagnostic.** `reachedBy` is templated into `"... and %s reaches
it, ..."`, so the new site string produces:
> *"and table T's field s **keys an array by it reaches it**, putting E in a table closure"*

`internal/check/check.go:1824`. Suggest `"%s %s's field %s, which keys an array by it,"`
→ *"and table T's field s, which keys an array by it, reaches it, putting E in a table
closure"*. Cosmetic, but it is a user-facing refusal in a repo with a diagnostics gate.

**Nit 2 — the runtime guard is debug-only.** Same binary, `-DNDEBUG`:
```
about to index slot 0
NO ASSERT
exit=0
```
That is the ordinary C++ contract for `assert` and matches the owner's words, so it is
not a defect — but §2.4 says *"asserted otherwise"* without noting that a release build
has no runtime guard and `Slot<Key>()` is the form that holds everywhere. One clause.

Neither nit blocks. Ship it.

Issue #262, opened 2026-09-02
The C# tables backend: the first port, driven by the space dogfood (Unity reads config, options and render tables)


gafferongames:
Owner ruling, 2026-09-02 (Glenn, live): "we need the C# port of tables in schema to verify the dogfood. approved."

**Why C# first.** Unity decodes Config.bin, Options and the per-frame render tables in C# (space: unity/Assets/UnitySpaceGame/Config.cs, Options.cs, SchemaBin.cs). The C++ half of space#441 runs end to end without it; Unity cannot follow until this exists. Real load picks the first port.

**Scope.** Mirror the C++ reference (SPEC-TABLES.md is the contract; `ports never invent contracts`): the FIXED class first — storage structs, `<Name>Measure/Save/Load` over caller-owned buffers, the report, reflection descriptors, `?T` optional fields (#260), enum-keyed arrays (#255), name-hashed enums and union arms (#254), u32 lengths and counts, the JSON walk (#253) if landed. The variable class (pointers, arena, region, cooked) is a named follow-on for C#; the backend refuses a pointered unit by name until then. Allocation doctrine (owner, #248): types never allocate; a fixed table with a union may allocate for the arm if the language needs it — C# has no native union, so decide the form (struct with all arms inline vs one reference per arm) by measurement on the space render tables, and state it. Doctrine line: fixed tables as close to their equivalent type as the wire allows; a bench leg beside the type.

**Gates.** Golden wire bytes shared with C++ (the same instance, byte-identical Save; Load of C++-written bytes); the evolution corpus (V1/V2, P1/P2, the three-way optional/pointer/value case) in C#; the C# smoke pattern the estate already uses (serialize.cs); negative control: a sabotaged codec red under the shared goldens.

Gate item for schema#248, after the fixed-class language items (#260 #255 #253 #257); before space#441 closes.

gafferongames:
Landed: #266 merged at 69bc4ec — the C# tables backend, fixed class, golden-locked to C++ (18 shared goldens; 12,497 hostile buffers zero divergence; `?T` and `[E]T` seams filled; slot-0 guard survives release). The C# variable class and text form are named follow-ons in §15; the bench leg is #270.

PR #266, opened 2026-09-02
C# tables backend: the fixed class, golden-locked to C++ (#262)

gafferongames:
The C# TABLE-wire backend, FIXED class, golden-locked to the C++ reference. Closes the language item schema#262 asks for and unblocks the Unity half of the dogfood (space#441): Config.bin, Options and the per-frame render tables can now be decoded in C# from the same bytes the C++ side writes.

`internal/codegen/cpptable` is the reference and this port mirrors it decision for decision — the framing, the elision rules, the clamps, the storage-invariant refusals and every report event. It invents no contract; where C# forces a different spelling the reason is at the site.

## What it emits

`<Base>Table.cs` beside the packet `<Base>.cs`, for every file of a unit that declares tables, and nothing at all for a unit that declares none:

- **storage classes** for the `table` declarations,
- **`<Name>Measure(value)` / `<Name>Save(value, Span<byte>)` / `<Name>Load(value, ReadOnlySpan<byte>, TableReport)`** — name-first on the unit's `Schema` partial class (C# has no namespace-level functions), plus the `SaveBody`/`LoadBody` pair the closure recurses through,
- **`TableReport`**, **`TableWriter`/`TableReader`** (ref structs over the caller's span), and the f32/f64 bit-pattern pair — once per unit, in the protocol-id home file,
- **`TableEnumId` / `TableEnumValue`** per enum in the closure, emitted by the file that declares it,
- **reflection descriptors**: `<Name>TableType()` returning field name, wire id, kind, `IsArray`/`Counted`, array bound, declared range, branch guard, the nested table's descriptor, and the enum/union vocabulary with the wire id each variant rides under.

A unit whose closure declares a pointer is **refused by name**: *"the C# table backend's variable class is a named follow-on"*. The arena, the region, the cooked form and the pointer surface are §15 work.

`?T` optional fields and `[E]T` enum-keyed arrays are **implemented**, mirroring the C++ reference on main (#265) — see "The seams" below.

**Rebased onto `fd92d5e`** (current main, past #280's baseline fix) and force-pushed. Nothing but the rebase changed: the C# emitter, the name registry, the C# test leg, the csproj, `compiler/builtin.go`, the checker changes, `test/tables/main.cpp`, the Makefile, `USAGE.md` and **all 18 goldens are byte-identical** to the reviewed `8b7b4aa`. The only delta is main's own #279 addition to `SPEC-TABLES.md` §16 arriving. The C++-side conflicts from the earlier rebase were resolved by taking main's — the JSON walker's descriptor columns and main's `TableReport`, which carries `duplicate`, so C#'s `Duplicate` mirrors a reference instead of running ahead of one.

- **`SPEC-TABLES.md` "Backend status"** — now *"C++, and C# for the FIXED class"*, with the C# pointer refusal named.
- **`SPEC-TABLES.md` §11** — a VARIABLE-LENGTH table under the C# backend, refused by name.
- **`SPEC-TABLES.md` §15** — "beyond C++ and C#", and the C# variable class and text form as the named follow-ons.
- **`USAGE.md`** — the C# save/load example beside the C++ one, including what C# cannot express about the keyed accessor.

Nothing else in the three files moved.

## The seams (`?T`, `[E]T`, and the report's fifth counter)

**`?T` — presence decides, never content.** A `<Name>Present` bool beside the value; a present optional writes its body **even when entirely default**, an absent one writes nothing, and load sets presence from the fact the field rode. Descriptors mark it `Optional`.

The three-way `T` / `?T` / `*T` set is pinned as goldens: **byte-identical over non-default content**, and the **empty end's asymmetry** pinned too (`chain_value_empty` is 2 bytes; `chain_optional_empty` and `chain_pointer_empty` carry a body). C# refuses P2 (pointered), so the leg reads the pointer's **C++-written** bytes back through both spellings it does carry — `*T` bytes decode as `T` and as a present `?T`, and an elided `T` reads as **absent** through `?T`, which is the right answer in each direction.

**`[E]T` — slots by name.** `E.Max + 1` slots in a generated `TableKeyed<T>`; wire **kind 16**, body opening `(elem kind u8, N u32)` then `(variant id u16, L u32, element)` pairs in ascending ordinal, per-slot elision, no `None` key on the wire, a stored key of `0` **stopping the body** as framing damage (not counted unknown — `0` names nothing), an unknown key skipped by its length and counted. `Skip` gains 16.

Kind 16 is what makes the keyed↔positional edit a *reported* one, and the corpus proves it: `ledger` is `[Grade.Max + 1]int32` in V1 and `[Grade]int32` in V2, and each direction reads **one kind mismatch** and leaves the array at its declared defaults. The V1/V2 middle-insert/removal corpus is pinned with **identical report counts on both sides** (forward: 2 unknown, 2 kind mismatch; reverse the same) — `Beta` slides from ordinal 2 to 3 and still lands in its own slot, and the test asserts it did **not** land in the slot its old ordinal names.

Descriptors carry `KeyTypeName` / `KeyName` / `KeyId`, with **`KeyId(0) == 0`** and **`KeyName(0) == "None"`** marking slot 0 invalid — §8's mechanism as #273 states it, so a walker stepping `[0, ArrayBound)` skips it with no rule about slot indices.

**What C# cannot express**, stated at the site, in `USAGE.md`, and here:

- **No compile-time key.** C# generics take no non-type parameter and the language has no `constexpr`, so `Slot<ShipType::Bomber>()` has **no C# spelling** and the compile-time refusal is simply unavailable.
- **No non-boxing generic enum→int.** Measured, not assumed: `(int)(object)key` in a `where TKey : struct, Enum` generic **throws `InvalidCastException`** for any enum whose backing is not exactly `int` (it is an unbox, not a conversion), and `Convert.ToInt32` boxes on every index — per-frame garbage in the consumer this port exists for.

So the indexer takes the **slot index** and the caller writes the cast: `fleet.Ships[(int)ShipType.Bomber]`. The `None` refusal survives as the runtime guard — and unlike C++'s `assert` it is **not** compiled out in release, so the C# guard is the stronger of the two at runtime and the weaker at compile time. Generated codecs walk `.Slots` directly and never pay for it.

**`TableReport.Duplicate`.** The page (§4) says every backend carries the counter and that a wire read always leaves it zero. C# carries it; the leg asserts it stays zero. **C++ on main does not have it yet** — the JSON walk (§16) that raises it is unported — so this is the one place C# follows the page ahead of the reference, deliberately.

## A red on main, found here and fixed on main

While rebasing onto `0762afb` this branch hit a failure it did not cause: `TestCorpusBaselinesAreCurrent/examples` was red on a clean checkout of main, because #267 added `tables/examples/Guarded.schema` (declaring `table Patrol`) without regenerating the committed baseline #263 had just introduced. Bisected to `0762afb` (`526239c` green), reported rather than absorbed — regenerating a committed projection inside an unrelated PR is exactly what #263 exists to prevent. Fixed on main by **#280**, with a reason, where it belonged. Nothing about it remains in this PR.

## A hole in main, closed here

`table TableKeyed` **passes `check` on main today and emits a C++ header that does not compile** (`error: redefinition of 'TableKeyed' as a different kind of symbol`) — the exact §11 class the name registry exists to close, introduced with #265 and reproduced here. `TableKeyed` is now registered in `internal/tablenames`, which claims it for **both** backends. The registry's scan reaches it through its own corpus (`runtimeSrc`), kept separate from `tableSrc` because a keyed array legitimately emits a C++ class **template** and the zero-cost gate greps for one — that separation is itself a finding: widening the shared corpus broke an unrelated gate on the first try.

## Decisions, labelled as mine

**Storage is a sealed class with public fields, not a struct** — `string(N)`/`bytes(N)` as a pre-allocated `byte[N]` beside an `int` length, arrays as a pre-allocated `T[N]` beside an `int` count, a union as its tag beside one pre-allocated arm. This is not a free choice: a table's closure contains plain `type` declarations whose C# classes **the packet emitter already wrote**, and the table codecs decode into those very classes. One unit, one spelling. (It also keeps the read path allocation-free — a C# struct cannot hold an inline array, so a struct form would have to reallocate or hand-manage every buffer, and it would still not be blittable, so §9's relocatability is unreachable in C# either way.)

**The reset is verb-first `TableReset(X value)`, overloaded on the value type — not `<X>Reset`.** SPEC-TABLES.md §11 freezes the 23 name-first suffixes a closure member claims; a port must not quietly mint a 24th. `TableReset` joins `TableEnumId`/`TableEnumValue` in the verb-first family, which claims nothing from a declaration's name.

**Four C++ descriptor columns are dropped**: `TableFieldInfo`'s `offset`, `elem_size` and `count_offset`, and `TableTypeInfo`'s `size`. All four are memory facts with no C# meaning — a C# field has no `offsetof` and a C# class has no meaningful `sizeof`, and a walker reaches storage through the language's own reflection. Every other column is there name for name.

**The cross-file DAG rule (§11) moved to the front end.** It first lived in `cpptable` and fired only for `--lang cpp`, which made a unit legal under one target and illegal under another — the trap the rule exists to prevent. It now lives in `internal/check`, so every target refuses the same units, and it reports at the declaration that closes the cycle. The backend copy is gone.

## The union form, measured

C# has no native union, so the owner licensed two candidates (#248). Measured on a **representative shape — the space per-frame render table**: a 256-record frame, each record carrying a union of three blittable arms, decoded from the wire. .NET 10, Release, arm64 (M-series), median of 7 runs × 20 000 frames, two sittings:

| form | ns/frame | alloc/frame | construction footprint per frame |
|---|---|---|---|
| **A** — all arms inline, `[StructLayout(LayoutKind.Explicit)]` overlap | 1056 | 0 B | 16 440 B |
| **B** — tag + one arm per variant, pre-allocated (**shipped**) | 1051 / 1068 | 0 B | 47 160 B |
| **C** — one reference per arm, allocated **on read** (the literal option b) | 1561 / 1575 | **7 504 B** | 47 160 B |

**B/A = 0.996 and 1.011 across the two sittings — indistinguishable.** C is **1.48–1.49× slower and allocates 7.5 KB per frame**, which is 60 Hz garbage in the Unity client, so it is out.

A and B tie on the read path. A's only win is a 2.9× smaller resident footprint, and it is a **one-time construction** cost, not a per-frame one. Against that, A costs: explicit layout is only legal while every arm stays blittable, so the day an arm gains a string, a nested table or an array the form has to change (to sum-of-arms, losing the footprint win anyway); it cannot decode into the packet emitter's existing arm classes; and it diverges from the C# union spelling the estate already ships for packets.

**Shipped: B.** Zero allocation on the read path, ties A on speed, and it is the spelling the C# packet backend already uses — so a union field means one thing in a C# unit, whether the value came off the table wire or the packet wire.

## Repairs from the cold review ([#266 review](https://github.com/mas-bandwidth/schema/pull/266#issuecomment-5503569791))

The review's five conditions, all addressed on this branch.

**F1 — a legal schema emitted C# that did not compile.** `internal/check` claimed the C++ table runtime's names and **none** of the C# ones, so `enum TableReset { A, B }` beside any table passed `check` and then failed `dotnet build` with **26 CS0119** (`'Schema.TableReset(Foo)' is a method, which is not valid in the given context` — the enum lands in expression position). Reproduced, then fixed: the claim list is **front-end law, not one target's inventory**, and every backend's runtime spelling is now claimed for all of them. That also closes the pre-existing `TableEnumId`/`TableEnumValue` hole C++ has carried since #254.

The gate that keeps the list honest **does not hardcode the names**, and after the delta pass it does not read declaration syntax either. `internal/tablenames` is the one registry — the checker claims from it, so the second copy that let the C# names go unclaimed no longer exists — and `TestTableRuntimeNamesAreClaimed` holds it honest against the emitted output in **both directions**: an emitted `Table*` name nobody registered fails, and a registered name nothing emits any more fails too, so the list cannot rot into claims nothing needs. Every registered name is then asserted refused when a unit declares a table, and kept when it does not.

The scan recognises **no declaration shape at all**: it strips line comments, collects every `Table*`-prefixed identifier that remains, and requires the whole set to be registered. Over-collection is the safe direction — a false hit costs a registration or a reworded comment, a miss costs a legal schema that does not compile. The registry records which backends' text carries each name, and marks the two **scoped** spellings (a field descriptor's `Table` column, and `TableRef`'s unrelated C# meaning) which claim nothing because they are reached through their owner.

Controls, all red: the delta pass's five missed shapes (`public enum`, non-ref `public struct`, `public static readonly`, a generic method with `<` before `(`, non-sealed `public class`), the two the old regexes did catch, both rot directions, and — with the claim removed from the checker — all 24 subtests.

**F2 — `--lang cs` accepted a unit `--lang cpp` refused.** The cross-file DAG rule moved from `internal/codegen/cpptable` to `internal/check`, the backend copy removed. All four now refuse identically and at a source line:

```
$ schema check A.schema B.schema        # and --lang cpp, --lang cs, --lang go
B.schema:3:1: Beta closes a cross-file table reference cycle: A -> B -> A — a unit's
table files form a DAG by reference, so if a declaration in one file reaches one in
another, nothing there may reach back; move a declaration so the cross-file graph is
acyclic (SPEC-TABLES.md §11)
```

Fixtures in `internal/check/tables_test.go`: the two-file cycle, a cycle closed through a **union arm**, and three controls that must stay legal — an acyclic cross-file graph, same-file recursion through a pointer, and a table-free unit (which is not graphed at all). Negative-controlled by commenting out the call: both refusal cases go red, all three controls stay green.

**F3 — the labelled decision undercounted.** Four columns, not three: `TableTypeInfo.size` too. Corrected above and in the generated runtime's own comment.

**F4 — the union numbers are now reproducible.** The harness lands at `bench/tools/cs-union-form` (`dotnet run -c Release`). It is a **decision probe, not a bench-standard leg** — no CSV row, not in `run.sh`, not in `make test`, measures no `bench/corpus/` shape — and its README says so, including that nothing in it may be divided against a bench CSV row. It reproduces from its committed location: B/A **0.996**, C/A **1.490**. `make shape-gate` stays clean with **no new ledger entry** (the gate caught one corpus identifier in a comment on the first run and it was reworded rather than exempted).

The bench leg #262 actually names as a gate item — *a fixed table beside its equivalent type on the ledger*, the codec half the zero-cost gate does not hold — is **not** built here and does not evaporate: filed as **#270** with the owner's doctrine quoted, and named from both the probe's README and `bench/README.md` so neither can be mistaken for it.

**F5 — the stale `// C++-only today` comment** above the claimed-suffix list, sitting exactly where F1 needed edited.

**F8 (a note, not a condition, but two lines):** `tables-cs-standalone` asserted a file count before its loop, so the gate can no longer pass vacuously on a glob that matches nothing.

F6 (bare `dotnet` off PATH) and F7 (`-0.0` eliding against a `0.0` default, byte-identical in both backends) are inherited estate items, not introduced here, and are left as the review filed them.

## Gates

**Shared golden wire — the load-bearing one.** C++ is the reference writer: `test/tables/main.cpp` pins eight instances' encodings into `testdata/wire/tables/*.bin` (`make update-goldens` re-pins deliberately), and `test/cs-tables` builds the same instances value for value, byte-compares its own `Save`, then **loads those very C++-written files**. The set covers every wire kind, both array shapes, enum arrays, unions, flags, guards, cross-file nesting, all-default elision, and extents past 65535 (u32 lengths and counts).

Beside it, mirroring the C++ test block for block: exact capacity (measure == save, one byte short refuses), storage-invariant refusals, bounded elements (a count the body cannot cover keeps the prefix and never fabricates from the next field's bytes), all-default, guards, **the evolution corpus in both directions with the report counts**, mid-vocabulary enum and union inserts both ways, the repeated-id unnameable-variant case, changed array bounds both ways, unnameable enum values refused on write and read, flags staying positional, wide extents, clamping, framing damage, reflection, cross-file nesting.

Two more, in the Makefile:
- **`tables-cs-standalone`** — the C# twin of the C++ "compiled without `-I$(SERIALIZE)`" build: no generated `<Base>Table.cs` may name the serialize runtime.
- **`tables-cs-refuses-pointers`** — `tables/pointers` under `--lang cs` must fail, naming the follow-on.

Everything rides `make test` and therefore CI unchanged; the compiler-side contracts are pinned in `compiler/tables_test.go` (`TestCsEmitsTableSources`, `TestCsRefusesPointeredTables`, and the seven remaining backends still refusing by name).

## Negative controls for the seams

Three more sabotages, each reverted, each turning the **shared goldens** red:

1. **`?T` made content-decided** (a present all-default optional elides). Red on `keyed_config` (225 written vs 234 pinned), `chain_optional_empty` (2 vs 11), and the asymmetry assertion against the pointer's bytes.
2. **The keyed body written positionally** — kind 14, slot index in place of the variant id: the spec's own named control. Red on `keyed_config`, `v1_seams` and `v2_seams` (**same length, different bytes** — a length check alone would have missed it), on both evolution report counts, and squarely on *"Beta rode by NAME — a positional encoding would have put it in Omega's slot"*.
3. **Key 0 treated as unknown instead of malformed.** Red on both halves of the rule: the body no longer stops, and the event is miscounted.

## Negative control (the original two)

Two sabotages of the emitter, each reverted:

1. **Nested-table elision flipped** (`body > 2` → `body >= 2`, so an all-default nesting always rides). First run: **green** — the golden set had no instance with an all-default nested table, so the gate could not see it. That is the control doing its job: a `profile_elide` instance was added to both sides, and the sabotage then went **red** (`profile_elide — 55 written, 46 pinned`) while every self-consistent round-trip check in the C# test stayed green. Exactly the class of bug a single-language test cannot catch.
2. **The f32 codec** (`TableFloatToBits` xor 1). Red on `root_full`, `loadout_full`, `archive`, `v1_cfg`, `v2_cfg` and on four round-trip float assertions — and correctly still green on `root_default`, `profile_elide` and `wide_blob`, which carry no f32.

## Extra verification

An adversarial unit outside the committed corpus (a unit whose first file declares no table at all, so the runtime lands in a member-less home file; an empty table; nested branch guards with an `else`; `int64` min and `uint64` max defaults; flags, enum arrays, `bytes` and `string` inside a closure `type`; a fixed array of tables containing a string) was generated for both backends, populated identically, and saved: **byte-identical, populated and all-default.**

## Results

18 shared goldens (8 new for the seams), **regenerated against main's C++ after the rebase and byte-identical** — the JSON walk moved no wire byte, as expected · `go test ./...`, `go vet ./...`, `gofmt`, `golangci-lint` and `modernize` clean · `schema_test`, `schema_test_guard`, `schema_test_tables` pass · `tables-zero-cost`, `tables-cs-standalone`, `tables-cs-refuses-pointers`, `shape-gate` pass · `test/cs-tables` passes (`TreatWarningsAsErrors`) · `make check`, `make fmt` clean. The full nine-language `make test` needs the pinned Dart/JDK/BEAM toolchains this scratch clone does not carry — CI runs it.


gafferongames:
## Cold adversarial review — verdict: **MERGE-WITH-CONDITIONS**

Held behind the spec-of-record PR by owner ruling regardless (#248: *"a SPEC-ONLY PR is now in flight… It gets a cold read and LANDS FIRST. Every implementation PR (#263, and the coming #260/#255, #253, #262) is HELD from merge until it lands, then rebases onto it"*). Nothing below changes that; the conditions are what to fix before it rides.

I had no part in writing this. Fresh clone, `gh pr checkout 266`, branch never edited — every sabotage below ran against a **copy** of the generated tree in scratch.

The port is good. The behavioural core is the strongest thing in the PR and I could not break it: I fed **6 459 hostile buffers** to the C++ reference and to the C# backend and compared load return, all four report counters, `Measure`, `Save`, and the **full re-encoded bytes** of the loaded value. **Zero divergence.** The read path allocates **0 B**, everywhere I could point a probe. Both of the PR's own negative controls reproduce verbatim. The findings below are all at the edges — but two of them are real, and one of them lets a legal schema emit C# that does not compile.

---

# FINDINGS

## F1 — MUST FIX. The C# table runtime's seven unit-level names are unclaimed, so a legal schema generates C# that does not compile

`internal/check/check.go:2168` claims the table runtime's namespace-level names when a unit declares a table. It claims the **C++** spelling of the float helpers:

```go
for _, gen := range []string{"TableReport", "TableWriter", "TableReader",
    "TableTypeInfo", "TableFieldInfo", "table_bits_to_float",
    "table_float_to_bits", "table_bits_to_double", "table_double_to_bits",
    ...
```

The C# emitter puts **seven** verb-first names on `Schema`, and the checker claims **none** of them:

```
$ grep -on "static [a-z]* Table[A-Za-z]*\b" internal/codegen/cstable/*.go | sed 's/.*static //' | sort -u
bool TableEnumId
bool TableEnumValue
double TableBitsToDouble
float TableBitsToFloat
uint TableFloatToBits
ulong TableDoubleToBits
void TableReset
```

Repro (`enum TableReset` — a name this PR newly makes dangerous):

```
package p

enum TableReset { A, B }

table Foo
{
    g TableReset
    v float32 = 1.5
}
```

```
$ schema generate --lang cs --out ocs .      # exits 0, emits X.cs + XTable.cs
$ dotnet build
ocs/X.cs(42,29): error CS0119: 'Schema.TableReset(Foo)' is a method, which is not valid in the given context
ocs/X.cs(44,29): error CS0119: ...
ocs/X.cs(46,29): error CS0119: ...
ocs/XTable.cs(261,22): error CS0119: ...
ocs/XTable.cs(262,22): error CS0119: ...
ocs/XTable.cs(263,22): error CS0119: ...
```

`--lang cpp` on the same unit compiles clean (`c++ -fsyntax-only` → 0). The break is expression position: `TableReset.A` resolves to the method group, not the type. A `table TableReset` survives (type position only) — it is **enums** that break.

Same failure, same shape, for all four float helpers:

```
enum TableFloatToBits  → error CS0119: 'Schema.TableFloatToBits(float)' is a method, ...
enum TableBitsToFloat  → error CS0119: 'Schema.TableBitsToFloat(uint)' is a method, ...
enum TableDoubleToBits → error CS0119: 'Schema.TableDoubleToBits(double)' is a method, ...
enum TableBitsToDouble → error CS0119: 'Schema.TableBitsToDouble(ulong)' is a method, ...
```

And `TableEnumId` / `TableEnumValue` break **both** targets — that pair is pre-existing from #254, not this PR's (C++: `XTable.h:181: error: must use 'enum' tag to refer to type 'TableEnumId' in this scope`), but the same one-line fix closes it.

This is the exact invariant §11 exists to hold: *"the rest are refused here, so **no legal schema can reach a non-compiling header through this door**."* Today it does.

**The verb-first decision itself is correct and I confirmed it is load-bearing** — `table Foo` beside `table FooReset` is accepted today by both targets, so minting `<X>Reset` as a 24th name-first suffix would have been a silent collision. The fix is not to reverse the decision. It is to add the seven C# names to the `len(c.tables) > 0` block at `check.go:2168`, beside the snake_case twins already there.

The PR body's own sentence is the bug, stated: *"`TableReset` joins `TableEnumId`/`TableEnumValue` in the verb-first family, **which claims nothing from a declaration's name**."* It must claim — at unit level, the way `TableReport` already does.

## F2 — MUST FIX (code or page). `--lang cs` accepts a unit `--lang cpp` refuses

§11 states the cross-file table DAG rule as **language-neutral**: *"a unit's table files form a DAG by reference… a port whose modules import by file inherits the same acyclicity, **and the rule is stated so it does not have to rediscover the reason**."* The check lives in a backend: `internal/codegen/cpptable/cpptable.go:492`.

Two-file cycle (`A.schema` declares `AlphaHolder { b Beta }`, `B.schema` declares `Beta { a Alpha }`):

```
=== CYCLE, cpp ===
schema: table include cycle: A -> B -> A — the generated Table headers would include
each other; move a declaration so the cross-file reference graph is acyclic (SPEC-TABLES.md)
exit=1
=== CYCLE, cs ===
exit=0     A.cs  ATable.cs  B.cs  BTable.cs
```

And the C# output is fully correct — I compiled it and round-tripped it **both** directions:

```
measure=34 save=34
load=True v=7 w=42 unknown=0 kind=0 clamped=0 malformed=False
beta: v=11 w=99
```

So the PR's technical argument holds: C# does not import by file, and the emitter genuinely needs no acyclicity. **But the consequence is a unit that is legal under one target and illegal under another** — in a repo whose whole proposition is one declaration across nine languages, that is a trap, and someone will author against C# and discover it at `--lang cpp`.

Pick one, in this PR: enforce it in `internal/check` for every target (my recommendation — §11 already states it as neutral, so this is the cheaper reconciliation), or amend §11's DAG paragraph to say the rule binds only ports whose modules import by file. Present-state doctrine says page and code agree; right now they do not.

## F3 — accuracy. The labelled decision "three descriptor columns dropped" undercounts

Four are dropped. `offset`, `elem_size`, `count_offset` on `TableFieldInfo` — and **`size`** on `TableTypeInfo` (`uint32_t size; // sizeof the storage struct`), which has no C# twin either:

```
=== C++ TableTypeInfo ===            === C# TableTypeInfo ===
const char * name;                   public string Name;
uint32_t size;          <-- dropped  (absent)
int32_t num_fields;                  public int NumFields;
const TableFieldInfo * fields;       public TableFieldInfo[] Fields;
```

The judgment is right — all four are memory facts. But *"Every other column is there name for name"* is not true, and a decision labelled as the author's should be exact.

**Non-blocking, for #253's C# leg when it comes:** dropping offsets is correct, but the C# descriptor then gives a reflection-driven walker only the **schema** field name (`Name = "version_note"`) while the storage member is `VersionNote` with `VersionNoteLength` beside it. The C++ walker gets a direct offset and needs no convention. A C# walker must re-derive the C# spelling from a mapping that lives in the emitter and is nowhere in the descriptor. #253's negative control is literally *"sabotage the walk's offset arithmetic"* — the C# equivalent needs something to sabotage. A `MemberName` column, or a generated accessor, is what closes it. Note for #253, not a condition here.

## F4 — the union measurement is unreproducible from the tree, and the bench leg #262 asked for is absent

The four-decision section rests on a table of numbers (`B/A = 0.996 and 1.011`, `C` allocating `7 504 B/frame`). No harness is committed:

```
$ git diff main...HEAD --stat -- bench/
(empty)
$ grep -n -i "table" bench/BENCH-STANDARD.md
(nothing)
```

There is no tables bench leg anywhere in `bench/` — not for C# and not for C++. #262 names one as a gate item (*"a bench leg beside the type"*), and #248's doctrine comment is explicit: *"Zero-cost gate already holds the storage half; **the bench leg holds the codec half**."*

What I **could** verify independently, and did:

- **Form B allocates nothing on the read path.** Measured, not asserted — see the allocation table below.
- **Form B is genuinely the C# packet backend's union spelling.** `Effect` (tag + pre-allocated `Buff` + pre-allocated `Debuff`) is emitted by the **packet** emitter into `Tables.cs`; the table codecs decode into it. So the "one unit, one spelling" claim is true, not aspirational.
- **Form C obviously allocates** and is out on the doctrine alone (60 Hz garbage in Unity).

What I cannot reproduce or refute: the A-vs-B ratio and C's exact figure. The **decision** survives without them — it rests on packet-emitter consistency and on A's explicit-layout blittable-only constraint, both of which I verified — but the table should either arrive with its harness or be restated as an uncommitted sitting. Either close the bench-leg gate item here or defer it explicitly on #262 so it does not quietly evaporate.

## F5 — stale comment, in the exact place F1 needs edited

`internal/check/check.go:2160`:

```go
// C++-only today, so only the CamelCase spellings claim.
```

False on this branch. The PR did the present-state pass on SPEC-TABLES.md, USAGE.md and `compiler/builtin.go`'s comment and missed this one — which sits directly above the list F1 has to extend.

## F6 — non-blocking, pre-existing. dotnet is the one unpinned toolchain

Dart/JDK/BEAM/Rust go through `$(DART)`/`$(JAVA)`/`$(MIX)`/`$(RUSTUP_BIN)` against `dist/`. All three C# legs — including the new `cd test/cs-tables && dotnet run` — call bare `dotnet` off PATH, and CI's `setup-dotnet` pins `8.0` while every csproj (`test/cs`, `test/cs-ludicrous`, and the new `test/cs-tables`) targets `net10.0`, so the leg runs on whatever SDK the runner happens to carry. **The new leg follows the existing pattern exactly, which is the right call for a port** — flagging it as an estate item, not a condition on this PR.

## F7 — non-blocking observation, shared with C++

`-0.0` does not survive the table wire when the field's declared default is `0.0`: it compares equal, elides, and loads back `+0.0`.

```
ProfileConfig  precision = 0x8000000000000000  ->  resave = 0000   (2 bytes: elided)
WeaponConfig   damage    = 0x80000000          ->  resave = a9150a 00000080 0000  (kept: default is 21.0)
```

Byte-identical in C++ and C#, so this is a property of §3's elision rule, not a port defect. Worth a clause in §3 if anyone cares.

## F8 — non-blocking. `tables-cs-standalone` can pass vacuously

```make
@for f in build/tables-generated-cs/*/*Table.cs; do \
    if grep -n "Serialize" $$f; then ... exit 1; fi; \
done
```

If the glob ever matches nothing — a renamed stamp target, a moved `--out` — the loop runs zero times and the gate prints success. It works today and I negative-controlled it (below); a file-count assertion first would make it honest.

---

# VERIFIED GOOD

## 1. Behavioural equivalence with the C++ reference — 6 459 hostile buffers, zero divergence

Differential harness: same bytes into both backends; compare `load`, `unknown`, `kind_mismatch`, `clamped`, `malformed`, `measure`, `save`, **and the full hex of the re-encoded loaded value** (a stronger oracle than the report counts — any value difference shows up in the resave).

Corpus: directed cases for every §4 event on `RootConfig`/`WeaponConfig`/`LoadoutConfig`/`ProfileConfig` — unknown id across all 11 skippable kinds incl. kind 0 and kind 99, kind mismatch (incl. i32↔i16 and container↔scalar), clamp high/low/both, `bits(6)` overflow, repeated id last-wins (incl. last-is-mismatch and both union orders), unnameable enum → `None`, unnameable enum **inside an array**, unnameable union arm → empty, explicit arm id 0, arm body malformed, arm length past end, guard off with the guarded field present, guard **after** the guarded field, array count past the body, count past the reader's bound, count short of it, wrong element kind, nesting depth 8/64/256/1000/2000 through both unknown-table and unknown-union skippers, truncation, missing terminator, trailing garbage, `0xFFFFFFFF` lengths — **plus 6 000 deterministic mutations** of four seeds (byte substitution, bit flip, truncation, extreme-byte, 4-byte `0xFF` splice) **plus 378 extreme-value instances** (int8/16/64 min and max, uint8/16/32/64 max, 12 f64 bit patterns incl. both NaN signs, both infinities, both zeros, subnormals, max/min normal; 12 f32 patterns through a fixed array and through a defaulted scalar; empty/one-byte/at-bound/bound+1/non-UTF-8/embedded-NUL strings; empty and at-bound bytes; every counted and fixed array at its exact bound; all-default nested table both elided and explicitly present; union `None` and each arm both defaulted and valued; all 64 flags bits individually, all set, none set; all 256 `bits(6)` values).

```
$ diff <(sort cpp.out) <(sort cs.out) ; echo $?
0
$ diff <(sort cpp2.out) <(sort cs2.out) ; echo $?
0
```

Event coverage actually exercised (from the C++ side; C# identical line for line):

```
unknown=0 4543 | unknown=1 1531 | unknown=2 4 | unknown=3 3
kind=0    5481 | kind=1     600
clamped=0 4829 | clamped=1  765 | clamped=2 436 | clamped=3 51
malformed=0 2607 | malformed=1 3474
load=0 2195 | load=1 3886
```

**Oracle negative-controlled twice**, on a copied generated tree:

```
SABOTAGE A: penetration clamp bound 10 -> 11             divergent lines: 116
SABOTAGE B: unnameable-union-arm `r.Report.Unknown++` removed   divergent lines: 236
```

So the harness reaches both the scalar clamp path and the union arm path, and reported zero because there is nothing there.

The subtle one is right: a repeated union field id whose **last** occurrence is unnameable resets the arm explicitly rather than leaving an earlier decoded arm standing (both orders tested, both byte-identical to C++).

## 2. Allocation — 0 B on every codec path, probe negative-controlled

`GC.GetAllocatedBytesForCurrentThread` around 1 000 iterations over warm instances with a reused `TableReport`:

```
RootConfig.Load (root_full, warm instance)     total=  0 B   per-op= 0.000 B
RootConfig.Measure                             total=  0 B   per-op= 0.000 B
RootConfig.Save                                total=  0 B   per-op= 0.000 B
RootConfig.Load (root_default)                 total=  0 B   per-op= 0.000 B
LoadoutConfig.Load (loadout_full)              total=  0 B   per-op= 0.000 B   <- union + enum arrays + flags
LoadoutConfig.Measure / .Save                  total=  0 B   per-op= 0.000 B
ProfileConfig.Load (profile_elide)             total=  0 B   per-op= 0.000 B
ProfileConfig.Measure / .Save                  total=  0 B   per-op= 0.000 B
ArchiveConfig.Load (archive, cross-file)       total=  0 B   per-op= 0.000 B
ArchiveConfig.Measure / .Save                  total=  0 B   per-op= 0.000 B
WideBlob.Load (wide_blob, 210 KB)              total=  0 B   per-op= 0.000 B
WideBlob.Measure / .Save                       total=  0 B   per-op= 0.000 B
WeaponConfig.Load (malformed)                  total=  0 B   per-op= 0.000 B   <- the error path too
Schema.RootConfigTableType()                   total=  0 B   per-op= 0.000 B   <- descriptor is cached

NEGATIVE CONTROL: new byte[64] per op          total=88000 B  per-op=88.000 B
NEGATIVE CONTROL: box an int per op            total=24000 B  per-op=24.000 B
```

The doctrine line holds: a fixed table with a union allocates nothing in C#, so the language never needed the carve-out.

*(Micro-note, no action: the descriptor's lazy init `if (info != null) return info;` is unsynchronised. Benign — idempotent, reference store is atomic, CLR does not reorder stores — but C++'s function-local static is thread-safe by language and C#'s is thread-safe by argument.)*

## 3. Goldens reproduce, and the C# leg really consumes the C++-written files

Eight instances. `SCHEMA_UPDATE_WIRE_GOLDENS=1 ./build/schema_test_tables` regenerates them byte-identically:

```
$ diff /tmp/gold_before.txt /tmp/gold_after.txt && echo "GOLDENS REGENERATE BYTE-IDENTICAL"
GOLDENS REGENERATE BYTE-IDENTICAL
$ git status --porcelain testdata/
(clean)
```

The C# leg's `GoldenWire` byte-compares its own `Save` against `testdata/wire/tables/<name>.bin` and then loads those very files back — verified at `test/cs-tables/src/Program.cs:82` and the eight call sites.

The extreme-value classes named as "goldens the set lacks" I covered through the differential path above instead of by adding pins, which exercises write **and** read: all-default nested table (elided and explicit), empty/max-length strings, arrays at bound, union `None` and each arm, flags all bits, int8/int64 min, uint64 max, NaN/inf/-0.0 bit patterns. All byte-identical. If you want them as standing pins rather than as this review's transcript, that is a corpus addition, not a defect.

## 4. Both of the PR's negative controls reproduce — exactly

**Control 2 (f32 codec `^ 1`).** Note the helper is emitted **once per unit**, so all three units' copies must be sabotaged; sabotaging only the `examples` unit gives 3 failures, not 5. With all three:

```
FAILED: table wire golden root_full    — C# bytes must equal the C++-pinned bytes (398 written, 398 pinned)
FAILED: table wire golden loadout_full — ... (152 written, 152 pinned)
FAILED: table wire golden archive      — ... (414 written, 414 pinned)
FAILED: table wire golden v1_cfg       — ... (181 written, 181 pinned)
FAILED: table wire golden v2_cfg       — ... (184 written, 184 pinned)
FAILED: old reader: nested factor
FAILED: new reader: nested factor
FAILED: union insert: arm payload
FAILED: union insert moved tag: lands correctly
cs tables test: FAILED
```

Five goldens plus four round-trip float assertions, and correctly green on `root_default`, `profile_elide`, `wide_blob`. As claimed.

**Control 1 (nested-table elision `body > 2` → `body >= 2`).** Verbatim:

```
FAILED: table wire golden profile_elide — C# bytes must equal the C++-pinned bytes (55 written, 46 pinned)
cs tables test: FAILED
```

And I reproduced **phase A**, which is the part that makes the control worth anything — same sabotage, with only the `profile_elide` write golden removed:

```
cs tables test passed
```

The entire 1 216-line conformance leg goes green on a broken elision rule. That is an honest control and it did real work; the `profile_elide` instance earns its place.

## 5. §11's 23 suffixes are respected; the verb-first choice is load-bearing

The C# emitter's name-first spellings are exactly `Measure`, `SaveBody`, `Save`, `LoadBody`, `Load`, `TableInfo`, `TableType` — all seven inside the claimed 23. No 24th minted.

And the choice is not cosmetic:

```
foo_and_fooMeasure    cpp: REFUSED   cs: REFUSED
foo_and_fooReset      cpp: OK        cs: OK        <- an `<X>Reset` suffix would have collided silently
table_named_TableReport / TableWriter / TableReader / TableFieldInfo / TableTypeInfo / Schema
                      cpp: REFUSED   cs: REFUSED
```

Correct reasoning, correct outcome — the gap is F1: the verb-first names it moved to instead were never claimed.

## 6. Deviation (a) is exactly what the PR says it is

```
Attachment / Buff / Debuff / Effect / Grade   ->  ONLY in the packet Tables.cs
RootConfig / WeaponConfig / Loadout / Profile ->  ONLY in TablesTable.cs
```

One declaration, two wires. No double definition, no layout coupling — and C++ does the identical thing (`struct Attachment` in `Tables.h`, consumed from `TablesTable.h`). The claim that the packet emitter's existing classes forced the storage shape is true, not a rationalisation. Form B being the packet backend's own union spelling is confirmed at `Tables.cs:57` (`EffectType Type` + pre-allocated `Buff` + pre-allocated `Debuff`).

## 7. The seam is honest — `?T` / `[E]T` cannot be mis-generated because they cannot be parsed

Fetched `origin/optional-and-enum-keyed`'s `tables/examples/Keyed.schema` (#265's corpus) and ran it against this branch:

```
$ schema generate --lang cs  --out out .
schema: Keyed.schema does not parse; ... Keyed.schema:31:14: unexpected character "?"
$ schema generate --lang cpp --out outcpp .
schema: Keyed.schema does not parse; ... Keyed.schema:31:14: unexpected character "?"
```

With the optional field deleted, `[E]T` is refused identically by both, at the checker rather than the parser:

```
Keyed.schema:37:14: Weapon is not a constant   (identical for --lang cs and --lang cpp)
```

Both constructs are refused at the front end, before any backend sees them, so no backend can mis-generate either. The seam comment at the top of `internal/codegen/cstable/codecs.go` names the three fold-in sites and they are the right three.

## 8. Gates

Full nine-language `make test` green locally (I symlinked the pinned `dist/` Dart/JDK/BEAM toolchains into the review clone):

```
MAKE TEST EXIT=0
```

Inside it:

```
tables C# standalone gate: generated Table sources name no runtime
tables C# refusal gate: a pointered unit is refused by name
tables test passed
cs tables test passed
```

`tables-cs-standalone` negative-controlled — appending `// Serialize.WriteBits(x);` to a generated Table source:

```
STANDALONE GATE FAILED: the serialize runtime leaked into build/tables-generated-cs/examples/TablesTable.cs
```

`TestTablesMoveNoProtocolId`, `TestCsEmitsTableSources`, `TestCsRefusesPointeredTables` all pass. All seven remaining backends refuse with the updated wording (`— tables are C++ and C# only today, and the <lang> table backend is a named follow-on; generate with --lang cpp or --lang cs …`). `make check` and `make fmt` clean, no working-tree drift. `go build ./...`, `go vet ./...`, `gofmt -l .` clean.

The C# leg is wired into `make test` and therefore CI the same way the packet C# leg is — same `dotnet run` shape, same `TreatWarningsAsErrors` refuser, same sibling-checkout convention, generated sources in `build/` and never in `generated/`. Hermetic apart from F6, which it inherits rather than introduces.

## 9. Provenance

Five commits, every one authored **and** committed by `Rowan Claude <rowan@mas-bandwidth.com>`, every one carrying `Co-Authored-By: Claude Fable 5.1`. No strays.

---

# CONDITIONS

1. **F1** — claim the seven C# table-runtime names in `internal/check/check.go:2168` (and close the pre-existing `TableEnumId`/`TableEnumValue` hole for C++ in the same line). This one is not optional: a legal schema currently emits C# that does not compile, which is the precise thing §11 promises cannot happen.
2. **F2** — reconcile the cross-file DAG rule: enforce it in `internal/check` for every target, or amend §11 to say it binds only file-importing ports. Page and code must agree.
3. **F3** — correct the labelled decision to four columns (`TableTypeInfo.size` too), or restore `size`.
4. **F5** — the stale `// C++-only today` comment at `check.go:2160`.
5. **F4** — either land the bench harness behind the union table, or restate the numbers as an uncommitted sitting and carry the bench-leg gate item forward explicitly on #262 so it does not evaporate.

F6, F7, F8 are notes, not conditions.

The wire behaviour, the allocation profile, the golden discipline and the negative controls are all as advertised, and the three labelled decisions are each defensible on their stated reasoning — F1 is a gap in the machinery around a **correct** decision, not a wrong decision. Fix the five and this is a clean port.

gafferongames:
## Delta pass — verdict: **MERGE**

Fresh clone, `gh pr checkout 266` at `a537234`, branch never edited (`git diff origin/cs-tables HEAD` empty, `git status` clean). I re-verified **only the deltas**, reproducing each myself rather than reading the diff. All five conditions are closed, and F8's hardening rode along.

Still held behind **#269** (the spec-of-record PR — owner ruling on #248: implementation PRs land after it and rebase onto it) and behind the pending `?T`/`[E]T` seam fill (#265). Both remain unreachable today — I re-ran #265's `Keyed.schema` against this head and both targets still refuse identically at the parser (`Keyed.schema:31:14: unexpected character "?"`), so nothing can mis-generate through the seam.

---

## F1 — CLOSED, and closed wider than I asked

All seven names now refused, **both targets**, at the declaration line:

```
enum_TableReset        cpp: REFUSED  cs: REFUSED
enum_TableEnumId       cpp: REFUSED  cs: REFUSED
enum_TableEnumValue    cpp: REFUSED  cs: REFUSED
enum_TableFloatToBits  cpp: REFUSED  cs: REFUSED
enum_TableBitsToFloat  cpp: REFUSED  cs: REFUSED
enum_TableDoubleToBits cpp: REFUSED  cs: REFUSED
enum_TableBitsToDouble cpp: REFUSED  cs: REFUSED

X.schema:3:1: declaration TableReset collides with the generated TABLE-wire runtime
(SPEC-TABLES.md) — both generate the symbol TableReset; rename at the source
```

**C++'s own pre-existing hole is closed too** — `enum TableEnumId` has emitted a non-compiling header since #254 (`XTable.h:181: error: must use 'enum' tag to refer to type 'TableEnumId' in this scope`) and is now refused at check, for every target. Making the list front-end law rather than one target's inventory was the right call and it bought that for free.

Scoping is correct — refused in the other declaration forms when the unit declares a table, and the namespace stays open when it does not:

```
table_named_TableReset            REFUSED     tablefree_enum_TableReset    ACCEPTED
type_named_TableReset             REFUSED     tablefree_enum_TableEnumId   ACCEPTED
flags_named_TableFloatToBits      REFUSED
const_named_TableReset            REFUSED
```

### Trying to defeat `TestTableRuntimeNamesAreClaimed`

It scans 12 names and all 12 pass. I added hypothetical new runtime names to the C# emitter in a **throwaway copy** of the tree (never the branch) and it caught both shapes the emitter actually uses, with an actionable message naming the fix site:

```
(a) public static long TableChecksum(ulong v)   FAIL  <-- caught
(b) public sealed class TableCursorX            FAIL  <-- caught
      tables_test.go:538: a declaration named TableChecksum was accepted — the generated
      table runtime emits that name, so the unit cannot compile; claim it in
      internal/check's unit-level table list
```

That is the drift-catcher working, and it is a real improvement over a hardcoded list.

**One honest limit, non-blocking.** The scan is two regexes over the emitted text — `^\s*public (?:sealed class|ref struct) (Table\w+)` and `^\s*public static \S+ (Table\w+)\(` — so it sees the two idioms in use and not others. Five plausible shapes slip past:

```
(c) public enum TableFlagsX { A, B }                  PASS  <-- MISSED
(d) public struct TableSpanX (non-ref struct)         PASS  <-- MISSED
(e) public static readonly int TableLimitX = 64;      PASS  <-- MISSED
(f) public static int TableCastX<T>(int v)            PASS  <-- MISSED  (generic: `<` before `(`)
(g) public class TableThingX (not sealed)             PASS  <-- MISSED
```

The `len(names) < 12` floor only fires if the scan *loses* names, not if it fails to see a new *shape*. The test's own message — *"the scan, not the claim list, is what broke"* — shows the author already knows the scan is the fragile half, and this does not reopen F1: the seven real names are claimed and independently verified on both targets. Worth hardening whenever a port next touches this, either by widening to `(?:class|struct|enum|interface|delegate)` and `[(<;=]`, or by having the emitter export its runtime-name list as a Go var that both the emitter and the checker read — then an unlisted name is impossible rather than merely likely to be caught.

## F2 — CLOSED. The DAG rule is front-end law and every target inherits it

Two-file cycle, **all nine targets**, identical diagnostic at the closing declaration:

```
cpp     B.schema:3:1: Beta closes a cross-file table reference cycle: A -> B -> A — ...
cs      B.schema:3:1: Beta closes a cross-file table reference cycle: A -> B -> A — ...
c / go / rust / js / dart / java / elixir   — same line, same text
```

**Union-arm cycle** — a ring closed *only* by a union arm reaching back across files (`Pick.a -> LeafA`):

```
cpp  B.schema:14:1: Mid closes a cross-file table reference cycle: A -> B -> A — ...
cs   B.schema:14:1: Mid closes a cross-file table reference cycle: A -> B -> A — ...
```

Negative-controlled — retarget that one arm to a local type and the same shape becomes legal, so the arm walk is load-bearing and not incidental:

```
cpp  ACCEPTED (correct: the only cross-file edge is A -> B)
cs   ACCEPTED
```

The legal shapes all still pass:

```
LEGAL 1  one-directional cross-file (the committed corpus, Nested -> Tables)   cpp ACCEPTED   cs ACCEPTED
LEGAL 2  same-file recursion through pointers (tables/pointers)                cpp ACCEPTED
         cs still refuses for the RIGHT reason: "unit declares variable-length tables (...)"
LEGAL 3  diamond A->B, A->C, B->D, C->D                                        cpp ACCEPTED   cs ACCEPTED
LEGAL 4  three-file chain A->B->C                                              cpp ACCEPTED   cs ACCEPTED
```

Multi-hop cycles name the whole ring, not just the closing edge:

```
C.schema:3:1: C3 closes a cross-file table reference cycle: A -> B -> C -> A — ...
```

**Deterministic** — the sorting is real, not decorative. 30 runs of each cycle case, `sort -u`:

```
1 unique line   (3-file cycle)
1 unique line   (2-file cycle)
```

The backend copy is gone (80 lines removed from `cpptable.go`, no `include cycle` string left in any table backend). The one remaining `include cycle` in `internal/codegen/cpp/cpp.go` is the packet emitter's own rule for `type` headers — a different, pre-existing check, correctly untouched.

## F3 — CLOSED, and I was wrong to hedge

Source and PR body both now say four (`TableFieldInfo`'s `offset`, `elem_size`, `count_offset` plus `TableTypeInfo`'s `size`). I checked the "every other column is there name for name" claim properly this time rather than eyeballing it: C# carries `TableRef` as a `Func<TableTypeInfo>` **and** `Table` as a property over it —

```csharp
public Func<TableTypeInfo> TableRef;
public TableTypeInfo Table { get { return TableRef == null ? null : TableRef(); } }
```

— so C++'s `table` column *is* present name for name, with the factory beside it for descriptor-graph init order. The sentence is literally true. Nothing outstanding.

## F5 — CLOSED

`check.go`'s stale `// C++-only today, so only the CamelCase spellings claim.` is replaced with the accurate rule, and it now points at `checkClaimedNames` as the reconciliation site — which is where a future port will actually need to go.

## F4 — CLOSED. The numbers reproduce, and I want to record how

**First contact did not reproduce**, and that is worth stating plainly:

```
A 1377 ns/frame   B 1764 ns/frame   C 2334 ns/frame
B/A ratio 1.281   C/A ratio 1.694
```

`load1` was **15.72** — from my own `make test` minutes earlier, plus a leftover `VBCSCompiler`. That is the repo's own doctrine firing exactly as written (*"one bench at a time per machine: check for sibling bench processes and wait for a quiet window"*, *"discard contaminated runs whole"*). I killed the compiler server, waited ~10 minutes for load to fall under 2.5, and took three sittings:

```
===== SITTING 1 (load1=2.49) =====
A 1062   B 1056   C 1581      B/A 0.995   C/A 1.489
===== SITTING 2 (load1=2.45) =====
A 1061   B 1060   C 1584      B/A 1.000   C/A 1.493
===== SITTING 3 (load1=2.45) =====
A 1060   B 1059   C 1580      B/A 0.999   C/A 1.490
```

Against the README's claim (B/A 0.996 and 1.011; C 1.48–1.49x; C allocating 7 504 B/frame): **reproduced to three significant figures**, on a different machine, from the committed location. The allocation and footprint numbers are exact, not approximate:

```
alloc/frame:      A 0 B          B 0 B          C 7504 B
construction:     A 16440 B      B 47160 B
```

So B ties A on the read path and C is 1.49x slower with 7.5 KB/frame of garbage — the decision's stated basis holds, and it is now checkable by anyone with `dotnet run -c Release`.

**Correctly fenced as a probe, not a leg**, which was the part I most wanted to see done right:

```
grep -c cs-union-form bench/run.sh   ->  0
grep -c cs-union-form Makefile       ->  0
CI references                        ->  none
new ledger / LOCK / SHAPE-GATE.allow / corpus entries  ->  none
make shape-gate: clean — 19 file(s) carry shape knowledge, every one on the ledger
```

The README says so itself in its own words (*"nothing here may be divided against a bench CSV row"*), and the gate's one catch was reworded rather than exempted — no new exemption row, which is the right way to answer a shape gate.

**#270 is real and does not paper over the gate item.** Filed with the owner's doctrine quoted verbatim, and named from both `bench/README.md:312` and the probe's own `README.md:54`, each saying explicitly that the probe does not close it. The thing #262 asked for — a fixed table beside its equivalent type on the ledger — is still owed and is now impossible to lose.

## F8 — closed in passing

`tables-cs-standalone` no longer depends on the glob being non-empty.

---

## Full battery, re-run on this head

The delta changed **no generated behaviour** — the emitted C++ is byte-identical to the pre-delta build and the emitted C# differs only in the F3 descriptor comment:

```
diff -r .../tables-generated/examples          ->  (identical)
diff -r .../tables-generated-cs                ->  only the descriptor comment block
```

I re-ran the differential anyway rather than reasoning from that:

```
6459 hostile buffers -> both backends
cpp=6459  cs=6459
divergent lines: 0
```

(Same corpus as the cold review: every §4 event directed, 6 000 deterministic mutations of four seeds, 378 extreme-value instances; comparing load return, all four report counters, measure, save, and the full re-encoded bytes.)

```
MAKE TEST EXIT=0                       # full nine-language, pinned dist/ toolchains
tables test passed
tables C# standalone gate: generated Table sources name no runtime
tables C# refusal gate: a pointered unit is refused by name
cs tables test passed
go test ./... — 16 ok, 0 FAIL

--- PASS: TestTablesMoveNoProtocolId          # independence
--- PASS: TestTablesMoveNoGeneratedPacketByte # independence
--- PASS: TestTableFileDag                    # new, F2
--- PASS: TestTableRuntimeNamesAreClaimed     # new, F1 (12 subtests)
--- PASS: TestCsEmitsTableSources
--- PASS: TestCsRefusesPointeredTables

make check / make fmt: clean, no drift · gofmt -l: clean · go vet: clean
```

**Provenance:** seven commits, every author and every committer `rowan@mas-bandwidth.com`, every one carrying the Claude co-author trailer. No strays.

---

## Verdict

**MERGE.** F1–F5 and F8 are closed, and two of them closed better than the finding asked: F1 became front-end law covering every backend's spelling (which retired a C++ defect standing since #254), and F2's rule moved to the front end where §11 always said it belonged, with a negative-controlled union-arm case and proven determinism. F4's numbers reproduce to three significant figures from the committed harness on a different machine, correctly fenced as a probe, with the real bench obligation filed as #270 rather than quietly absorbed.

The one thing I could still defeat is the *scan* behind F1's gate, not the claim it guards — five C# declaration shapes the regexes do not see. Non-blocking, worth hardening the next time a port touches that file.

Held behind **#269** and the `?T`/`[E]T` seam fill.

gafferongames:
## Delta read — head `92d070c` · verdict: **MERGE**

Fresh clone, `gh pr checkout 266`, branch never edited (`git diff origin/cs-tables HEAD` empty, `git status` clean). Lands after #263.

Both seams are filled and I could not break them. Headline: **12,497 hostile buffers through both backends — 0 divergent lines**, all three new negative controls reproduce exactly, and the branch closes a **second** non-compiling-header hole that main carries today. One non-blocking judgment call below on `TableReport.Duplicate`.

---

## 1. The 18 goldens — verified against a C++ writer, and read by a third party

First, the premise. The branch changes **no C++ byte**, so "the goldens are the C++ reference's own bytes" is not an assumption:

```
$ diff -r <cpp generated from 85becbf> <cpp generated from 92d070c>
IDENTICAL
```

Then the pins themselves:

```
goldens pinned: 18
tables test passed
$ SCHEMA_UPDATE_WIRE_GOLDENS=1 ./build/schema_test_tables && diff <before> <after>
18 GOLDENS REGENERATE BYTE-IDENTICAL from the C++ writer
$ git status --porcelain testdata/     (clean)
cs tables test passed
```

**16 of the 18 are shared**; `chain_pointer` and `chain_pointer_empty` are C++-only, correctly, because the C# variable class is a follow-on. Worth stating plainly since the brief said 18 — the count is 18 pinned, 16 cross-checked.

I did not take either test's word for the wire. I wrote a **decoder from SPEC-TABLES §3/§3.2 alone** — no schema code, no generated code — and ran it over all 18. My first pass errored on the three keyed files because I had guessed the kind-16 framing; §3.2 says `L`, then `element kind (u8)`, `N` (u32) = **slots present**, then N pairs of `variant id (u16)`, `L (u32)`, body. With the spec's framing:

```
=== archive 414  chain_optional 40  chain_optional_empty 11  chain_pointer 40
=== chain_pointer_empty 11  chain_value 40  chain_value_empty 2  keyed_config 234
=== keyed_default 2  loadout_full 152  profile_elide 46  root_default 2  root_full 398
=== v1_cfg 181  v1_seams 197  v2_cfg 184  v2_seams 195  wide_blob 210133
                                                    ... all 18: OK
```

**The empty-end asymmetry, read out of the C++-written bytes.** The populated end is byte-identical across all three spellings:

```
chain_value      df300c05000000636861696eb20b0d130000009491040700000064bc0c0300000074697000000000
chain_optional   df300c05000000636861696eb20b0d130000009491040700000064bc0c0300000074697000000000
chain_pointer    df300c05000000636861696eb20b0d130000009491040700000064bc0c0300000074697000000000
```

The empty end diverges, and `test/tables/main.cpp:3136` says exactly why (*"a by-value nesting at its defaults writes nothing, while a PRESENT optional and a non-null pointer write their body anyway"*):

```
chain_value_empty      0000                          (2 B — all-default by-value nesting ELIDES)
chain_optional_empty   b20b0d0200000000000000        (11 B — present + all-default RIDES)
chain_pointer_empty    b20b0d0200000000000000        (11 B — identical to the optional)
```

And the dual holds from the other side: an **absent** optional writes nothing (`Chain__c_absent` resaves 11 B carrying only `name`). Presence is information the wire must carry, so it defeats default-elision; a value nesting has no presence to carry, so it does not. Pinned correctly.

**The keyed body**, decoded by the third-party decoder — nested keyed inside keyed, `N` counting slots present, scalar and table element kinds both:

```
id=0x9ae1 keyed L=57 elem=table N=2 { key=0xbb03 {spawn_count=8, banner="red"} ... }
id=0xeff5 keyed L=120 elem=table N=2 { key=0xb534 { health=250, turrets: keyed N=2 {...} } }
id=0xdcf3 table L=34 { id=0x443f keyed L=25 elem=i32 N=2 { key=0xbb03 {10} key=0x6e0f {30} } }
```

## 2. The three new negative controls — reproduced, each on a copy of the generated tree

**NC1 — `?T` made content-decided.** Exactly as described, plus three assertions the PR body did not enumerate:

```
FAILED: table wire golden keyed_config — (225 written, 234 pinned)
FAILED: chain_optional_empty: presence decides, not content
FAILED: table wire golden chain_optional_empty — (2 written, 11 pinned)
FAILED: and a non-null pointer writes the same body — the empty end's asymmetry is shared
FAILED: optional: present-at-default RIDES
FAILED: optional: a field that rode is PRESENT
```

**NC2 — the keyed body written positionally** (kind 14, slot index for the variant id). The load-bearing detail reproduces: **same length, different bytes** — a length check alone would have missed it:

```
FAILED: table wire golden keyed_config — (234 written, 234 pinned)   <-- same length
FAILED: table wire golden v1_seams     — (197 written, 197 pinned)   <-- same length
FAILED: table wire golden v2_seams     — (195 written, 195 pinned)   <-- same length
FAILED: keyed evolution: Beta rode by NAME — a positional encoding would have put it in Omega's slot
FAILED: keyed evolution: two Gamma slots V2 cannot name — counted, never misplaced
FAILED: keyed evolution reverse: Omega and Sigma are names V1 does not have
FAILED: keyed vs positional reverse: a keyed body under a positional field is a KIND MISMATCH
      ... 14 assertions red in total
```

**NC3 — key 0 treated as unknown instead of malformed.** Red on both halves of the rule, as claimed:

```
FAILED: key 0: a None key is framing damage, not an unknown variant
FAILED: key 0: and it is NOT counted unknown — 0 names nothing, it is damage
```

Baseline before and after each: `cs tables test passed`.

## 3. C# cannot express a compile-time key — the guard survives release, and nothing boxes

The contrast is real. C++ carries **both** halves:

```cpp
T & operator[]( E key ) { assert( key != E::None ); ... }          // compiled out under NDEBUG
template <E Key> T & Slot() { static_assert( Key != E::None, ... ) } // compile-time, no C# spelling
```

C# has no `Slot<Key>()` (generics take no non-type parameter) and its indexer takes `int`. So the question is whether the runtime guard actually survives. Probed in a **Release** build, optimizations on, `DEBUG` undefined:

```
configuration: RELEASE (optimizations on)
DEBUG symbol defined at compile time: False
OK  get [0] threw: slot 0 is None's and is never valid: None is the null key of an enum-keyed array
OK  set [0] threw
OK  [(int)Team.Blue] = 9
RefuseNone present in the RELEASE assembly: True
RefuseNone [Conditional] attributes: 0 (must be 0)
indexers on TableKeyed<T>: 1
  indexer param types: Int32
generic methods on TableKeyed<T>: 0 (must be 0)
Slots field public: True
  window 0/1/2: indexer alloc over 100000 reads = 0 B / 0 B / 0 B
  indexer WRITE alloc over 100000: 0 B
GUARD PROBE: ALL OK
```

No `[Conditional]`, no `#if`, present in the release assembly, and it throws. **No enum-taking indexer and no generic method at all** on `TableKeyed<T>` — so there is no `Convert.ToInt32(TEnum)` / boxing form on offer; the caller writes `(int)ShipType.Bomber` and the USAGE text says so. Generated codecs reach `value.Teams.Slots[i]` directly and never pay the guard (confirmed in the emitted source).

*(My first allocation reading showed 6 216 B — that was a one-off tier-up inside the measurement window, my error, not the code's. With a 200 000-iteration warm-up and three successive windows it is flat zero.)*

## 4. The `TableKeyed` hole — real on main, closed here for both targets

**On main (`85becbf`)** `table TableKeyed` is accepted and the generated C++ header does not compile:

```
XTable.h:215:8: error: redefinition of 'TableKeyed' as different kind of symbol
XTable.h:222:5: error: use of class template 'TableKeyed' requires template arguments; argument
                deduction not allowed in non-static struct member
XTable.h:223:16: error: use of class template 'TableKeyed' requires template arguments
XTable.h:255:41: error: use of class template 'TableKeyed' requires template arguments
```

**On the branch (`92d070c`)**, refused at check for both targets, at the declaration:

```
cpp  X.schema:5:1: declaration TableKeyed collides with the generated TABLE-wire runtime (SPEC-TABLES.md)
cs   X.schema:5:1: declaration TableKeyed collides with the generated TABLE-wire runtime (SPEC-TABLES.md)
```

That is the second §11 hole this PR has closed in the C++ reference (after `TableEnumId`/`TableEnumValue`), and both fell out of making the claim list front-end law rather than one target's inventory. The registry is earning its keep.

**And the registry closed my last round's open note.** I flagged that the old regex scan saw only two declaration idioms; the new `internal/tablenames` + identifier scan catches all five shapes that previously slipped:

```
(c) public enum TableFlagsX          FAIL <-- caught     (was: MISSED)
(d) public struct TableSpanX         FAIL <-- caught     (was: MISSED)
(e) public static readonly TableLimitX FAIL <-- caught   (was: MISSED)
(f) generic TableCastX<T>()          FAIL <-- caught     (was: MISSED)
(g) public class TableThingX         FAIL <-- caught     (was: MISSED)
(h) a name only in a // comment      pass  <-- correct: a comment declares no symbol
```

and the reverse direction holds too, with a message that tells you which way to fix it:

```
tables_test.go:575: internal/tablenames says the C# backend defines TablePhantomX, but nothing in
the emitted C# names it — drop the registration or fix the backend; a claim nothing needs takes a
name away from every schema for free
```

## 5. The differential, extended — 12,497 buffers, 0 divergence

The cold review's 6,459 re-run on this head, **plus 6,038 new** for the seams: 38 directed keyed/optional cases and 6,000 deterministic mutations of `KeyedConfig`, `TurretConfig` and `P3.Chain`. Compared: load return, all four **shared** report counters, measure, save, and the full re-encoded bytes.

```
total buffers: 12497     cpp lines=12497     cs lines=12497
DIVERGENT LINES: 0

unknown   9601 =0 | 2880 =1 | 13 =2 | 3 =3
kind     11314 =0 | 1183 =1
clamped  10821 =0 | 1187 =1 | 438 =2 | 51 =3
malformed 7268 =1 | 5229 =0        load=0 4501 | load=1 7996
```

The new §3.2 rules all land identically, and the directed cases show they land *correctly*, not merely identically:

```
k_key0_first            malformed=1  resave=2 B      (body stops at once, nothing kept)
k_key0_last             malformed=1  resave=37 B     (prefix before key 0 IS kept)
k_key0_middle_then_more malformed=1  resave=37 B     (parent reads on past the length)
k_unknown_key           unknown=1  malformed=0       (foreign name, not damage — known slot kept)
k_all_unknown_keys      unknown=2  resave=2 B
k_dup_key               clean, resave 1 slot         (last-wins, silent)
k_dup_key_triple        clean, resave 1 slot
k_descending            clean, 3 slots               (reader does not rely on ascending)
k_as_array14            kind=1                       (keyed field, kind 14 on the wire)
arr_as_keyed16          kind=1                       (array field, kind 16 on the wire)  <-- both directions
k_wrong_elem_kind       kind=1
k_N_over / k_N_huge / k_truncated_pair / k_slot_len_past_body   malformed=1
unknown_kind16          unknown=1                    (Skip gained 16 — unknown field skipped by L)
o_absent                gunner not written
o_present_default       11 B, the empty body rides
o_present_valued        22 B
```

**Allocation on the new shapes** — the keyed array was the obvious new risk, and it is clean:

```
KeyedConfig.Load / Measure / Save (keyed_config)        0 B/op
KeyedConfig.Load (keyed_default, all-default)           0 B/op
HullConfig.Load (malformed)                             0 B/op
TurretConfig.Load / Measure / Save (?T present)         0 B/op
P3.Chain.Load (?Link present)                           0 B/op
Schema.KeyedConfigTableType()                           0 B/op
NEGATIVE CONTROL: new byte[64]                         88 B/op
```

## 6. Battery, docs, provenance

```
MAKE TEST EXIT=0                       # full nine-language, pinned dist/ toolchains
tables test passed · cs tables test passed
tables C# standalone gate: generated Table sources name no runtime
tables C# refusal gate: a pointered unit is refused by name
go test ./... — 16 ok, 0 FAIL

--- PASS: TestTablesMoveNoProtocolId            (independence)
--- PASS: TestTablesMoveNoGeneratedPacketByte   (independence)
--- PASS: TestTableFileDag
--- PASS: TestTableRuntimeNamesAreClaimed
--- PASS: TestCsEmitsTableSources · TestCsRefusesPointeredTables

make check / make fmt: clean, no drift · gofmt -l: clean · go vet: clean
```

**Docs — confined to the four page changes, confirmed by hunk count.** Every hunk in the two changed files, and nothing else in the tree's `.md` beyond the bench READMEs already reviewed:

```
SPEC-TABLES.md  @@ -76   backend status line
                @@ -1315 §11 C# variable-class refusal
                @@ -1615 §15 lines
USAGE.md        @@ -579  "C++ only today" -> C++ and C#
                @@ -659  the C# example block
```

Three hunks in SPEC-TABLES, two in USAGE, zero elsewhere. Nothing else moved.

**Provenance:** 9 commits, every author and committer `rowan@mas-bandwidth.com`, 9 co-author trailers, no strays.

---

## One judgment call: `TableReport.Duplicate` — non-blocking, but I would drop it

It never fires. Not once across 12,497 buffers, including cases built specifically to trigger it — a keyed body with the same key twice and three times, and repeated field ids:

```
Duplicate distribution over all 12,497:   12497 × duplicate=0
k_dup_key  duplicate=0    k_dup_key_triple  duplicate=0    o_repeated_twice  duplicate=0
buffers that set it at all: 0
```

There is no write site anywhere — the only occurrence in the emitter is the declaration at `cstable.go:533`, and the only occurrences in generated code are five field declarations.

That is **documented and intended**, and I want to be fair about it: the comment says it is the text form's counter and the wire never raises it (§16.2), SPEC-TABLES §1781 does define `report.duplicate`, and the leg pins the invariant (`Check(report.Duplicate == 0, "…a wire read leaves it zero")`). So it is a reservation, not a dead lie.

The reason I would still drop it: **neither backend implements §16.** There is no `Json`/`Text` emitter in `cpptable`, and no `duplicate` anywhere in the C++ tree — the C++ `TableReport` has four members, C#'s has five. So the two report types are no longer mirror images, and the divergence is in the port, for a feature that does not exist on either side yet. There is no forcing function: adding a public field to a C# class when §16 lands costs nothing, and if it lands in the reference first the two stay identical the whole way. Against that: it is inert, documented at the site, and pinned by test.

Either drop it until §16 lands (my preference — the reference does not have it), or mirror it into the C++ `TableReport` in this PR so the two shapes stay equal. Not a blocker; a coin-flip I would call the other way, and worth a sentence in the PR body either way so the next port does not read it as licence to add report members ahead of the reference.

---

## Verdict

**MERGE.** Both seams are filled correctly and provably: the keyed wire matches a decoder written from the spec alone, the three-way set's empty-end asymmetry is pinned and reads correctly out of C++-written bytes, key 0 is damage rather than a foreign name in both halves of the rule, kind 16 is a mismatch against 14 in **both** directions, and 12,497 hostile buffers produce zero divergence with zero allocation. The three new negative controls reproduce exactly, including NC2's same-length/different-bytes case that a weaker gate would have missed. The C# accessor's honest limitation is documented and its runtime guard genuinely survives release. And the PR closes a second non-compiling-header hole that main carries today — while the new registry closes the scan blind spot I left open last round, in both directions.

`Duplicate` is the only thing I would change, and it does not hold the merge. Lands after #263.

Issue #257, opened 2026-09-02
Packing moves into schema: a directory tree that mirrors a root table becomes one table instance — `schema pack`, un-opinionated by construction


gafferongames:
Owner, 2026-09-02 (Glenn, live), on enum-keyed arrays (#255): "It also may make it possible to move the table reading and packing entirely down into schema now." / "which is a WIN".

**Why it is un-opinionated now.** What made pack an opinion was linking instances into a collection by an enum key, and the key field inside each instance. With #255 the link IS the declaration (`ships [ShipType]ShipConfig`), and with #253 the JSON form of the root table IS the packed content: `{ "ships": { "Fighter": {...}, "Bomber": {...} } }` reads through the generic walk into one root instance, and `ConfigSave` writes Config.bin. No key field, no manifest, no collection concept.

**What schema adds (proposal, Rowan's, for the owner's one-line ruling on the directory rule):** `schema pack --root Config --out Config.bin <dir>` assembles ONE JSON document from a directory tree that MIRRORS the root table, then runs #253's walk and Save:

- a directory named after a field holds that field's value;
- for an enum-keyed array, one file per variant, `<Variant>.json`;
- for a bounded array, files in name order become elements (or one `<field>.json` holding the array);
- for a nested table, either `<field>.json` or a directory of its fields;
- a plain `<field>.json` at any level is the field's value verbatim;
- the root may also just be one `<Root>.json`.

The rule is structural only — the tree is the table — so it prescribes nothing about content. Output is the root table's wire bytes and nothing else; an envelope (magic, content hash, protocol id) is the caller's few lines, per SPEC-TABLES.md §1, or a documented `--envelope` flag if one recurring shape earns it (decide in the dogfood, not here). The inverse, `schema unpack`, writes the tree back from a bin via ToJson, which is the editor/tool round trip §1 promises.

**Consequence.** replicant's data engine (37k lines: table, pack, four codegens) retires whole; replicant keeps delta and objects; space vendors schema for tables and replicant for the wire. space#441 phase 2 becomes: schema pack replaces replicant pack, SchemaBin.h shrinks to the envelope.

**Held by test:** tables/examples gains a directory corpus that packs to bytes identical to Save of the hand-built instance; unpack→pack is byte-stable; hostile tree (unknown dir, two files for one variant, a variant name the enum lacks) is refused or counted per #253's rules. Gate item for schema#248; queued after #253 and #255.

gafferongames:
[...]

PR #267, opened 2026-09-02
Tables: JSON in and out of one table over the reflection descriptors — §16 (#253)

### LINKED ISSUE #253: Tables: JSON in and out of ONE table instance, driven by the reflection descriptors — the un-opinionated half of packing

gafferongames:
Closes #253. Gate item for #248.

**Rebased onto 85becbf; SPEC-TABLES.md, SPEC.md and USAGE.md are taken from main
WHOLESALE** — this branch's own §16 and every other doc edit was dropped, and the
three files are byte-identical to main. The page (#269, #273) is the source of the
rules; this PR is the C++ backend meeting them.

One PR: the reflection columns the walk needs, ONE generic walker emitted per
generated header, the `json = "key"` attribute, `?T` and `[E]T` in the text form,
the corpus and hostile batteries, three gates, and a nested-guard corpus table.

## What this implements, against the page

**One walk, not a codec per table.** The header carries §8's descriptors and one
walker over them; `<Name>FromJson`, `<Name>ToJsonMeasure` and `<Name>ToJson` are
wrappers that name a descriptor. `make tables-json-walk` holds it: the walker's
source is byte-identical in all **14** generated headers of the corpus, whose units
disagree about packages, tables, kinds, keyed arrays, optionals and pointer modes.

**Reflection gained** flags bit names, union arm payloads (tag offset/width, each
arm's payload offset and descriptor), a per-type `reset` hook, each field's text
key, and `bits(N)`'s implied range. Merged with #265's `optional`/`present_offset`
and `key_type_name`/`key_name`/`key_id` columns.

**`?T`** — presence of the KEY is the presence. A key sets the field present
whatever its value, including one the walk then counts as a kind mismatch; a JSON
`null` is the exception the page names and reads as ABSENT, putting the value back
at its defaults so a repeated key ending in `null` leaves nothing behind. An absent
optional writes no key; a PRESENT all-default one still writes, because presence is
the payload and not content.

**`[E]T`** — an object keyed by variant NAME. The walk enumerates
`[0, array_bound)` and skips the slot whose key id is `0`, which is exactly the
mechanism §8 states, so `"None"` names no slot, reads as an unknown key, and is
never emitted. An absent key keeps that slot's defaults; a repeated field key
resets every slot first, so last-wins replaces the object outright.

**The `json` key** moves no byte on either wire — `TestJsonKeyMovesNoWire` pins the
projection, the protocol id and every table-wire field id, and the committed
`generated/` tree does not move.

## Where the code does NOT match the page — one gap, and it is the owner's call

**§16.1 states a builder surface for the VARIABLE-LENGTH class**
(`SceneFromJson( builder, text, text_bytes, &report )`), and §16.2 carries a
`pointer *T` row. **This backend does not implement either.** A variable-length
table gets no text form; the refusal is by absence and it is loud — the header
carries a named comment and there is no function to call, so nothing reads a
pointered table as an empty one.

It is implementable, but not as a line item on this PR, and the reason is a
collision between two of this document's own gates:

- the one-walk gate wants a single walker byte-identical in **every** header;
- the zero-cost gate (§2.2) forbids `Builder` / `TableRef` / `TableRegion` from
  appearing in a **pointer-free** header.

A builder-aware walk must name those symbols, so one walker cannot satisfy both.
The way through is a SECOND walker emitted only in `anyVariable` units, plus
conditional descriptor columns carrying per-target `Emplace`/`At` as function
pointers — the same conditional treatment `is_pointer` already gets. That is a
round of roughly the size of this one, not a patch.

**§15 no longer lists it as a follow-on**, so the page currently promises a surface
nothing provides. Either §16.1's builder line and §16.2's pointer row move back to
§15 as a named follow-on, or this work grows the second walker. **Not mine to
decide — flagging the sentence.**

Two further §16.2 rows describe constructs the compiler cannot yet declare, so
there is nothing to implement or test for them: `*string` / `*bytes` (§2.5, refused
by the parser) and a `table` union arm (§2.6, refused by the checker).

## Held by test

`make test` runs all of it, plus three gates: `tables-json-walk` (one walker, 14
headers), `tables-zero-cost` (unchanged and green), `tables-json-negative-control`.

- Every table in `tables/examples` round-trips `ToJson` → `FromJson` → `Save`
  byte-identical to the original's `Save` — the keyed and optional shapes, the
  cross-file nesting, the 70000-element extents, the all-default instance and the
  evolution pair included.
- **Pinned texts** — known instance, known literal text — for the corpus root, and
  for both new constructs. This is the test a round trip cannot replace: reader and
  writer share `enum_name`, so a vocabulary error round-trips green. Verified
  honest by sabotaging bit-name emission.
- Hostile: wrong JSON types at every kind, malformed number tokens, both-direction
  overflow at every integer width, float widths, `bits(N)`, multi-byte string
  clamping, over-long arrays, bad base64, two-key unions, `"None"` as a keyed key,
  lone surrogates, thirteen framing breaks, nesting past the cap.
- Guards in all four states of a nested `if/else` (`Guarded.schema`, added), and a
  text whose keys precede their guards.
- A mutation fuzz asserting the page's honesty clause directly: **no text that
  reads clean may produce an instance the writer would refuse.**
- **Negative control**, checked in both directions — it fails when built against
  the unsabotaged walker, so it cannot pass by accident.

Whole suite clean under `-fsanitize=address,undefined -fno-sanitize-recover=all`.

## Header compile cost, banked

`+11%` to `+15%` on an empty TU including one corpus table header (arm64 clang,
three interleaved same-sitting pairs; an independent reviewer measured +10.9% at an
earlier revision). The three number-conversion includes are ~+2% of it; the walker
body is the rest. One sitting was discarded whole for contamination rather than
averaged in. The page states the walk rides in every table header with no opt-out;
this is what that costs.

## Review

The adversarial review's conditions F1–F12 are all addressed; the response is in
[this comment](https://github.com/mas-bandwidth/schema/pull/267#issuecomment-5503612840)'s
reply and the commits that cite them. F2 (JSON has one number type), F3 (RFC 8259
number grammar) and F5 (valid-UTF-8 output) became rules on the page and are
implemented as the page now states them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ build/schema_test_guard: build/guard-generated/.stamp test/guard/main.cpp
 # The tables corpus (SPEC-TABLES.md): the tabledemo unit plus the
 # two-generation evolution pair (tblv1/tblv2), generated at build time into
 # build/ — test-only, never part of the committed generated/ tree.
-build/tables-generated/.stamp: bin/schema $(SCHEMAS_TABLES) $(SCHEMAS_TABLES_POINTERS) test/tables/V1.schema test/tables/V2.schema test/tables/P1.schema test/tables/P2.schema test/tables/P3.schema
+build/tables-generated/.stamp: bin/schema $(SCHEMAS_TABLES) $(SCHEMAS_TABLES_POINTERS) test/tables/V1.schema test/tables/V2.schema test/tables/P1.schema test/tables/P2.schema test/tables/P3.schema test/tables/JsonKeys.schema
 	@mkdir -p build/tables-generated
 	./bin/schema generate --lang cpp --out build/tables-generated/examples tables/examples
 	./bin/schema generate --lang cpp --out build/tables-generated/pointers tables/pointers
@@ -208,6 +208,7 @@ build/tables-generated/.stamp: bin/schema $(SCHEMAS_TABLES) $(SCHEMAS_TABLES_POI
 	./bin/schema generate --lang cpp --out build/tables-generated/p1 test/tables/P1.schema
 	./bin/schema generate --lang cpp --out build/tables-generated/p2 test/tables/P2.schema
 	./bin/schema generate --lang cpp --out build/tables-generated/p3 test/tables/P3.schema
+	./bin/schema generate --lang cpp --out build/tables-generated/jsonkeys test/tables/JsonKeys.schema
 	@touch $@
 
 # The ZERO-COST GATE (SPEC-TABLES.md): a table with no pointer in its by-value
@@ -227,6 +228,45 @@ tables-zero-cost: build/tables-generated/.stamp
 	done
 	@echo "tables zero-cost gate: value-only tables carry no pointer machinery"
 
+# The GENERIC-WALK GATE (SPEC-TABLES.md §16): the text form is ONE walk over
+# the reflection descriptors, not a per-table codec — that is the property
+# which makes it schema's rather than a packer's. The walker's source must
+# therefore be the SAME BYTES in every generated header of the corpus, whose
+# units disagree about packages, tables, kinds and pointer modes. The package
+# name lives in the guard and the namespace, outside the markers, so this is a
+# strict byte comparison with nothing normalised away.
+.PHONY: tables-json-walk
+tables-json-walk: build/tables-generated/.stamp
+	@rm -rf build/json-walk && mkdir -p build/json-walk
+	@for f in build/tables-generated/*/*Table.h; do \
+		out=build/json-walk/$$(echo $$f | tr / _); \
+		awk '/---- json walk: begin ----/,/---- json walk: end ----/' $$f > $$out; \
+		if [ ! -s $$out ]; then echo "GENERIC-WALK GATE FAILED: no walker in $$f"; exit 1; fi; \
+	done
+	@first=""; for f in build/json-walk/*; do \
+		if [ -z "$$first" ]; then first=$$f; else \
+			cmp -s $$first $$f || { echo "GENERIC-WALK GATE FAILED: the walker in $$f is not the walker in $$first"; exit 1; }; \
+		fi; \
+	done
+	@echo "tables generic-walk gate: one walker, byte-identical in $$(ls build/json-walk | wc -l | tr -d ' ') generated headers"
+
+# The NEGATIVE CONTROL for the walk (SPEC-TABLES.md §16.5). A green round-trip
+# suite proves nothing until the suite is shown capable of going red: the
+# walker's field-offset arithmetic is sabotaged by one field width, and the
+# round trip must FAIL on the first table with two fields. Attachment is that
+# table — two four-byte fields at offsets 0 and 4 — so the sabotage swaps them
+# and stays inside the struct.
+.PHONY: tables-json-negative-control
+tables-json-negative-control: bin/schema test/tables/json_negative_main.cpp
+	@rm -rf build/json-sabotage && mkdir -p build/json-sabotage
+	./bin/schema generate --lang cpp --out build/json-sabotage tables/examples
+	@sed -i.bak 's|const uint8_t \* storage = (const uint8_t \*) base + f->offset;|const uint8_t * storage = (const uint8_t *) base + ( f->offset ^ 4 );|' build/json-sabotage/TablesTable.h
+	@grep -q 'f->offset ^ 4' build/json-sabotage/TablesTable.h || { echo "NEGATIVE CONTROL: the sabotage did not apply"; exit 1; }
+	@mkdir -p build
+	$(CXX) -std=c++17 -Wall -Wextra -Werror -ffp-contract=off \
+		-Ibuild/json-sabotage test/tables/json_negative_main.cpp -o build/schema_test_json_negative
+	./build/schema_test_json_negative
+
 # Deliberately compiled WITHOUT -I$(SERIALIZE): the generated Table headers
 # carry no serialize dependency, and this build proves it stays that way.
 build/schema_test_tables: build/tables-generated/.stamp test/tables/main.cpp
@@ -235,6 +275,7 @@ build/schema_test_tables: build/tables-generated/.stamp test/tables/main.cpp
 		-Ibuild/tables-generated/examples -Ibuild/tables-generated/pointers -Itest/tables \
 		-Ibuild/tables-generated/v1 -Ibuild/tables-generated/v2 \
 		-Ibuild/tables-generated/p1 -Ibuild/tables-generated/p2 -Ibuild/tables-generated/p3 \
+		-Ibuild/tables-generated/jsonkeys \
 		test/tables/main.cpp -o $@
 
 build/schema_test_random: generated/cpp/.stamp test/random_main.cpp
@@ -390,6 +431,8 @@ test: build/schema_test build/schema_test_guard build/schema_test_tables build/s
 	./build/schema_test_guard
 	./build/schema_test_tables
 	$(MAKE) tables-zero-cost
+	$(MAKE) tables-json-walk
+	$(MAKE) tables-json-negative-control
 	./build/schema_test_random
 	./build/schema_test_ludicrous
 	cd test/c && ../../build/schema_test_c
--- a/compiler/tables_test.go
+++ b/compiler/tables_test.go
@@ -145,8 +145,12 @@ func TestGeneratedTableCodeAllocatesNothing(t *testing.T) {
 			t.Errorf("generated table code contains %q — generated codecs must not allocate", banned)
 		}
 	}
+	// every `new` in the header is a PLACEMENT new — `new ( address ) T{}`,
+	// which allocates nothing: the read path's in-place prefill, and the
+	// descriptor's reset hook (SPEC-TABLES.md §8) which is the same prefill
+	// behind a function pointer. An allocating new has no parenthesis after it.
 	for line := range strings.SplitSeq(table, "\n") {
-		if found := strings.Contains(line, "new "); found && !strings.Contains(line, "new ( &") && !strings.Contains(line, "<new>") {
+		if found := strings.Contains(line, "new "); found && !strings.Contains(line, "new (") && !strings.Contains(line, "<new>") {
 			t.Errorf("generated table code contains a non-placement new: %q", line)
 		}
 	}
@@ -213,17 +217,23 @@ func tableHeader(t *testing.T, src string) string {
 
 // TestZeroCostForValueOnlyTables is the mode's whole justification, at the
 // grain the property actually holds: a UNIT whose tables are all fixed-size
-// emits none of the pointer machinery — no builder, no arena, no reference
+// emits none of the POINTER machinery — no builder, no arena, no reference
 // type, no lifecycle surface, not one extra descriptor column, not one extra
 // include (SPEC-TABLES.md §2.2). Per TABLE the guarantee is narrower and
 // TestPointerSurfaceEmitted states it exactly.
+//
+// What the gate is NOT about: the reflection surface and the text form that
+// walks it ride in every table closure's header by design, fixed class
+// included (SPEC-TABLES.md §16.1), and <cstdlib> is one of the text form's
+// three number-conversion includes rather than the arena's — which is why it
+// left this list when the walk landed.
 func TestZeroCostForValueOnlyTables(t *testing.T) {
 	header := tableHeader(t, tableSrc)
 	for _, leak := range []string{
 		"TableArena", "TableSlot", "TableWorker", "TableRef", "TableRegion",
 		"kTableSegment", "kTableSlab", "kTableMaxDepth", "is_pointer", "variable",
 		"Builder", "LayoutId", "OpenWalk", "PackMeasure", "LoadMeasure",
-		"Cook", "Open", "<atomic>", "<cstdlib>", "template",
+		"Cook", "Open", "<atomic>", "template",
 	} {
 		if strings.Contains(header, leak) {
 			t.Errorf("pointer machinery leaked into a value-only unit: %q", leak)
--- a/internal/check/check.go
+++ b/internal/check/check.go
@@ -508,6 +508,7 @@ var valuedAttr = map[string]bool{
 	"max":        true,
 	"resolution": true,
 	"was":        true,
+	"json":       true,
 }
 
 func (c *checker) enumMax(e *ast.MaxExpr) (*big.Int, bool) {
@@ -1151,6 +1152,9 @@ func (c *checker) resolveField(owner string, f *ast.Field, inTable bool) *ir.Fie
 		c.errf(f.Pos, "field %s: was is a table-wire concept — it aliases a renamed field's wire id, and only table fields have wire ids; a `type`'s wire is positional, so a rename there moves no bit (SPEC-TABLES.md)", f.Name)
 		return nil
 	}
+	// `json` outside a table CLOSURE is refused in checkTables, where the
+	// closure is known: a `type` a table reaches has a text form and may
+	// carry the attribute, and only membership decides it.
 
 	// the fixed and 128-bit families mirror serialize's own surface exactly
 	// (SPEC §4.3, runtime-first): fixed(I, F) and int128 are RANGED — the
@@ -1339,6 +1343,21 @@ func (c *checker) resolveAttrs(f *ast.Field, out *ir.Field) {
 				continue
 			}
 			out.WasName = lit.Value
+		case "json":
+			// the text form's key (SPEC-TABLES.md §16.3): the one attribute
+			// the JSON walk adds, so a declaration can meet an existing text.
+			// Table fields only — enforced below, where the owner's kind is
+			// known — and it moves no wire byte.
+			lit, ok := a.Value.(*ast.StringLit)
+			if !ok {
+				c.errf(a.Pos, `json takes the field's text key as a quoted string, e.g. json = "type" (SPEC-TABLES.md §16.3)`)
+				continue
+			}
+			if lit.Value == "" {
+				c.errf(a.Pos, "json = \"\" names nothing — json records the key this field reads and writes in the text form (SPEC-TABLES.md §16.3)")
+				continue
+			}
+			out.JsonKey = lit.Value
 		case "round":
 			// refused by name: rounding is not an attribute — it is the one
 			// fixed-point rule, half away from zero, everywhere (SPEC §4.3)
@@ -1726,6 +1745,22 @@ func (c *checker) checkTables() {
 			c.errf(pos, "table %s: the name collides with a member of the generated %sBuilder — a member function hides the type name it shares, and the header would not compile; rename the table (SPEC-TABLES.md §6.2)",
 				name, name)
 		}
+		// the TEXT form's keys (SPEC-TABLES.md §16.3): two fields of one
+		// closure member whose keys collide are indistinguishable in a JSON
+		// object, exactly as colliding ids are on the wire — refused once,
+		// naming both, whether the collision comes from a `json` attribute
+		// or from an attribute meeting a plain field name.
+		seenKey := map[string]*ir.Field{}
+		for _, f := range st.Fields {
+			key := ir.TableFieldJsonKey(f)
+			if prev, dup := seenKey[key]; dup {
+				c.errf(pos, "%s %s: fields %s and %s collide on the JSON key %q — rename one, or give one a different json key (SPEC-TABLES.md §16.3)",
+					what, name, describeTableJsonField(prev), describeTableJsonField(f), key)
+				continue
+			}
+			seenKey[key] = f
+		}
+
 		seen := map[uint16]*ir.Field{}
 		for _, f := range st.Fields {
 			if f.Type.Pointer {
@@ -1774,6 +1809,30 @@ func (c *checker) checkTables() {
 		}
 	}
 	c.checkTableVariantIdentity(names)
+	c.checkJsonKeysInClosure()
+}
+
+// checkJsonKeysInClosure refuses `json = "key"` on a field no table closure
+// reaches (SPEC-TABLES.md §16.3). The text form is the table closure's — a
+// `type` a table nests has one and may carry the attribute; a type nothing in
+// a closure reaches has no text form for a key to name.
+func (c *checker) checkJsonKeysInClosure() {
+	for _, name := range sortedKeys(c.structs) {
+		if c.tableClosure[name] {
+			continue
+		}
+		st := c.structs[name]
+		pos := ast.Pos{}
+		if d, ok := c.astDecls[name]; ok {
+			pos = d.DeclPos()
+		}
+		for _, f := range st.Fields {
+			if f.JsonKey != "" {
+				c.errf(pos, "type %s: field %s carries json = %q, but no table reaches %s — the text form is the table closure's, and a type outside one has none (SPEC-TABLES.md §16.3)",
+					name, f.Name, f.JsonKey, name)
+			}
+		}
+	}
 }
 
 // sortedKeys gives a map's keys in a stable order, so diagnostics do not
@@ -2039,6 +2098,15 @@ func declKindName(d ast.Decl) string {
 func describeTableField(f *ir.Field) string {
 	if f.WasName != "" {
 		return fmt.Sprintf("%s (was %q)", f.Name, f.WasName)
+	}
+	return f.Name
+}
+
+// describeTableJsonField names a field in a text-key diagnostic, showing the
+// `json` attribute where one carries the key.
+func describeTableJsonField(f *ir.Field) string {
+	if f.JsonKey != "" {
+		return fmt.Sprintf("%s (json %q)", f.Name, f.JsonKey)
 	}
 	return f.Name
 }
--- a/internal/check/check.go
+++ b/internal/check/check.go
@@ -2538,6 +2538,7 @@ var tableGeneratedVerbs = []string{
 	"LoadMeasure", "LoadMeasureBody", "LoadBuilder", "TableType", "Builder",
 	"At", "Root", "Emplace", "Pack", "PackMeasure", "OpenWalk",
 	"Cook", "CookMeasure", "Open", "LayoutId", "TableFields", "TableInfo",
+	"FromJson", "ToJson", "ToJsonMeasure",
 }
 
 // tableBuilderMembers are the member names of a generated <Name>Builder. A
--- a/internal/check/tables_test.go
+++ b/internal/check/tables_test.go
@@ -43,6 +43,16 @@ func TestTableRefusals(t *testing.T) {
 			src: "package t\ntable Tab { speed float32 | was = \"\" }\n"},
 		{name: "was takes a quoted string", want: "was takes the field's old name as a quoted string",
 			src: "package t\ntable Tab { speed float32 | was = velocity }\n"},
+		{name: "json outside a table closure is refused by name", want: "the text form is the table closure's",
+			src: "package t\ntype P { a int32 | json = \"b\" }\n"},
+		{name: "json keys collide inside one table", want: "collide on the JSON key",
+			src: "package t\ntable Tab {\n    a int32 | json = \"b\"\n    b int32\n}\n"},
+		{name: "two json attributes naming one key", want: "collide on the JSON key",
+			src: "package t\ntable Tab {\n    a int32 | json = \"k\"\n    b int32 | json = \"k\"\n}\n"},
+		{name: "json takes a quoted string", want: "json takes the field's text key as a quoted string",
+			src: "package t\ntable Tab { a int32 | json = b }\n"},
+		{name: "json with an empty string", want: "json = \"\" names nothing",
+			src: "package t\ntable Tab { a int32 | json = \"\" }\n"},
 		{name: "effective wire ids collide (was aliases a live sibling)", want: "collide on table-wire id",
 			src: "package t\ntable Tab {\n    a int32 | was = \"b\"\n    b int32\n}\n"},
 		{name: "a table nesting itself is a composition cycle", want: "type composition cycle",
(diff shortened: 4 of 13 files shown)


gafferongames:
## Adversarial review — MERGE WITH CONDITIONS

Cold review, fresh clone at `7f9dcbf`, no part in writing this. Everything below
was run; every claim has a repro. **Held regardless** behind #269 (the spec of
record, which already carries a §16 — see F10) and behind #265 (the seams).

The core claim is true. The generic-walk gate is real, the walk is genuinely one
walk, nothing on either wire moved, and a text I authored from §16.2 alone —
never from the code — round-tripped byte-identically on every row of the table.
The defects below are all in one place: **the READ path is more permissive than
the page says, in ways that lose or invent data silently.**

---

# FINDINGS

### F1 — HIGH. The reader accepts a value the writer refuses, and reports the text clean.

```
$ cat > p.cpp   # ProfileConfig.precision is float64, no declared range
  const char * t = "{\"precision\": 1e400}";
  ProfileConfigFromJson(v, t, strlen(t), &r);

P1 float64 1e400        ok=1 unknown=0 kind=0 clamped=0 dup=0 malformed=0
   stored precision = inf   ToJsonMeasure = -1
P1 float32 array 1e400  ok=1 unknown=0 kind=0 clamped=0 dup=0 malformed=0
   ratings[0]=inf ratings[1]=-inf  ToJsonMeasure=-1
```

`FromJson` returns **true** with an all-zero report — §16.2: *"Silence means the
text matched the schema exactly"* — and stores ±inf. `ToJsonMeasure` then returns
-1 for that instance forever. Decision 4 (the writer refuses a non-finite float)
is not mirrored on the read side, so §16 has a reachable state where an instance
built from a text the walk called clean can never be written back. That is the
one invariant §16.1 sells.

One line at the conversion site: a non-finite result is `clamped` to the field's
range (or `kind_mismatch`), never stored.

### F2 — HIGH. Integers in JSON's other legal spellings are silently dropped.

```
READ  exponent form for an integer  1e3    ok=1 kind=1 clamped=0 -> experience=0
READ  exponent form for an integer  2E0    ok=1 kind=1 clamped=0 -> heading=0
READ  integral float form           2.0    ok=1 kind=1 clamped=0 -> heading=0
READ  plain                         2      ok=1 kind=0 clamped=0 -> heading=2
```

JSON has ONE number type. `2.0` **is** the integer 2 and `1e3` **is** 1000, and
every JSON library that round-trips numbers through a double emits them that way
— including this walker's own float writer, which emits `"precision": 1e+21`.
§16.3 exists so *"a declaration can meet an existing text"*; this is the class of
existing text most likely to arrive, and each such field silently keeps its
default. §16.2 only warns that *"a fraction is a kind mismatch"* — `1e3` is not a
fraction and the page says nothing.

Accept a non-integral token whose VALUE is integral; keep a true fraction as
`kind_mismatch`, exactly as decision 1 ruled.

### F3 — MEDIUM-HIGH. The number scanner is not a JSON number grammar, and garbage becomes a clamped value.

```
P2 integer 1-2      ok=1 kind=0 clamped=2 malformed=0   heading=32767
P2 integer 5+       ok=1 kind=0 clamped=2 malformed=0   heading=32767
P2 float 1.2.3      ok=1 kind=0 clamped=0 malformed=0   precision=1.2
P2 float 1e5e5      ok=1 kind=0 clamped=0 malformed=0   precision=100000
P2 float --3        ok=1 kind=0 clamped=0 malformed=0   precision=0
P2 integer +7 / 007 ok=1 ... accepted as 7
P2 float .5 / 3.    ok=1 ... accepted
```

`TableJsonScanNumber` grabs any run of `[0-9+-.eE]` and hands it to a digit loop
that reads `'-'` as a digit value of 18446744073709551613, overflows, and reports
a **clamp**. §16.2: *"`malformed` is reserved for text that is not JSON"*. `1-2`
is not JSON. A typo in the authoring file this section exists for becomes int16
MAX with a clamp count, not a diagnostic — the single worst failure mode for a
config pipeline. Validate the token against JSON's number production before
conversion; anything else is `malformed`, which is what the page already promises.

### F4 — MEDIUM. "Duplicate keys: last wins" is false for arrays, and it is wire-visible.

```
P3 dup fixed array   dup=1   podium = Gold,Silver,Gold      (last-wins: Gold,None,None)
P3 dup fixed float   dup=1   ratings = 9.5 2.5 3.5 4.5      (last-wins: 9.5 0 0 0)
P3 dup counted array dup=1   grades_count=1 -> Gold,Silver,Gold
```
repro: `{"podium": ["Bronze","Silver","Gold"], "podium": ["Gold"]}`

§16.2 states last-wins generally and carves out only *"a nested table or a union
arm"* as re-establishing; it also says a short fixed array *"leaves the tail at
its defaults"*. Both are violated: the second occurrence overlays a prefix onto
the first occurrence's tail. The nested-table path already does the right thing
via the new `reset` hook — the array path just needs the same treatment.

### F5 — MEDIUM. `ToJson` can emit a text that is not valid JSON.

Two reachable routes:

```
P6 surrogate pair + lone low + lone high   ok=1  all counters 0
   name bytes: f0 9f 98 80 20 ed b0 80 20 ed a0 80
   text = { "name": "😀 ��� ���", ... }
P6 stray lead byte 0xC3 + '('              ok=1  name bytes: 61 c3 28 20 62
```

(a) `"\udc00"` is valid JSON input; the walk encodes it as CESU-8 `ed b0 80`,
which is not valid UTF-8, and writes it back as raw bytes. (b) any non-UTF-8
already in the field — which this PR deliberately preserves (19a88ea) — is
written unescaped. RFC 8259 §8.1 requires a JSON text to be valid UTF-8, so both
outputs are unreadable by a conforming parser, while §16.2's row says `string(N)`
is *"UTF-8"*. The round trip is closed under schema's own reader only. Owner's
call: either refuse a lone surrogate on read, or drop "UTF-8" from the row and
say the form is byte-transparent.

### F6 — MEDIUM. The §16 round-trip oracle cannot see a vocabulary error at all.

I sabotaged bit-name emission in a throwaway generated header
(`FlagNamePerks( (int) v )` → `( (int) v + 1 )`) and ran both oracles:

```
### against the SABOTAGED walker ###
PR oracle (ToJson->FromJson->Save): ok=1 wire-identical=GREEN
   text says: "perks": [ "Cloaked", "???" ]
PAGE-authored text vs hand-built:   wire-identical=RED (unknown=1)

### against the HONEST walker ###
PR oracle: GREEN     PAGE-authored: GREEN
```

`check_json_round_trip` stays green while the writer emits `["Cloaked","???"]`
for `Shielded|Turbo`, because reader and writer share `enum_name`. The suite is
saved only by the pre-existing §8 reflection assert at `test/tables/main.cpp:758`
— nothing in §16's own battery pins a spelling. (`test_json_key_attribute` is the
honourable exception: it reads a literal text and asserts the emitted key. Every
other §16 test round-trips through the writer.) **One pinned-text test — `ToJson`
of a known instance equals a known literal text — closes the whole class** (enum,
flags, union arm, None). It is also the test the negative control's own §16.5
sentence implies and does not deliver.

### F7 — MEDIUM (latent). Decision 4's refusal is enforced by a bound, not by the name.

`TableJsonWriteScalar` refuses on `bit > f->enum_max` / `tag > f->enum_max`, then
calls `enum_name` and writes whatever comes back. When the two disagree it writes
`"???"` — a spelling the reader counts as `unknown`. Unreachable in an honest
build (enum_max and the name function are generated together and agree for dense
enums, flags and unions), but it is not the check §16.2 claims: *"a set bit no
variant names is REFUSED on write."* Check the name.

### F8 — MEDIUM. The walk is unconditional and costs +10.9% header compile, unmeasured.

`<cstdio>`, `<cstdlib>` and `<clocale>` plus ~470 lines of inline walker now ride
in **every** table header, fixed class included, with no opt-out. Interleaved,
same sitting, median of 9, arm64 clang, an empty TU including `TablesTable.h`:

```
main       median 0.0466s  min 0.0455 max 0.0490   x1.000
branch     median 0.0517s  min 0.0504 max 0.0541   x1.109
```

The zero-cost gate is **not** weakened — its grep list is byte-unchanged and green,
and moving `#include <cstdlib>` out of the `anyVariable` branch touched the include,
not the gate. But this repo rules header cost on measured evidence (CLAUDE.md,
union vs variant: *"0.17 -> 0.27 is not trivial for me"*), and §16.1 states the
design choice while naming no number. Bank the number; say in §16.1 what it costs.

### F9 — LOW / ON THE PAGE. Six guesses I had to make writing a text from §16 alone.

All six proved right, which is the good news; four are not on the page:
1. **what None WRITES as for an enum** (`"None"` — §16.2 only says an unknown name
   *reads* as None);
2. **what `null` does anywhere** — the draft's only `null` was on the pointer row,
   deleted with it, and nothing replaced it (it is `kind_mismatch`);
3. **whether base64 padding is required, optional, or emitted** (both read; the
   writer pads);
4. **what an empty flags mask writes as** (`[]`).
(5) a `bits(N)` over its width clamps and (6) an over-long fixed array counts
`clamped` are inferable from the Bounds paragraph, but the kind rows do not say so.

### F10 — LOW. Rebase hazard: #269 already carries a §16 that contradicts this one.

`origin/spec-design-of-record` §16 is the **pre-decision #253 draft**, still saying:

```
49: | integers, `bits(N)` | number | integral; a fraction is malformed |
62: | pointer `*T` | ... `SceneFromJson( builder, text, bytes, &report )` |
80: **The report is the wire's report** (§4): ...
```

— superseded by decisions 1, 5 and 2 respectively, plus the guard sentence
(decision 3). A textual merge leaves the spec self-contradicting: **#267's §16
must REPLACE #269's §16 wholesale, not merge into it.** #269 also already assigns
§17 to packing and §18 to the baseline, so the PR body's *"if #263 lands first,
this section renumbers to §17"* is stale — §16 stays §16, #263 becomes §18. Noted,
not a defect.

### F11 — LOW. `duplicate` makes `TableReport` asymmetric across backends.

Checked directly against `origin/cs-tables`. **No descriptor-column conflict
exists**: cs-tables does not touch `TableFieldInfo`/`TableTypeInfo` at all, and
#267's three new columns (`json`, `arms`, `reset`) are entirely new. A trial
`git merge-tree origin/tables-json-walk origin/cs-tables` conflicts in exactly two
mechanical places:

```
CONFLICT (content): Merge conflict in Makefile
CONFLICT (content): Merge conflict in test/tables/main.cpp
```
(the `test:` target list and the test roster — whichever lands second rebases).

What does diverge is `TableReport`: cs-tables emits `Unknown/KindMismatch/
Clamped/Malformed` and no `Duplicate`. That struct is §4's cross-language
contract, and a C++ **wire** read now returns a counter no other backend has and
that no wire path ever raises. Say in §4 that it is text-form-only, so the
ninefold port knows.

### F12 — LOW / COVERAGE. The corpus has one flat guard.

`tables/examples` has only `if has_loadout` — no `else`, no nesting — so
`TableJsonGuardHolds`'s `active && !has_target` parse is never reached by the
suite. I built one (`Deep` with nested if/else) and it is correct in all four
states, elision matching the declaration exactly, and an all-keys-before-guards
text landing byte-identical to the hand-built instance:

```
active=0 target=0 | speed=0 target_id=0 wander=0 note=1 | wire 22 vs 22 IDENTICAL
active=1 target=0 | speed=1 target_id=0 wander=1 note=0 | wire 27 vs 27 IDENTICAL
active=1 target=1 | speed=1 target_id=1 wander=0 note=0 | wire 31 vs 31 IDENTICAL
all-keys-before-guards vs hand-built: wire IDENTICAL
```

It passes today — add the table and it stays passing.

---

# VERIFIED GOOD — each proven, none assumed

**1. The generic-walk gate is real, and there is no per-table codec.**
```
tables generic-walk gate: one walker, byte-identical in 11 generated headers
```
Grepping every generated header for JSON code outside the walker markers returns
**only** the three wrappers per table, each a single `return TableJsonRead/Write(
&value, <Name>TableType(), ... )`. Nothing table-specific. The gate's strictness
holds up: the package name really does live only in the guard and namespace.

**2. Round-trip FROM THE PAGE — 27 texts, every §16.2 row, all identical.**
Written from §16.2 alone, `FromJson`'d, then `Save` compared against `Save` of the
hand-built instance: every scalar width incl. int64 min / uint64 max / -0.0 /
float32 max and min-normal; string AT 32 and one past; base64 at all three padding
lengths, unpadded, and empty; enum by name, `None`, and an unknown name; flags
empty and two-bit; fixed `[3]` under/at/over; bounded `[..4]` under/at/over;
nested table; union each arm, `{}`, absent, and two keys; `bits(6)` at and over;
the guarded group in all three configurations. **0 divergences.**

**3. No allocation on read — instrumented, not inspected.** Counting `operator new`
and, via a DYLD malloc interposer, every libc-internal malloc:
```
RootConfig FromJson (6409 byte text):  operator-new calls during read = 0
WideBlob   FromJson (648053 byte text): operator-new calls during read = 0
READ 480-digit float token             mallocs = 0
WRITE Measure+ToJson: pass 0 mallocs=7, passes 1-3 mallocs=0   (one-time libc init)
```
The walker's whole C-runtime surface is `memcpy memcmp strlen strcmp strncmp
strchr snprintf strtof strtod localeconv` — no `malloc`, `new`, `strdup`, `std::`.
Steady state is allocation-free in **both** directions, which is stronger than
§16.1 claims.

**4. Clean under ASan/UBSan.** The whole tables suite and my entire hostile battery
built `-fsanitize=address,undefined -fno-sanitize-recover=all`: no report. Every
framing break behaves — unterminated string, BOM, comment, bare value, array root,
raw control byte, unknown escape, short `\u`, two texts, trailing garbage all
refuse with `malformed`; 1MB of nesting under an unknown key hits the depth cap and
refuses rather than smashing the stack. The stray-lead-byte class is closed: `0xF8`
followed by five bare continuations rides as seven separate bytes, bounded, no
overrun.

**5. The negative control is genuine in both directions.** Sabotaged: passes.
Built against the honest walker it fails, by hand:
```
FAIL json negative control: a walker with its offsets sabotaged still round-tripped
      Attachment byte-identically. exit=1
```

**6. Not one wire byte moved.**
- `Tables.h`/`Nested.h`/`Wide.h` and all three `*Wire.h`: **byte-identical** to main.
- Committed `generated/` tree: unchanged after regeneration.
- `TablesTable.h` diffed with the walker, the descriptor rows and the wrappers
  removed: every remaining hunk is additive (the `duplicate` field, the two new
  descriptor structs, the three new columns). **No codec line moved.**
- `TestTablesMoveNoProtocolId` and `TestJsonKeyMovesNoWire` green; the latter pins
  the projection, the protocol id, every table-wire field id, and that the id is
  still the hash of the SCHEMA name.

**7. Measure is exact under pretty-print, by construction.** Measure and write are
one code path; the only asymmetry (`overflow`) cannot fire when `buffer == NULL`.
Confirmed empirically on every instance: `measure == write` at exact capacity, one
byte short always -1, a roomy buffer byte-identical to the exact one.

**8. Decision 3 (guards) round-trips.** Guard false with the guarded key present
gives the same wire as a plain guard-false instance; the JSON→JSON leg is stable
(226 → 226 bytes, identical); order-free placement is real.

**9. Decision 5's refusal is by name, and is better founded than the PR states.**
Every variable-length member gets a named comment in the header, and the fixed
members of a pointer unit (`Meta`, `Settings`) still get the text form. The cost
argument is understated: threading a builder through the walk would have to name
`TableRef`/`Builder`/`TableRegion` in every header, which fails `make
tables-zero-cost` outright — not merely the byte-identity gate. And the omission
costs #257 nothing: a packed root (`ships [ShipType]ShipConfig` of fixed tables) is
the fixed class, so `ConfigFromJson` will exist.

**10. Decision 1 is well-founded.** On the table wire, a float where an int is
declared IS `kind_mismatch`; `1.5` is the float shape; the mapping is consistent.
(One wording nit: the §16.2 prose narrows `kind_mismatch` to *"the wrong JSON
type"*, which the fraction row then widens. Add a clause.)

**11. Docs are present-state.** USAGE's new section teaches the surface by example
with no history and no working notes; the reflection section correctly gained
flags and `arms()`. §15's follow-on line exists. Arrays of unions are already a
named follow-on and the compiler refuses them by name — the walker's element-shape
gap for them is unreachable, not a bug.

**12. Provenance clean.** Nine commits, each titled as a present-state statement,
each citing #253, all authored `Rowan Claude <rowan@mas-bandwidth.com>`. The PR
body attributes the §16 draft to Rowan and the six reversals to the owner, and
labels decision 5 as the piece of ruled scope not built.

---

# VERDICT: MERGE WITH CONDITIONS

The architecture is right and the gate that matters is real. Conditions, in order:

1. **F1** — a non-finite conversion result never reaches storage silently.
2. **F2** — accept `2.0` and `1e3` for an integer field when the value is integral.
3. **F3** — validate the JSON number production; non-JSON is `malformed`, as §16.2
   already says.
4. **F4** — reset an array field on entry so "last wins" is true.
5. **F6** — one pinned-text test. It is the cheapest gate in the PR and it closes
   every vocabulary and key spelling at once.
6. **F5** — an owner ruling, then one line of page or one line of code.
7. **F8** — bank the +10.9% and name it in §16.1.
8. **F9/F10/F11/F12** — page and coverage; cheap, and F10 must be handled at rebase
   or the spec of record contradicts itself.

F7 is latent and can ride a later pass.

**HELD regardless** behind #269 (the spec of record — which today carries the
superseded draft of this very section) and behind #265 (the seams the walker's two
SEAM comments name). Nothing here should land before those two settle.

gafferongames:
## Delta review — MERGE

Cold re-review, fresh clone at `384479f`, branch never touched
(`git status --porcelain` empty throughout). Everything below was run.

**F1–F9 and F12 are repaired, and I could not break the repairs.** Both seams
match §16.2 row for row, including the hostile cases I invented rather than the
ones the PR anticipated. One gap remains and it is **not this PR's** — recorded
at the end so it does not get lost.

### The docs claim, corrected

`git diff origin/main -- SPEC-TABLES.md SPEC.md USAGE.md` is **not** empty — but
the reason is good news, not bad. §16 has **landed on main** (spec of record), and
the PR now carries **no doc change at all**:

```
$ git diff --stat origin/main...HEAD -- SPEC-TABLES.md SPEC.md USAGE.md
(empty)
$ git log --oneline HEAD..origin/main
7a24516 SPEC-TABLES: §18.3 states the pairing's second question ... (#274)
```

The branch is one unrelated doc commit behind main; `mergeable: MERGEABLE`. My
earlier **F10 is resolved**: main's §16 is the corrected text, §17 is packing,
§18 is the baseline, and there is no competing copy to reconcile. The review
target has therefore changed — main's §16 is now the contract, and it is a much
stronger one than the branch used to carry. I reviewed against that.

---

# (1) The number table — 45 tokens, 0 failures

Every row asserted against §16.2 **Numbers** as it stands on main.

```
 1 2      -> heading  want=value  ok=1 k=0 c=0 m=0 got=2
 5 2.0    -> heading  want=value  ok=1 k=0 c=0 m=0 got=2
 6 1e3    -> experience want=value ok=1 k=0 c=0 m=0 got=1000
 8 1.5e3  -> experience want=value ok=1 k=0 c=0 m=0 got=1500
11 2.5    -> heading  want=kind_mismatch ok=1 k=1 c=0 m=0
15 -9223372036854775808 -> timestamp want=value ok=1 k=0 c=0 m=0
16 18446744073709551615 -> epoch     want=value ok=1 k=0 c=0 m=0
19 128    -> tilt     want=clamped ok=1 c=1 got=127
24 1e400  -> precision want=kind_mismatch ok=1 k=1 got=0
26 1e39   -> ratings  want=kind_mismatch ok=1 k=1 got=0        (float32 width)
29 1-2    -> heading  want=malformed ok=0 m=1
33 +7 / 34 007 / 35 .5 / 36 3. / 37 1e / 38 1e+ / 40 0x10
41 Infinity / 42 NaN / 43 01 / 44 -   -> all malformed
45 null   -> heading  want=kind_mismatch ok=1 k=1

rows=45  failures=0
```

Integral acceptance is exact (`2`, `2.0`, `1e3`, `1.5e3`, `1.0e2`, `2E0`, `-1e3`);
genuinely fractional in an integer field is `kind_mismatch`; RFC 8259's grammar is
enforced including the four cases the page names *and* the ones it doesn't
(`+7`, `007`, `.5`, `3.`, `1e`, `1e+`, `0x10`, `Infinity`, `NaN`, `01`, `-`).
**No token anywhere produces an infinity.** F1/F2/F3 closed.

*(Nit, not a defect: the 40-digit overflow reports `clamped=2` — saturate then
width-clamp. The page says "clamps and counts" without fixing the count.)*

# (2) F4, F5, F6, F7, F1

```
F4 fixed array dup   dup=1 podium = Gold,None,None        (want Gold,None,None)
F4 fixed float dup   dup=1 ratings = 9.5 0 0 0            (want 9.5 0 0 0)
F4 dup union arm     dup=1 type=Debuff amount=3           (whole value re-established)

F5 lone surrogates   name bytes: 61 ef bf bd 62 ef bf bd 63   (want exactly that)
F5 ill-formed in STORAGE (0xC3 0x28 0xF8 0x80) -> whole emitted text is valid UTF-8: YES
   and it re-reads clean

F1 {"precision": 1e400}  clean=0 measure=226      <- no longer clean, still writable
F1 {"precision": 1e308}  clean=1 measure=231
   (4 texts read clean; every one writable)
```

**F6/F7 together — the sabotage sweep.** I sabotaged three vocabularies
independently in throwaway generated headers and ran the whole suite:

```
=== SABOTAGE: unionarm (case 1: "buff" -> "BUFF") ===
FAIL json pinned text: ToJson does not spell what the page says
tables test: 3 failure(s)

=== SABOTAGE: enum (EnumName(Grade(v ^ 1))) ===
FAIL main.cpp:3697: size == (int64_t) strlen( expected )      <- the pinned text
=== SABOTAGE: flags (FlagNamePerks((int) v + 1)) ===
FAIL RootConfig: ToJsonMeasure refused
FAIL ProfileConfig: ToJsonMeasure refused
FAIL LoadoutConfig: ToJsonMeasure refused
```

This is the finding I care most about, and it is the one that moved furthest. On
the old head the flags sabotage round-tripped **GREEN** while the writer emitted
`["Cloaked","???"]`. Now it is **refused outright** — which is §16.3's
*"refused by the NAME, not by a bound, so a vocabulary that disagrees with its own
extent cannot slip through as a placeholder"*, implemented exactly. F7 closed as a
by-product of the pinned text. The pinned text catches the other two directly. The
whole vocabulary class is now sealed.

# (3) The two seams, against the page's rows

```
?T absent  ToJson:  { "damage": 10, "cooldown": 0.5 }            <- no "gunner" key
?T present all-default ToJson: ... "gunner": { "reaction": 0.2, "tracking": false }

?T key absent              -> absent    ok=1 present=0 k=0
?T key present, object     -> present   ok=1 present=1 k=0
?T key present, {}         -> present   ok=1 present=1 k=0
?T key present, WRONG type -> present   ok=1 present=1 k=1     <- presence IS the key
?T key present, null       -> ABSENT    ok=1 present=0 k=0
?T dup then null: present=0 reaction=0.2 (default restored) dup=1
```

```
[E]T ToJson: { "teams": { "Red": {...}, "Blue": {...}, "Green": {...} }, ... }
   "None" never emitted.   round trip wire 44 vs 44 IDENTICAL

[E]T keyed by variant name        ok=1 unknown=0 Red=5
[E]T "None" key -> unknown        ok=1 unknown=1 Red=4 (default)
[E]T unknown variant name         ok=1 unknown=1 Red=4
[E]T absent key keeps defaults    ok=1 unknown=0 Red=4
[E]T null on a keyed slot         ok=1 k=1                      (my own case)
[E]T duplicate keyed field        ok=1 dup=1 Red=1 Blue=4       (ALL slots reset)
[E]T array where object declared  ok=1 k=1                      (my own case)

seam probe failures = 0
```

Every row of §16.2's `?T` and `[E]T` entries is honoured, including the two the
page states most sharply: *"a key present sets the field present, whatever its
value"* (proven with a kind-mismatched value: `present=1, k=1`) and the keyed
duplicate resetting **every** slot, not just the ones the repeat names.

*One nit:* a repeated slot key **inside** an `[E]T` object is last-wins but
reports `duplicate=0`. The value is right; the counter is silent. §16.2's
duplicate rule reads as being about a table's field keys, so this may be intended
— worth one sentence either way.

# (4) §16 conformance, sentence by sentence — one gap, and it is not yours

I read main's §16 start to finish. Everything is honoured **except one thing**,
and it is a single gap wearing four hats:

> §16.1: *"A VARIABLE-LENGTH table reads through its builder, because that is
> where its storage comes from (§6.5): `SceneFromJson( builder, text, text_bytes,
> &report );`"*

```
$ grep -n "FromJson" build/tables-generated/pointers/GraphTable.h
6975:// no ListNodeFromJson and no ListNodeToJson exist to call.
6987:// no SceneFromJson and no SceneToJson exist to call.
```

The same gap covers three §16.2 rows that only ever occur in the variable class:
`*string`, `*bytes` (§2.5 makes both make their holder variable) and
`pointer *T | object, or null`.

**This is not a regression and blocking the PR does not fix it.** Main today has
*no* text form at all; this PR takes conformance from nothing to everything but
the variable class, and the fixed members of a pointer unit (`Meta`, `Settings`)
**do** get the form. But §15's named follow-ons no longer carry the
variable-class text form line the old draft had, so the spec of record now states
a surface in present tense that nothing implements and nothing tracks. **Land the
PR, then either implement it or give §15 its line back** — the second is one
sentence.

Two more §16 sentences that are unreachable for reasons upstream of this PR:

- §16.2 union row: *"A `table` arm (§2.6) is the same object form."* The compiler
  refuses a table union arm, **identically on main**:
  ```
  Arm.schema:9:11: Payload is a table, not a union payload — a payload is a
  declared `type`; tables nest in tables directly
  ```
  §2.6 is spec'd and unimplemented; §16.2's clause is a consequence, not a cause.
- §16.2 *"compressed floats"* — verified working (they are a ranged float32 to the
  walk): `{"x": 0.337}` round-trips wire-identical, `5.0` clamps to 1 with
  `clamped=1`.

# (5) Gates, independence, tree

```
tables generic-walk gate: one walker, byte-identical in 14 generated headers
json negative control: the round trip changed the wire — red, as required
ok  github.com/mas-bandwidth/schema/v2/internal/check
ok  github.com/mas-bandwidth/schema/v2/compiler
generated/ after regeneration: (empty git status = byte-unchanged)
tables test passed
```

**14 headers**, not 13. My own page-authored battery (27 texts across every §16.2
row, written from the page and never from the code) is still **0 divergences** on
this head, ASan/UBSan clean. Framing still exact: unterminated string, BOM,
comment, bare value, array root, raw control byte, unknown escape, short `\u`,
two texts, trailing garbage all `malformed`; 1MB of nesting under an unknown key
hits the cap and refuses. `null` is `kind_mismatch` for a plain kind.

**Still allocation-free**, including the new seams:
```
RootConfig  FromJson (6409 byte text):   operator-new during read = 0
WideBlob    FromJson (648053 byte text): operator-new during read = 0
KeyedConfig FromJson (?T + [E]T):        operator-new during read = 0
```

§16.5's held-by-test list is delivered item for item — the roster is 22 tests
including `test_json_pinned_text`, `test_json_number_grammar`,
`test_json_no_infinity_reaches_storage`, `test_json_duplicate_arrays_last_wins`,
`test_json_nested_guards`, `test_json_optional_presence`, `test_json_keyed_arrays`
and `test_json_pinned_keyed_and_optional`. **F12 closed** by
`tables/examples/Guarded.schema` (nested + negated branches now in the corpus).

**F8 is now an owner ruling**, not a PR decision — main's §16 preamble states
*"there is no opt-out"* outright. Only the number is unbanked: re-measured on this
head, interleaved, same sitting, median of 9, arm64 clang, empty TU including
`TablesTable.h` — **main 0.0480s → branch 0.0541s, x1.128** (up from x1.109 as the
walker grew with the seams). Worth one line in the round log; not a condition.

**F11 resolved on the spec side** — main §4 now says
*"`duplicate` is the TEXT FORM's counter and the wire never raises it"*, and §15's
port line names the text form, so #266's C# `TableReport` gains it when it ports.

# (6) The UAF at `test/tables/main.cpp:2642` — confirmed pre-existing

Reproduced on **origin/main**, unmodified, under ASan:

```
==15962==ERROR: AddressSanitizer: heap-use-after-free
    #0 test_keyed_variable_oracle() main.cpp:2642
freed by:  TableArenaShutdown ... DepotBuilder::Lock() ... main.cpp:2628
alloc by:  TableArenaGrabSlab ... DepotBuilder::DepotBuilder() main.cpp:2596
```

`git blame` puts it in **#265** (*"Tables: optional fields `?T` and enum-keyed
arrays `[E]T`"*). It is a **test** bug, not a library one: the oracle keeps the
arena pointer `root` and reads `root->spare_present` after `builder.Lock()` has
shut the arena down and freed the slab — by design. Not this PR's, and this PR
does not touch it.

With a one-line local patch to a scratchpad copy (snapshot `spare_present` before
`Lock()`; the branch was never edited), **this PR's full tables suite is clean**:

```
$ c++ -fsanitize=address,undefined -fno-sanitize-recover=all ... main_patched.cpp
$ ./pr_san
tables test passed
exit=0
```

Worth its own one-line PR against main.

---

# VERDICT: MERGE

Every finding I raised is repaired, and repaired at the root rather than at the
symptom — the flags sabotage refusing outright rather than round-tripping green is
the tell. My page-authored oracle, which is the one that does not share the
writer's vocabulary, is green across every §16.2 row including both new seams, and
I could not construct a hostile case that got past it.

Not conditions on this merge, but do not lose them:

1. **§16.1's `SceneFromJson( builder, ... )` and the `*string` / `*bytes` /
   `pointer *T` rows are stated and unimplemented, and §15 no longer tracks it.**
   Either implement or restore the follow-on line. This is the only place the spec
   of record over-promises what ships.
2. One line for the repeated-slot-key duplicate count inside an `[E]T` object.
3. Bank the x1.128 header cost in the round log.
4. Fix the #265 UAF on main separately — it is real and it hides under a
   non-sanitized `make test`.

PR #269, opened 2026-09-02
SPEC-TABLES: the design of record for the tables campaign — spec first (#248)

gafferongames:
Spec only, three files — `SPEC-TABLES.md`, `SPEC.md` §4.2's grammar block, `USAGE.md`. No compiler, emitter, corpus or test change, and no generated byte moves.

**Spec first always** (owner, live). The tables campaign written as one language, landing ahead of the implementation PRs rather than trailing them. Every section is meant to be implementable from the page alone by someone who does not have this repo — a property the cold read tested directly by writing a Go encoder from §3/§3.1/§3.2 and matching the generated `Save` byte for byte on the first attempt.

## The load-bearing change: the keyed array body gets its own wire kind

`[E.Max + 1]T` stays legal and positional (the owner's ruling), so a table field can be edited between the two spellings. They were both kind `14` with different layouts. Measured by the cold read, keyed writer → positional reader:

```
Load returns true.  unknown=0 kind_mismatch=0 clamped=0 malformed=0
slots = {262144, 65536, -1157431296}
```

Every counter zero, every value garbage — a silent, data-corrupting edit, and the **third** member of a silent class §4.1 claimed had exactly two. **The keyed body is now kind `16`.** The edit becomes a reported `kind_mismatch`, §4.1's claim becomes true, and §18.2's compile-time refusal is the belt to that braces. §3's skip rules gain `16` (still three rules), and §3.2 states ascending-ordinal write order with readers key-driven.

## Sections added

- **`## The performance ladder` / `## What allocates, and what never does`** (the opening) — the owner's ladder and the language-keyed union carve-out.
- **§2.5 `*bytes`/`*string`**, **§2.6 union arms as tables**, **§4.1 the silent class**, **§13.2/§13.3 rulings in the owner's words**, **§16 the text form**, **§17 packing**, **§18 the baseline**.

## The cold read's ten must-fixes — all closed

| # | fix | where |
|---|---|---|
| D1 | positional↔keyed is a silent corrupting edit | §2.4, §3, §3.2, §4, §4.1, §11, §18.2 — closed by kind `16` |
| D2 | byte-identity claim overbroad | §2.3, §2.5, §3, §3.1 + USAGE — scoped to non-default content, empty-end asymmetry named |
| D3 | §17 not implementable (compiler cannot run generated code) | §17.1 pack's own IR-driven engine + the goldens; §16 preamble states rules, not one implementation |
| D4 | report contradiction on duplicates | `duplicate` counter in §4 and §16.2 |
| D5 | guard rule order-dependent, drops data | §16.2 — write elides, read is order-free and infers nothing |
| J1 | a fraction was "malformed" | §16.2 — `kind_mismatch`; `malformed` reserved for not-JSON |
| J4, J6 | writer refusals and text shape unstated | §16.3 (the `-1` refusals), §16.1 (pretty-print as contract) |
| J7 | §8 missing columns the walk needs | reset hook, text key, `bits(N)` implied range, union tag offset/width; flags bound corrected to **highest declared bit index** |
| O1 | variable-class narrowing | page keeps `SceneFromJson(builder, …)`; §15 does **not** list it as a follow-on; open-for-owner item 1 |
| D6, D7 | §17's licensing sentence, §11's element-kind row, history narration | §13.3 quotes the ruling's second clause; §17 preamble; §11; §12 and §13 de-narrated |

## Owner rulings folded in this pass

- **`[E]T` slot 0** (#255): "the 0th entry in the array is not valid, and is 'null', so you really only store n-1 entries"; and "accessing the 0th entry is an error/assert". Storage stays `E.Max + 1` (owner: "memory is cheap so you could just leave the array as-is, and just put an error/assert on index by zero"), slot 0 exists and is never valid, indexing it is a `static_assert` where expressible and an assert otherwise. `None` never rides; a stored key `0` is **malformed**, not unknown, because `0` is the reserved id no name can fold to. `"None"` is an unknown key in the text form; no `None.json` in pack; §8 marks slot 0 invalid. Unbiased indexing throughout — `ships[ShipType::Bomber]`.

## The delta read's five (`7e54c0d`) — all closed

| # | finding | fix |
|---|---|---|
| N1 | §3's elision list omitted guarded-false fields while §16.2 cited the rule as if §3 carried it — an encoder from §3 alone wrote a field the reference does not | a bullet in §3's elision list; §16.2 now genuinely defers |
| N2 | §4 never said an array's ELEMENT kind is checked, though the code checks it — a faithful reader would decode `[3]int32` into `[3]float32` and report nothing | §3's `14`/`16` rows state element kind is identity, not only framing; §4's kind-mismatch bullet covers it |
| N3 | adding an `if` guard is report-clean and round-trip-lossy, and §4 named guards nowhere | §4 gains the bullet; §4.1 states why it is **not** a third silent edit — see below |
| N4 | two `§16.3` refs stale after §16.3 was inserted; "field names" vs keys | → `§16.4`; §16.2 opens on field KEYS |
| N5 | §2.4 put the `static_assert` in `operator[]( E )`, where a runtime key cannot carry one | storage is a keyed-array type over `T[E.Max + 1]`; `operator[]( E )` asserts, `Slot<Key>()` static_asserts; a two-line snippet shows both |

**N3, the one that needed a call: stated, not made a refusal.** Adding or removing a guard reads correctly in both directions with every counter zero; the loss is on write-back, when the new reader elides the field on its next save. So §4.1's enumeration honestly stays at two — a silent edit is one that *decodes* wrong — and the page says so explicitly rather than leaving a reader to work out why the class is adjacent but not a member: *"A person whose file came back wrong needs the two above; a person whose tool rewrote a file needs this one."* Making it a §18 refusal would be a **new policy row** that no ruling covers, and a guard is a legitimate edit that a blanket refusal would block, so it goes to the owner instead (open item 11) rather than onto the page.

## From the #265 delta read (`817a775`)

Three things §2.4 owed a porter:

- **The type-body carve-out.** `per_team [Team]int32` inside a `type` generates a plain `int32_t per_team[4]` — no wrapper, no keyed accessor, no `None` guard — because the type wire is positional and there is no key to check. Only the TABLE-wire encoding is keyed, and the keyed storage type is generated for table bodies alone. The previous text would have had a porter emit the wrapper for a `type`.
- **A key enum is in the table closure's vocabulary**, even when no field has its type. It rides by variant name like any other, so `| max = K` headroom and variant-id collisions are refused for it — and the diagnostic names the KEYING FIELD as the reaching edge, because that edge is the only reason the rule applies and a person looking at the enum alone would not see it. Stated in §2.4, §5 and §11.
- **`NDEBUG` compiles the runtime guard out.** "Asserted otherwise" was doing too much work: the assert in `operator[]( E )` is absent in exactly the build a shipped game runs. `Slot<Key>()` is the form that holds in release, costs nothing, and cannot be disabled — preferred wherever the key is written literally, which is most places.

## Reconciliations

1. **Two branches both claimed §16** → §16 text form, §17 pack, §18 baseline. §1–§15 frozen (cited from code comments and issues).
2. **#264's `L == 4` union-arm discriminator is NOT here** — a wire-v2 construct; v1 has no node table, so a table arm rides inline in both classes, framed as a `type` arm.
3. **`*bytes` framing pinned to `bytes(N)`'s exactly** (kind `14` u8 array; `*string` kind `12`) — joins the wire-identity family, no new kind.
4. **Backend status stays `C++`** — the cold read judged this right; #266 carries its own status edit.
5. **§2.2's zero-cost gate rescoped to the pointer machinery** (#265 decision 6). The cold read verified the gate empirically: pointer-free unit with `?T` and `[E]T` passes, gaining exactly the five language columns, with `is_pointer`/`variable` absent — so §8's "only the two POINTER columns are conditional" is literally true.
6. **26 suffixes**, matching `tableGeneratedVerbs` — verified identical, same order.
7. **#253's §16.4 folded into §8**; **§16.6's replicant claim superseded by #257**; **guard-presence closed** per #253's own comment.

### Implementation-taught — the page adopted what a branch learned

Each of these is the page catching up to a build, per the gate's own provision:

- **`origin/tables-json-walk` (#267), six places.** A fraction is `kind_mismatch` not `malformed` (malformed stops the whole read — one bad value must not); `duplicate` as a real counter on `TableReport`; the guard rule as shipped (order-free read, eliding write) because the drafted rule silently lost fields whose guard key came later in the object; the writer's `-1` refusals as §5's save-failure rule carried into text; pretty-printing stated because measure==write and `unpack`→`pack` byte-stability are uncheckable without it; and §8's four missing columns plus the flags off-by-one (`enum_max` is the highest bit **index**, so a walker written from the old page looped one short).
- **From the #267 review**, also adopted: JSON has one number type, so `2.0` and `1e3` read into an integer field (the class of existing text §16.3 exists to meet); RFC 8259 number-token validation, so `1-2` is `malformed` rather than a clamp to int16 MAX; last-wins applies to a whole value including arrays; the writer always emits valid JSON/UTF-8; and the four page gaps a from-the-page author had to guess (how `None` writes, what `null` does per kind, base64 padding, empty flags).
- **`origin/tables-baseline` (#263), the repairs.** §18 now matches `DefaultTokenPolicy` row for row, including the referent rule (`RuleRefs` per vocabulary — table bodies by field id, enums and unions by name, flags by bit position), the always-refuse referent drop, the declaration-rename member matching with its warn, and the repairable-baseline behaviour of `--update`. §18.1 states that an optional's presence is **recorded and judged on nothing**, so nobody later adds a `RuleFixed` row and breaks the `T` ↔ `?T` move.
- **`origin/optional-and-enum-keyed` (#265): no divergences.** The cold read found full byte-level agreement with §3/§3.1/§3.2 on both scratch units. Its other seven decisions were checked against the rulings and kept verbatim.

## Open for the owner

1. **The text form's variable-class narrowing.** #267 ships no `FromJson` for a variable-length table — refusal by absence. The ruling was "one table at a time" with **no class restriction**, so the page states the design (`SceneFromJson( builder, … )`) and §15 does not list it as a follow-on. Either the walker builds it or the ruling narrows; the page must not decide it. Already on #248 as item (e). *(Consistency note either way: §16.2 carries rows for `*string`, `*bytes` and pointer `*T`, and a `*bytes` field alone makes its holder variable-length — so under the narrowing those three rows describe a table with no `FromJson` to call, and §17 needs a matching sentence about whether `pack` is fixed-class only.)*
2. **Wire v2's five decisions** (#251 / draft #264) — untouched; the page is v1 throughout.
3. **`*string`** is not ruled; "byte buffer as primitive" was.
4. **`schema pack` carries a second implementation of the wire** — a Go text reader and table-wire encoder inside the compiler, golden-locked to the generated backends. It follows from the codegen philosophy (interpreters stay in tooling) and it is the only way `pack` can exist in a Go compiler, but it is a second implementation of §3 in the tree and every port inherits the goldens obligation.
5. **`duplicate` on `TableReport`** — a new column on the wire's own cross-language report struct, for an event the wire never raises. Nine backends will each carry it; cheaper nodded now than ported nine times and changed.
6. **The text form's writer refusals** (`-1` for an unnameable enum value, an unnamed flags bit, an invalid union tag, a non-finite float).
7. **The lone-surrogate / non-UTF-8 decision.** The review named two options: refuse on read, or drop "UTF-8" from the row and call the form byte-transparent. The page takes a third that keeps both halves — storage stays byte-transparent, and the WRITER substitutes `U+FFFD` for any ill-formed sequence so the emitted text is always valid JSON. Mine, not ruled.
8. **The baseline's new refusal classes** — the positional↔keyed spelling change (from D1), and the flags rename-in-place refusal (from #263's text, not the owner's words directly; it follows the same reasoning as the reorder refusal).
9. **§18's `--reason` flag spelling** is #263's; the ruling was that the update is never bare.
10. **Should adding or removing a GUARD be a baseline finding?** (N3.) It is the one edit that reads clean and loses data on write-back, which is precisely the save-game cycle §18 exists for — but it is a legitimate everyday edit, so a REFUSE would be heavy-handed and a WARN is the proportionate shape. Either way it is a new `DefaultTokenPolicy` row that #263 does not implement, so the page names the hazard (§4, §4.1) and stops there. If you want it, #263 gains the row and the obligation below becomes real.
11. **The text-form walk costs +10.9% header compile time**, unconditional, fixed class included. Measured by the #267 review, interleaved, median of 9, arm64 clang, empty TU including `TablesTable.h`: main 0.0466s → branch 0.0517s. The zero-cost gate is **not** weakened (grep list byte-unchanged and green). But this repo rules header cost on measured evidence — the union-vs-variant precedent, *"0.17 -> 0.27 is not trivial for me"* — so the number is banked here rather than buried. The page states that the walk rides in every table header with no opt-out; whether it should stay unconditional is yours.

## Rebase obligations recorded (not this PR's)

- **#263**: render the `key` token (§18.2's keyed refusal and D1's baseline half both need it) and a presence token; renumber its §16 → §18 in `internal/baseline/diff.go` and `cmd/schema/main.go`. **Also: its own §16 REFUSE list row still reads "swapped for one whose identities do not ride" while the prose below it already states the STAND IN rule** — the list and the prose contradict each other on the branch, and "identities ride" is exactly the weaker test the delta review replaced. One-line repair, and it is the row a reader checks first. *(Conditional, on open item 10: a guard-change policy row.)*
- **#267**: its §16 **replaces** this one wholesale rather than merging; §16.4/§16.5/§16.6 renumber (this page folds "what reflection gains" into §8).
- **#266**: "the 23 name-first suffixes" → 26; the `TableReset` decision is unchanged.
- **#265**: the `[E]T` slot-0 ruling and kind `16` land there.

CI: green (docs-only; nothing in the build reads these files). Head is `817a775`.

### DIFF
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -15,24 +15,70 @@ arena, read through a region and a root pointer, never held by value).
 A fixed-size table pays for none of the variable-length machinery: that
 property is a gate, held by test, not an intention (§2.2).
 
-Tables are schema's declarations for **data that crosses builds**: config
-files, asset archives, tool output, editor state — and just as much,
-**messages**: bytes produced by one build of one program and consumed by
-another build of another, whether the trip is a disk file read years later
-or a datagram read milliseconds later by a peer that deployed last week.
-The hardcoded wire (`type`, SPEC.md) is the same-build contract, guarded by
-the protocol id: same-or-refuse. Packets are its loudest user, not its
-definition — any data whose writer and reader share a schema build belongs
-there. The table wire is the opposite contract: **any reader reads any
-data**, and the differences are reported, never fatal.
+Tables are schema's declarations for **data that outlives the build that
+wrote it**: config files, asset archives, tool output, editor state,
+**save games** — a file written by a build nobody has any more and read by
+a build its writer never saw, years apart — and just as much, **messages**:
+bytes produced by one build of one program and consumed by another build of
+another, whether the trip is a disk file read years later or a datagram
+read milliseconds later by a peer that deployed last week. Save games and
+tool output are one case, not two, and the whole of this document serves
+it. The hardcoded wire (`type`, SPEC.md) is the same-build contract,
+guarded by the protocol id: same-or-refuse. Packets are its loudest user,
+not its definition — any data whose writer and reader share a schema build
+belongs there. The table wire is the opposite contract: **any reader reads
+any data**, and the differences are reported, never fatal.
 
 The two contracts never mix. Flatbuffers- and protobuf-class evolution
 ideas apply here, to tables — they do not apply to `type`, whose wire is
 hardcoded under the protocol id, and nothing in this document changes that.
 
+## The performance ladder
+
+**Tables are LESS performant than types**, and that is the trade they exist
+to make: types buy speed with a hardcoded, same-build wire; tables buy
+tolerance, reflection and evolution with a neutral byte wire, length
+prefixes and — in one narrow case — an allocation. A chosen cost is
+licensed. Unexplained slowness is still a defect, in every rung.
+
+- **A type is a raw struct.** Its generated storage IS the struct a person
+  would have written, and its codec is measured against a hand-written
+  serialize of that struct. The fastest-correct mission applies to types
+  without qualification, and the standing ledger holds them to it.
+- **A fixed-size table is a raw struct with a tolerant byte codec around
+  it**, and is expected to run as close to its equivalent type as the
+  neutral wire allows. That is a bench obligation, not an intention: a
+  fixed table sits beside its equivalent type on the ledger, and the gap
+  is explained or closed. The zero-cost gate (§2.2) holds the storage
+  half of it; the bench leg holds the codec half.
+- **The variable-length class pays for exactly what it buys** — pointers,
+  an arena on the building side, a region on the reading side, a
+  reference indirection per edge. This is where "less performant" lives,
+  and it is deliberate.
+
+## What allocates, and what never does
+
+- **A type never allocates**, in any backend, on any path. This rung does
+  not move.
+- **A fixed-size table with no union allocates nowhere, in any backend.**
+  It is a struct; measure, save and load work over caller-owned buffers.
+- **A fixed-size table WITH a union may allocate for the arm, and only
+  where the LANGUAGE needs it.** The carve-out is keyed on the backend
+  language, never on the table's class: a backend with no native union
+  builds a pseudo-union — one reference per arm, the set arm non-null,
+  allocated on read and write — and a backend with a native union (C++)
+  allocates nothing for the same declaration.
+- **The variable-length class allocates by nature**, and assuming
+  otherwise is foolish: the arena on build, the region on load. C++ keeps
+  the caller-owned form — `LoadMeasure` sizes the region and the caller
+  supplies it, `Builder` grows storage the caller owns — which is
+  allocation with the caller holding the pointer, not its absence. Other
+  backends allocate inside their runtime and say so. No port contorts
+  itself toward zero allocation for a variable-length table.
+
 **Backend status: C++.** Every other backend refuses a unit that declares
 tables, by name, with this document cited. Per-language backends are named
-follow-ons.
+follow-ons (§15).
 
 ## 1. Purpose
 
@@ -72,7 +118,7 @@ table ShipConfig
 A table body is a type body — the field grammar of SPEC §4.2, hosted by
 `table`: bare and ranged integers, `bits(N)`, `bool`, floats and
 compressed floats, enums, flags, strings, bytes, bounded arrays, unions,
-`if` branches, and declared types as field groups. Two additions:
+`if` branches, and declared types as field groups. Six additions:
 
 - **Tables nest.** A named table is a field type (above); nesting is by
   value, and a bounded array of tables is a collection. A table may not
@@ -83,6 +129,14 @@ compressed floats, enums, flags, strings, bytes, bounded arrays, unions,
 - **Tables point** (§2.1). `next *Node` declares a POINTER to a table.
   Recursion THROUGH a pointer edge is legal and expected — the by-value
   cycle rule exempts pointer edges, because a pointer carries no size.
+- **Fields may be OPTIONAL** (§2.3). `settings ?GunnerSettings` is present
+  or absent by value, with no pointer and no change of mode.
+- **An array may be ENUM-KEYED** (§2.4). `ships [ShipType]ShipConfig` has
+  exactly one slot per variant, indexed by the variant.
+- **A blob is a node** (§2.5). `data *bytes` declares an unbounded byte
+  buffer at exactly its used size, and `*string` is its sibling.
+- **A union arm may be a table** (§2.6), which is what makes an evolvable
+  message set expressible.
 - **`was` — the rename attribute** (§5).
 
 ### 2.1 Pointers
@@ -104,22 +158,16 @@ table Scene
 **A pointer is for recursion, sharing and size — not for optionality.**
 Every field on this wire is already optional: absence is the reader's
 default (§4), so nothing has to be pointed at to be left out. The spelling
-for an OPTIONAL SECTION is the guarded field:
+for an OPTIONAL SECTION is `?T` (§2.3):
 
 ```
 table Scene
 {
-    has_settings bool
-    if has_settings
-    {
-        settings Settings   // an optional section: off the wire when the guard is false
-    }
+    settings ?Settings   // an optional section: off the wire when it is absent
 }
 ```
 
-The guard is a field like any other, the section rides only when it is
-true, and a reader whose writer left it out gets the declared defaults —
-with no pointer, no allocation and no change of mode. Reach for `*` when
+No pointer, no allocation and no change of mode. Reach for `*` when
 the structure genuinely needs it: a table that refers to ITSELF (a chain, a
 tree), a large subtree you would rather not carry by value, or one node
 several parents name (sharing is a named follow-on, §15 — v1 writes a
@@ -130,8 +178,10 @@ choice is a real one.
 The spelling is C's, deliberately: it reads as what it is. The rules,
 each refused by name (§11):
 
-- **A pointer targets a `table`, and only a `table`.** `*SomeType`,
-  `*SomeEnum`, `*SomeUnion` are refused: value-semantics data has no
+- **A pointer targets a `table`, or one of the two buffer primitives.**
+  `*Node` names a declared table; `*bytes` and `*string` name an unbounded
+  buffer at its used size (§2.5). Everything else is refused — `*SomeType`,
+  `*SomeEnum`, `*SomeUnion` — because value-semantics data has no
   independent identity to point at. Nest it by value instead.
 - **A pointer is declared inside a table body, and nowhere else.** A
   `type` body refuses one: types remain value semantics, and that is the
@@ -151,22 +201,32 @@ pointers in the struct. Its meaning depends on the form it sits in
 The compiler works out which class a table belongs to; the schema never
 says. The rule is a least-fixed-point over BY-VALUE edges:
 
-- A table is **VARIABLE-LENGTH** if it declares a pointer, or if anything
-  it nests by value — directly, or as a bounded array — is
-  variable-length.
+- A table is **VARIABLE-LENGTH** if it declares a pointer — including
+  `*bytes` and `*string` (§2.5) — or if anything it nests by value is
+  variable-length. "Nests by value" reaches through every by-value edge
+  there is: a plain nested table, an element of a bounded array, an
+  element of an enum-keyed array, a member of a guarded (`if`) group, an
+  optional's value (§2.3), and a UNION ARM that is a table (§2.6).
 - Every other table is **FIXED-SIZE**.
 
 Pointer edges do not propagate the mode: a table that is merely POINTED
 AT stays fixed-size if it holds no pointer of its own. It gains an
 allocation and a resolution entry, and nothing else.
 
-**A fixed-size table pays nothing for any of this**, and that is a gate,
-not a hope: for a unit whose tables are all fixed-size, the generated
-output is byte-for-byte what it was before pointers existed — no
-builder, no arena, no reference type, no lifecycle surface, not one extra
-descriptor column, not one extra branch in a codec, not one extra
-`#include`. The build fails if a single symbol of the pointer machinery
-appears in a pointer-free unit's generated header.
+**A fixed-size table pays nothing for the VARIABLE-LENGTH machinery**, and
+that is a gate, not a hope: in a unit whose tables are all fixed-size the
+generated output carries no builder, no arena, no reference type, no
+lifecycle surface, not one descriptor column that exists to describe a
+pointer, not one branch in a codec that exists to follow one, and not one
+extra `#include`. The build fails if a single symbol of the pointer
+machinery appears in a pointer-free unit's generated header.
+
+The gate is scoped to that machinery and to nothing else. A LANGUAGE
+FEATURE the fixed class itself can use — an optional's presence companion,
+an enum-keyed array's key columns, the wire id a variant rides under — is
+emitted in every unit, whatever its mode, because a fixed-size table can
+declare it and a tool walking a fixed-size table has to find it. The gate
+asks "did the pointer world leak in?", never "did the descriptor grow?".
 
 The assumption behind the split, stated so nobody quietly designs against
 it: size and mode correlate in practice. Value-only tables are assumed
@@ -182,6 +242,184 @@ follow-on. **Extents have no wire ceiling**: lengths and counts ride as u32
 (§3), so the only limit is the language's own — a string, bytes or array
 extent lives in int32 storage (SPEC §4.3), and that cap is what a too-large
 extent is refused against.
+
+### 2.3 Optional fields: `settings ?GunnerSettings`
+
+```
+table ShipConfig
+{
+    health   float32 = 100.0
+    settings ?GunnerSettings   // present or absent, by value
+    tier     ?int32            // scalars too: present, and the value
+}
+```
+
+`?T` declares a field that is PRESENT or ABSENT. Its storage is the value
+plus a generated `<field>_present` bool, so the holder stays FIXED-SIZE:
+an optional costs one bool, one branch and no allocation, and a table of
+optionals is still a plain struct (§2.2's gate is untouched).
+
+**PRESENCE decides whether the field rides, never content.** An absent
+optional is not written; a present one is ALWAYS written, even when its
+value is entirely default — the pointer's rule (§3.1), for the same
+reason: otherwise "absent" and "present with nothing to say" would be one
+value on the wire. On load, a field that rode sets `present = true`; a
+field that did not leaves `present = false` with the value at its declared
+defaults.
+
+The framing is exactly the non-optional field's, which makes `?T`, `*T`
+and a plain `T` nesting **one field on the wire**: a schema may move a
+field among the three and no byte moves, in either direction. The corpus
+holds all six directions and the byte-identity of the three writers.
+
+`?` applies to a nested table, a nested `type`, an enum, a `flags` mask
+and any scalar. It is refused, by name, on:
+
+- **a `type` body** — a type's wire is positional and every field always
+  rides, so there is no absence to express;
+- **a pointer** — a pointer is already optional, and null rides exactly
+  as an absent optional does;
+- **a union** — its `None` arm IS the absence, and an empty union already
+  elides;
+- **an array, a string or `bytes`** — each already carries a count or
+  length whose zero is emptiness, and a second presence bit beside it
+  would be two answers to one question. Wrap it in a table and make that
+  optional; the general form is a named follow-on (§15).
+- **a specified default** — presence is the only default an optional has.
+
+### 2.4 Enum-keyed arrays: `ships [ShipType]ShipConfig`
+
+```
+enum ShipType { Fighter, Bomber, Scout }
+
+table Fleet
+{
+    ships [ShipType]ShipConfig   // one slot per variant, indexed by the variant
+}
+```
+
+An array bound that NAMES A DECLARED ENUM is an enum-keyed array. The two
+spellings never overlap, because an enum is declared: `[Name]` naming a
+constant is the fixed array it has always been, and `[Name]` naming an
+enum is keyed.
+
+- **Storage** is `T ships[ShipType.Max + 1]`, indexed directly by the enum
+  value — slot 0 is `None`'s. Every slot exists, so there is no count
+  companion, and `ships[int( ShipType::Bomber )]` just works.
+- **On the TABLE wire the slots ride by NAME** (§3.2): the array body
+  carries `(variant id, element)` pairs, so inserting, removing or
+  reordering variants leaves every surviving slot in its own home. This is
+  the whole point of the construct: an ordinal-indexed array is the last
+  positional vocabulary the table wire had, and it failed silently.
+- **On the TYPE wire the spelling IS `[E.Max + 1]T`** — positional,
+  bitpacked, same-build, the protocol id moving exactly as that spelling
+  moves it. One construct, two wires, the same rule every kind follows.
+  The two spellings project identically and share one protocol id, and
+  that is held by test.
+- **Fixed-size when `T` is**, so the zero-cost gate holds.
+
+Refused by name: a bound naming a `flags` declaration (a mask holds any
+set of bits at once, so it names no single slot); a bounded keyed array,
+`[..E]` or `[A..E]` (a keyed array is COMPLETE by construction); and an
+element that is a pointer, as for any array (§15).
+
+### 2.5 Byte buffers: `data *bytes`
+
+```
+table Asset
+{
+    format  AssetFormat
+    data    *bytes        // an unbounded blob, stored at exactly its used size
+    label   *string       // the sibling: a string node at its used length
+}
+```
+
+`bytes(N)` is fixed inline storage of N bytes in every instance. `*bytes`
+is a POINTER to a byte buffer that has no declared bound and occupies
+exactly the size it is given. The distinction is what makes large content
+expressible: in a table of a million nodes, a `bytes(65536)` field costs
+64 KB per node whether it is used or not, because the region packs storage
+verbatim, while a `*bytes` field costs the four-byte reference plus what
+each node actually holds.
+
+- **Storage** is the four-byte relocatable reference (§2.1) beside a u32
+  used length. Declaring one makes the holder VARIABLE-LENGTH (§2.2), like
+  any other pointer.
+- **In the arena**, `builder.AllocBytes( n )` returns a node of exactly `n`
+  bytes; `builder.AllocString( n )` is the sibling.
+- **In a region** it is packed at its used length, its reference
+  self-relative like every other (§6.3).
+- **On the wire it is framed exactly as `bytes(N)` is** — kind `14`, an
+  array of element kind `6` (u8) at its used count — and `*string` is
+  framed exactly as `string(N)` is, kind `12`. No new kind, no new skip
+  rule, and one useful consequence: `bytes(N)` and `*bytes` are ONE FIELD
+  ON THE WIRE, as `T`, `?T` and `*T` are (§3.1), so a field that outgrows
+  its inline bound moves to a blob and no byte changes. A null blob is
+  absent, exactly as a null pointer is.
+- **In the cooked form**, `Open`'s walk bounds the blob by its length as it
+  bounds every node, so a pointer into a mapped file IS the asset: no copy,
+  no parse (§7).
+
+Two sentences that must travel together: **a tolerant wire load COPIES the
+blob** — a gigabyte on the wire path is a gigabyte read — and **the cooked
+path is the zero-copy one.** Both are true and neither is the whole story.
+
+schema does not interpret the bytes. A format tag beside the blob is an
+ordinary field the user declares. A pattern falls out and is worth naming
+though it is no construct: a sub-document, or a rarely needed arm, can be
+stored as its own wire bytes inside a `*bytes` and decoded only when
+something asks for it — which keeps a very large file's resident memory
+proportional to what is touched rather than to what exists.
+
+Refused by name: `*bytes` or `*string` outside a table body; a specified
+default on one (a fresh reference is null); an array of them, as for any
+pointer (§15); `?` on one, because null already IS absence (§2.3).
+
+### 2.6 Union arms may be tables
+
+Inside a TABLE closure a union's arm may name a `table`, not only a
+declared `type`:
+
+```
+table OpenDocument  { path string(256), line uint32 }
+table SaveDocument  { path string(256), force bool   }
+
+union ToolBody
+{
+    open OpenDocument
+    save SaveDocument
+}
+
+table ToolMessage
+{
+    sequence uint32
+    body     ToolBody
+}
+```
+
+That is an evolvable MESSAGE SET by construction, and it is the case this
+document's opening promises. Arms ride under their own name hashes (§5),
+so adding a message is adding an arm; a reader that lacks the arm reads
+the union as empty, skips the body by its length and counts `unknown`
+(§4); arms may be removed and reordered freely. A message may therefore
+carry a nested table, a collection, or anything else a table body holds —
+which is what a message set of documents needs and what a `type`-only
+payload could never express.
+
+- **A union declared for the TYPE wire keeps refusing table arms.** Types
+  are value semantics and their wire is positional; a table arm is a
+  table-closure construct and is refused by name outside one.
+- **Mode derivation runs through arms** (§2.2): an arm that is a
+  variable-length table makes the union's holder variable-length. A union
+  of fixed-size table arms leaves its holder fixed-size, and the zero-cost
+  gate holds.
+- **On the wire an arm's body is framed exactly as a nested table's is**
+  (§3, kind `15`): `arm id (u16)`, then `L`, then `L` bytes of table body.
+  A `type` arm and a `table` arm are the same three bytes of framing, so a
+  reader needs no new rule and an arm may change between the two forms
+  without the framing moving.
+- **A backend without a native union may allocate for the arm** (the
+  ladder, above): the carve-out is the language's, not the table's.
 
 ## 3. The wire
 
@@ -210,11 +448,22 @@ schema's codebase:
   | `12` string | `L`, then `L` bytes. No terminator; no encoding imposed |
   | `13` table | `L`, then `L` bytes of table body (fields, then the u16 zero terminator) |
   | `14` array | `L`, then the array body: `element kind (u8)`, `N`, then the elements |
-  | `15` union | `arm id (u16)`, and when it is not 0, `L` then `L` bytes of table body |
+  | `14` array, ENUM-KEYED | `L`, then the array body: `element kind (u8)`, `N` = the number of SLOTS PRESENT, then N pairs of `variant id (u16)`, `L (u32)`, `L` bytes of element (§3.2) |
+  | `15` union | `arm id (u16)`, and when it is not 0, `L` then `L` bytes of arm body |
 
   **The union carries NO outer length** — its `arm id` sits where the other
   three containers put theirs, and the length that follows frames the arm
   alone. It is the one payload whose framing a skipper has to know (below).
+  An arm's body is a table body whether the arm names a declared `type` or
+  a `table` (§2.6): fields, then the u16 zero terminator, the same bytes a
+  kind `13` payload carries.
+
+  **Three spellings that add no row.** A `?T` optional field is framed
+  exactly as the non-optional `T` (§2.3); `*T` is framed exactly as a
+  by-value `T` (§3.1); and `*bytes` and `*string` are framed exactly as
+  `bytes(N)` and `string(N)` (§2.5). Each family is ONE FIELD ON THE WIRE
+  under several declaration spellings, which is what lets a schema move a
+  field among them without moving a byte.
   - **Array elements.** For a scalar element kind the elements sit back to
     back at that kind's fixed width. For element kind `13` (table) each
     element is preceded by its own `L`. `bytes(N)` rides as an array of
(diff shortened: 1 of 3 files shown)


gafferongames:
# Cold read — #269, SPEC-TABLES as the design of record

Read whole and in order: CLAUDE.md, this PR body, every comment on #248, then #253 #255 #256 #257 #258 #259 #260 #262 with comments, then SPEC-TABLES.md on the branch start to finish, then the SPEC.md §4.2 and USAGE.md diffs. Then three checks: faithfulness against the owner's actual words, implementable-from-the-page against the branches that implement each construct, and internal consistency.

**Method for the implementable check** (it matters for what follows): built `bin/schema` from `origin/optional-and-enum-keyed`, generated C++ for two scratch units exercising `?T` present / `?T` absent / `?T` present-at-its-default / `[E]T` with table elements / `[E]T` with scalar elements / `bytes(N)` / an enum / a flags mask / a union with a `type` arm / an all-default nested table — then wrote a Go encoder **from §2.3, §2.5, §3, §3.1 and §3.2 alone**, reading no emitter source, and byte-compared it against the generated `Save`. Both units matched byte for byte on the first attempt. Scratch deleted.

Verdict at the bottom: **LAND-AFTER-FIXES**.

---

## DEFECTS

### D1 — §4.1's "exactly two, there is no third" is false, and §2.4 invites the third

§4.1 says the point of the subsection is that "a person reading a save game that came back wrong needs to know there is no third". There is a third, and it is worse than the two named.

`[E.Max + 1]T` stays legal and positional — §2.4 says so, §13.3 quotes the owner making that "the user's choice". So a field can be edited between the keyed and the positional spelling, and §2.4 encourages reading them as interchangeable: *"One construct, two wires, the same rule every kind follows."* On the **type** wire they are one field. On the **table** wire they are two different encodings, and the page never says so.

Measured, `[Team]int32` writer → `[Team.Max + 1]int32` reader, same field name, same id, same kind:

```
v2 (keyed) bytes: 3f440e2300000004030000000000040000000100000003bb040000000900000030f604000000070000000000
positional reader on keyed bytes: ok=1 slots={262144,65536,-1157431296}
                                  unknown=0 kind_mismatch=0 clamped=0 malformed=0
```

Load returns true, every counter is zero, every value is garbage. That is a **silent, data-corrupting edit** — the exact class §4.1 exists to enumerate. The other direction (positional writer → keyed reader) loses the whole field and reports only `malformed=1`, with nothing saying why.

Fix, three places:
- **§2.4** gains the sentence: the two spellings are one field on the TYPE wire and *different encodings on the table wire*, so changing spelling on a table field is a wire break, not a refactor.
- **§4.1** names it as the third member of the silent class, and its answer is the same machinery answer the other two get.
- **§18.2**'s refuse list gains "an array changed between the positional and the keyed spelling" — the baseline catches it for free once #263 renders the `key` token (the policy row is already `RuleFixed`, and `diffTokens`' second loop refuses an added fixed token).

### D2 — the `T`/`?T`/`*T` byte-identity claim is overbroad, and so is `bytes(N)`/`*bytes`

Three places state a universal:

- §2.3: *"a schema may move a field among the three and no byte moves, in either direction. The corpus holds all six directions and the byte-identity of the three writers."*
- §3.1: *"A schema may change any of them into any other and no byte moves … the byte-identity of the three writers."*
- USAGE: *"move a field among the three and no byte changes, in either direction."*

False for all-default content, and the page says why two paragraphs later. §3: *"an all-default nested table [is] not written at all"*; then *"a present optional, a non-null pointer and a non-null blob are ALWAYS written, even when the value is entirely default."* So a by-value `T` at its defaults elides and a present `?T` at its defaults rides. Measured on the scratch unit: a `?Inner` present-and-all-default rode as `b3190d0200000000` (9 bytes); the identical `plain Inner`, all-default, wrote **0 bytes**.

The corpus test the page cites is scoped accordingly — `test_optional_three_way_evolution` on `origin/optional-and-enum-keyed` sets `link.value = 66` before comparing the three writers. It proves the particular; the page states the universal.

Same shape for §2.5's *"`bytes(N)` and `*bytes` are ONE FIELD ON THE WIRE … a field that outgrows its inline bound moves to a blob and no byte changes"*: an empty `bytes(N)` elides, a non-null zero-length `*bytes` rides (§3 says exactly that).

Fix: state each family as **one framing with one named asymmetry** — content decides elision for the by-value spellings, presence decides for the pointer-shaped ones — and scope the byte-identity claim, and the corpus obligation it hands the implementer, to non-default content. As written, the page hands an implementer a test that cannot pass.

### D3 — §17.1 is not implementable as written: the compiler cannot run generated code

§17.1: *"Each file's content is read by §16's walk."* But §16's walk is generated C++ — §16.1: *"the generated header carries the descriptors and one walker over them"*, gated on *"the walker's source is the same bytes in every generated header"* — and `schema pack` is a Go command inside the compiler. There is no path from the one to the other, so §17 as written cannot be built.

Rowan's decision, labelled, on #257 (per the codegen philosophy — generate code for language independence and zero allocation, **interpreters stay in tooling**): `schema pack` carries its own engine, an IR-driven text-form reader and table-wire encoder in Go inside the compiler, implementing §16's rules and §3's wire from the same IR the emitters use, held to the generated C++ by goldens.

What §17 must say:
- pack reads each file **under §16's rules**, with its own tooling-side engine over the IR — not by calling generated code;
- the goldens are what make the two one wire: for every corpus tree, `schema pack` bytes == C++ `Save` of the hand-built instance, and `schema pack` → C++ `Load` → C++ `ToJson` → `schema unpack` is byte-stable.

And the related half: **§16 must not read as though the generated C++ walker is the only implementation of the text form.** Today §16.1's "Both are ONE generic walk" and the byte-identical-walker gate read as the *definition* of the text form rather than as C++'s way of holding it. §16 states RULES; the Go engine and every backend implement those rules, and goldens make them one. One sentence in §16.1 fixes it, and it also pre-answers what the C# port (#262/#266) owes.

### D4 — §16.2 contradicts itself on the report, and the fix is a fifth counter

*"Duplicate keys: last wins, counted"* stands beside *"The report is the wire's report (§4): `unknown`, `kind_mismatch`, `clamped`, `malformed`."* None of those four honestly holds a repeat, so the page tells an implementer to count into nothing.

`origin/tables-json-walk` resolved it and the resolution is right: a fifth counter, `duplicate`, and it sits on **`TableReport` itself** (verified in the emitter, `int32_t duplicate = 0;` beside the four), with the note that the wire never raises it — a body carrying an id twice is legal input whose last occurrence wins, silently, by §3. **§4 must gain the counter too, not just §16**, because §4 is where the page enumerates the report.

### D5 — §16.2's guard rule is order-dependent and drops data

*"a guarded group's fields are placed only when the guard reads true"* makes the read depend on key order inside a JSON object, which is nobody's contract: a text whose guard key follows the fields it guards silently loses them. The landed rule is sound and the page should carry it: **write** elides a guarded-false field, exactly as the wire does; **read** places every key it can name and infers nothing; a field placed with its guard false is elided again on the way out, so the wire never sees it.

### D6 — §17 does the thing the ruling it cites says not to do, and the quotation is truncated at that clause

§13.3 quotes the text-form ruling as *"the general idea of reading JSON file into a table can move into schema… that's not opinionated. but only one table at a time?"* The ruling on #253 continues: *"So we separate packing from table reading JSON, **the actual linking up by enums has to be replicant only since that is opinionated**."*

§17's directory rule is that linking — one `<Variant>.json` per enum-keyed slot. What licenses it is #257's argument: with `[E]T` the link IS the declaration, so it stopped being an opinion. That argument appears in neither §13.3 nor §17. A reader who checks the cited ruling finds it saying the opposite of §17. Fix: §17 carries the one sentence, and §13.3 either quotes the clause or marks the elision.

### D7 — two smaller ones

- **§11's baseline refusal list** omits "an array's element kind changed", which §18.2 has and `elem: RuleFixed` implements. USAGE's "What refuses" list omits both that and the key-enum swap.
- **Present-state:** §12's *"the constructs that used to force a pointer no longer do"* is history narration, new in this PR, in a document CLAUDE.md holds to present state only. (§13's *"The fenced/append-only draft this document previously carried is dead"* is pre-existing on main — same class, worth taking out while the file is open.)

---

## DIVERGENCES — page vs each implementation branch

### `origin/optional-and-enum-keyed` (#265) — none

Full byte-level agreement with §3/§3.1/§3.2 as the page states them, on both scratch units, including: `fold16(fnv1a32(name))` with the 0→1 rebound; the `3 + 4 + 5 + n` array framing; the union's *missing outer length*; **"the length rides for EVERY element kind, scalars included"** in a keyed body; presence-not-content elision; and the elision set (default scalar, empty string, empty array, all-default fixed array, empty union, all-default nested table). The page-side items above (D1, D2) are the page's, not this branch's.

### `origin/tables-json-walk` (#267) — six, and the page moves on five

The page's §16 is #253's **draft** with amendments, not what #267 built. #267's PR body names six places the build disagreed; each is a place the page must move, because the build's reasoning is sound and this is the "implementation taught better" case the gate provides for.

| # | page says | code does | who moves |
|---|---|---|---|
| J1 | a fraction in an integer field is **malformed** | `kind_mismatch`, skipped and counted | **PAGE** — `malformed` stops the whole read and keeps a partial instance, the analogue of §4's framing damage; one bad value must not do that |
| J2 | "counted" with no counter to count into | `report.duplicate`, on `TableReport` | **PAGE** (see D4; §4 and §16 both) |
| J3 | guarded fields placed only when the guard reads true | write elides, read is order-free and infers nothing | **PAGE** (see D5) |
| J4 | *silent* on what the writer does with a value it cannot name | `ToJsonMeasure` returns **-1** for an enum value no variant names, a flags bit no variant names, an invalid union tag, and a non-finite float | **PAGE gains it** — it is §5's save-failure rule carried across, and a caller now has to know `-1` has two meanings |
| J5 | `SceneFromJson( builder, text, bytes, &report )` for a variable-length table | **no `FromJson` is emitted at all** — refusal by absence, named as a §15 follow-on | **OWNER** — see O1 |
| J6 | *silent* on the text's shape | pretty-prints: one entry per line, two-space indent | **PAGE gains it** — load-bearing, because §16.1 promises measure==write byte for byte and §17.2 promises `unpack` → `pack` byte-stable, and neither is checkable while the shape is unstated |

Two more from reading the branch beside its PR body:

- **J7 — §8 is missing columns §16's walk cannot work without.** #267 emits, and §8 does not name: the per-type **reset hook** (a generic walker filling a value has to establish the defaults an absent field takes and holds no type to spell), each field's **text key**, and a `bits(N)` field's **implied `[0, 2^N − 1]` range**; and §8's union entry should carry the **tag's offset and width**, not only the arm payload descriptors. §8 presents itself as the complete unconditional set, so these belong in it. Separately, §8 says a flags field carries *"the bit count"* while the code carries `enum_max` = the highest declared **bit index** — a walker written from the page gets the loop bound wrong by one.
- **J8 — numbering.** #267 numbers §16.4 "What reflection gains" / §16.5 "Held by test" / §16.6 "What this is not", and cross-references §16.6 from §11 and §15. This page folds "what reflection gains" into §8 (reconciliation 7) and lands §16.4 = Held by test, §16.5 = What this is not. #267 renumbers on rebase — worth naming so it is not lost.

**One consistency trap either way on J5.** §16.2 carries rows for `*string`, `*bytes` and pointer `*T`. A `*bytes` field makes its holder VARIABLE-LENGTH (§2.5), so if variable tables have no text form, all three rows describe a table that has no `FromJson` to call. Whichever way the owner rules, that has to be closed — and §17 needs the matching sentence about whether `schema pack` is fixed-class only.

### `origin/tables-baseline` (#263) — the lists match; two implementation obligations

Checked §18.2 row by row against `DefaultTokenPolicy` and the three vocabulary diffs, and they agree exactly:

- refuse: `kind`, `elem`, `default`, `key` = `RuleFixed` (change, add and remove); `diffFlags` refuses insert, remove, reorder **and rename-in-place**;
- warn: `bound`, `size` = `RuleShrink` (shrink warns, grow passes); `diffEnums` / `diffUnions` warn on a removed variant or arm;
- pass: `type`, `array`, `was` = `RulePass`, which is what makes `bounded ↔ fixed` and a `was` rename silent, and what lets `bytes(N)` → `*bytes` drop its `size` token without a finding.

On the rename-in-place refusal (PR body open item 5): the code refuses it with exactly the reasoning the body gives — *nothing distinguishes a rename from a new meaning on a spent bit, so the author says which.* Page and code agree; it is the owner who has not ruled, which the body already says. Correct as listed.

- **B1 — IMPLEMENTATION MOVES.** §18.2 refuses "an enum-keyed array's key enum swapped", but the branch's renderer does not emit a `key` token yet (`diff.go`: *"the renderer needs `add("key", ...)` the day `[E]T` lands"*). After #269 lands, #265 lands before #263, so #263 owes that line or §18's refusal is a page rule with no gate behind it. Same line is what closes D1.
- **B2 — IMPLEMENTATION MOVES.** §18.1 says the projection records "presence of an optional"; the branch has no such token. It must render one once `?T` exists, and the page should say the fact is **recorded and judged on nothing** (which is what §18.2's PASS list implies), so nobody adds a `RuleFixed` row for it later and breaks the `T` ↔ `?T` move.
- **B3 — bookkeeping.** #263 numbers itself §16 in `internal/baseline/diff.go`'s header and in `cmd/schema/main.go`'s doc comment. Both become §18 on rebase.

### `origin/cs-tables` (#266) — "Backend status: C++" is acceptable

Reconciliation 4 is the right call: the page states backends by what is in the tree at the moment it lands, §15 names the C# backend with its scope and reason, and #266 carries its own status edit. Judged **fine**. One carry-over: #266's reasoning for `TableReset` being verb-first cites "the 23 name-first suffixes"; the count is now 26 and the decision is unchanged (`TableReset` still claims nothing from a declaration's name) — its prose needs the number, not the decision, updated on rebase.

---

## OPEN FOR THE OWNER — additions the PR body missed

The body's six are right. Five more:

1. **The text form was narrowed away from a ruled scope, and the page is silent.** #267 cut VARIABLE-LENGTH tables from the text form — no `FromJson` emitted, refusal by absence — judging it not cheap on top of the arena. The ruling on #253 was "one table at a time" with **no class restriction**, and #248 already carries this as "FOR GLENN'S RETURN (e)": *"it is a narrowing — confirm or send it back."* The page states the opposite (a builder-taking `FromJson`, with a concrete signature), and none of the six open items mentions it. **The design of record cannot land silent on a narrowing of a ruled scope.** This is the one addition I would not land without.
2. **`schema pack` carries a second implementation of the wire.** D3's fix means a Go text reader and table-wire encoder inside the compiler, golden-locked to the generated C++. It is the right call and it is labelled as Rowan's on #257, but it is a second implementation of §3 in the tree and every port inherits the goldens obligation. One line, beside open item 4's envelope question.
3. **`duplicate` is a new column on `TableReport`**, the wire's own report struct, for an event the wire never raises. Small, but it is public surface that nine backends will each carry. Cheaper to have it nodded now than ported nine times and then changed.
4. **The text form's writer refusals** (`-1` for an unnameable enum value, an unnamed flags bit, an invalid union tag, a non-finite float) extend §5's save-failure rule into the text. Sound, and Rowan's, and on neither the page nor the open list.
5. **If D1 is closed in the baseline**, refusing the positional↔keyed spelling change is a new refusal class. It fits the owner's *"if the compiler helped you be careful — that would be good"* squarely, but it is a new row and the owner should see it named.

---

## VERIFIED GOOD

- **§3, §3.1 and §3.2 are implementable from the page alone.** A Go encoder written from those sections and nothing else produced bytes identical to the generated C++ `Save`, first attempt, on both scratch units — covering `?T` present, `?T` absent, `?T` present-at-default (scalar, flags and table), `[E]T` with table elements, `[E]T` with scalar elements, `bytes(N)` at a used length, an enum riding as its name hash, a flags mask as u64, a union with a `type` arm, and an all-default nesting eliding. This is the strongest thing in the read: the section does what the PR body promises — implementable by someone who does not have this repo.
- **The 26-suffix list is exactly `tableGeneratedVerbs`** on `origin/tables-json-walk` — same 26, same order. Reconciliation 6 is correct.
- **§2.2's rescoping is honest, and the gate still holds what the page says.** Ran the `tables-zero-cost` grep pattern against a pointer-free unit carrying `?T` and `[E]T`: passes. The descriptor grew by exactly the five language columns #265 decision 6 adds (`optional`, `present_offset`, `key_type_name`, `key_name`, `key_id`); `is_pointer` and `variable` are **absent** from the pointer-free unit and present in the pointer corpus. So §8's *"only the two POINTER columns are conditional"* is literally true, and the gate does ask "did the pointer world leak in?" rather than "did the descriptor grow?". Reconciliation 5 is the right call and the page's wording matches the test.
- **§18's refuse/warn/pass lists match `DefaultTokenPolicy` row for row** (detail above).
- **Numbering and cross-references.** §16 text / §17 pack / §18 baseline per reconciliation 1; every `§N` and `§N.M` in the page resolves to a heading that exists; SPEC.md references are all prefixed `SPEC §` and none collide with the local numbering.
- **§13.2 and §13.3 quotations are faithful.** Checked one by one against #248, #255, #256, #257, #258, #259, #260. The `[E]T` entry splices two separate utterances but marks it with an ellipsis. The performance-ladder entries are quoted exactly, and the one place the page *interprets* — "which this document reads as a bench obligation" — says so on the page. D6 is the one exception.
- **SPEC.md §4.2's grammar edits are minimal and right**: `?` as a prefix on `Type` **before** the array bound (which is what makes `?[..N]T` refusable by name), `*` taking `bytes`/`string`, the enum-bound note on `Bound`, the table-arm note on `UnionVariant`. Nothing else in the grammar moved.
- **The doctrine in the opening is the owner's**, ladder and allocation rungs both, and the one carve-out is keyed on the LANGUAGE rather than the table's class, which is the correction the owner made explicitly.

---

## VERDICT: LAND-AFTER-FIXES

The wire sections are the best part of the page and they hold up under a from-scratch reimplementation. Everything below is on the page, not on the branches, except where marked.

**Must fix before landing:**

1. D1 — §2.4, §4.1 and §18.2: the positional↔keyed spelling change is a silent, corrupting edit; §4.1's "there is no third" is false as it stands.
2. D2 — §2.3, §3.1, §2.5 and USAGE: scope the byte-identity claim; name the elision asymmetry.
3. D3 — §17.1 and §16.1: `schema pack`'s own IR-driven engine and the goldens that make it one wire with the generated backends; §16 states rules, not one implementation.
4. D4 — §4 and §16.2: the `duplicate` counter.
5. D5 — §16.2: the guard rule (write elides, read is order-free).
6. J1 — §16.2: a fraction is `kind_mismatch`, not `malformed`.
7. J4, J6 — §16: the writer's refusals, and pretty-printing as the one formatting opinion.
8. J7 — §8: the reset hook, the text key, the `bits(N)` implied range, the union tag's offset and width; and "bit count" → highest declared bit index.
9. O1 — the PR body gains a seventh open item: the text form's variable-class narrowing.
10. D6, D7 — §17's licensing sentence, §11's element-kind row, and the two history sentences.

**Not blocking, carry as rebase obligations:** B1 and B2 on #263; B3 and J8's renumbering; #266's "23" → "26".

Nothing here is a DO-NOT-LAND. The page is the right shape and lands first correctly; six of the ten fixes are the page catching up to what #267 already learned, which is precisely the mechanism the gate describes.

gafferongames:
# Delta read — #269 at 2541a52

Fresh clone, `gh pr checkout 269`. Re-ran every check from the cold read plus the eight the delta asks for. **All ten must-fixes are closed**, and the `[E]T` slot ruling is faithfully carried. Five new findings, all single-sentence page edits; three of them are implementable-from-the-page failures, so:

**Verdict: LAND-AFTER-FIXES** — N1, N2, N3, N4, N5 below. Nothing structural, nothing that touches the landing order.

---

## 1. The from-the-page encoder, re-run against the new §3/§3.2 — **MATCH**

`origin/optional-and-enum-keyed` has moved to a8a48d4 ("the owner's None-slot ruling, the keyed body's own wire kind, and the review's F1–F5"), so a real byte comparison was possible. Built its `bin/schema`, generated a unit carrying `?T` present / `?T` absent / `?T` present-at-its-default / `[E]T` **of tables** / `[E]T` **of scalars** / `bytes(N)` / enum / flags / a union with a `type` arm / an all-default nesting — then wrote a fresh Go encoder from §2.3, §2.4, §2.5, §3, §3.1 and §3.2 **alone**, reading no emitter source.

```
45e40407000000b0f30e080000000603000000aabbcc72d207e7a3c52f090200000000000000
b3190d0200000000009d160400000000d8f410230000000d0200000003710900000009d60405
00000000004c6d0900000009d6042a000000000049eb100f0000000401000000866d04000000
630000003ae30fe6ea09000000e0320a0000204000000000
```

Byte-identical, first attempt. **No guess was required at any byte.** Every decision was licensed by a sentence: kind `16` and its `L`-then-body framing, `N` = slots present, the `(variant id, L, element)` pair with the length riding for scalar elements too, per-slot elision, "a `None` key NEVER RIDES", "slots are written in ascending variant ordinal", and the whole unchanged elision set. I set `Gamma` before `Alpha` in the driver deliberately — the page's ascending-ordinal sentence is what tells you the writer re-orders, and it does.

The zero-cost gate pattern still passes on that header (`TableKeyed`, kind 16, `?T`, and no pointer machinery), and its 11 static asserts hold.

## 2. §4.1's "exactly two" — TRUE, with two gaps that have to close first

The three claims the delta rests on, measured against the revised branch:

```
keyed(16) bytes:       7739100f0000000401000000866d04000000630000000000
  16 -> 14 reader: ok=1 arr={0,0,0} unknown=0 kind_mismatch=1 clamped=0 malformed=0
positional(14) bytes:  77390e1100000004030000000000000063000000000000000000
  14 -> 16 reader: ok=1 slots={0,0} unknown=0 kind_mismatch=1 clamped=0 malformed=0
  key 0 patched:   ok=1 slots={0,0} unknown=0 kind_mismatch=0 malformed=1
```

Both directions report `kind_mismatch`; a stored key of `0` is `malformed`. Exactly what §3.2 and §4 say. D1 is closed at the wire, and §4.1 now states honestly *how* the would-be-third was made reportable rather than pretending it never existed.

I then enumerated every edit class the page can express and checked each lands somewhere. Reported or safe: field add / remove / reorder / rename-with-`was` / rename-without; field kind change; array bound; string and bytes capacity; enum variant add, remove, reorder, rename; union arm the same; `T`↔`?T`↔`*T`; `bytes(N)`↔`*bytes`; `[E]T`↔`[E.Max+1]T`; a keyed array's key enum swapped (old keys unnameable → `unknown`); enum or flags storage width; `bits(N)` width; ranged-integer bounds; `type`↔`table` nesting and a union arm changing between the two; a `json` key; a declaration or package rename. Silent: the two named. **Two classes have no home on the page**, and both are N-findings below: an array's ELEMENT kind (N2) and a guard added or removed (N3).

## 3. The `T`/`?T`/`*T` and `bytes(N)`/`*bytes` asymmetry — CORRECT in all four places

§2.3 ("one framing with three spellings … over non-default content"), §3 (the "What differs inside a family is ELISION, and only at the empty end" paragraph), §3.1 ("only the byte-identity claim is scoped … the byte-identity of the three writers over non-default content"), §2.5 ("no byte changes **for any non-empty value**"), and USAGE in all three of its sites. The corpus obligation is scoped with the claim, which was the sharp half of the original defect — the page no longer hands an implementer a test that cannot pass. The direction analysis is right too: an elided field reads as absent, null or the declared default, each of which is correct, so nothing misdecodes at the empty end.

## 4. §16 vs `origin/tables-json-walk` — every rule now matches or is listed

| rule | page | shipped code | state |
|---|---|---|---|
| fraction in an integer field | `kind_mismatch`; integral-VALUED numbers (`2.0`, `1e3`) accepted; a non-number token (`1-2`, `1.2.3`) is `malformed`, RFC 8259 grammar | same (the read-path repair, #248 item g) | **matched** |
| `duplicate` | §4 and §16.2, on `TableReport`, "the wire never raises it" | `int32_t duplicate = 0;` beside the four | **matched**, and open item 5 |
| guards | write elides, read order-free, re-elided on the way out | same | **matched** |
| writer refusals | §16.3, `-1` for an unnameable enum value, an unnamed flags bit, a bad union tag, a non-finite float; the two meanings of `-1` spelled out | same | **matched**, and open item 6 |
| pretty-print | §16.1, "part of the contract", one entry per line, two-space indent, with the reason (measure==write, `unpack`→`pack`) | `line(depth)`: newline + two spaces per level | **matched** |
| `null` | `kind_mismatch` except `?T` (absent) and pointer (null) | — | page-first, one of the four gaps |
| valid UTF-8 out, `U+FFFD` substitution | §16.3 | — | page-first, **open item 7** |
| variable-length tables | page KEEPS `SceneFromJson( builder, … )` and does **not** list it as a follow-on | code emits none | **page wins, code moves** — correctly left to the owner as open item 1 |

That last row is the right call and the right handling: the ruling was "one table at a time" with no class restriction, so the page states the design and the PR body says plainly that either the walker builds it or the ruling narrows, and that the page must not decide it. The consistency trap I flagged (`*bytes` / `*string` / `*T` rows describing a table with no `FromJson`) is carried in the same item, with the matching `pack` question.

## 5. §18 vs `origin/tables-baseline` at ddc6b2a — matches the repaired policy

The blocker repair did land (86b7370 "close the rename hole, judge referents, make `--update` the remedy it advertises"), and §18 tracks it row for row:

- refuse — `kind`, `elem`, `default`, `key` all `RuleFixed`; `flags-position` `RuleFixed`; a referent dropped or swapped without its identities riding (`enum`/`flags`/`union`/`type`/`payload` = `RuleRefs`);
- warn — `bound`/`size` `RuleShrink`, `enum-variant`/`union-arm`/`member` `RuleLoss`, and the paired-or-unpaired declaration rename;
- pass — `array`, `was`, and presence (no row, so judged on nothing).

§18.1's new "**Presence is RECORDED and judged on nothing**" paragraph and §18.3's three identity rules are exactly what the code implements. The keyed↔positional refusal rides on the `key` token appearing or vanishing, which catches both directions — the page could name that mechanism in a clause, but it is not wrong without it.

`schema tables-baseline --update --reason "…" <unit>` still matches `cmd/schema/main.go`. The `key` renderer remains the one implementation obligation, and the PR body records it.

## 6. §17 and §16.1 neutrality — closed

§17.1 states the own-engine decision with its reason (the compiler is Go, the walk is the target language's, building the bridge would make packing depend on a C++ toolchain) and the three goldens that make two implementations one wire, plus the inheritance clause for every future backend. §16's "**This section states RULES, not one implementation**" paragraph does the other half, and §16.5's gate is correctly narrowed to "within a backend that holds the form as a single walker". §17's new "Why this is not the opinion the text form's ruling kept out" paragraph closes the faithfulness gap I raised about the truncated #253 ruling — the argument is now on the page where a reader checking the quote will find it.

## 7. Present-state, numbering, the flags bound

- Both history sentences are gone: §12 is rewritten in present tense and §13's "the fenced/append-only draft … is dead" is deleted.
- Numbering §16 / §17 / §18 holds; every `§N.M` resolves to a real heading except the two in N4; `§4.2` and `§4.3` are all prefixed `SPEC §`.
- **The flags bound correction is right**: §8 now says "bounded by **the highest declared BIT INDEX** — not a count, so a walker loops `[0, enum_max]` inclusive", which is what the code carries. §8 also gained the reset hook, the text key, the `bits(N)` implied range, the tag's offset and width with each arm's offset, and the slot-0-invalid mark from the ruling. J7 fully closed.

## 8. The open-for-owner list — complete

The ten cover everything I flagged (my O1–O5 are items 1, 4, 5, 6 and 8), and two are self-reported additions I had not raised: the `U+FFFD` writer substitution as a third option between the review's two (item 7), and the **+10.9% header compile time** for the unconditional walk, banked with its measurement against the union-vs-variant precedent rather than buried (item 10). The rebase obligations are recorded for #263, #265, #266 and #267. **No omission found.** N1–N3 below are page fixes rather than owner questions — unless someone proposes making a guard change *reportable*, which would be a new construct and would then belong on the list.

---

## NEW DEFECTS

### N1 — §3's elision list omits guarded fields, and §16.2 cites the rule as if it were there

§16.2: *"**Writing** elides a field whose guard is false, exactly as the wire does (§3)."* §3's elision bullet names a defaulted field, an empty string or array, an all-default fixed array, an empty union and an all-default nested table. **Guards are not in it**, and the word "guard" does not appear in §3 at all.

Measured — a guarded-false field with a value in storage:

```
unguarded g=0,x=5:    8bad04050000000000
  re-saved by guarded: 0000
```

The wire writer does elide it. So an encoder written from §3 alone writes a field the reference implementation does not, and §16 forward-references a rule its own referent does not carry. One bullet in §3's elision list fixes both.

### N2 — §4 never says an array's ELEMENT kind is checked, and the implementation checks it

```
[3]int32 bytes:         77390e1100000004030000000100000002000000030000000000
  elem i32 -> f32 reader: ok=1 arr={0,0,0} unknown=0 kind_mismatch=1 clamped=0 malformed=0
```

The code is right. The page is not: §4's kind-mismatch bullet is about *the field's* kind, and the one array-shaped case it names is the `16`-against-`14` one. §3's kind `14` and `16` rows present the element kind as part of the body's framing, never as part of identity. **A reader written faithfully from the page would decode `[3]int32` bytes into a `[3]float32` field and report nothing** — which is the same silent-corruption class as the original D1, one level down, and it would make §4.1's claim false again. Corroboration that it matters: §18.2 refuses an element-kind change and `elem` is `RuleFixed`.

Fix: one clause in §4's kind-mismatch bullet ("…and an array whose ELEMENT kind differs is the same event"), and a word in §3's `14`/`16` rows saying the element kind is identity, not only framing.

### N3 — adding a guard is report-clean and round-trip-lossy, and §4 does not name guards at all

Same probe, read side:

```
-> guarded reader (g false): ok=1 g=0 x=5 unknown=0 kind_mismatch=0 malformed=0
   re-saved by guarded:      0000
```

A build that adds `if g { x }` around an existing field reads an old file correctly — `x = 5`, every counter zero — and then writes the value away on the next save. For the save-game case §18 is built around (load, edit, store), that is silent loss with a clean report. §4 has no guard bullet and §4.1 does not mention the class.

It is arguably not a member of the silent class as §4.1 defines it — the *read* is correct and the loss is on write-back — and that distinction is exactly what the page should say, in one line, rather than leave to the reader. §4.1's own standard is "a person reading a save game that came back wrong needs to know"; a person whose tool round-tripped the file needs to know this one.

### N4 — two cross-references to §16.3 are stale

Inserting "16.3 What the writer refuses" pushed the `json` key to §16.4, and two references did not follow:

- §8, the text-key paragraph: *"the `json = "…"` attribute's value, else the field's own name (§16.3)"* → §16.4;
- §16.2, the Numbers paragraph: *"meeting an existing text is what §16.3 exists for"* → §16.4.

(Nit in the same area: §16.2 opens *"an object whose keys are field names"* while §16.4 makes keys a separate thing from names — "field keys" is the accurate word, and the branch's own §16 used it.)

### N5 — §2.4 puts the compile-time refusal somewhere it cannot live

§2.4: *"reaching it is an ERROR — a compile-time refusal where the language can express one (a `static_assert` on a constant index in C++), an assert otherwise — **through the generated keyed type's `operator[]( E )`**."*

`operator[]( E )` takes a runtime value, so it can only assert. The shipped shape puts the two in different members:

```cpp
T & operator[]( E key ) { assert( key != E::None ); ... }
template <E Key> T & Slot() { static_assert( Key != E::None, "..." ); ... }
```

A port reading §2.4 would try to put the static_assert in `operator[]` and fail. Related, one sentence up: §2.4 says storage **is** `T ships[ShipType.Max + 1]`, while the shipped storage is a keyed-array type wrapping that array — which is precisely what gives the accessors a home. Both are one edit: storage is a generated keyed-array type over `T[E.Max + 1]`, its `operator[]( E )` asserts, and a compile-time accessor beside it refuses a constant `None` outright. USAGE already says it correctly ("a compile-time refusal on a constant index, an assert otherwise") without misattributing it.

---

## VERIFIED CLOSED from the cold read

D1 (kind `16`, both directions reported, key `0` malformed, §4.1 rewritten and honest about the mechanism) · D2 (all four sites plus USAGE, with the corpus obligation scoped) · D3 (§17.1's engine and goldens, §16's rules-not-implementation paragraph) · D4 (`duplicate` in §4 and §16.2 with the wire-never-raises-it note) · D5 (order-free guards in the text form) · D6 (§17's licensing paragraph) · D7 (§11's element kind, USAGE's lists, both history sentences removed) · J1 (numbers, and better than asked — the integral-value rule) · J4 (§16.3) · J6 (pretty-print as contract) · J7 (§8, including the flags bound correction) · O1–O5 (open items 1, 4, 5, 6, 8). The `[E]T` slot ruling is faithful to #255's three comments, including the owner's "memory is cheap so you could just leave the array as-is" closing the storage question in favour of `E.Max + 1`.

## The fix list

1. **N1** — a guarded-false field joins §3's elision list.
2. **N2** — §4's kind-mismatch bullet covers an array's element kind; §3's `14`/`16` rows say the element kind is identity.
3. **N3** — §4 (and a line in §4.1) names adding or removing a guard, and says why the clean read is not a silent-class membership while the write-back loss still deserves the warning.
4. **N4** — two `§16.3` → `§16.4`; "field names" → "field keys" in §16.2's opening.
5. **N5** — §2.4's storage and accessor sentence.

All five are page-only and none disturbs the structure, the numbering or the landing order. With them in, this is a LAND.

PR #272, opened 2026-09-02
schema pack / unpack: the directory tree that mirrors a root table, an IR-driven Go engine golden-locked to C++ (#257)

### LINKED ISSUE #257: Packing moves into schema: a directory tree that mirrors a root table becomes one table instance — `schema pack`, un-opinionated by construction

gafferongames:
Closes #257. **Rebased onto `main` @ `56d2a3b`** (force-pushed) — #269's page, #265's `?T`/`[E]T`, #273/#274/#279, #263's baseline, #267's JSON walk, #266's C# tables backend, #282's keyed-duplicate rule, #264 and **#278's sanitized tables leg** are all in `main`, so this stacks on nothing.

**Second cold read: MERGE-WITH-CONDITIONS. All four conditions and both nits are in, plus the coverage rows it listed.**

- The `json = "key"` hunks this branch carried verbatim are **gone**: they are `main`'s now, and this PR touches **neither `ir/` nor `internal/check/` nor `internal/codegen/` nor `generated/`**.
- The kind vocabulary is `main`'s: the engine reads `ir.TableKind*`, `ir.TableScalarKind`, `ir.TableFieldKind`, `ir.TableElemKind`. Two copies of that mapping would be two wires.
- `internal/tablenames` is main's registry of names the *generated runtimes* define; this PR emits no generated code, so it registers nothing and changes nothing there.
- `SPEC-TABLES.md`, `SPEC.md` and `USAGE.md` are `main`'s **wholesale**; only the four sentences §17 still lacked are added on top (listed below), and the diff removes no line of main's prose.
- The Makefile's `test` target is shared with #266: both sets of steps and prerequisites ride, C# first.
- `tables/examples/tables.baseline` is main's (#280's regeneration included) with one entry appended by `--update --reason` for the tables `Pack.schema` adds — new declarations only, so nothing already written moves.

## What lands

`schema pack --root T --out F <dir>` assembles ONE instance of a root table from a directory tree that MIRRORS it and writes **the root's wire bytes and nothing else**; `schema unpack` is the inverse. The engine is §17.1's: the packer carries its own implementation of §16's rules and §3's wire, inside the compiler, over the same IR the emitters use, because the compiler is a Go program and cannot run the code it emits.

`internal/tabletext` (instance model, §16 reader and writer) · `internal/tablewire` (§3 encoder and decoder) · `internal/tablepack` (the directory rule, §17.4's refusals) · `compiler.Pack` / `Unpack` / `UnpackOneFile` / `TableReport` and the two CLI verbs. `cmd/schema` still imports only `compiler` and `ir`.

## The second read's conditions

| | condition | repair |
|---|---|---|
| **G1** | float32 double-rounded — Go read through `float64`, C++ uses `strtof` | one correctly-rounded conversion at the field's OWN width: `ParseFloat(token, 32)` when single. Corpus rows for the upper band (both signs), the midpoint itself, and the smallest subnormal (both signs) |
| **G2** | `unpack` pruned only inside its own shape, so switching shapes left a tree `pack` refuses | the prune covers the ROOT'S WHOLE SHAPE — both `<field>` spellings **and** `<Root>.json` — so either form written over the other packs back clean. Fixtured in both directions |
| **G3** | a duplicate whose repeat is REJECTED wiped the first value | **re-establishment happens on PLACEMENT, not on seeing the key**: the unconditional reset is gone, and each kind re-establishes exactly where the reference does — array and keyed object before placing, nested table and union arm at their own entry, `bytes` before decoding, and a SCALAR not at all. Rows for a scalar, a float, a nested table, an array element and a keyed slot, plus the placeable and wrong-shape controls |
| **G4** | §17 promised a byte-stability a legal instance does not have | §17.3 and §17.5 carry §16.3's carve-out: the U+FFFD substitution can push a string past its bound, that lap is COUNTED rather than byte-identical (and §17.4 makes it a nonzero exit), and the fixed point is reached in one lap. Row: `str-illformed-at-bound` |
| **nit** | `pack` left a `.bin` behind when it exited 1 | the report gate runs BEFORE the write, and the write goes through a temporary renamed into place — a refused run leaves the directory as it found it |
| **nit** | the drivers' post-`Load` check ignored `report.duplicate` | both `hostile_main.cpp` and `pack_main.cpp` compare it |

**One more divergence found while mirroring that loop, and fixed:** §16.2's *presence of the KEY is the presence* means a `?T` given a value the walk will not place is still **PRESENT** — the reference sets it on the mismatch path too. Go only set it on the placed path. Row: `optional-wrong-shape`.

### #282's rule, flipped
A KEYED OBJECT'S KEYS ARE KEYS: a variant named twice is a duplicate, counted on the **resolved slot** and **before the shape check**, so a repeat the walk then refuses is still a repeat. A key naming no slot — an unknown variant, or `"None"` — is `unknown` **each time** and never a duplicate. Four rows: the repeated slot, the repeated unknown, the repeated `"None"`, and the mixed case (`unknown 2, duplicate 1`).

### Coverage rows the read listed
Trailing commas · nesting past the depth cap · a duplicate **array**, **table** and **union** key · the `json = "key"` rename addressed by its declared name (`unknown`) · base64 unpadded and empty.

**The two-sided gate now reads `pack hostile-value gate: 67 trees agree on report and wire across both engines`** — every one of those rows checked against the generated `FromJson` for report *and* bytes, not just against my own expectation.

### Sanitized twins, following #278

#278 stood the tables leg up under ASan+UBSan. Both of this PR's C++ drivers get the same treatment for the same reason: each reads a file size off disk, sizes its own buffers from it, and compares byte ranges whose lengths came out of a file or a manifest — bounds and lifetime questions `-Werror` cannot see. `PACK_INCLUDES` / `PACK_CXXFLAGS` are shared between each build and its twin so the two can never drift into covering different code, and `-fno-sanitize-recover=all` makes it a gate rather than a printout.

```
pack golden: schema pack == Save and schema unpack == ToJson for both roots      (plain)
pack golden: schema pack == Save and schema unpack == ToJson for both roots      (asan+ubsan)
pack hostile-value gate: 67 trees agree on report and wire across both engines   (plain)
pack hostile-value gate: 67 trees agree on report and wire across both engines   (asan+ubsan)
```

**Clean — nothing to report.** Worth noting what that covers: the sanitized run drives the generated `FromJson`, `ToJson`, `Load` and `Save` over 67 hostile texts, which is breadth `test/tables/main.cpp` does not reach.

## The three goldens of §17.1 — all live

| | gate | result |
|---|---|---|
| 1 | `schema pack`'s bytes **==** the generated `Save` of the same instance built by hand | both corpus roots, byte-identical |
| 2 | `schema pack` → generated `Load` → report **all zero**, and the reload re-saves identically | both roots |
| 3 | **the text `schema unpack --one-file` writes == the text the generated `ToJson` writes** | both roots, byte-identical — the C `%g` shortest-round-trip float spelling, the pretty-print contract and every vocabulary spelling all agreeing at once |

`--one-file` is §17.2's last rule as an output shape (one `<Root>.json`, same instance, same writer, same bytes as the expanded tree); it exists because one whole text is what a backend's `ToJson` produces, which is what makes golden 3 a comparison of two texts rather than of a tree against an object.

## The gate that catches the review's class — now TWO-SIDED

`tables/pack/hostile-values/` — **43 trees, one per rule §16 states**, with a manifest both halves read. The same text goes through the packer **and** through the generated `FromJson`, and:

1. their **reports must agree counter for counter**;
2. their **wire bytes must agree byte for byte** — one text, one wire, two engines;
3. the bytes must then **`Load` clean and re-save identically**;
4. a case the manifest calls refused must be **refused by both**.

Negative control: relax one rule of the number grammar (accept a leading `+`) and the gate goes red.

It found real divergences on its first runs — the keyed-duplicate question that became #282, and then G1/G3 above once the second read's generator drove ~6,000 structured cases through it. That is the gate doing its job.

## The cold review's twelve findings

| | finding | repair |
|---|---|---|
| **F1** | `unpack` of a variable-length root recursed without bound | refused **by name on both verbs, before a byte is read or written**: `Decode` returns the refusal instead of folding it into the report, `Unpack` honours it, the writer spells a null pointer as `null`, and the write path carries §3.1's depth cap |
| **F2** | `bits(N)` never clamped | clamps at its IMPLIED `[0, 2^N − 1]`, counted, exactly as the descriptors carry it — and the corpus now packs a `bits(6)` at its maximum, with the over-range case in the hostile-value gate |
| **F3** | the tokenizer was neither RFC 8259's nor the backend's | RFC 8259's production, the backend's semantics: integral-in-value however spelled; a fractional value `kind_mismatch`; a magnitude the width cannot hold `kind_mismatch`; **no path stores an infinity**; `1-2 +7 007 01 .5 3. 1e 1e+ 1..2 1e2e3 5+ --3 1.2.3` malformed; `-0` into unsigned silent. The 37-row table is a fixture, and every row now runs against the backend too |
| **F4** | invalid UTF-8 written; a lone surrogate gave different wire bytes | ill-formed bytes write as U+FFFD; a lone surrogate escape READS as U+FFFD instead of manufacturing CESU-8. Golden 3 is the gate for the whole class |
| **F5** | `unpack` → `pack` not byte-stable into a populated tree | `unpack` PRUNES the entries it owns and did not write; the gate runs over a pre-populated tree |
| **F6** | §16/§17 were #269's first revision | the three files are `main`'s wholesale |
| **F7** | `null` on a `?T` was `kind_mismatch` | ABSENT on a `?T`, null on a pointer, `kind_mismatch` everywhere else |
| **F8** | `measure == save` was a tautology | `Measure` and the assertion deleted, with the reason where the function was |
| **F9** | the negative control sabotaged a tracked file | both negative controls build through `go build -overlay` from a copy under `build/` |
| **F10** | dot-entries skipped unreported | only a hidden **non-`.json` file** is passed over, and it is **named** |
| **F11** | the verbs reformatted `.schema` sources | `pack` and `unpack` load without format-in-place |
| **F12** | a non-silent report exited 0 | exits nonzero; `--tolerate` accepts it |

## The rest of the battery

Round trips: `unpack` → `pack` byte-stable for both roots, twice around, into a pre-populated tree, and through the one-file shape. Hostile trees (8) each refused naming the file and the reason. §16's rules as Go fixtures. §4's tolerance on the wire. A **pinned text** for the expanded per-field tree — the shape `ToJson` has no equivalent for. Fuzz: 7.9M execs clean over 45 s with a committed seed corpus, including the one real thing it found — the text form's fixed point is reached after ONE lap, because replacing an ill-formed byte with U+FFFD can push a `string(N)` past its bound, which is #267's stated cost and is now a named test.

## What §17 gained (the only page edits)

1. **§17.1** — `--one-file` is what makes the third golden a comparison of two whole texts; the hostile-value corpus is a two-sided differential (§17.5).
2. **§17.3** — the shape `unpack` writes, and why it PRUNES.
3. **§17.4** — a hidden non-JSON file is passed over and named; a non-silent report is a nonzero exit, `--tolerate` accepting it; neither verb writes to the unit's sources.
4. **§17.5** — the hostile-value corpus, and byte-stability *into a tree that already holds one*.

`USAGE.md` gains the matching paragraph in its existing pack section.

## Decisions, mine, labelled

1. **`unpack` writes the EXPANDED tree and prunes** what it owns and did not write; what it does not own is left for `pack` to name. `--one-file` is the opt-in for §17.2's last rule.
2. **A variable-length root is refused by name on both verbs** — §16.1 makes its text form a builder surface, a named follow-on.
3. **Invalid UTF-8 is byte-transparent on READ** and replaced on WRITE, both mirroring the backend.
4. **Only a hidden non-`.json` FILE is passed over**, and it is named.
5. **`--tolerate`, not `--strict`**: silence is the default contract, so the flag is the one that accepts noise.
6. **A repeated keyed slot key mirrors the reference** (above), pending the owner's word.


gafferongames:
## Cold adversarial review — #272

Reviewed at `c63f954` in a fresh clone, built and run on arm64 macOS 26.6.2, Apple clang 21.0.0, go 1.27.0. The oracle for §16 is #267 at its head (`origin/tables-json-walk` @ `b12175e`), generated and built independently; the oracle for the wire is this branch's own generated `PackTable.h` / `TablesTable.h`. The spec of record is `origin/spec-design-of-record` @ `2541a52`. This lands after #269, #265 and #267 by owner ruling.

**Verdict: DO-NOT-MERGE as it stands.** The shape is right and most of the engine is a faithful port — eight rich differential cases agree byte for byte on wire, report *and* text. But four defects break the PR's own central claim ("one implementation in the compiler and one in every generated backend are made ONE WIRE by goldens"): one of them hangs the tool on a one-line input, one makes `schema pack` write bytes this branch's own C++ reader refuses to accept clean, and two make the same authoring text produce different wire bytes on the two sides. None is a redesign; all are small and local.

---

# Findings

## F1 — `schema unpack` of a variable-length root recurses without bound. Hang, then OOM, from a zero-byte input.

Decision 2 ("a variable-length root is refused by name") is implemented on `pack` only. `unpack` refuses nothing and walks straight into unbounded recursion.

```
$ ./bin/schema unpack --root Scene --in /dev/null out/ tables/pointers
# ... never returns. GOTRACEBACK=all + SIGQUIT after 6s:
goroutine 1 gp=0x235b880221e0 m=nil [runnable]:
tabletext.(*Model).writeValue(0x235b881823b0, 0x235b8b9ff888, 0x235b8bd310e0, 0x7b74)
	internal/tabletext/write.go:65
tabletext.(*Model).writeScalar(...)
	internal/tabletext/write.go:181
tabletext.(*Model).writeField(...)
	internal/tabletext/write.go:115
```

`0x7b74` is the `depth` argument: 31 604 levels and climbing, with `out/name.json` and `out/version.json` already on disk. Three lines:

- `internal/tablewire/decode.go:21-24` — `Decode` *does* call `refuseVariable`, then **throws the diagnostic away** and only sets `report.Malformed`.
- `internal/tablepack/unpack.go:32` — `tablewire.Decode(m, inst, wire, &report)`: the return value is discarded, so `Unpack` proceeds to write the tree regardless.
- `internal/tabletext/write.go:176-181` — `if st := StructOf(f); st != nil { inst := cell.Tab; if inst == nil { inst = m.New(st) }; ... }`. A **null pointer materialises a fresh instance and writes it**. For `next *ListNode` that is infinite. There is no `null` case in the writer at all and no depth cap on the write path (`maxJsonDepth` is checked only on read).

Even without recursion this is wrong: the page's `pointer *T` row says "object, or `null`; `null` is a null pointer". Today a non-recursive `*Settings` at null would be written as an object of defaults.

## F2 — `bits(N)` is never clamped. `schema pack` writes bytes this branch's own C++ reader will not read clean, and `pack → Load → Save` is not byte-stable.

`internal/tabletext/value.go:371-382` — `StorageBytes` for `ir.TBits` returns 4 or 8 **bytes**; the bit width is never a bound. `readScalar`'s last clamp (`read.go:783`) therefore does nothing for `bits(N)`.

```
$ mkdir -p tree/weapons && echo '{ "channel": 64 }' > tree/weapons/0-a.json
$ ./bin/schema pack --root RootConfig --out t.bin --verbose tree tables/examples
packed t.bin: 24 bytes
report: silent — the data matched the schema exactly
                                          vvv
  cd 91 0e 0f 00 00 00 0d 01 00 00 00 06 00 00 00 66 73 06 40 00 00 00 00
```

Fed to `RootConfigLoad` from `build/tables-generated/examples/TablesTable.h` (this branch's own emitter):

```
load ok=1 unknown=0 kind_mismatch=0 clamped=1 malformed=0
resave=24 (in=24)
RESAVE: ***DIFFERS***
  in : ... 66 73 06 40 00 00 00 00
  out: ... 66 73 06 3f 00 00 00 00
```

So gate 2 of the PR's own golden ("a C++ `Load` of pack's bytes reports **all zero**") is false for any tree the corpus does not happen to cover, and `pack` reported *silent* while C++ `FromJson` of the same text reports `clamped=1`. `tables/pack/root/weapons/0-cannon.json` uses an in-range channel, which is why the golden is green.

## F3 — the number tokenizer is neither RFC 8259's nor #267's. 18 of 37 tokens diverge; several place silent garbage.

`internal/tabletext/read.go:291-315` (`scanNumber`) lifts any run of `[0-9+-.eE]` containing a digit, and sets `integral=false` on `.`/`e`/`E` alone. `tokenInteger` (`read.go:926-966`) then does `token[i] - '0'` on whatever bytes are in it — byte arithmetic, wrapping.

Same text to both readers, `ProfileConfig`:

| token | C++ #267 (and the page) | Go #272 |
|---|---|---|
| `1e3`, `1e2`, `1E2`, `1e+2`, `2.0`, `1.0e1`, `0.0`, `0e0` | integer placed (1000, 100, …) | **`kind_mismatch`, value dropped** |
| `1-2` | `malformed` | **places 2632** |
| `+7` | `malformed` | **places 7** |
| `007`, `01` | `malformed` | **places 7 / 1** |
| `.5`, `3.`, `1.2.3` | `malformed` | `kind_mismatch` |
| `--3` | `malformed` | `clamped`, value 0 |
| `1e`, `1e+`, `1..2`, `1e2e3` | `malformed` | **+Inf stored, report silent** |
| `1e400`, `1e1000000`, `9e99999`, `1.79769…e309`, `1e39` into a `float32` | `kind_mismatch`, "never stored as infinity" | **+Inf stored, report silent** |
| `-0` into an unsigned field | silent (fixed in #267 `b12175e`) | **spurious `clamped`** |

`saturateFloat` (`read.go:971-976`) returns `±Inf` on any `ParseFloat` error, so *every* malformed float token becomes an infinity. That collides head-on with the writer, which correctly refuses one — end to end:

```
$ mkdir -p tree/profiles && echo '{ "precision": 1e400 }' > tree/profiles/0-a.json
$ ./bin/schema pack --root RootConfig --out inf.bin --verbose tree tables/examples
packed inf.bin: 31 bytes
report: silent — the data matched the schema exactly      <-- the report lied

$ ./bin/schema unpack --root RootConfig --in inf.bin out/ tables/examples
schema: float +Inf has no JSON spelling — the writer refuses it rather than
losing it silently (SPEC-TABLES.md §16.2)
exit=1
```

The page states the invariant this breaks in as many words: *"The read path cannot produce one (above), so an instance that came from a text is always writable."*

`1-2 → 2632` and `01 → 1` are the ones that matter most: a typo in an authoring file becomes a value, silently, with a clean report.

## F4 — the writer emits invalid UTF-8, and a lone surrogate escape produces different WIRE bytes.

`internal/tabletext/write.go:265` passes stored bytes through verbatim. `internal/tabletext/read.go:246` takes a lone lead surrogate as itself and encodes it as CESU-8.

```
input: { "name": "\uD800x" }
C++ #267 wire: df 30 0c 04 00 00 00 ef bf bd 78 00 00      (U+FFFD)
Go  #272 wire: df 30 0c 04 00 00 00 ed a0 80 78 00 00      (CESU-8)   <-- different bytes
```

```
input: { "name": "a\xffb" }   (raw invalid UTF-8; wire agrees, 61 ff 62)
C++ ToJson: 231 bytes — "a�b" as U+FFFD
Go  Write : 229 bytes — 0xff passed through, so the emitted text is not valid UTF-8
```

Page §16.3 and #267's §16.2 both require it: *"The writer always emits valid JSON and valid UTF-8 … a lone surrogate escape READS as U+FFFD rather than manufacturing CESU-8."* This is also the golden the branch does not carry — page §17.1's third one, *"every text `schema unpack` writes is byte-identical to the one the backend's `ToJson` writes"* — which is precisely the gate that would have caught it.

## F5 — `unpack` → `pack` is not byte-stable into a directory that already holds a tree.

`unpack` writes but never prunes (`internal/tablepack/unpack.go`): an absent `?T` and a guarded-out field write no file — and do not delete the stale one. Unpacking a newer `.bin` over yesterday's checkout resurrects both.

```
table Outer { keep int32, extra ?Inner, flag bool, if flag { guarded int32 } }

$ ./bin/schema pack --root Outer --out a.bin treeA sch    # keep=1, extra present, flag=true
$ ./bin/schema unpack --root Outer --in a.bin work sch    # work/ = keep,extra,flag,guarded
$ ./bin/schema pack --root Outer --out b.bin treeB sch    # keep=2 only -> 9 bytes
$ ./bin/schema unpack --root Outer --in b.bin work sch    # SAME dir
$ ls work            # extra.json, flag.json, guarded.json, keep.json  <-- two stale
$ ./bin/schema pack --root Outer --out c.bin work sch
b.bin : 337a04020000000000                                (9 bytes)
c.bin : 337a040200000079b50d090000003ddf040900000000000000 (25 bytes)
```

Page §17.3 and this branch's §17.2 both say flatly "`unpack` → `pack` is byte-stable". It is, only into an empty directory — which is not the workflow the verb exists for. Either prune what would not be written, or refuse a non-empty target without a flag; and move the round-trip gate onto a **pre-populated** tree.

## F6 — decision 7 is not accurate, and merging as-is would revert #269.

§16/§17 here are verbatim from #269's **first** revision, `d9d3748`, not its head `2541a52` ("close the cold read's ten fixes, and the enum-keyed slot ruling", landed the same day). `diff` of §16 against `d9d3748` is exactly the three claimed additions and nothing else — but against the page's head it is 100+ lines short. What this branch's copy would take back out:

- the **Numbers** paragraph (`2.0`/`1e3` are integers; `1-2`, `5+`, `1.2.3`, `--3` are `malformed`);
- the **`null`** paragraph (`kind_mismatch` everywhere except `?T` → absent and pointer → null);
- **§16.3 "What the writer refuses"** in full — the -1 rules and the valid-UTF-8/U+FFFD obligation;
- the rewritten **Guarded fields** rule (reading places every key whatever the order);
- last-wins **as a whole value**, and `report.duplicate` in the report list;
- the **pretty-print contract** (one entry per line, two-space indent) — the thing that makes byte-comparison of texts checkable at all;
- `"This section states RULES, not one implementation"` — the paragraph #269 wrote *for this engine*;
- **§17.1 "The engine"** as its own subsection, with its **third** golden (unpack's text == the backend's `ToJson`);
- the §16.2 rows for `None` writing as `"None"` / `{}` / `[]`, `"None"` as a keyed key, base64 padding, `bits(N)`'s implied bound, float-width `kind_mismatch`.

The numbering shifts by one as a result: what this branch calls §17.1–§17.4 is §17.2–§17.5 on the page, and the Makefile comment cites "§17.4" for what the page calls §17.5.

Worse, two of the pasted sentences contradict **this branch's own code**:

- "integers, `bits(N)` … integral; **a fraction is malformed**" — `read.go:770-776` counts `kind_mismatch` (which is right, and is what the page says).
- "a guarded group's fields are placed **only when the guard reads true**" — `readTable` places every key it can name regardless of guard, and I confirmed the two key orders pack to identical bytes. Right code, stale prose.

Take §16/§17 from `2541a52`, re-apply the three additions on top, and re-check every rule in this review against it.

## F7 — `?T` reads `null` as `kind_mismatch`; the page says ABSENT.

`Shape()` never returns `'z'`, so `null` fails the shape check for every kind including `?T` and pointers.

```
$ echo '{ "gunner": null }' > tree/ships/Fighter.json
$ ./bin/schema pack --root PackConfig --out t.bin --verbose tree tables/examples
report: unknown 0, kind_mismatch 1, clamped 0, duplicate 0, malformed false
```

Page: *"`null` is `kind_mismatch` for every kind except the two where absence is a value: a `?T` reads `null` as ABSENT, and a pointer reads it as null."* Not testable against #267 (it has no `?T`), so this branch is the only implementation — and the page wins.

## F8 — `measure == save` is a tautology, and the test that asserts it cannot fail.

`tablewire.Measure` is `len(Encode(...))` (`encode.go:43-49`) and `wire_test.go:55-61` asserts `size != len(wire)`. Decision 5's reasoning is sound — §9's battery belongs on the generated code — but by this repo's own doctrine ("a gate that cannot be made to fail is not a gate") the assertion is scaffolding. Delete the test and `Measure`, or make `Measure` an independent size computation. Do not keep a green line that proves nothing.

## F9 — `make tables-pack-negative` sabotages a tracked source file in place.

`Makefile:271-284` copies a sabotaged `internal/tablewire/encode.go` over the real one and restores it on the next shell line. Ctrl-C during the `go build` leaves the working tree sabotaged, and `make -j` racing another Go target compiles the sabotaged source. It works (I ran it; it goes red correctly and restores cleanly), but the negative control should build from a copy under `build/`, not from the source tree.

## F10 — dot-entries are skipped unreported, including `.json` files that name real fields. (judgement: keep, but narrow it)

Decision 4 is right about `.DS_Store`. But `readDir` (`pack.go`) drops *every* dot-entry, so `.version.json` naming a real field is silently ignored — the one hole in "a tree that does not mirror the table is reported rather than guessed at". Skip only entries that are not `*.json`, or count the skip and print it under `--verbose`.

## F11 — `pack`/`unpack` reformat the unit's `.schema` files in place.

Inherited from `loadUnit`, but new for these verbs: `schema unpack --root Outer --in b.bin work sch` printed `formatted sch/Stale.schema`. A verb whose job is to assemble a config tree writing to schema sources is surprising; worth an exemption or a note.

## F12 — a non-silent report still exits 0.

`pack` printed `report: unknown 4, kind_mismatch 0, clamped 10, ...` and exited 0, writing the `.bin`. Per §17 that is "reported, never fatal" and defensible — but a build pipeline has no way to fail on it. A `--strict` would cost four lines.

---

# Verified good

Everything below I ran and confirmed on this branch.

**The wire.** `make tables-pack` and `make tables-pack-negative` both pass. Kind 16's layout is identical in the Go encoder (`encode.go:280-337`), in the generated C++ (`PackTable.h:900-923`) and in both §3.2 texts: `id u16, kind 16, L u32, elem-kind u8, N u32`, then `(variant id u16, L u32, body)` ascending by ordinal, defaults elided per slot, None never riding. I decoded `39 2d | 10 | 14 00 00 00 | 0d | 01 00 00 00 | 7a d9 | 09 00 00 00 | …` by hand off a hostile tree and it is exactly that. Nothing in `generated/` or `internal/codegen/` is touched by this PR — the engine really is tooling only.

**Instances the corpus lacks.** I packed each of these and round-tripped through the branch's generated C++ `Load` → `Save`; all byte-identical, all reports zero: a present all-default `?T` (rides, 31 bytes); an `[E]T` with only the `Max` slot present; an empty bounded array `[]`; a union `None` `{}`; an empty flags mask `[]`; `int64` min; `uint64` max; `float64` max; a `float32` denormal; negative zero (survives as `00 00 00 80`); a `string(32)` filled to exactly 32 bytes ending in a multi-byte character; base64 padded and unpadded; a 70 000-byte base64 body clamped into `bytes(16)`.

**§16 against #267.** Eight rich differential cases — nested unions with both arms, flags by name, enums by name, fixed and bounded arrays, clamped counts, unknown enum and flag names, a two-key union, exponent float formatting (`1e21`, `1e-7`) — agree with #267 byte for byte on **wire, report and emitted text**. String clamping at a multi-byte boundary agrees exactly, including the odd interior-drop case (`30×'a'` + emoji + `'Z'` → both keep `'Z'`); that is a shared spec question, not a divergence. Duplicate keys agree, `duplicate` counter included. Raw invalid UTF-8 rides through to the wire identically (decision 3 holds on the *read* side; only the writer is wrong, F4).

**§17's directory rule.** Every case I could think of behaves and names the file: `<Root>.json` alone; `<Root>.json` beside a field file (refused); a nested table as `<field>.json` vs as a directory (byte-identical, `511b0d09…`); a bounded array as `<field>/` files-in-name-order vs one `<field>.json`; both at once (refused, "one value, one place"); a lower-case variant filename (refused, not a variant); a variant claimed by a file and a directory (refused); `None.json` (refused, naming the file and §3.2); a scalar given a directory (refused); a file naming no field, a directory naming no field, a `.txt` stray (all refused); an empty tree (2 bytes, silent). All eight committed hostile trees refuse with a file name and a reason. **Symlink loops cannot recurse** — a symlink is not `IsDir()`, so it lands in the "names no field" arm; I tried both a `loop` directory symlink and a `loop.json` one.

**Depth and recursion on the read path.** 127 levels accepted, 129 refused as malformed, 10 000 refused without a stack overflow — §3.1's cap of 128 honoured. A hostile `*Node` tree cannot recurse through `pack`: every variable-length root in `tables/pointers` is refused by name.

**Round trips.** `pack → unpack → pack → unpack → pack` is byte-identical for both corpus roots across both laps, and the two unpacked trees are identical. The tolerance path is right: a `.bin` carrying a field the schema has since dropped unpacks with `unknown 1` and the surviving fields written. The report aggregates once per tree — six files with four unknown keys and ten clamps produced one line.

**Fuzz.** `FuzzReadText` for 120 s: **19 951 335 execs, 55 new interesting, PASS, no crashes**. `go build ./... && go vet ./... && go test ./...` all clean; `gofmt -l` clean; the `internal/publicapi` external-module gate passes and `cmd/schema` still imports only `compiler` and `ir`.

**Provenance and hygiene.** Both commits authored and committed `Rowan Claude <rowan@mas-bandwidth.com>`. No Python, no scratch files, no `.bak`/`.orig`, working tree clean before and after every gate. Decision 6 verified mechanically: **every** line the PR adds to `ir/ir.go`, `ir/table.go` and `internal/check/check.go` appears verbatim in `origin/tables-json-walk`, and `internal/codegen/cpptable/json.go` is untouched.

---

# What I would want before this merges

1. F1 — `Unpack` honours `Decode`'s return; the writer emits `null` for a null pointer and carries the same depth cap the reader has.
2. F2 — clamp `bits(N)` at `2^N − 1` and count it. One line in `StorageBytes` or `readScalar`, plus a corpus file with an over-range `bits` value so the golden covers it.
3. F3 — replace `scanNumber`/`tokenInteger`/`saturateFloat` with RFC 8259's grammar and #267's semantics: integral-valued tokens however spelled, a genuinely fractional value `kind_mismatch`, everything else `malformed`, and no path that stores an infinity.
4. F4 — U+FFFD on write, U+FFFD for a lone surrogate escape on read.
5. F5 — `unpack` prunes, and the byte-stability gate runs over a pre-populated tree.
6. F6 — rebase §16/§17 onto `2541a52` and re-derive the three additions.
7. F7 — `null` on `?T` reads as absent.
8. Add page §17.1's **third** golden (unpack's text vs the backend's `ToJson`) once #267 is in — it is the gate that catches F4's whole class, and F3's too.
9. F8–F12 as noted.

The corpus is what let three of these through: two clean trees prove the happy path and nothing else. A `tables/pack/hostile-values/` tree — an over-range `bits`, a malformed number token, a lone surrogate, a `null` on a `?T` — packed and compared against C++ `FromJson` of the same texts would have caught F2, F3, F4 and F7 in one gate.

gafferongames:
## Second cold read — #272 at `02a36da`

Fresh clone, rebased head `02a36da` on `0762afb`, arm64 macOS 26.6.2, Apple clang 21.0.0, go 1.27.0. Both C++ oracles built from **this branch's own** generated headers (`build/tables-generated/examples`) — since #267 and #269 are now in `main`, one generator serves both the wire and the text sides. All four new gates green on my bench:

```
pack golden: schema pack == Save and schema unpack == ToJson for both roots
pack negative control: eliding a present ?T turns the golden red
pack hostile-value gate: 43 trees agree on report and wire across both engines
pack hostile-value negative control: a leading + packs once the grammar is relaxed
```

**Verdict: MERGE-WITH-CONDITIONS.** All twelve findings from the first read are genuinely closed — I re-ran every repro and each one is fixed at the root, not papered over. The hostile-value gate is the real thing: two-sided, manifest-driven, with a working negative control, and it is exactly the gate that would have caught F2/F3/F4/F7. Against it I ran roughly 6,000 structured differential cases and found **two more divergence classes**, both narrow, both fully characterised, both one-line fixes plus a corpus row. They are wire divergences between the two engines — the class this PR exists to close — so they land before merge, not after. Everything else is wording and nits.

---

# The twelve, re-run

| | repro at `02a36da` | |
|---|---|---|
| **F1** hang on a variable-length root | `unpack --root Scene --in /dev/null out/ tables/pointers` → `schema: Scene is VARIABLE-LENGTH … pack a fixed-size root`, exits at once | **fixed** |
| **F2** `bits(N)` unclamped | `{"channel":64}` → `…66 73 06 3f…`, `clamped 1` | **fixed** |
| **F3** number grammar | 43 hostile tokens (`1-2`, `+7`, `007`, `01`, `.5`, `3.`, `1e`, `1e+`, `1..2`, `1e2e3`, `1e400`, `9e99999`, `-0`, `1e3`, `2.0`, `1e2`, `0e0`, …): **0 of 43 diverge** | **fixed** |
| **F4** invalid UTF-8 on write / CESU-8 | 15 string cases (lone lead and lone trail surrogate, surrogate pair, raw `0xff`, overlong, truncated, CESU-8 input, every escape): **0 of 15 diverge**, wire *and* text | **fixed** |
| **F5** unpack does not prune | 9-byte bin unpacked over a 4-file tree → `flag.json keep.json`, repack `337a04020000000000` — byte-identical | **fixed** |
| **F6** stale §16/§17 | §16 byte-identical to `main`; §1–§15 identical; §18+ identical; exactly four §17 hunks (§17.1 `--one-file`, §17.3 pruning, §17.4 three rules, §17.5 the two additions) | **fixed** |
| **F7** `?T: null` | `null` → absent (2 bytes), `{}` → present-all-default (11 bytes), valued → 15 bytes; all three agree with the backend | **fixed** |
| **F8** tautological `Measure` | `grep -rn "func Measure\|tablewire.Measure" internal/ compiler/ cmd/` → nothing | **fixed** |
| **F9** in-place sabotage | 0 in-place `cp` into tracked files, 3 x `go build -overlay`; `git status` clean before and after both negative controls | **fixed** |
| **F10** dot-entries swallowed values | `.DS_Store` passed over and NAMED under `--verbose`; `.version.json` → `".version" names no field` | **fixed** |
| **F11** `.schema` rewritten | md5 of a deliberately unformatted `.schema` unchanged across `pack` and `unpack` | **fixed** |
| **F12** exit 0 on a dirty report | exit 1 without `--tolerate`, 0 with it, 0 on a clean tree | **fixed** |

§17.1's third golden I checked independently rather than trusting the driver: `unpack --one-file` text vs `PackConfigToJson` / `RootConfigToJson` of a `Load` of the same bytes — **byte-identical, 1096 and 2153 bytes**.

---

# New findings

## G1 — the float32 conversion double-rounds. Go reads through `float64`, C++ uses `strtof`.

`internal/tabletext/read.go` `placeFloat` does `strconv.ParseFloat(token, 64)` and then `float32(value)`. The generated walk's `TableJsonTokenDouble` (`TablesTable.h:1183-1199`) branches on width: `if ( single ) { return (double) strtof( work, NULL ); }`. One correctly-rounded conversion versus two, so they part company on decimals that lie between FLT_MAX and the float32 rounding midpoint yet round to the *midpoint* as a double.

```
{ "ratings": [340282356779733661637539395458142568447,0,0,0] }
C++: kind_mismatch=0  wire=30  …0a 04 00 00 00 ff ff 7f 7f…      (FLT_MAX, clean)
Go : kind_mismatch=1  wire=2   00 00                             (dropped)
```

Not one magic string — the whole band, both signs:

```
340282356779733661637539395458142568447   DIVERGE
340282356779733661637539395458142568440   DIVERGE
340282356779733661637539395458142560000   DIVERGE
340282356779733661637539395458000000000   DIVERGE
340282356779733661637539000000000000000   DIVERGE
-340282356779733661637539395458142568447  DIVERGE
340282356779733661637539395458142568448   agree (the midpoint itself: both reject)
340282350000000000000000000000000000000   agree (rounds clear of it: both accept)
```

And the same class at the subnormal end, where it is **silent on both sides and only the bytes differ**:

```
{ "ratings": [7.0064923216240853547e-46,0,0,0] }
C++: report clean, wire=30  …0a 04 00 00 00 01 00 00 00…   (smallest subnormal)
Go : report clean, wire=2   00 00                          (rounded to zero, elided)
```

The page's own words are the tie-break: *"Numbers are converted exactly, at every integer width and both float widths."* Go must move — `strconv.ParseFloat(token, 32)` when `single`, which is what `strtof` is.

## G2 — `unpack` prunes only inside its own shape, so switching shape leaves a tree `pack` refuses.

§17.3's new paragraph is right and the expanded case works (F5). But `--one-file` writes no field files, so by that rule every field entry should go; and the expanded form does not remove a stale `<Root>.json`, which names the root rather than a field. Both directions:

```
$ ./bin/schema unpack            --root PackConfig --in pc.bin m1 tables/examples
   m1: global.json reserves.json ships thresholds version.json
$ ./bin/schema unpack --one-file --root PackConfig --in pc.bin m1 tables/examples
   m1: PackConfig.json global.json reserves.json ships thresholds version.json
$ ./bin/schema pack --root PackConfig --out m1.bin m1 tables/examples
schema: …/m1: PackConfig.json is the whole root, and 5 other entries beside it
claims it too — a root is one file or one tree of fields, never both (§17.1)
```

```
$ ./bin/schema unpack --one-file --root PackConfig --in pc.bin m2 …   # m2: PackConfig.json
$ ./bin/schema unpack            --root PackConfig --in pc.bin m2 …
   m2: PackConfig.json global.json reserves.json ships thresholds version.json
   -> pack refuses, same message
```

The tool writes a tree its own sibling verb rejects. It is the same class F5 named, and the fix is the same rule applied to both shapes: when writing `--one-file`, prune the field entries; when writing expanded, prune `<Root>.json`.

(Prose slip in that message: *"5 other entries beside it **claims** it too"*. It reads correctly only at 1.)

## G3 — a duplicate key whose repeat is REJECTED wipes the first value in Go and keeps it in C++.

When the repeat's JSON shape matches but the value is refused at the value level — a fraction in an integer field, a magnitude a float width cannot hold, or a malformed token — Go has already called `m.reset(fv)` and the earlier value is gone. C++ leaves it standing. **Identical reports, different wire bytes.**

```
{ "tilt": 100, "tilt": 5e-324 }
C++: kind_mismatch=1 duplicate=1   wire=6   3c 02 02 64 00 00     (100 stands)
Go : kind_mismatch=1 duplicate=1   wire=2   00 00                 (wiped)
```

Every scalar kind, nested tables, bounded-array elements, and enum-keyed slots:

```
{ "precision": 1.5, "precision": 1e400 }              C++ 1.5      Go default
{ "experience": 5, "experience": 2.5 }                C++ 5        Go default
{ "badge": 9, "badge": 1e309 }                        C++ 9        Go default
{ "primary": { "damage": 3, "damage": 1e400 } }       C++ 3        Go default
{ "attachments": [ { "slot": 2, "slot": 1.5 } ] }     C++ 2        Go default
{ "thresholds": { "Hard": 64, "Hard": 0.1 } }         C++ 64       Go default
{ "thresholds": { "Hard": 64, "Hard": +1 } }          C++ 64       Go default   (malformed repeat)
{ "tilt": 100, "tilt": "x" }                          agree        (shape check fires first)
{ "tilt": 100, "tilt": 7 }                            agree        (repeat is placeable)
```

C++ is the one the page describes. §16.2 ties re-establishment to *placing* — *"a repeated table or union key re-establishes the whole value at its defaults **before placing it**"* — and says a value the walk refuses is *"skipped, never coerced"*. A repeat that is never placed should leave the field alone. Go resets unconditionally; the fix is to reset into a scratch and commit on success, or restore on a value-level rejection.

### How I found them

A structured generator (schema-shaped keys, hostile values, duplicates, trailing commas) over five roots:

- **3,000 cases with duplicate keys: 108 diverge** — 53 carry a float32 double-rounding token (G1), the remaining 55 are all G3.
- **3,000 cases with field-key duplicates suppressed: 56 diverge** — 50 G1, and the 6 survivors are all G3 inside an enum-keyed object (the generator only deduplicated field keys, not slots).
- Nothing outside those two classes in 6,000 cases.

Clean sweeps, no divergence at all: **56** integer width-boundary cases across every width (min, max, plus and minus one, times ten, 24-digit overflow); **21** base64 edges (empty, `"A"`, `"AA"`, `"QQ="`, `"A==="`, unpadded, over-long, embedded spaces, URL-safe alphabet); **15** UTF-8 and surrogate cases; **43** numeric tokens; nesting at depths 1, 100, 126, 127, 128, 129, 200 and 2,000 plus real four-level nesting; **30** structural cases (trailing commas, duplicate array/table/union/flags/enum keys, `"None"` and unknown and wrong-shape keyed keys, the `json = "key"` rename addressed both ways, empty text, array root, missing colon, comment, trailing rubbish, fixed and bounded array over- and under-run, guard false with the field given, guard order swapped).

## G4 — §17 promises byte-stability that a legal instance does not have. (wording, and one corpus row)

The U+FFFD rule and a string bound interact, and the two engines agree exactly — this is not a divergence, it is a stated invariant that is not true:

```
lap 0   { "name": <32 x 0xff> }  ->  wire 32 x ff, report CLEAN, both engines
        writer emits 32 x U+FFFD  =  96 bytes of text
lap 1   that text back in        ->  wire 10 x U+FFFD (30 bytes), clamped=1, both engines
lap 2   fixed point
```

§16 already owns this honestly: *"for a string holding invalid UTF-8 the round trip is not byte-identical, because the alternative is emitting a text no conforming parser can read."* But §17.3 and §17.5 say "`unpack` → `pack` is byte-stable" flat, and this PR strengthens §17.5 to "INTO A TREE THAT ALREADY HOLDS ONE" without adding the carve-out. Judgement: **the behaviour is right and should not change** — the writer must emit valid UTF-8, the bound is the bound, and the new nonzero-exit rule means a pipeline actually sees the `clamped` on lap 1. §17 just needs §16's clause carried across, and the corpus needs a `str-illformed-at-bound` row so the fixed point is pinned rather than assumed.

## Nits

- **`pack` writes the `.bin` and then exits 1** on a non-silent report. In a Make rule the output is left newer than its prerequisites, so the second `make` sees it up to date and the failure vanishes. Write after the gate, or remove on refusal.
- **`test/tables/hostile_main.cpp:158`** checks `unknown|kind_mismatch|clamped|malformed` after `Load` but not `report.duplicate`. Harmless today (the wire never raises it) and one word to fix; the same omission is in `pack_main.cpp`.

## Coverage the 43 do not reach

Present and good. Absent, and each is one file: **trailing commas**; **nesting past the depth cap**; a **duplicate array / table / union key** (only a scalar `dup-key` is pinned, and it is the placeable kind — the one that hides G3); the **`json = "key"` rename addressed by the declared name** (must be `unknown`); **base64 padding variants** beyond not-base64 and overlong; and rows for **G1** and **G3**.

Also worth recording: `keyed-duplicate-slot` currently pins `0,0,0,0,false` — both engines agree with each other and both disagree with §16.2, which counts the repeat. That is the known divergence being fixed C++-side, but the manifest line will have to move with it, so it deserves a `# known: §16.2 says duplicate=1, tracked in <PR>` comment beside that row now.

---

# Verified good

Everything above the findings, plus: `go build ./... && go vet ./... && go test ./...` clean; `gofmt -l` clean; `FuzzReadText` **150 s, PASS, no crashes**; the Go half of the hostile gate reads the *same* `cases.txt` the C++ half does (`pack_test.go:229`), so the manifest cannot drift between the two sides; `tables/pack/pinned/` is a real pinned-text gate (`pack_test.go:414`), which is the "a round trip alone cannot see a vocabulary error" test §16.5 asks for; `--one-file` round-trips byte-stably for both roots across two laps and packs to the same bytes as the expanded shape; all four commits authored and committed `Rowan Claude <rowan@mas-bandwidth.com>`; no Python, no scratch files, no `.bak` or `.orig`; working tree clean before and after every gate including both negative controls.

# Conditions

1. **G1** — `ParseFloat(token, 32)` for a single-width field, plus `num-float32-double-rounding` rows for the upper band and the subnormal one.
2. **G3** — do not reset a field until the repeat is known to be placeable, plus a `dup-key-rejected` row (field key *and* keyed slot).
3. **G2** — `unpack` prunes across both shapes, plus a test that unpacks each shape over the other and packs the result.
4. **G4** — carry §16's invalid-UTF-8 clause into §17.3/§17.5, and add `str-illformed-at-bound`.
5. The two nits, and the coverage rows above as you see fit.

The first four are small and the gate they extend already exists. Nothing here is a rework, and the repair pass this branch just went through is exactly what I would want to see — land these and it is a merge.

gafferongames:
## Third cold read — #272 at `91b4ea6`

Fresh clone, head `91b4ea6` on `56d2a3b`, arm64 macOS 26.6.2, Apple clang 21.0.0, go 1.27.0. Oracles rebuilt from this branch's own generated headers.

**Verdict: MERGE.** G1, G2, G3 and G4 are closed; I re-ran the two 6,000-case differential sweeps that found them and they are now **zero for zero**. The keyed-duplicate flip, the `optional-wrong-shape` row, the temp+rename fix and the `report.duplicate` comparison all check out against the merged C++. The sanitized twins are real and I proved the gate flag aborts rather than recovers. One non-blocking cleanup below.

---

# The four conditions, re-run

## G1 — float32 double-rounding: closed

`placeFloat` now takes the width, so the single conversion matches `strtof`. The whole band that diverged, both signs, plus the midpoint and the two sides of it:

```
340282356779733661637539395458142568447   AGREE  accepted, clean   (was: C++ FLT_MAX, Go kind_mismatch)
340282356779733661637539395458142568440   AGREE  accepted, clean
340282356779733661637539395458142560000   AGREE  accepted, clean
340282356779733661637539395458000000000   AGREE  accepted, clean
340282356779733661637539000000000000000   AGREE  accepted, clean
-340282356779733661637539395458142568447  AGREE  accepted, clean
340282356779733661637539395458142568448   AGREE  kind_mismatch=1   (the midpoint: both reject)
340282356779733661637539395458142568449   AGREE  kind_mismatch=1
3.40282357e38                             AGREE  kind_mismatch=1
```

Subnormal end, where it used to be a **silent** byte divergence — `7.0064923216240853547e-46`, `7.006492321624085354e-46`, `7.00649232162408535e-46`, `3.503246160812042775e-46`, `1e-45`, `1.4e-45`, `2.8e-45`, `7e-46`: **8 of 8 agree**. Rows `num-float32-upper-band`, `num-float32-midpoint`, `num-float32-subnormal` pin all three regions.

## G2 — pruning across both spellings: closed, both directions

```
expanded  -> --one-file :  global.json reserves.json ships thresholds version.json
                        -> PackConfig.json                      repack byte-identical
--one-file -> expanded  :  PackConfig.json
                        -> global.json reserves.json ships thresholds version.json
                                                                repack byte-identical
```

And §17.3's "an entry naming NO field is left exactly where it is" holds — I dropped `README.txt`, `.git/` and `.DS_Store` into a tree and switched shape; all three survived, only the tool's own entries moved.

## G3 — re-establishment on placement: closed

The unconditional reset is gone. Every reduction that diverged now agrees, and the placeable cases did not regress:

```
{ "tilt": 100, "tilt": 5e-324 }                        AGREE  km=1 dup=1
{ "precision": 1.5, "precision": 1e400 }               AGREE  km=1 dup=1
{ "experience": 5, "experience": 2.5 }                 AGREE  km=1 dup=1
{ "badge": 9, "badge": 1e309 }                         AGREE  km=1 dup=1
{ "primary": { "damage": 3, "damage": 1e400 } }        AGREE  km=1 dup=1
{ "attachments": [ { "slot": 2, "slot": 1.5 } ] }      AGREE  km=1 dup=1
{ "thresholds": { "Hard": 64, "Hard": 0.1 } }          AGREE  km=1 dup=1
{ "thresholds": { "Hard": 64, "Hard": +1 } }           AGREE  dup=1 malformed  (malformed repeat)
{ "tilt": 100, "tilt": 5e-324, "tilt": 3 }             AGREE  km=1 dup=2
{ "ratings": [1,2,3,4], "ratings": [9] }               AGREE  dup=1
{ "primary": {...}, "primary": {...} }                 AGREE  dup=1   (re-establish)
{ "primary": { "effect": buff }, "primary": { "effect": debuff } }  AGREE  dup=1
{ "perks": ["Shielded"], "perks": ["Turbo"] }          AGREE  dup=1
```

Ten `dup-key-*` rows pin the kinds.

## G4 — the carve-out, and its claim is true

§17.3 now says the lap is counted (`clamped`, a nonzero exit) and that **the fixed point is reached in one lap**. That is a stronger statement than I asked for, so I checked it rather than took it — 32 × `0xff` in a `string(32)`, through the real tree:

```
lap0 bin 6f96599920f3
lap1 bin ea2943769493   text 0fad47050ec1     (96 bytes of U+FFFD, then clamped)
lap2 bin ea2943769493   text 32b656251f63
lap3 bin ea2943769493   text 32b656251f63     <- text agrees from lap2 on
```

Bins identical from lap 1, texts identical from lap 2 — exactly what the paragraph claims, and it holds under `--one-file` too. `str-illformed-at-bound` pins it.

---

# The delta's own additions

**The keyed-duplicate flip (#282's rule).** Counted on the resolved slot, before the shape check; unknown and `"None"` repeats stay `unknown`. Verified against the merged C++, all agreeing:

```
{ "ships": { "Fighter": {"health":1}, "Fighter": {"mass":2} } }        dup=1
{ "ships": { "Fighter": {}, "Fighter": 3 } }                           km=1 dup=1   (before the shape check)
{ "ships": { "None": {}, "None": {} } }                                unknown=2 dup=0
{ "ships": { "Nope": {}, "Nope": {} } }                                unknown=2 dup=0
{ "ships": { "Fighter": {}, "None": {}, "Nope": {}, "Fighter": {} } }  unknown=2 dup=1
{ "thresholds": { "Easy": 1, "Easy": 2, "Easy": 3 } }                  dup=2
```

**`optional-wrong-shape`.** A `?T` given an unplaceable value stays PRESENT — 11 bytes, the same as `{}` — with `kind_mismatch=1`, while `null` is absent (2 bytes). That is §16.2 read correctly on both halves at once: *"presence of the KEY is presence: a key present sets the field present, whatever its value"*, with `null` as the one stated exception. Good catch, and it composes rather than colliding.

**The nits.** A refused pack now leaves **no** `.bin` at all, an existing good output is preserved byte-for-byte under a failing pack, and no `.tmp` residue. `report.duplicate` is compared in both drivers (`hostile_main.cpp:77,159`, `pack_main.cpp:342,394`).

**The sanitizers really gate.** `-fsanitize=address,undefined -fno-sanitize-recover=all` — I built a deliberate signed-overflow probe with the same flag set and it aborted (**exit 134**, `SUMMARY: UndefinedBehaviorSanitizer`), so the flag is doing what the comment says rather than printing and continuing. The twins carry 344 and 295 asan/ubsan symbols, and both gates run plain and sanitized, clean.

---

# Evidence

**The two sweeps that found G1 and G3, re-run unchanged:**

| sweep | before | at `91b4ea6` |
|---|---|---|
| 3,000 structured cases with duplicate keys | 108 diverge | **0** |
| 3,000 with field-key duplicates suppressed | 56 diverge | **0** |

**Every earlier battery, re-run, all zero:** 43 numeric tokens; 56 integer width boundaries across every width; 18 base64 edges; 12 UTF-8 / surrogate / overlong / truncated cases; depths 1, 100, 126, 127, 128, 129, 200, 2000.

**The tree entry points, which my probe does not reach.** `pack` reads a `<field>.json` through `ReadValue` and a `<Variant>.json` through `ReadElement`, not `Model.Read` — different code paths from everything above, and the reset rewrite touches them. I packed six hostile values both as one `<Root>.json` and as the equivalent field/variant files: **same bytes every time**, including the rejected duplicate in a field file, the rejected duplicate in a variant file, the duplicate keyed slot, `optional-wrong-shape`, the float32 band and the illformed-at-bound string. The bytes then load clean in the backend and re-save byte-identically.

**Gates:** all four green, each run twice (plain and sanitized) — `pack golden`, `pack negative control: eliding a present ?T turns the golden red`, `pack hostile-value gate`, `pack hostile-value negative control: a leading + packs once the grammar is relaxed`. Round trips byte-stable for both roots in both shapes across two laps. All eight hostile *tree* refusals still name the file and the reason. `FuzzReadText` 150 s: **19,438,951 execs, PASS**.

**Docs:** §16 byte-identical to `main`; §1–§15 identical; §18+ identical; four §17 hunks and nothing else, plus the 8-line USAGE paragraph.

**Hygiene:** `go build ./... && go vet ./... && go test ./...` clean; `gofmt -l` clean; all five commits `Rowan Claude <rowan@mas-bandwidth.com>`; no Python, no strays, no TODO/FIXME residue in the new packages; working tree clean before and after every gate including both negative controls.

---

# One non-blocking cleanup

`tables/pack/hostile-values/cases.txt` lines 14 and 15 are the **same case twice**, differing only in column width — the old-alignment row survived the realignment for the longer new names:

```
14:array-past-bound···········RootConfig··packs·0,0,2,0,false
15:array-past-bound·············RootConfig··packs·0,0,2,0,false
```

```
$ grep -v '^#' cases.txt | grep -c .        67
$ awk '{print $1}' … | sort -u | wc -l      66
$ ls -d tables/pack/hostile-values/*/ | wc -l   66
```

Nothing goes unchecked — the tree is simply packed and asserted twice — but the gate counts manifest rows, so it prints "67 trees" over 66. The number is the instrument's own claim about its coverage, and this repo's line is that the harness rots too. Delete line 14 and the banner reads 66. Not worth holding the merge for; fold it into the next touch.

---

Three reads, thirty-odd findings, and every one of them is now either fixed at the root or pinned by a row in a gate that would catch it again. The hostile-value corpus has gone from not existing, to 43 trees, to 66 covering every rule §16 states — and the two classes it did not cover when I last looked are the two it now pins hardest. Ship it.

PR #285, opened 2026-09-02
Tables JSON walk: emit the walker into a generated <Base>Table.cpp, compiled once (#283)

### LINKED ISSUE #283: Tables JSON walk: emit the walker into a generated <Base>Table.cpp, compiled once — the header carries declarations only
Owner ruling, 2026-09-02 (Glenn, live), on the unconditional walker's +11–15% header compile cost: "It would be best if it were written to a corresponding cpp file so it doesn't need to be included every time."

**Change.** `schema generate --lang cpp` emits, for each table unit file, a `<Base>Table.cpp` beside `<Base>Table.h` holding the generic JSON walker (tokenizer, reader, writer) and the per-table `<Name>FromJson` / `<Name>ToJsonMeasure` / `<Name>ToJson` definitions; the header declares them. The generic-walk gate moves to the .cpp (one walker, byte-identical across the corpus's generated .cpp files). The reflection descriptors stay in the header (constant data the wire codecs and tools already rely on). A project that uses the text form compiles the generated .cpp; a project that never does pays nothing — no walker, no number-conversion includes in the header.

**Page (spec first).** SPEC-TABLES.md §16.1 currently states the walk rides in every table closure's header; it changes to the .cpp rule with the owner's words in §13; the +11–15% number and open item (f) on #248 retire. USAGE.md's JSON example shows the .cpp in the build line. The test harness and Makefile compile the generated .cpp for the tables leg (plain and sanitized).

**Gates.** All existing JSON batteries green unchanged; header compile time measured before/after (expect the walker's share gone); zero-cost gate unchanged; the C# backend is unaffected (no text form).

gafferongames:
Closes #283.

Owner's ruling: *"It would be best if it were written to a corresponding cpp file
so it doesn't need to be included every time."*

A table unit file now emits **`<Base>Table.cpp` beside `<Base>Table.h`**. The
header declares the three functions per table and keeps the reflection
descriptors; the `.cpp` holds the generic walk, the definitions, and the three
number-conversion includes the walk needs. A translation unit that includes the
header for the wire codecs or the descriptors pays nothing for a form it never
calls.

## The measurement

Same empty-TU method the `+11–15%` came from — arm64 clang, interleaved, one
sitting, n=15 after warmup. Min and median agree, so the scheduler is not in it:

```
prewalk (no walk in the tree)   min 0.0478s   median 0.0493s
before  (walk in header)        min 0.0538s   median 0.0559s
after   (walk in .cpp)          min 0.0485s   median 0.0498s

after vs before   -9.8%    <- the walker's share, gone
after vs prewalk  +1.5%    <- the header is back to its pre-walk cost
```

Three sittings agreed (−10.8%, −9.9%, −9.8%); two earlier ones had a contaminated
arm (spreads to 83%) and are reported here only as corroboration of the
direction, not averaged in. The header drops **3656 → 1986 lines**.

## The split, and the line that does not move

Two further owner sentences decide the shape, and §13.5 records all three:

> *"As we get more complex stuff in types/tables, needing a .cpp file now makes
> more sense."*
> *"I think it's OK for types to remain header only."*

So the split is **by wire**, not by convenience:

- **The type wire stays header-only.** A type is a struct and its codec, both
  folded into the caller — nothing to compile once and link. `SPEC.md` §6.1 says
  so where it describes the generated C++ layout.
- **A table unit gains `<Base>Table.cpp` for its RUNTIME.** The JSON walk is what
  lives there today. Storage structs, wire codecs and descriptors stay in the
  header, because they are respectively inlineable, template-parameterised over a
  context, and constant data a tool reads directly.
- **What may follow it, and on what evidence.** Any further table runtime that is
  neither a template nor constant data is a candidate, decided on measurement when
  one exists. **The arena and the cooked form are explicitly not moved** — their
  surfaces are templates over a sink or a context, so they have no single
  definition to emit, and no number has been taken for them.

Page changes: `SPEC-TABLES.md` §6.1 (file layout), §13.5 (the three rulings), §16
preamble and §16.1 (the `.cpp` rule and its build line), §16.5 (the gate reads
`.cpp`); `SPEC.md` §6.1; `USAGE.md`'s JSON example gains the compile step and says
the cost is opt-in. The `+11–15%` note needed no retiring — it never survived
#269's rewrite of §16.

## Gates

The **generic-walk gate moved with the walker**: it compares the `.cpp` files and
still finds one walker byte-identical across all **15**. Both negative controls
sabotage and compile the `.cpp`. Every consumer of the text form now compiles the
generated unit, through two shared variables so a build and its sanitized twin
cannot drift:

- `TABLES_JSON_SOURCES` — the tables leg and its ASan/UBSan twin (#278)
- `PACK_JSON_SOURCES` — the pack golden, the hostile-value gate, and both twins
  (#272, rebased onto it)

All green: tables (plain + sanitized), `tables-json-walk`, `tables-zero-cost`,
both JSON negative controls, `tables-pack` (both engines), `tables-hostile-values`
(67 trees), the two C# gates, `go test ./...`, `make check`.

One test states the property it always guarded rather than the file list it
happened to see: `TestTablesMoveNoGeneratedPacketByte` now allows a table to grow
`Table.h` **and** `Table.cpp`, and still fails if any packet file appears or moves.


gafferongames:
## Review — MERGE

Cold review, fresh clone at `44209af` (45e720f is the base — the pack PR),
branch never touched (`git status --porcelain` empty throughout). Up to date with
main, `mergeable: MERGEABLE`. Everything below was run.

**The measurement reproduces, the split is clean, and every gate is green.** Two
small defects and one banking gap, none of them blockers.

---

# FINDINGS

### D1 — a stale cross-reference in the emitter

`internal/codegen/cpptable/cpptable.go:534`:

```go
// The TEXT form's runtime, in its own translation unit (owner's
// ruling, SPEC-TABLES.md §13.3): the generic walk and this file's
```

The ruling is **§13.5** ("Header versus translation unit, ruled"). **§13.3** is
"The fixed-class constructs, ruled" — a different section, and the one place in
this PR where a comment points at the wrong page. One character.

### D2 — an all-VARIABLE table unit emits a dead `.cpp`

The emission guard is `if len(members) > 0`, and `members` counts variable-length
tables — which get no definitions. A unit whose tables are *all* pointered
therefore emits a full walker with nothing to call it:

```
$ cat AllVar.schema      # table Node { value int32, next *Node }  +  table Root { head *Node }
$ ./bin/schema generate --lang cpp --out out AllVar.schema
AllVar.h  AllVarTable.cpp  AllVarTable.h  AllVarWire.h

  1614 lines
  FromJson/ToJson definitions: 0
  walker present: 1
$ c++ -std=c++17 -Wall -Wextra -Werror -c AllVarTable.cpp -o out.o    # clean
$ nm -g out.o | grep -c ""
0                                    <- exports nothing at all
```

Not a correctness bug — the object is empty and the link is unaffected. But it is
**picked up by the globs this PR introduces**:

```
Makefile:346: TABLES_JSON_SOURCES = $$(ls build/tables-generated/*/*Table.cpp)
Makefile:393: PACK_JSON_SOURCES   = $$(ls build/tables-generated/examples/*Table.cpp)
```

so a glob-driven build compiles it every time: **median 0.07s per build, n=9, for
an object that exports nothing.** The corpus does not show it because every
pointers unit happens to carry fixed members too (`GraphTable.cpp` 6 definitions,
`MarksTable.cpp` 3, `PartsTable.cpp` 6). The guard wants "at least one FIXED
member", not "at least one member". Note the fix must be *not emitting the file* —
emitting an empty one would fail `tables-json-walk`, which requires a walker in
every `*Table.cpp`.

### D3 — §13.5 banks the BEFORE number and not the AFTER

§13.5 records *"measured at +11% to +15% compile time"* — the cost that motivated
the ruling — and that is the half my earlier review asked for. The half that
proves the ruling worked is only in the PR body:

```
$ git diff --stat origin/main...HEAD -- ROUND-LOG.md notes/
(empty)
```

CLAUDE.md's optimization doctrine asks for *"paired before/after (median-of-7,
same sitting)"*. The pair exists and it is good; the tree carries one side of it.
One line in `ROUND-LOG.md` closes this, and §13.5 is arguably the wrong home for a
number anyway.

---

# VERIFIED GOOD

### (1) The measurement — reproduced independently, two sittings

Three trees, all generating the same `Tables.schema` (`git diff 526239c HEAD --
tables/examples/Tables.schema` is **empty**, so this is apples-to-apples):
pre-walk `526239c`, before `45e720f`, after `44209af`. Empty TU including
`TablesTable.h`, arm64 clang, interleaved in one sitting, n=15 after two warmups.

```
sitting 1
prewalk    min 0.0459s  median 0.0466s  spread 3.5%
before     min 0.0518s  median 0.0529s  spread 23.5%
after      min 0.0467s  median 0.0475s  spread 18.5%
  after vs before   median -10.3%   min -9.9%
  after vs prewalk  median  +1.8%   min +1.7%
  before vs prewalk median +13.4%   min +12.9%

sitting 2
  after vs before   median -10.7%   min -10.7%
  after vs prewalk  median  +0.8%   min +0.8%
  before vs prewalk median +12.9%   min +12.8%
```

The reported **−9.8%** and **+1.5%** both sit inside my range, and min tracks
median in every row, so the noisy `before`/`after` spreads are not carrying the
result. `before vs prewalk +12.9–13.4%` also confirms §13.5's "+11% to +15%"
against my own earlier independent numbers on this axis (+10.9%, +12.8%).

Line counts are exact — **3656 → 1986** — and the residual is accounted for:
pre-walk was **1894**, so the +92 is #267/#265's descriptor columns (`json`,
`arms`, `reset`, `optional`, keyed) plus the three declarations per table, not the
walk. The walk is gone completely:

```
$ grep -c "TableJson" build/tables-generated/examples/TablesTable.h
0
```

and the three number-conversion includes went with it — header keeps
`cstdint/cstring/cstddef/new/type_traits/cassert`, the `.cpp` takes
`cstdio/cstdlib/clocale`.

### (2) The gate moved with the walker; both controls sabotage the `.cpp`

```
tables generic-walk gate: one walker, byte-identical in 15 generated .cpp files
json negative control: the round trip changed the wire — red, as required
keyed duplicate negative control: the repeat went uncounted — red, as required
```

Both rules now `sed` the `.cpp` and pass it on the compile line; both still assert
the sabotage applied before building.

### (3) Every battery green

```
./build/schema_test_tables            tables test passed
./build/schema_test_tables_asan       tables test passed        (ASan+UBSan twin)
tables-pack                pack golden: schema pack == Save and unpack == ToJson (both roots, both engines, + asan twin)
tables-pack-negative       eliding a present ?T turns the golden red
tables-hostile-values      67 trees agree on report and wire across both engines (+ asan twin)
tables-hostile-negative    a leading + packs once the grammar is relaxed
tables-cs-standalone       generated Table sources name no runtime
tables-cs-refuses-pointers a pointered unit is refused by name
test/cs-tables$ dotnet run cs tables test passed
go test ./...              20 ok, 0 failures
make check                 green
```

`TABLES_JSON_SOURCES` / `PACK_JSON_SOURCES` do keep each build and its sanitized
twin on one source list — the drift they exist to prevent is genuinely prevented.

### (4) The link contract behaves exactly as §16.1 says

```
(a) COMPILE ONLY, header alone:                          compile OK
(b) LINK without the generated .cpp:
      "tabledemo::ProfileConfigFromJson(tabledemo::ProfileConfig&, char const*,
       long long, tabledemo::TableReport*)", referenced from: _main
      "tabledemo::ProfileConfigToJsonMeasure(tabledemo::ProfileConfig const&)"
    ld: symbol(s) not found                              link exit=1
(c) LINK WITH build/.../TablesTable.cpp:                  runs -> "1 226"
```

Fully qualified, names the exact function, unmissable. USAGE gains the
`c++ -c ShipTable.cpp` step and states the cost is opt-in. *(It does not say what
forgetting looks like — the symbol above is clear enough that I would not add it,
but one clause would make the doc self-contained.)*

### (5) The page is present-state and the split is stated where it belongs

`SPEC-TABLES.md` §6.1 (file layout, "The TYPE wire has no such file and does not
want one"), §13.5 (all three owner verbatims — the ruling, "as we get more complex
stuff…", and "I think it's OK for types to remain header only"), §16 preamble
(walk is a translation unit; "no opt-out at the DECLARATION level… paying for it
is a build decision, not an include" — a careful distinction, and the right one),
§16.1 (the build line), §16.5 (gate reads `.cpp`); `SPEC.md` §6.1 carries the
header-only ruling for the type wire. §13.5's "arena and cooked form are not
moved — their surfaces are templates over a sink or a context, and no number has
been taken for them" is the honest form of that claim.

### (6) Only the walk moved, and nothing else shifted

Header still carries the descriptors and the wire codecs (`TableTypeInfo` 25,
`TableFieldInfo` 9, `offsetof` 46, `Measure(` 38, `Save(`/`Load(` 7 each).
Diffing main's header against this one **with the walk block excluded** shows
exactly three kinds of change and no others: the three number-conversion includes
leave, the `#ifndef …_SCHEMA_TABLE_JSON` guard leaves, and each `inline`
three-line wrapper definition becomes a three-line declaration. `<cstdlib>` is
correctly **restored** to the `anyVariable` branch it belongs to (the arena's).

```
generated/ after regeneration:  (empty git status = byte-unchanged)
generated/cpp/*.cpp:            none — examples/ declares no tables
tables-zero-cost rule:          BYTE-IDENTICAL to main
```

`TestTablesMoveNoGeneratedPacketByte` now states the property instead of the file
list, and still fails if a packet file appears or moves. Provenance clean: three
commits, present-state titles, all `Rowan Claude <rowan@mas-bandwidth.com>`, all
citing #283.

### A property this PR earns and does not claim

Moving to a real TU raises an ODR question the PR body never addresses, so I
checked it. All 49 walker symbols are **weak external** (vague linkage):

```
$ nm -m TablesT.o | grep -i tablejson | grep -oE "weak external|external" | sort | uniq -c
  49 weak external
```

so six same-package `.cpp` files (`Tables`, `Nested`, `Wide`, `Guarded`, `Keyed`,
`Pack`, all `namespace tabledemo`) link cleanly — verified by linking all six
objects together and running the result. **ODR requires those definitions to be
token-identical, and that is precisely what the generic-walk gate now proves,
byte for byte, across all 15.** The gate has quietly become the ODR guarantee as
well as the one-walk guarantee.

And because this is a game-facing project, I checked **unity builds** — six
same-package `.cpp` files `#include`d into one TU:

```
$ c++ -std=c++17 -Wall -Wextra -Werror -c unity.cpp     # exit 0, runs
```

which works only because the `#ifndef <PKG>_SCHEMA_TABLE_JSON` guard survived the
move into the `.cpp`. That guard is load-bearing, not vestigial — worth a comment
saying so before someone removes it as dead weight in a `.cpp`.

---

# VERDICT: MERGE

The owner's ruling is implemented exactly, the number is real and reproduces on an
independent measurement, and nothing moved that should not have. Not conditions:

1. **D1** — `cpptable.go:534` says §13.3, means §13.5.
2. **D2** — gate the `.cpp` on a FIXED member, not on any member; today an
   all-pointered unit ships 1614 dead lines that the build's own glob compiles.
3. **D3** — bank the after-number in `ROUND-LOG.md`; §13.5 has the before.
4. Optional: one line in the `.cpp` saying its include guard is what makes unity
   builds legal.

Issue #290, opened 2026-09-02
Dogfood finding: §2.4's "prefer Slot<Key>(), which is most places" is ZERO places in a data-driven engine — reword the guidance


gafferongames:
From the space dogfood (space#443): keys in a data-driven engine are runtime values (enum fields read from data, loop indices), so the compile-time accessor `Slot<Key>()` was used zero times and every access went through `operator[]( E )`, whose guard compiles out under NDEBUG. Consequences: (1) §2.4's guidance sentence is wrong for the primary consumer — reword to say the compile-time form is for literal keys and the runtime form is the common case; (2) the runtime guard being compiled out in release is therefore the NORMAL case, which the page should say plainly, and which makes the iteration surface (the sibling issue) the real safety: iterate valid slots and the None question does not arise. Spec-only, one paragraph.

Issue #289, opened 2026-09-02
Dogfood finding: iteration over an enum-keyed array — every call site wants `for ( auto & ship : root.ships )` instead of `for ( int i = 1; …) root.ships[(ShipType) i]`


gafferongames:
From the space dogfood (space#443, "What the dogfood taught the language"): the one construct the branch actively wished for. Every consumer of a keyed array iterates the valid slots, and today spells it as an integer loop from 1 to Max with a cast at each site — the slot-0 rule re-derived by hand everywhere. The keyed-array type should offer iteration over VALID slots (1..Max) yielding (key, element) — a range-for in C++, an enumerator in C#, the equivalent per port — so slot 0 never appears at a call site and the idiom dies. Spec first (§2.4 gains the iteration surface); C++ and C# together; then space adopts it.

gafferongames:
Dispatched with #290 and subtraction item 6: keyed iteration over valid slots lands in C++ and C# together, and `Slot<Key>()` is REMOVED (Rowan's decision under the owner's delegation: measured zero uses in the dogfood; iteration is the real safety; "perfect only when there is nothing left to take away").

Issue #288, opened 2026-09-02
Expressiveness gate 2 — THE RENDER DATA: using tables or types, the per-frame render data must be implementable end to end (C++ producer, C# consumer pointing at it, no hand mirror)


gafferongames:
Owner requirement, 2026-09-02 (Glenn, live): "New requirement just dropped. Using tables, or types, we should be able to implement the render data."

**What the render data is** (space): the client hands the host engine, every frame, a block of fixed-stride sections of fixed-size records (RenderHeader, RenderCamera, RenderShip, RenderTurret, …, declared today as tables in Render.replicant, generated in C++, consumed in Unity by POINTING at the memory through blittable structs — no parse, no copy). Today the C# side is a HAND-WRITTEN mirror (space#444; schema#287) and the layout contract is hand-kept.

**The gate.** Beside §12's Config.bin/Assets.bin gate, a second: the render data expressed as schema declarations with nothing left over, and BOTH sides generated — the C++ struct and a blittable C# struct with byte-identical layout, size and every offset asserted on both sides by generated code (the layout id's facts, made a cross-language contract), plus fixed-stride sections (replicant#39's strided arrays: a declared, generated, enforced stride). The C# consumer points at the C++-written block and reads it as is. Whether that is a fixed TABLE (struct semantics, tolerant wire beside it) or a TYPE (raw struct, same-build) is the design question this issue decides on the page first — the owner said "tables, or types".

**Spec first:** SPEC-TABLES.md §12 gains the second gate; the layout-contract construct (blittable emission, stride, the cross-language asserts) is designed on the page, cold-read, then built — C# first (#287), then space's render path moves onto it and the hand mirror dies (#444). Gate item for schema#248, after the Config/Assets dogfood lands.

gafferongames:
Owner, same minute (verbatim): "*including* the blittable arrays with stride." / "probably with tables, I would imagine. since render data is BIG" / "so it's more table like than for example, a type." → Scope explicitly includes the strided blittable arrays; the design lean is TABLES (size makes it table-shaped): the render block as fixed tables laid out at a declared stride and pointed at — the fixed class's region form, per frame. The page decides; this is the owner's lean, not a ruling closing the type option.

gafferongames:
[...]

Issue #287, opened 2026-09-02
PROMINENT — Unity's render path runs on a HAND-WRITTEN struct mirror; the generated C# render classes are dead code. The C# backend needs a blittable struct form so the layout contract is generated


gafferongames:
Owner, 2026-09-02, on render being generated: "That was always the intent." It is: space's render structs are generated on both sides (game/Generated/cpp/Render.h used by RenderManager; game/Generated/cs/Render.cs copied into Unity's CodeGen). What survives by hand is Unity's interop mirror — unity/Assets/UnitySpaceGame/Network/ClientInterop/Render.cs declares `public struct RenderShip` beside the generated `public sealed class RenderShip` — because the C# emitter produces CLASSES and the per-frame path across the dll boundary needs a BLITTABLE struct to read the fixed-stride sections without copying.

**The gap:** the C# backend (types and tables) emits no blittable struct form. **The follow-on:** an emission mode (attribute or flag) producing `[StructLayout(LayoutKind.Sequential)]` blittable structs whose layout matches the C++ struct byte for byte (static-asserted on both sides: sizeof/offsetof pinned, the same instrument RenderManager's static_asserts are), so the hand mirror retires and the layout contract is generated, not maintained. Related: replicant#39 (strided arrays for the render tables). After the dogfood; a C# port item.

gafferongames:
**Owner asked for this to be prominent — he was not aware of it.** Verified by command in space (main dbd97226 + #442):

- Unity's render path (`NetworkRenderSyncSystem.cs`, `NetworkRenderSyncSystemJobs.cs`) uses `UnitySpaceGame.Network.ClientInterop.RenderShip` etc. — the HAND-WRITTEN blittable structs in `unity/Assets/UnitySpaceGame/Network/ClientInterop/Render.cs` (last touched 2026-08-12, when the RenderX types moved into Render.schema), with explicit `using RenderShip = …ClientInterop.RenderShip` aliases.
- The GENERATED `unity/Assets/CodeGen/Schema/Render.cs` and `RenderTable.cs` are referenced by NOTHING outside CodeGen. Dead code carried into every Unity build.
- So the per-frame render layout contract is held by hand on the C# side and by static_asserts on the C++ side; nothing generates the C# half. A field added to a render table in the schema reaches the C++ struct and never the Unity struct until someone edits the mirror.

[...]

PR #291, opened 2026-09-02
SPEC: the block form — the render data as an ordinary table (#288, spec-first)

gafferongames:
Spec-only, against `main` (da9a87f). No compiler code. `make check` green.

**Reshaped after the cold read** ([findings](https://github.com/mas-bandwidth/schema/pull/291#issuecomment-5505463965), verdict DO-NOT-LAND as drafted) and the owner's direction on this thread. The first draft's `section` construct is gone.

## What the page says now

A fixed table may be marked **`| block`**. That gives the same declaration a **third projection** beside its wire (§3) and its cook (§7): one in which the table's own bounded arrays are laid out of line at a fixed pitch, and the instance at the front of the block carries, per array, an `(offset_of, count, stride)` triple — what the table already knows about its own rows, written where the other side can read it.

```
table RenderFrame | block
{
    version uint64
    cameras [..1]RenderCamera
    ships   [..MaxShips]RenderShip
    lasers  [..MaxLasers]RenderLaser
}
```

**Nothing there is new.** Ordinary bounded arrays of ordinary fixed tables; `RenderFrame` keeps `Measure`/`Save`/`Load` and its cook. The rule the dropped keyword was carrying is two sentences: **depth one, bounded only** (the marked table's own `[..N]` arrays move; fixed, keyed, and anything inside a row stay), **elements are structs**, **the pitch is `sizeof`**, **the wire is untouched**.

§19 is the two halves the owner named — *"a builder issue in C++ and a reflection issue in C#"*: §19.1 the builder (layout from counts, wide fill), §19.2 the reflective read (§8's descriptors carry the projection's layout, which is what retires the mirror; the generated blittable struct is the typed fast path beside them).

## Reconciliations — what changed from the draft, and why

| draft | now | why |
|---|---|---|
| a `section` field spelling | no construct; `\| block` on the table | *"why even have a 'section'?"* / *"header just collapses into stuff that a table knows about its own rows"* |
| block root has **no** table wire (decision 10) | ordinary table, full wire | the draft's argument was **circular** (D8): a wire form has no triple to ride. A bounded array of fixed tables already rides as kind `14` / element kind `13`. No kind spent, §3 untouched |
| a *fourth form* | the **third form** of a fixed table | one declaration, three projections |
| §2.2 gains a section-mode paragraph | §2.2 **unedited** | a bounded array of fixed structs is already a by-value edge; a form is not a mode |
| §6 gains a "third life" | removed | a block-form table has exactly the fixed class's life |
| `\| stride` triggers the form | `\| block` **alone** | the case has nine strided arrays and **zero** declared strides — a trigger the primary case never fires is not a trigger |
| `\| stride = N` in v1 | **dropped**, §15 follow-on | the only consumer that exists cannot read a pitch that is not `sizeof`; and D1's `<=` check already absorbs the edit headroom was for |
| block root refused as nested / pointed-at / union arm | **not** refused | its by-value form is an ordinary table; only the block form needs a base, as a cook needs a file's |

## The read's defects, all carried

- **D1** — the tolerant open checks `sizeof(element) <= the pitch it READ`, never an equality against its own constant. The draft's check refused the very edit it was named to absorb.
- **D2** — the block-form table's **own** field offsets are in the baseline and the digest. A scalar inserted before an array moves every triple; the draft judged that nowhere.
- **D3** — the C# asserts are named to the **managed unmanaged-struct model**, not the marshalling model, and **a `bool` in a row is one byte**. The two models disagree; an unnamed model can pass on one measurement and garble on the other. A `bool` row is pinned in the corpus.
- **D4** — storage lifetime stated: a block is valid until the next `Begin` on that storage, which invalidates every row pointer; double-buffering is two storages and is the caller's.
- **D5** — the multi-threaded fill is an **obligation** (*"Hard requirement"*), held by a refuser — no allocation, no lock, no atomic in the generated fill path, build fails if one appears — beside a real multi-threaded fill in the corpus. The draft stated a permission a serial backend would have satisfied.
- **D6** — `<Table><Field>Span` claimed in §11; 27 suffixes → **36**.
- **D7** — the digest sees layout, not meaning; it does not replace a semantic version field.
- **D8** — §14 records the reconciliation, not the circular rejection.
- **R2–R5** — `Begin`'s refusal names the array, its count and its maximum (clamping is the producer's job; counts are gathered before `Begin`, which is single-threaded); the shared-bound looseness stated with its number (7.51 MiB vs a 10 MiB budget, ~25% unoccupiable) and a §15 follow-on; the 64-byte guarantee stated as **per-array**, with the intra-array boundary case named; §12.1's no-copy claim scoped to **the boundary**.

Present-state fixes: §19's "its weight has changed" narration and the "person who remembers" sentence are gone; provenance lives in §13.6.

## Ranking and provenance folded in

§12.1 now ranks the path from the owner's words: the wire and the cook are **read-hot and write-cold** (built offline by tooling and a distributed build cache — §7), while the block form is hot on **both** sides at 60 Hz or better — *"It is the hottest path for tables to be sure"* — **pre-cooked at build**, with two measured numbers: the per-frame C++ write and the per-frame C# read.

§7 gains how a cooked artifact is addressed: the pair **(asset hash, cooked-version hash)**, the latter being the build's table protocol id — which on this page is the layout id. A cook and a block never share one.

## Verification

- The nine-row worked layout in §19.1 is checked to the digit by a scratch Go program written from the page's own rule (deleted): starts **192 / 320 / 26752 / 84352 / 92992 / 95872 / 495872 / 559872 / 572672**, used extent **577,472** — identical to the hand-written layout the form must replace, at every one of the nine. The prologue is free in this shape (152 and 168 both round to 192).
- `make check` green over the whole corpus. No schema sources, no Go, no goldens touched.

## Open for the owner

One item: **whether declared stride headroom (`| stride = N`) should return in v1** rather than staying a §15 follow-on. The page takes the read's overturn — the evidence is nine arrays and zero declared strides, plus a consumer that loses its cast path on any pitch that is not `sizeof`, plus D1's `<=` check already absorbing the append it existed for. §15 carries the owner's sentence and the reason.

Everything else the draft left open is now ruled or decided: spelling (no construct), stride derivation (`sizeof`, always), the table wire (kept), nesting (depth one), block-relative offsets (kept — the artifact reads the instance by value, which a self-relative delta would not survive), the trigger (`| block` alone), and the C# model (managed).

Related: #287 (the typed fast path), #289 / #290 (the iteration lesson §19.2 applies).

### DIFF
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -50,11 +50,25 @@ licensed. Unexplained slowness is still a defect, in every rung.
   neutral wire allows. That is a bench obligation, not an intention: a
   fixed table sits beside its equivalent type on the ledger, and the gap
   is explained or closed. The zero-cost gate (§2.2) holds the storage
-  half of it; the bench leg holds the codec half.
+  half of it; the bench leg holds the codec half. **A fixed table and a
+  `type` are semantically the same thing — a struct.** What separates them
+  is the wire each rides and the versioning that comes with it, and nothing
+  else; the constructs below are the table wire's, not a different idea of
+  what a value is.
 - **The variable-length class pays for exactly what it buys** — pointers,
   an arena on the building side, a region on the reading side, a
   reference indirection per edge. This is where "less performant" lives,
   and it is deliberate.
+
+**Where the effort goes is decided by which SIDE is hot, and the forms
+differ.** The wire and the cooked form are **read-hot and write-cold**: they
+are written once, offline, by a tool or a build pipeline (§7's pipeline
+paragraph says where that runs and how its output is addressed), and read by
+the game. Optimise their readers; their writers are a build cost. **The
+BLOCK FORM (§19) is the one form hot on BOTH sides** — written every frame in
+one language and read every frame in another — so it is the hottest path
+tables have, and the fastest-correct mission bites hardest there. §12.1 is
+the gate that decides whether it earns its place.
 
 ## What allocates, and what never does
 
@@ -83,10 +97,10 @@ follow-on. Every other backend refuses a unit that declares tables at all, by
 name, with this document cited. The remaining per-language backends are named
 follow-ons (§15).
 
-**The SECTIONED BLOCK (§2.7, §19) is specified and unimplemented.** No backend
-emits it yet: a unit declaring a section is refused by name, with §19 cited,
+**The BLOCK FORM (§2.7, §19) is specified and unimplemented.** No backend
+emits it yet: a table marked `| block` is refused by name, with §19 cited,
 never emitted with the block surface missing. C++ and C# take it together,
-because the construct is an ABI between two languages and one language alone
+because the form is an ABI between two languages and one language alone
 cannot hold the gate it exists for (§12.1).
 
 ## 1. Purpose
@@ -127,7 +141,7 @@ table ShipConfig
 A table body is a type body — the field grammar of SPEC §4.2, hosted by
 `table`: bare and ranged integers, `bits(N)`, `bool`, floats and
 compressed floats, enums, flags, strings, bytes, bounded arrays, unions,
-`if` branches, and declared types as field groups. Seven additions:
+`if` branches, and declared types as field groups. Six additions:
 
 - **Tables nest.** A named table is a field type (above); nesting is by
   value, and a bounded array of tables is a collection. A table may not
@@ -146,11 +160,12 @@ compressed floats, enums, flags, strings, bytes, bounded arrays, unions,
   buffer at exactly its used size, and `*string` is its sibling.
 - **A union arm may be a table** (§2.6), which is what makes an evolvable
   message set expressible.
-- **A field may be a SECTION** (§2.7). `ships section [..MaxShips]RenderShip`
-  declares a strided array stored out of line in one block, with an
-  `(offset, count, stride)` triple in the declaring table — the header-plus-
-  sections shape of a per-frame render block, made language (§19).
 - **`was` — the rename attribute** (§5).
+
+And one addition that is not a field spelling at all: **a fixed table may be
+marked `| block`** (§2.7), which gives it a third FORM beside its wire and its
+cook — the same declaration laid out so another language points at its rows
+(§19). It adds no field type, no wire kind and no byte on the wire.
 
 ### 2.1 Pointers
 
@@ -231,15 +246,11 @@ Pointer edges do not propagate the mode: a table that is merely POINTED
 AT stays fixed-size if it holds no pointer of its own. It gains an
 allocation and a resolution entry, and nothing else.
 
-**A SECTION is by-value storage, so a block root is FIXED-SIZE** (§2.7). A
-section field's storage is the sixteen-byte `(offset, count, stride)`
-triple, which holds no pointer and is blittable like every other by-value
-field, so a table that declares sections is a plain struct and the mode
-derivation has nothing to propagate: the ELEMENT of a section is required
-to be fixed-size in the first place (§11), and the strided records live in
-the block rather than in the header. A section edge is therefore not a
-by-value edge for the derivation above — it reaches no variable-length
-table, because it cannot name one.
+**`| block` does not enter this derivation at all** (§2.7). A block-form
+table declares nothing new: its fields are the bounded arrays of fixed
+structs it always had, which are by-value edges under the rule above, so it
+derives FIXED by that rule unchanged. The marker selects a FORM, and a form
+is not a mode.
 
 **A fixed-size table pays nothing for the VARIABLE-LENGTH machinery**, and
 that is a gate, not a hope: in a unit whose tables are all fixed-size the
@@ -256,13 +267,13 @@ emitted in every unit, whatever its mode, because a fixed-size table can
 declare it and a tool walking a fixed-size table has to find it. The gate
 asks "did the pointer world leak in?", never "did the descriptor grow?".
 
-**The BLOCK machinery takes the same gate, on the same terms** (§19): a
-unit that declares no section carries no block storage type, no `Begin`, no
-`Open`, no strided iterator and no layout constants — the build fails if
-one symbol of it appears — while the section's descriptor COLUMNS (§8) ride
-in every unit, because they describe the language and a fixed-size table
-can declare a section. Machinery is gated; columns are not, which is the
-rule the paragraph above already states.
+**The BLOCK machinery takes the same gate, on the same terms** (§19): a unit
+in which no table is marked `| block` carries no block storage type, no
+`Begin`, no `Open`, no row iterator and no layout constants — the build fails
+if one symbol of it appears. No marker, not one symbol. The descriptor
+COLUMNS (§8) the block form reads ride in every unit as every other column
+does, because they describe the language. Machinery is gated; columns are
+not, which is the rule the paragraph above already states.
 
 The assumption behind the split, stated so nobody quietly designs against
 it: size and mode correlate in practice. Value-only tables are assumed
@@ -526,78 +537,67 @@ payload could never express.
 - **A backend without a native union may allocate for the arm** (the
   ladder, above): the carve-out is the language's, not the table's.
 
-### 2.7 Sections: `ships section [..MaxShips]RenderShip`
+### 2.7 The block form: `table RenderFrame | block`
 
 ```
-table RenderFrame
+table RenderFrame | block
 {
-    cameras section [..1]RenderCamera
-    ships   section [..MaxShips]RenderShip
-    lasers  section [..MaxLasers]RenderLaser
+    version uint64
+    cameras [..1]RenderCamera
+    ships   [..MaxShips]RenderShip
+    lasers  [..MaxLasers]RenderLaser
 }
 ```
 
-A SECTION is a **strided array of a fixed-size element, stored out of line
-in one block**, with an `(offset, count, stride)` triple in the declaring
-table. A table that declares one is a **BLOCK ROOT**: its instance is the
-HEADER at the base of a block, the sections follow it in that same block,
-and §19 is the form — how the block is laid out, filled by many threads,
-and pointed at from another language.
+**`| block` declares no construct.** Every field above is an ordinary field
+of an ordinary fixed table: a scalar and three bounded arrays of structs,
+which this document has always had. The marker says that this table has a
+THIRD FORM beside its wire (§3) and its cook (§7) — one in which its own
+bounded arrays are laid out of line at a fixed pitch, so a consumer in
+another language points at their rows instead of parsing them (§19). One
+declaration, three projections of it; **it's tables all the way down.**
 
-**What the construct is FOR is speed across a language boundary.** A
-section's records sit at a DECLARED PITCH that both generated sides carry as
-a constant and the header carries as data, so a consumer written in another
-language points at a record and reads it: no marshalling, no copy, no parse,
-no per-element call. That is the whole reason the pattern exists in
-hand-written form and the whole reason it becomes a declaration; everything
-below is in service of it.
+**The rule, in full, because it is the one thing a reader has to know that
+the declaration does not spell:**
 
-- **Storage in the header is SIXTEEN BYTES**: `offset (u64)`, `count (u32)`,
-  `stride (u32)`, in that order, with no interior padding. It is by value, it
-  holds no pointer, and it is blittable, so a block root is FIXED-SIZE
-  (§2.2) and its header is one memcpy-able struct like any other table.
-- **The offset is BLOCK-RELATIVE** — the byte distance from the block's
-  BASE, not from the slot that holds it. A region reference is self-relative
-  (§6.3) because nothing there hands a walker a base; a block's consumer is
-  handed the base by construction, and a header read BY VALUE — which is how
-  a blittable consumer reads it — keeps working, where a self-relative delta
-  would not survive the copy (§14).
-- **The bound is a MAXIMUM, and the spelling is `[..N]`.** `N` is the ceiling
-  the block's STORAGE is sized from (§19.1); `count` is the runtime fact, and
-  it is the number the producer declares before the block is laid out. Every
-  other array spelling is refused in section position (§11): a
-  fixed `[N]T` would make the count a constant, and a keyed `[E]T` is
-  complete by construction and has no count at all.
-- **The STRIDE is derived, and may be declared.** Derived, it is `sizeof` of
-  the element rounded up to the element's own alignment — which for a
-  standard-layout struct (§9) is `sizeof` itself, so the common declaration
-  needs no attribute and the dogfood's own render records express with none.
-  `| stride = N` declares it instead, and `N` is refused when it is smaller
-  than `sizeof` or is not a multiple of the element's alignment (§11).
-  Declaring a larger stride buys HEADROOM — bytes past the record's end,
-  inside its pitch — and it costs the consumer's fast path, which is the
-  honest trade §19 states.
-- **The ELEMENT is a fixed-size `table` or a declared `type`**, and nothing
-  else. A pointer, a `*bytes`, a `*string`, a variable-length table and a
-  table that itself declares a section are each refused by name (§11): a
-  block is one flat pointer-free extent, and a record whose meaning depends
-  on anything outside it is not a record a consumer can point at.
-- **Sections do not nest.** A section's element declares no section, and a
-  block root is not nested by value, pointed at, made a union arm or used as
-  an array element (§11). One block, one header, one level of sections. A
-  block root's section offsets are meaningful only when it sits at a block's
-  base, and refusing every other placement is what makes that true by
-  construction rather than by care.
-- **`?` and `*` are refused on a section** (§11): a section with a count of
-  zero already IS an empty one, and a block holds no pointers.
+- **DEPTH ONE, BOUNDED ONLY.** In the block form, every **bounded** array
+  field **of the marked table itself** — `[..N]T` — is laid out of line, in
+  declaration order, each at its own pitch. Everything else stays exactly
+  where it is: a fixed `[N]T`, an enum-keyed `[E]T`, and **every array at any
+  depth inside an element**. An element is one flat record or it is not a
+  record another language can point at, and one level is what makes "which
+  arrays move" answerable by reading one declaration.
+- **THE INSTANCE IS WHAT MOVES, not the schema.** The block form generates
+  the table's struct as a PROJECTION: each out-of-line array's inline storage
+  — `T[N]` and its count companion — is replaced, AT THAT FIELD'S POSITION,
+  by `(offset_of u64, count u32, stride u32)`, sixteen bytes with no interior
+  padding. Every other field keeps its by-value storage at its natural offset.
+  That projection is what sits at the front of the block, and nothing declares
+  it: it is what the table already knows about its own rows, written down
+  where the other side can read it.
+- **THE ELEMENTS ARE STRUCTS.** A bounded array takes the out-of-line form
+  only when its element is a fixed-class `table` or a declared `type` — the
+  same thing, semantically (the ladder, above). A variable-length table, a
+  `*T`, a `string` or `bytes` element is refused by name in a block-form table
+  (§11): none of them has a fixed pitch, and striding what has no pitch means
+  nothing.
+- **THE PITCH IS `sizeof`.** A row's stride is the element's `sizeof`, rounded
+  up to its alignment — which for a standard-layout struct (§9) is `sizeof`
+  itself. It is derived, always, and no declaration adjusts it in this
+  version; declared headroom is a named follow-on (§15) with the reason it is
+  not here. The stride still RIDES in the triple, because it is the pitch the
+  consumer indexes with and it must come from the data, never from the
+  consumer's own constant (§19.2).
+- **A TABLE, NOT A NEW KIND OF TABLE.** A block-form table keeps everything
+  an ordinary fixed table has: `Measure`, `Save` and `Load` over the tolerant
+  wire (§3), its cook (§7), its reflection descriptors (§8), its place as a
+  nested field or an array element in someone else's table. `| block` adds a
+  form; it removes nothing. That is what makes "is the render data a table?"
+  answerable with one word.
 
-**A section has NO WIRE KIND, and a block root therefore has no table wire.**
-§3's set is closed, so a section's triple cannot ride, and nothing pretends
-otherwise: a block root gets no `Measure`, no `Save` and no `Load`, and the
-BLOCK is its serialized form. Its ELEMENTS are ordinary fixed-size tables and
-keep their whole table wire, so `RenderShip` saves and loads tolerantly like
-any other record — it is only the header that is wire-less. That is the
-fourth form's honest boundary and §19 states what it costs.
+**`| block` is refused on a `type`, and on a table whose closure is not all
+structs** (§11). A type has no table wire to be a third form beside, and a
+table that reaches a pointer has no fixed pitch anywhere in it.
 
 ## 3. The wire
 
@@ -689,12 +689,15 @@ schema's codebase:
   **A kind a reader does not know at all is not skippable** and is framing
   damage, which is why the set is closed and why a new kind is a wire
   change rather than an addition.
-- **A SECTION has no kind and never rides** (§2.7). The set above is closed,
-  so the `(offset, count, stride)` triple has no framing here and a block
-  root has no wire form at all — the BLOCK (§19) is its serialized form. A
-  section's ELEMENTS are ordinary fixed-size tables and their own wire is
-  untouched, so the records a block carries are savable and loadable
-  tolerantly wherever something wants that; only the header is wire-less.
+- **The BLOCK FORM moves no byte on this wire, and spends no kind** (§2.7,
+  §19). A block-form table's bounded arrays ride here exactly as every other
+  bounded array of tables does — kind `14`, element kind `13`, the LIVE
+  count — and `| block` is not a wire fact at all: it selects a second
+  projection of the same declaration, the way a cook (§7) is a third. The
+  `(offset_of, count, stride)` triple is a block-form artifact and has
+  nothing to ride here, because a wire form has no triple. So a tool can
+  `Save` and `Load` a frame, and a diff of two frames is an ordinary table
+  diff.
 - **Field ORDER within a body is not part of the contract.** This
   implementation writes fields in declaration order, and a reader must not
   rely on it: every field is found by its id, so any order decodes the same
@@ -1363,16 +1366,19 @@ The mode (§2.2) decides the shape of the API, because the two classes are
 genuinely different things at runtime and neither should be contorted to
 look like the other.
 
-**A block root has a THIRD life and it is §19's.** A table that declares a
-section (§2.7) is fixed-size, so its header is a value like any other, but
-its instance lives at the base of a block rather than on its own and it has
-no wire form at all. The surface below is the wire's; the block's is stated
-where the block is.
+The block form (§19) is not a third life: a block-form table is fixed-size
+and has exactly the life below. What changes is how one instance is laid
+out, not what kind of thing it is.
 
 ### 6.1 Fixed-size: a value
 
-A fixed-size table is a struct. Its whole surface is three free
-functions, name first:
+A fixed-size table is a struct. **A fixed table and a `type` are the same
+thing semantically — both are structs — and what differs between them is the
+wire each rides and the versioning that comes with it.** Everything below is
+that wire's surface, not a different notion of a value; and a third form of
+the same struct is §19's.
+
+Its whole surface is three free functions, name first:
 
 ```cpp
 ChatMessage msg;
@@ -1588,6 +1594,15 @@ load time does not demand the accelerator, the game just uses the generic
 table: cooking is a choice made per asset, never a requirement of the
 format.
 
+**A cooked artifact is CONTENT-ADDRESSED by a pair — the hash of its source
+asset, and the cooked-version hash, which is this build's table protocol id
+and on this page is the LAYOUT ID (below)** — and that pair is the key the
+runtime searches for, the key a distributed build cache produces under and
+serves from. It is why the cooking side is a build cost rather than a
+runtime one: the work happens offline, once per (asset, layout), and the
+game does a lookup. That is the fact the performance ladder cites when it
+calls the wire and the cook read-hot and write-cold.
+
 The example pair is the whole rule. A huge data file naming every mesh in
 the game, or every texture — that is a cook. A configuration file small
 enough that the cost of loading it does not matter is the wire, and stays
@@ -1729,11 +1744,19 @@ for it. Here the tolerant wire stays the format of record and the cooked
 form is a build-locked accelerator beside it, produced only where load
 time asks for one. The two-form split is the design.
 
+**A COOK and a BLOCK are different accelerators and must not share a layout
+id.** Both are build-locked projections of one declaration, but they lay it
+out differently — a cook writes the by-value form verbatim, a block writes
+the projection with its rows out of line (§2.7) — so one id covering both
+would let a runtime accept one where the other was written. Each carries its
+own id, computed over its own facts, and neither is checked against the
+other's.
+
 **And prior art gets one MEASURED sentence, from the case §19 exists for.**
 The render data this document's second gate is held to (§12) was built with
 flatbuffers once and the build was abandoned — in the owner's words, *"I
 used to use flatbuffers to build render data, but it was too slow because it
-was not parallizable."* That is the specific failure the sectioned block is
+was not parallizable."* That is the specific failure the block form is
 shaped against: a per-frame producer that has to go wide cannot afford a
 builder with a serialization point in it, whatever the read side costs. A
 cook does not answer it either — a cook is produced from a builder by a
@@ -1810,15 +1833,22 @@ listing.
 offset beside the reference, so a walker reads the blob's extent the way
 it reads an array's count (§2.5).
 
-**A SECTION field carries its element's descriptor and the three offsets of
-its triple** — `section` marks the field, `table` names the ELEMENT's
-descriptor, and `offset_offset`, `count_offset` and `stride_offset` name the
-three members inside the sixteen-byte triple (§2.7) — so a tool handed a
-block and its header's descriptor walks every section without knowing the
-spelling that produced it, reading the pitch from the data rather than
-assuming it. The declared MAXIMUM rides beside them in the column an array's
-bound already uses, and a table's own **`block`** flag says it is a block
-root, so a tool can tell a header from an ordinary fixed table (§2.7).
+**A BLOCK-FORM table carries a second set of positions for the same fields,
+and this is what makes the block form READABLE BY REFLECTION** (§19.2). A
+table's own **`block`** flag says it has the form; each out-of-line array
+field then carries, beside the offset its by-value storage has, the
+**projection offset** of its `(offset_of, count, stride)` triple and the
+offsets of the three members inside it, with the ELEMENT's own descriptor
+already in the column a nested table uses. Every other field carries its
+projection offset too, because the projection is a different struct from the
+by-value one (§19.3) and a walker over a block needs the positions that
+struct actually has.
+
+That is the whole mechanism behind the block form's read side: a consumer
+with the descriptors reads the triples out of an instance and points at rows,
+with no hand-written struct per table and no knowledge of the spelling that
+produced any of it. The descriptors are constant data, so this costs a
+lookup, not a parse.
 
 These columns exist in every unit, whatever its mode — they describe the
 LANGUAGE, and a fixed-size table can declare all of them. Only the two
@@ -1882,18 +1912,20 @@ held by construction:
 - **Going wide on the BUILDING side** is §6.4: allocation is thread-local
   and nothing ever moves, so N workers fill one arena with no lock and no
   per-node atomic.
-- **The BLOCK is the strongest form both properties take** (§19), and it is
-  where the requirement for them came from. A block is one flat extent whose
-  every reference is a block-relative offset, so it relocates by plain
+- **The BLOCK FORM is the strongest form both properties take** (§19), and it
+  is where the requirement for them came from. A block is one flat extent
+  whose every reference is a block-relative offset, so it relocates by plain
   `memcpy` with no fix-up at all — no `Lock`, because it is born compacted.
-  And it goes wide with nothing to synchronize: its storage is sized from the
-  declared maxima and its layout is fixed in one pass over the SECTIONS
-  before any worker runs, so every record's address is settled ahead of the
-  scatter, N workers fill disjoint records, and there is no arena, no
-  allocation, no atomic and no per-node prologue pass. The
-  builder's four models (§14) all exist because a general structure cannot
-  know its bound; a block declares one, which is why item 3 there — reserve
-  the max and never resize — is exactly what a block does.
+  **And going wide there is an OBLIGATION, not a permission**: the layout is
+  settled from the counts in one pass over the table's out-of-line arrays
+  before any worker runs, so every row's address is known ahead of the fill;
+  N workers then own disjoint index ranges and write with no per-row
+  synchronisation and no lock. A generated fill path that allocates, locks or
+  takes an atomic does not conform, and §19.1 states that as a refuser rather
+  than as an aspiration. The builder's four models (§14) all exist because a
+  general structure cannot know its bound; a block-form table declares one,
+  which is why item 3 there — reserve the max and never resize — is exactly
+  what the block form does.
 
 ## 10. Independence from the hardcoded wire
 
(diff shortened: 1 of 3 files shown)


gafferongames:
Owner rulings, 2026-09-02 (Glenn, live): "1/2 = yes" — (1) the `section [..N]T` spelling stands; (2) stride derives by default (= sizeof) with `| stride = N` as declared headroom. Items 3 (no nesting), 4 (block-relative offsets), 5 (no table wire for a block root) remain open.

gafferongames:
Owner asks (verbatim): "is render data a table? are the render types being strided types or tables?" / "can render data root be a table?"

On the draft: the block root IS a table (declared `table`, fixed class), its sections are strided arrays of fixed tables; decision 10 withholds the TABLE WIRE from a block root. Rowan's recommendation for open item 5, toward the owner's lean: the block root is an ORDINARY fixed table — it has the table wire like any other (a section rides as a bounded array of tables with the present count; nothing new on §3's kind set), and the sectioned block is the table's THIRD FORM beside the wire and the cook: the wire for tools and versions, the block for pointing at across the C++/C# boundary. Then decision 10 dissolves, the zero-cost gate holds for units without sections as before, and "is the render root a table" is answered with one word. Elements: fixed tables (a plain `type` may also be an element by value, as anywhere). The cold reader should weigh this against the draft's rejection in §14.

gafferongames:
Owner (verbatim): "if it is, then we can make tables more powerful." — The direction for the revision after the cold read: the block root is a TABLE, and the sectioned block is a FORM any fixed table with strided sections can take (wire / cook / block), not a separate construct.

gafferongames:
Owner ruling (verbatim): "i would suggest that render data is a root table, with strided types or fixed tables being allowed. strided making no sense for non-fixed tables or types." → For the page: the block root is a ROOT TABLE; a section's element is a fixed-class TABLE or a plain TYPE (both fixed-size, blittable); a section whose element is variable-length (a table with a pointer in its closure) is REFUSED by name — no fixed pitch exists. `?T` elements and keyed arrays as elements: decide on the page (both are fixed-size; a keyed array inside a strided element is fine; an optional element is a record with a presence bool, fixed-size — allowed unless a reason appears).

gafferongames:
Owner (verbatim): "in fact, fixed tables and types are semantically the same (structs) they just have different wires and versioning stuffs." — the element rule in one line: a section holds structs.

gafferongames:
Owner (verbatim): "why even have a 'section'?" — Rowan's answer, for the revision: none is needed. With the root an ordinary fixed table, a 'section' is what the BLOCK FORM does with a bounded array — lays it out of line at a pitch. Declaration side: NO keyword; the only addition is the attribute `| stride = N` on a bounded array of structs (headroom the compiler cannot derive); `ships [..MaxShips]RenderShip | stride = 128`. The block form of a fixed table lays out every bounded-array field out of line at its derived-or-declared pitch, header first with (offset, count) per array; the by-value form keeps inline storage as today. A construct becomes an attribute plus a form — the language's usual shape.

gafferongames:
Owner (verbatim): "feels like the strided array and the section are the same thing, what you really need is a new concept of header with offset_of" — The concept for the page, named by the owner: THE HEADER. The block form of a fixed table generates a HEADER struct (both languages) carrying, per bounded-array field, (offset_of, count, stride) — offset_of as data — beside the root's own scalar fields; the producer lays the block out from the counts and writes the offsets; the consumer reads the header and points. No 'section' construct; strided array = bounded array of structs with an optional `| stride`; the header is the one new generated thing.

gafferongames:
Owner (verbatim): "header itself could be a table or type?" — Rowan's answer for the page: THE HEADER IS THE ROOT TABLE ITSELF. The user declares the root (`table RenderFrame { … scalars …, ships [..MaxShips]RenderShip | stride = 128, … }`); its block form generates that table's struct with each bounded array replaced by (offset_of, count, stride). Nothing separate is declared, because every header field is either a declared field or a fact derived from one. Root as `table` or `type` decides only which wire it also rides (the two are the same struct, per the owner's doctrine); the block form is available to both.

gafferongames:
Owner (verbatim), confirming: "yes, i guess header is root table expressed in some way that C++/C# can read header data, and get data about the strided arrays." — §19's opening line for the revision.

gafferongames:
Owner (verbatim): "so in this context header just collapses into stuff that a table knows about its own rows." — The last new noun retires: no 'header' concept. The block form of a fixed table writes into the instance what the table knows about its own rows — per bounded-array field, where the rows start (offset_of), how many are present (count), the pitch (stride) — so the other side reads it and points. §19 is titled and opened with that.

gafferongames:
Owner (verbatim): "It's tables all the way down." — §19's closing line.

Two questions the spec worker raised for the read, decided now (Rowan's, labelled): (1) WHAT SELECTS the block form: OPT-IN — a fixed table takes the block form when any bounded-array field carries `| stride` or the table carries a table-level marker (spelling for the page: `| block`); no attribute, no block machinery — the zero-cost gate stays law and "make tables more powerful" means the capability is one attribute away for any fixed table, never emitted unasked. (2) sizeof(header) ≠ sizeof(table): correct — the block form's struct is a PROJECTION of the table (arrays → (offset_of, count, stride) triples); the layout contract (§19.4) is asserted against the projection on both sides, stated plainly. Also: `| stride` is block-form-only (inline storage sits at sizeof pitch). The cold reader may overturn any of these with a reason.

gafferongames:
Owner (verbatim): "The way we currently read render data header in C++ and C# is just a builder issue in C++ and a reflection issue in C#" — §19's two halves: (1) THE BLOCK BUILDER (C++): reserve the block from the counts, write each array's (offset_of, count, stride), hand workers disjoint ranges — §6.4's builder in its block form; (2) THE REFLECTIVE READ (C#): the projection's descriptors (§8) carry where each triple sits and the element's layout, so a generic reader points at the rows — the descriptors replace the hand mirror. Both generated.

gafferongames:
Owner (verbatim): "And the builder must be able to be multi-threaded. Hard requirement." — §19's block builder: layout from counts up front, workers own DISJOINT ranges, no per-record synchronisation, no lock — stated as a contract (§6.4's rule in block form), held by a real multi-threaded fill in the corpus and by gate 2's timed bar against the hand-written scatter. Not a hope; the requirement the relocatable/multi-threaded builder was written for (§13.1's provenance).

gafferongames:
Owner: "OK sounds like we have converged." — Converged design for the revision: stride attribute + opt-in block form of a fixed table; the header is the table's block-form projection (what it knows about its own rows); elements are structs; §19.1 the block builder (multi-threaded, hard requirement), §19.2 the reflective read. The worker's flagged tension, decided (Rowan's): the DESCRIPTORS are the mechanism (the layout is data a consumer can read; they retire the mirror's maintenance) and the generated BLITTABLE STRUCT (#287) is the typed fast path a per-frame consumer in a job uses — both generated from the projection, the contract asserted either way. Closing line: "It's tables all the way down."

gafferongames:
## Cold adversarial read — #291

Read whole and in order: `CLAUDE.md`; `SPEC-TABLES.md` on `main` (§2.2, §2.4, §3, §6.3, §6.4, §7, §9, §12, §13, §18); the diff; the PR body; #288 with every comment; #287, #289, #290; then the artifact the construct must express — `game/Tables/Render.schema`, `game/Source/RenderManager.h`, `unity/…/ClientInterop/Render.cs`, `NetworkRenderSyncSystem{,Jobs}.cs`, `EntityPool.cs` at `space@schema-tables-2`. Layout checked by a scratch Go program written from §19's text alone (deleted). I had no part in writing the page.

Verdict first, then the evidence.

**DO-NOT-LAND as drafted.** Not because the page is bad — the layout, the wide-fill shape, the two-sided contract and gate 2 are sound and the arithmetic reproduces the dogfood exactly — but because the central construct has been overtaken by the owner's own direction on this thread ("it's tables all the way down" / "why even have a 'section'?" / "header just collapses into stuff that a table knows about its own rows"), and because seven defects below are *independent* of that redirection and survive into whatever replaces it. Land the reshaped page; carry these fixes into it.

---

# DEFECTS

### D1. §19.5's `BlockOpenCompatible` refuses the very edit §19.5 says it absorbs

§19.5 lists **"a stride declared larger"** as one of the four absorbed edits, and then defines the entry point that is supposed to absorb it as checking

> "each section's STRIDE against **this build's own constant** with `sizeof( element ) <= stride` beside it"

Checking the header's stride *against this build's own constant* is an equality test. A producer that declares a larger stride writes a header stride the older consumer's constant does not equal, and `BlockOpenCompatible` — the append-only path — refuses. Absorbed edit 4 is broken by the mechanism named to absorb it.

It also contradicts §19.3 one subsection earlier: *"The consumer reads OFFSETS and STRIDES from the header, never from its own constants. Its constants exist to be asserted against, not to index with."* The compatible check is precisely a case of indexing with its own constant.

**Fix:** the only check that is both correct and consistent is `this build's sizeof( element ) <= header.stride`, per array the consumer knows, plus magic / byte order / extent / alignment. Drop the equality entirely.

### D2. The header's own field offsets are recorded nowhere and judged by nothing

§18.1's new text scopes the layout lines to section elements: *"every table **a section names** records, per field, the BYTE OFFSET and SIZE."* The block root's own line is `block RenderFrame sizeof=32` plus one line per section. So the root's **scalar** fields have no recorded offsets, and §18.2's block rows scope every refusal to "a section element". Consequently:

- inserting a scalar before a section in the header, or reordering two header scalars, **moves every triple's offset** — and is not in the refuse list, not in the warn list, not in the pass list;
- a consumer reading the header **by value** (which is how the dogfood reads it — see V2) then reads every offset, count and stride at the wrong place and points at garbage.

This is not hypothetical: `space::RenderHeader` carries exactly such a field (`version uint64`) ahead of its nine triples. The header is a record too; its layout is the consumer's contract as much as an element's is.

**Fix:** the projection records the block root's own fields' offsets and sizes, and §18.2's REFUSES row covers "a field inserted, reordered, removed or retyped in the block root itself". The `layout_id` digest (§19.4) must cover them for the same reason.

### D3. §19.4 never names which C# layout model the asserts run against, and the two disagree on `bool`

§19.4 requires *"a generated check, run once, asserting each type's size and each field's offset"* against the C++ constants. C# has **two** answers: the interop/marshalling model (`Marshal.SizeOf` / `Marshal.OffsetOf`) and the managed unmanaged-struct model (`Unsafe.SizeOf<T>` / `UnsafeUtility.SizeOf<T>`). They disagree for exactly the field kinds this dogfood uses: **`bool` is 1 byte in the managed model and 4 bytes under default marshalling**, and `char` differs too. `space::RenderShip` carries two bools (`has_target_lock`, `predicted_explode`); the shipped Unity consumer sizes with `UnsafeUtility.SizeOf<T>()` and reads with `MemoryMarshal.Cast` — the managed model.

A two-language layout contract whose C# half does not say which model it asserts is a contract that can pass on one measurement and garble on the other.

**Fix:** name the model on the page — the managed/unmanaged-struct model (`Unsafe.SizeOf`, pointer arithmetic), because it is the one `MemoryMarshal`/`Span` actually indexes with — and state `bool` as one byte in a block record. Also state that trailing padding comes from `Size = N` while interior padding comes from generated fields; §19.4 today only mentions interior.

### D4. §19 never states the storage lifetime, so the double-buffer question is silent

`<Header>BlockStorage` is *"the producer's ALLOCATION — one max-sized, 64-byte-aligned extent, owned by the caller"*, and `Begin` *"stamps the prologue, writes every section's offset, count and stride"*. The page never says:

- what a second `Begin` on the same storage does to a `Block` already handed out;
- how long a consumer's record pointer is valid;
- whether the construct contemplates double-buffering at all.

This is the shape the case actually has: `RenderManager` owns **one** `uint8_t m_renderData[MAX_RENDER_DATA_BYTES]`, clears `m_hasRenderData`, refills it every frame, and Unity reads through `client_render_data()` into that same memory. A page whose whole purpose is a per-frame handoff across a boundary must state the lifetime rule.

**Fix:** one paragraph — a block opened over a storage is valid until the next `Begin` on that storage; `Begin` invalidates every pointer and every open block over it; double-buffering is two storages and is the caller's, exactly as the arena's slack is in §6.4.

### D5. The multi-threaded fill is stated as a *permission*, not as an *obligation* — a serial implementation conforms

The owner's words are a hard requirement ("And the builder must be able to be multi-threaded. Hard requirement"; "the hand-written code … is multi-threaded by design"; flatbuffers lost *"because it was not parallizable"*). Every sentence on the page states what the runtime **permits**:

- §19.2: *"disjoint index ranges into one section are **safe** concurrently"* — a safety guarantee to the caller;
- §12.1: *"N workers fill disjoint records with no lock…"* — a description of the gate's shape;
- §19.6's parallel-fill test asserts byte-identity against a serial fill under the sanitizer leg — that catches a **race**, not a **lock**.

Nothing on the page forbids a backend from emitting a `Begin` that allocates, or a fill path that takes a lock, or from being serial throughout. Such a port passes §19.6 trivially and is conforming as written.

**Fix:** state the conformance obligation as a refuser, on the model §2.2 already uses for the zero-cost gate: the generated fill path contains no allocation, no lock and no atomic, and the build fails if one appears; `Begin` is O(sections) and allocates nothing. Then §12.1's measured leg is the second refuser rather than the only one.

### D6. §11's generated-spelling set is short by one — `<Header><Field>Span`

§11 claims nine block spellings (`Block BlockStorage BlockBegin BlockBytes BlockMaxBytes BlockOpen BlockOpenCompatible BlockLayoutId Counts`) plus the per-field accessor rule (`RenderFrameShips`). §19.3 then uses `block.ShipsSpan`. That is a tenth generated spelling, claimed nowhere and refused nowhere — a table with a `ships` section and a field or declaration named `ships_span` collides silently. §11 exists to close exactly this.

### D7. §19.4 overclaims what the digest replaces

> "**It replaces a hand-bumped version field**: the case this construct comes from carries one today, incremented by a person who remembers."

A layout digest cannot replace that field, because the field is not a layout fact. `RenderVersion 6` in the dogfood records a **semantic** contract — RenderManager's own comment: *"RENDER FIELDS ARE REAL VALUES (the semantic contract, RenderVersion 6): Unity consumes these as real doubles — direct float casts, no divisor anywhere."* A units or meaning change that moves no offset moves no digest. Keep the prologue — `magic` and `layout_id` both earn their sixteen bytes — but say the digest replaces the *layout* half only; a semantic version stays a field the author declares and owns.

(Separately, and it strengthens decision 9 rather than weakening it: the shipped consumer reads `header.version` into a local and **never checks it**. The hand-bumped field is already dead weight; the digest is a real upgrade. Just don't claim it covers ground it cannot see.)

---

# The reshape: the draft against the owner's direction on this thread

The owner has, in sequence: "can render data root be a table?" → "if it is, then we can make tables more powerful" → "render data is a root table, with strided types or fixed tables being allowed. strided making no sense for non-fixed tables or types" → "fixed tables and types are semantically the same (structs)" → "why even have a 'section'?" → "the strided array and the section are the same thing, what you really need is a new concept of header with offset_of" → "header just collapses into stuff that a table knows about its own rows" → "It's tables all the way down."

### D8. §14's rejection of a table wire for a block root is not honestly priced — it is circular

Decision 10 and §2.7 rest on one argument: *"§3's set is closed, so a section's triple cannot ride."* That is circular. The triple is a **block-form artifact**; a wire form has no triple to ride. A bounded array of fixed tables already rides as kind `14` with element kind `13` and the live count — **no new kind is spent and §3 is untouched**. The alternative that dissolves decision 10 is never stated in §14 and therefore never priced, which is the one thing §14 exists to do. Under it, "is the render root a table?" is answered with one word, `Measure`/`Save`/`Load` survive on the root (a tool can dump a frame), §7's cook still applies, and §2.2 needs **no edit at all** — a bounded array of fixed structs is already by-value storage, so the block root already derives FIXED under the rule as it stands. The draft needs a whole new §2.2 paragraph explaining why a section is not a by-value edge; the reshape deletes that paragraph.

### The rule I would put on the page, with the keyword dropped

Dropping `section` loses exactly one thing a third party needs — **which arrays go out of line** — and it is recoverable in two sentences:

- **Depth and shape.** In the **block form** of a fixed table, every **bounded** array field **of the root itself** (`[..N]T`) is laid out of line, in declaration order, each at its own pitch. Every other array stays inline where it is: a fixed `[N]T`, an enum-keyed `[E]T`, and **every array at any depth inside an element** — an element must stay one flat record or it is not a record a consumer can point at. Depth one, bounded only.
- **The instance is the header.** The block form generates the root's struct as a **projection**: each out-of-line array's inline storage (`T[N]` plus its count) is replaced, at that field's position, by `(offset_of u64, count u32, stride u32)`; every other field keeps its by-value storage at its natural C ABI offset. That is the header, and nothing declares it — it is, in the owner's words, what the table knows about its own rows.
- **Elements are structs.** A bounded array takes the strided out-of-line form only when its element is a fixed-class `table` or a declared `type`. A variable-length table, `*T`, `string`, `bytes` — refused by name in a block-form table: no fixed pitch exists. That is "strided making no sense for non-fixed tables or types" made a refusal. `?T` and a keyed array *inside* an element are both fixed-size and stay allowed; nothing needs to say so beyond one line.
- **The wire is unchanged.** The root keeps its table wire — kind `14`, element kind `13`, live count. Nothing on §3's closed set. The cook (§7) still applies to the by-value form. The block form and the cook must not share a `layout_id`; say so.

### On the trigger — I would overturn "stride triggers the block form"

The proposed opt-in is "any bounded-array field carries `| stride`, or the table carries `| block`". **The stride half does not work, and the dogfood is the proof: space's render data has nine strided sections and zero declared strides.** Every stride in `RenderManager.h` is `sizeof`. A trigger the primary case never fires is not a trigger.

Make it the **table-level marker alone** (`| block`). One token; explicit; answerable by reading one line; §2.2's zero-cost gate stays law verbatim ("no marker, not one symbol"); and it restores the one piece of legibility the keyword was carrying — a reader can see which table's arrays move out of line. `| stride = N` then only *adjusts* a pitch inside a table already marked, and is **refused** on a table with no block form rather than silently inert. An inert attribute is a lie about the declaration.

Everything else in the two proposals I would keep: the block-form struct as a projection with its own `sizeof`, and the two-language contract asserted against the projection.

### What §19 should say under the reshaped form

Same six subsections, reframed, and almost all of the draft's text survives:

1. **§19 opening** — the block is the **third form a fixed table takes**, beside its wire (§3) and its cook (§7): one declaration, three projections. The four-form table becomes a three-form one; the "fourth form" framing and §6's "third life" paragraph both go.
2. **§19.1 layout** — unchanged and correct: base 64-aligned, projection struct at 0, each out-of-line array in declaration order at `max(64, alignof(element))`, record `i` at `offset_of + i*stride`, `BlockMaxBytes` at the declared maxima, `BlockBytes` from the counts. Add: the header itself follows the same C ABI rule as a record (§19.4 today scopes that rule to "a record" only), and say whether the tail bytes between the last record and the 64-byte rounding are specified — a handoff copies them.
3. **§19.2 building** — unchanged, plus D4's lifetime paragraph and D5's conformance refuser.
4. **§19.3 reading** — unchanged, plus D6's spelling.
5. **§19.4 contract** — unchanged, plus D2 (the root's own fields in the digest), D3 (name the C# model), D7 (drop the replaces-the-version claim).
6. **§19.5 evolution** — unchanged minus D1's equality check, and minus the "its weight has changed" narration (see below).

**Nothing in the draft is lost by the simplification, and three things are gained**: §2.2 needs no new paragraph; the "block root may not be nested / pointed at / made a union arm / used as an element" refusals mostly dissolve (the by-value form of that table is an ordinary nested table — only the *block form* requires a base, exactly as the cooked form requires a file's base); and the root keeps its wire, so a tool can dump a frame. The one genuine loss is that an author who wants a small inline array *beside* strided ones in the same root can no longer say so — the answer is to wrap it in a nested type, and the page should say that where a person will look for it.

---

# RISKS — decisions worth the owner's ruling, with numbers

**R1. Drop `| stride` from the first landing entirely.** Evidence: nine sections in the dogfood, **zero** declared strides. And the only consumer that exists **cannot read a strided section at all** — `NetworkRenderSyncSystem.ProcessSlice` errors and drops the section when `bytesPerItem != UnsafeUtility.SizeOf<T>()`, because it reads via `MemoryMarshal.Cast<byte,T>`, which requires stride == sizeof. Adopting headroom costs that consumer its cast path and forces the strided iterator. The owner has already said the property "may be obsolete now". *Recommendation: derive always in v1; name declared-stride headroom as a §15 follow-on.* If it stays, keep §19.3's honest statement that headroom costs the span.

**R2. `Begin` returning `false` on an over-max count drops a frame, silently, at 60Hz.** §19.2 shows `if ( !Begin(...) ) return;` and never states the producer's obligation or the diagnostic. Today `RenderManager` uses `core_assert( endOffset <= MAX_RENDER_DATA_BYTES )` — an assert, which is the worse failure mode, so `Begin` is an improvement — but the page should require the refusal to **name the section, its count and its maximum**, and should state plainly that clamping to the maximum before `Begin` is the producer's job. Also unstated: filling the `Counts` struct from more than one thread. One sentence ("counts are gathered before `Begin`, which is single-threaded") closes it.

**R3. `BlockMaxBytes` sums per-section maxima; the real bound is a shared pool.** Computed over the dogfood's real constants: **7,879,488 B (7.51 MiB)** against `MaxRenderDataBytes = 10 MiB` — it fits, comfortably, so allocate-max is affordable here and that is a genuine result. But note the sum is loose by construction: `ships(4096) + turrets(1024) + missiles(4096) + dynamic_props(4096) = 13,312` declarations all draw from one `MaxObjects = 10000` pool, so ~25% of that extent can never be occupied simultaneously. Not a defect — the allocation is once and never grown — but if a future case is tight, the page has no way to declare a shared bound, and §15 should say so rather than leaving the discovery to whoever hits it.

**R4. §19.1 justifies 64-byte section starts with "two workers filling adjacent sections must never share a line" — the common case is intra-section and unmentioned.** Every element in the dogfood except `RenderTurret`/`RenderLaser` (64) has a stride that is not a multiple of 64 (72, 88, 80), so two workers at a range boundary *within* a section do share one line. Bounded and harmless when ranges are contiguous runs — which is what `RenderManager` does (`numShips / numWorkers`) — but the guarantee should be stated for what it is: per-section, and contiguous runs are why the intra-section case costs one line per boundary rather than one per record.

**R5. §12.1's "no marshalling and no copy anywhere in the frame" is not true of the consumer it is measured against.** `ProcessSlice` copies every record into `EntityPool._renderDataBuffer` so Burst jobs can take a `NativeArray<T>`. The true and defensible claim is "no marshalling and no copy **at the boundary**". Worth fixing before the measured leg is run, because the bar in §12.1 is stated against this consumer.

---

# VERIFIED GOOD

**V1. The layout arithmetic is right, and it reproduces the artifact exactly.** From §19's text alone, over the dogfood's nine real sections and a representative frame, the page's model and `RenderManager`'s hand layout produce **identical section starts at every one of the nine** (192 / 320 / 26752 / 84352 / 92992 / 95872 / 495872 / 559872 / 572672) and an identical used extent (577,472). `RenderAlign` is `(x+63)&~63`, which is §19.1's `max(64, alignof)` with `alignof = 8`. §19.1's own worked example checks out to the digit: `sizeof = 64`, starts 64 / 192 / 524,480, `BlockMaxBytes = 655,552`. Element sizes reproduce the `static_assert` wall (72 / 88 / 64 / 72 / 72 / 80 / 80 / 64 / 80). The 16-byte prologue does **not** move the first section — `RenderAlign(152) = RenderAlign(168) = 192` — so the prologue is free in this schema.

**V2. Decision 4 (block-relative, not self-relative) is correct, and the artifact proves it.** The C# consumer does `header = MemoryMarshal.Read<RenderHeader>(span)` — a struct copy out of the mapped bytes. A self-relative delta does not survive that copy. §2.7 and §14 give exactly this argument, and it is the right one.

**V3. Decision 8 (`Sequential, Pack = 1, Size = sizeof`, generated interior padding) matches the hand mirror precisely.** `ClientInterop/Render.cs` already declares `[StructLayout(LayoutKind.Sequential, Pack = 1, Size = 88)]` etc. with `Size` equal to the C++ `sizeof`. The mirror needs no interior padding fields only because the schema orders fields largest-alignment-first *by hand* — the page's generated-padding rule is strictly more general and removes that discipline, which is the point. No contradiction with #287, which asks for `Sequential` blittable structs with sizeof/offsetof pinned on both sides; the page is a superset.

**V4. §19.4's "`Size` is the element's `sizeof`, never its stride"** is right for the stated reason and matches the artifact: the mirror's `Size` is the C++ `sizeof` and the stride rides in the header.

**V5. §18's block rows catch all three constructed breaks.** Insert-in-the-middle → refused (field offsets are recorded per element field). Stride shrunk → refused, and the §18.1 projection carries `stride=N derived|declared`, so "an element grew under a derived stride" is visible. Reorder → refused, including the two same-sized bools in `RenderShip` (offsets 80↔81 move). **§18 does not need a new stride row — it already grew one.** What it *does* need is D2's header row.

**V6. Zero-cost gate, refusal battery, negative controls, provenance.** §2.2's block-machinery gate is stated on the pointer machinery's terms. §19.6's negative controls are the right two (perturb one side's stride constant; perturb the compiler's own layout model) and honour "a layout test that shares its layout model with the code it checks proves nothing". §13.6 carries the owner's words verbatim and places each where it governs. §12.1's measured bar — two numbers, both required, paired in one sitting under the bench rules — is the right bar and is stated as a gate rather than a preference.

**V7. Gate 2 is expressible.** I wrote the whole dogfood out under the page's spelling: one block root, nine sections, ten element tables, two nested `type`s, six enums (`uint8_t`), one `flags` (`uint64_t`), bools and scalars — all fixed-class, all blittable. Nothing in `Render.schema` is inexpressible, and nothing `RenderManager` writes is unnameable except the two items above: the semantic `version` (D7 — expressible as a declared field, just not *replaced* by the digest) and the shared-pool bound (R3).

---

# The open list — omissions

The PR's five open items are honest as far as they go, and 1/2 are already ruled. Missing from it, each a decision the owner should see:

1. Which C# layout model the asserts run against (D3) — this one can garble silently and is not a taste question.
2. Whether the block root's own fields are covered by the baseline and the digest (D2).
3. Storage lifetime and double-buffering (D4).
4. `BlockOpenCompatible`'s check semantics (D1).
5. What a producer does when a count exceeds its maximum (R2).
6. Whether the multi-threaded fill is an obligation on the *implementation* or a permission to the *caller* (D5) — the owner has called it a hard requirement; the page reads as the latter.

# Present-state

Two spots narrate rather than state, against `CLAUDE.md`'s present-state law (§13.6 is the right home for provenance and does its job):

- §19.5 opens *"This is the secondary property, and **its weight has changed**. When the two sides were kept by hand, the stride's headroom **was** the versioning model…"* — write the present rule; the change of weight is §13.6's, and it is already there.
- §19.4's *"the case this construct comes from carries one today, incremented by a person who remembers"* — see D7; the sentence should state what the digest is, not what someone else's codebase does.

Cross-references in §19 all resolve (§2.2, §2.7, §3, §4, §6.1, §6.4, §7, §9, §11, §12.1, §14, §18, §18.2, §19.1–.5). The `schema-tables-baseline` rendering version moves 2 → 3 with §18.4's repair path named — correct.

---

**Verdict: DO-NOT-LAND as drafted.** Reshape to the table-form the owner has directed — no `section` construct, a table-level `| block` marker (not `| stride`) as the trigger, the block as the **third form of a fixed table** with its wire and cook intact, the header as the root's own block-form projection, elements as structs — and carry D1–D7 into that page, since every one of them is a defect in its own right and none is dissolved by the reshape. The layout model, the wide-fill shape, the two-sided contract, gate 2 and the §18 block rows all survive the move; that is most of the 780 lines, and it is good work. Re-read after the reshape.

gafferongames:
Owner (verbatim): "Yes, the render struct is interesting in that it is a structure that must be fast and effectively, pre-cooked on build. It must be fast to both write in C++ and read in C# since it is large and it is 60HZ or greater (we can render faster than we simulate)." — §12.1's frame: the block form is pre-cooked at build (layout fixed and known to both sides; nothing decided at frame time; the contract in generated asserts, not runtime checks); gate 2 measures BOTH the C++ write and the C# read per frame; the bar is set by 60 Hz-plus with rendering allowed to outpace simulation.

gafferongames:
Owner (verbatim): "It is the hottest path for tables to be sure." — §12.1 ranks it so: the block form is the hottest path tables have; the fixed-class performance doctrine (as close to its equivalent type as possible) bites hardest here, and gate 2's timed bar is the one that decides whether tables earn the render data.

gafferongames:
Owner (verbatim), placing the block form among the others: "The other paths for tables like save games, or especially assets -> cooked -> binary format for specific build version, these only need fast *readers*" — the wire and the cook are read-hot/write-cold; the block form is hot on both sides. State it in §12.1 and the opening's ladder.

gafferongames:
Owner (verbatim), the reason behind read-hot/write-cold: "Since the building will be done by tooling offline" / "(and generally, it will be done in a distributed build cache sort of way... this is how it is usually done)" — the wire's and the cook's writers run in tooling, offline, typically under a distributed build cache; their cost is a build cost; the game needs only fast readers of those forms. §7's pipeline paragraph and the ladder carry it.

gafferongames:
## Cold read, delta — 916d402 "the block form"

Fresh clone, `gh pr checkout 291` at 916d402. Re-derived the layout, the wire and the field ids from the revised text alone with scratch Go (deleted). Same reader as the 9d50fcd read.

**Verdict: LAND-AFTER-FIXES.** The reshape is right and the page is materially better than what I read two days ago: the construct is gone, the wire is intact and hand-verifiable, D1–D8 and R2–R5 all landed faithfully, and every layout number reproduces to the byte from the text alone. Eight defects remain — four I would fix before merge, four that are a line each. None of them is structural.

---

# What I reproduced

**The nine-array layout, from §19.1's rule and nothing else.** Projection = 168 (16 prologue + `version` + nine triples). Then, at the counts:

`cameras 192 · ships 320 · turrets 26,752 · missiles 84,352 · dynamic_props 92,992 · static_props 95,872 · cosmetic_props 495,872 · lasers 559,872 · explosions 572,672`, `BlockBytes = 577,472` (already 64-aligned, so the rounding is a no-op here), `BlockMaxBytes = 7,879,488 = 7.51 MiB`.

Identical to §19.1's table in every cell, and I checked each extent column against `count × pitch` as well. Same numbers the hand layout in the case this replaces produces, at all nine. `152` and `168` both round to `192`, so the prologue is free in this shape — as stated.

**The wire really is the ordinary wire.** Hand-encoded `RenderFrame{ version = 6, ships = [one row] }` from §3 alone: 62 bytes, kinds `{9, 14, 13}`, `L`/`N` framing as written, one `u16` terminator per body. No triple, no new kind, no `offset_of`, `| block` nowhere in it. §3's new bullet and §19.5's byte-identical-goldens test are both exactly right, and D8's circularity is properly dissolved rather than papered over.

**Expressiveness — nothing left over.** The whole render data writes as one `table RenderFrame | block` with `version uint64` and nine `[..Max]` arrays, over the ten element tables unchanged. The hand-declared `table RenderSection` **disappears** — the triple is generated, which is the reshape paying for itself. The semantic `version` survives as an ordinary field, and §19.3's "what the digest sees, and what it does not" is precisely the right home for why it must. Two nested `type`s, six `uint8_t` enums, a `uint64` flags, bools and scalars — all fixed-class, all blittable, all expressible. Nothing in the schema and nothing the producer writes is unnameable.

**The baseline catches the scalar-before-an-array case.** §18.1 now records `field version … offset=16` and `field ships … offset=24 out_of_line stride=32` with `block sizeof=40` — I checked that projection arithmetic too (16 + 8 + 16 = 40). Insert a scalar ahead of `version` and both recorded offsets move; §18.2 refuses "a field INSERTED before the end … in a block-form table ITSELF". And the PASSES row's qualifier — "past **every offset the baseline records**" — correctly closes the near-miss where a scalar is appended after the last scalar but *before* the arrays. D2 is carried properly, not just gestured at.

**Carried and correct:** D1 (`sizeof(element) <= the stride it read`, "never an equality, and never against its own pitch constant"), D3 (the managed model named, `Marshal.*` named as the one it is not, `bool` = one byte, and a `bool` row pinned in §19.5 so a port cannot pick wrong and pass), D4 (valid until the next `Begin`; double-buffering is two storages and is the caller's), D6 (27 + 9 = **36**, plus the per-array pair rule that claims `…Ships` *and* `…ShipsSpan`), D7, R2 (the refusal names the array, count and maximum; clamping is the producer's job), R3 (7,879,488 / 7.51 MiB stated with the shared-pool looseness, §15 holds the follow-on), R4 (64 buys a per-array guarantee; the intra-array boundary line is stated rather than implied), R5 ("no copy **at the boundary**", with the pool copy named as the consumer's design). §11's "What is NOT refused" is the right call and worth having written down.

---

# DEFECTS

### D9. §18.1's four new field ids are fabricated

I calibrated §5's fold on three ids the file already carried, and all three land exactly: `damage → 0x15a9`, `name → 0x30df`, and — the good one — `speed`'s `was=velocity` resolves through `id("velocity") = 0x2e46`, which both confirms the fold and demonstrates the `was` semantics. So the fold is provably this document's fold. The four ids added with the block example are not:

| name | page says | computes to |
|---|---|---|
| `version` | `0x4a1c` | **`0xe8e6`** |
| `ships` | `0x2b71` | **`0x2d39`** |
| `position` | `0x4c31` | **`0xdd45`** |
| `object_id` | `0x6b0e` | **`0xdc71`** |

In a file whose other ids are real, this is a trap: an implementer calibrating a hash against the spec's own example concludes their fold is broken. These came in with the first draft and survived the reshape unchanged. Regenerate them.

### D10. §19.4 flattens three edits that behave differently at runtime

"**Three edits are absorbed**" is stated without saying by which entry point, and they split:

- **Edit 3, a maximum raised** — absorbed by `BlockOpen` **itself**, because the declared maximum is *not* among §19.3's digest facts ("the projection's own fields and their offsets and sizes, each out-of-line array's element and pitch, and every row field's offset, size and kind"). The digest does not move; the id matches; the consumer reads the moved offsets. Correct, and quietly elegant.
- **Edits 1 and 2, a field appended to the table or to a row** — both move the digest (a new offset and size; a changed row size and pitch). `BlockOpen` **refuses**. They are absorbed only by `BlockOpenCompatible`.

That is by design — §14's "no silent bypass" — but §19.4 as written reads as though an old consumer on the default path keeps working across an appended field. Say which path absorbs which. And say in §19.3 that the declared maximum is deliberately **not** a digest fact: right now that is absence-by-omission, and a porter who folds it in silently breaks edit 3 on the default path with nothing to catch them.

### D11. §19.5 claims the refuser catches something it cannot

> "The test catches a race; the refuser catches a serial implementation that has no race because it never went wide."

It cannot. The parallelism lives in the **caller's** loop; the generated surface is `Begin` plus accessors. A `Begin` that is `O(arrays)` with no allocation, lock or atomic, and accessors that are one add, is byte-for-byte the same surface whether the backend author imagined a wide fill or not — there is no generated loop to be serial in.

What the refuser genuinely buys is exactly what §19.1 says it buys: no allocation, no lock and no atomic in the generated fill path, which is the property that *makes* a caller's wide fill possible. That is true, checkable as a symbol-presence gate on §2.2's model, and sufficient. Strike the second half of the §19.5 sentence; the obligation stands on the refuser plus §19.5's real multi-threaded fill plus §12.1's measured leg, and those three together are adequate.

### D12. The by-value form of a block-form table is unpriced, and §2.2's smallness assumption now has a counterexample it does not name

§2.7 insists "`| block` adds a form; **it removes nothing**" — `Measure`, `Save`, `Load`, the cook, its place as a nested field. Good, and it is the reshape's whole point. But a bounded array's storage is `T[N]` inline plus an `int32` count — verified in this repo's own output, `generated/cpp/Clauses.h`: `uint16_t items[12]; int32_t items_count;` — so the **by-value** struct of the flagship table is **~7,879,320 bytes (7.51 MiB)**. Meanwhile §2.2 still reads, unedited:

> "Value-only tables are assumed SMALL — messages, records, config fragments — so they are passed by struct and loaded directly, and no large-flat-struct machinery is warranted for them without a real case forcing it."

A real case is now forcing it. §14 answers this objection **for the projection** ("the instance stays small and copyable") and leaves the by-value form standing at the same size it rejected. Three consequences the page should own in one paragraph rather than leave a reader to hit:

- `Load` of a block-form table needs a multi-megabyte destination struct — the load path allocates nothing, so the caller owns that;
- §7's "the cook of a FIXED root table is … the struct's bytes behind the header" means a 7.5 MiB cook of mostly-unused array tails;
- the **wire** is small — elision writes only the live rows — while the **struct** is not, and those two facts sitting side by side is the thing to say plainly.

Then amend §2.2's assumption to name the block form as its exception. Nothing about the design changes; the page just stops asserting something its own flagship contradicts.

### D13. `BlockOpenCompatible` does fewer bounds checks than `BlockOpen`, on the path that trusts less

§19.4 gives the compatible path: magic, byte order, extent, alignment, and the pitch `<=`. `BlockOpen` additionally checks "each array's `offset_of` and extent inside the block". That check needs **no layout agreement at all** — it reads the instance's own numbers against the extent the caller passed — so there is no reason to drop it on the path that by definition takes bytes from a build that does not match. Add it; it costs a handful of comparisons on a path already doing them.

### D14. §15 presents as undecided something §2.7 and §11 have already decided

§15: "**DEPTH past one in the block form** — an out-of-line array inside an element, **or a block-form table as another's element**. It wants a decision about whose base a nested projection's offsets are relative to before it is anything."

But §11's "What is NOT refused" permits a block-form table as an array element, and §2.7's depth-one rule silently resolves it: its own arrays stay inline. So the case is decided and shipping, not open. Either refuse it in §11 for v1, or say in §15 that it is permitted today with the inner arrays inline and the follow-on is making them out-of-line. (Worth noticing what the current answer means concretely: a `[..N]RenderFrame` row is a 7.5 MiB row, legal and unremarked — see D12.)

### D15. One stale cross-reference

§13.6, on the multi-threaded producer: "which is the constraint the layout is shaped by (**§19.2**)". §19.2 is now "The reflective read"; the layout and the fill are §19.1. Left over from the earlier numbering. I walked all thirty §19.x references; this is the only one that misses — including §13.6's own "§19.1 is the builder half and §19.2 the reflective half" two bullets down, which is right.

### D16. §14 narrates the drafting history, against the present-state law

> "This feature **was first designed** as a `section` construct … That design **is recorded here because its central argument was WRONG** in an instructive way"
> "It **was first written down** as the near-miss alternative"
> "TAKEN, **overturning a stride-triggered opt-in**. The rejected proposal was …"

§14 is legitimately the rejected-alternatives section and every one of these belongs in it — but as **present-tense rejected alternatives**, not as an account of what a draft did two days ago and why it changed its mind. `CLAUDE.md` is explicit that provenance lives in git and `notes/`, and that the page reads for an implementer. The content is the honest pricing and should all stay; only the register moves. "A `section` construct — a field spelling whose storage is a triple, in a table with no wire — REJECTED: the argument is circular, because …" says the same thing and reads as a spec.

Worth noting the two I flagged last time are both properly gone: §19's "its weight has changed" and §19.3's "incremented by a person who remembers". So this is the residue of the reshape, not a habit.

---

# Notes, not defects

- **§19.1's "representative frame" is illustrative, not measured** — those counts are the example vector from the cold read, not a profiled frame. The load-bearing claim ("those starts are the same numbers the hand-written layout produces") is **count-independent**, because both walks are the same arithmetic over the same pitches; resting it on one invented frame understates it. Say the arithmetic agrees and the table is one instance of it.
- **"It's tables all the way down." appears three times** (§2.7, §13.6, §19's close). It governs in §13.6 and lands well in §19's close; the §2.7 use is the weakest of the three.
- **`Pack` and `Size` are interop-flavoured attributes** used here to set the *managed* layout. After §19.3 says "not the interop marshalling model", a porter may hesitate at the very next attribute; one clause ("`Pack` and `Size` set the managed layout too, which is why they are the mechanism here") closes it.
- **§19.3 states `sizeof(bool) == 1` in C++ as a fact** where it is really an *asserted* fact — implementation-defined by the standard, caught loudly by the C++ `static_assert`s rather than silently. Half a sentence.

---

**LAND-AFTER-FIXES.** D9 (fabricated ids), D10 (which path absorbs which, and the maximum's deliberate absence from the digest), D11 (a claim the refuser cannot support) and D12 (the unpriced by-value form) before merge. D13–D16 are one line each. The construct is gone, the wire is provably untouched, the layout reproduces from the text alone, and the owner's sequence is recorded where it governs — this is the page to build from.

gafferongames:
Delta-read fixes pushed — **`4f0d25a`**. All eight, none structural.

**D9 — every id on the page is now MACHINE-COMPUTED.** The four ids added with the block example were written rather than computed, and all four were wrong. Recomputed with the compiler's own `ir.FieldId` (a scratch `main` in-tree, deleted after):

| name | was | now |
|---|---|---|
| `version` | `0x4a1c` | **`0xe8e6`** |
| `ships` | `0x2b71` | **`0x2d39`** |
| `position` | `0x4c31` | **`0xdd45`** |
| `object_id` | `0x6b0e` | **`0xdc71`** |

Exactly the reader's four values. The three ids the file already carried verify against the same function untouched — `damage` `0x15a9`, `name` `0x30df`, and `speed`'s recorded `0x2e46`, which is `id("velocity")` through `was`. That last one is the good calibration the reader spotted, and it still holds.

**D10** — §19.4 now says which entry point absorbs which. A raised maximum is absorbed by `BlockOpen` **itself**; an appended field (to the table or to a row) moves the digest and is absorbed only by `BlockOpenCompatible`. §19.3 states the maximum's absence from the digest as a deliberate **exclusion**, with the rule behind it: *a fact both sides must AGREE on is in the digest; a fact a consumer READS is not.* A port can no longer fold it in and silently break edit 3 on the default path.

**D11** — struck. The refuser holds no allocation, no lock, no atomic on the fill surface — the property that *makes* a wide fill possible — and that is now the whole claim. The obligation rests on the refuser, the real multi-threaded fill, and §12.1's measured leg.

**D12** — the by-value form is priced, and §2.2's smallness assumption now names `| block` as the exception it declares: ~7.5 MiB inline for the flagship. Three consequences owned in place — `Load` needs that destination and the caller owns it, the cook is mostly-unused array tails, and the **wire is small where the struct is not**. §15 names the follow-on: a wire load straight into the projection.

**D13** — `BlockOpenCompatible` now checks everything `BlockOpen` does except the layout id, bounds included. **D14** — §15 and §11/§2.7 agree: a block-form table as another's row type is permitted today with its own arrays inline; the follow-on is making them out-of-line. **D15** — §13.6's producer ref → §19.1. **D16** — §14 is present-tense rejected alternatives; same content and pricing, register moved.

Notes taken too: the worked table is named as one instance of a **count-independent** arithmetic rather than a measured frame; `Pack`/`Size` are noted as setting the managed layout, right after the sentence naming the model they are not; C++'s `sizeof(bool) == 1` is stated as **asserted**; and the weakest of the three "tables all the way down" uses is dropped.

`make check` green; CI green.

gafferongames:
## Cold read, final — 4f0d25a

Fresh clone, `gh pr checkout 291`. Re-verified the ids against the compiler's own function rather than my own fold, re-derived the layout, and walked D10–D16.

**Verdict: LAND.**

---

### The ids — all seven, against `ir.FieldId` itself

I built a scratch module with a `replace` onto this checkout and called the compiler's own `ir.FieldId` / `ir.TableFieldId`, so the check no longer depends on my re-implementation of the fold. Every `id=0x` on the page — not just the four that moved:

| line | hashed from | `ir.FieldId` | page |
|---|---|---|---|
| `ShipConfig.damage` | `damage` | `0x15a9` | `0x15a9` |
| `ShipConfig.speed` | `velocity` *(the `was` alias)* | `0x2e46` | `0x2e46` |
| `ShipConfig.name` | `name` | `0x30df` | `0x30df` |
| `RenderFrame.version` | `version` | `0xe8e6` | `0xe8e6` |
| `RenderFrame.ships` | `ships` | `0x2d39` | `0x2d39` |
| `RenderShip.position` | `position` | `0xdd45` | `0xdd45` |
| `RenderShip.object_id` | `object_id` | `0xdc71` | `0xdc71` |

Seven for seven. Three checks beyond the table: `ir.TableFieldId(speed, was="velocity")` returns `0x2e46`, so the one row where the hash source is not the field's own name is right for the documented reason and not by luck — `FieldId("speed")` is `0xbc00`, which the page correctly does not show. No id on the page collides with the reserved node-table id `0xFFFF` or with the `0` terminator. And the compiler's function agrees with the independent hand fold I ran on 916d402, so the two derivations cross-check each other. D9 closed.

### D10 — the entry-point split, and the exclusion made explicit

Both halves landed, and the second is better than what I asked for. §19.4 now labels each edit: 1 and 2 "**The digest moves** … `BlockOpen` REFUSES and this is absorbed only by `BlockOpenCompatible`"; 3 "**The digest does NOT move** … absorbed by `BlockOpen` ITSELF, on the default path, with no waiver" — then says plainly that the default path absorbs exactly one of the three and why that is correct rather than an oversight.

§19.3 makes the exclusion positive: "A declared MAXIMUM is deliberately NOT a digest fact, and it is **excluded rather than merely absent** … a port that folded the maximum into the digest would break that absorption with nothing to catch it." And it generalises to a rule a porter can apply to the next fact: **"a fact both sides must AGREE on is in the digest; a fact a consumer READS is not."** That is the sentence that stops this recurring.

### D11 — struck, and replaced with the true claim

The false half is gone. What stands is exactly the defensible version: the refuser's whole claim is no allocation, lock or atomic in the generated surface, "the property that MAKES a caller's wide fill possible"; "the parallelism itself lives in the caller's loop, so **no gate here could assert it**"; and the obligation rests on the refuser, the fill test and §12.1's measured leg together. Correct, and it concedes the limit rather than papering it.

### D12 — priced, in the right place, with the consequence I was after

§2.2 now carries it where the assumption it breaks is stated: "**`| block` IS a case forcing it, and it declares that it is**" — the marker read as the author opting out of the smallness assumption, which is a better framing than an exception clause. All three consequences are there, including the one I most wanted: "**the WIRE is small where the struct is not** — elision writes only the live rows — so 'the file is a few hundred kilobytes and the struct is seven megabytes' is the normal state of affairs, not a defect." The ~7.5 MiB matches my computed 7,879,320 for the case at hand. §14's near-miss now cites §2.2 for the size instead of restating it.

The follow-on it grew out of — **a wire load straight into the projection**, so a tool can read a file and hand a block across without materialising the large struct — is a real derivation from the price, not a placeholder. Good addition.

### D13, D14, D15, D16

- **D13** — `BlockOpenCompatible` now "checks **everything `BlockOpen` checks except the layout id**", per-array `offset_of` and extent included, with the reason: those bounds read the instance's own numbers against the caller's extent and need no layout agreement, "so there is no reason to do fewer of them on the path that by definition takes bytes from a build this one does not match."
- **D14** — §15 no longer presents a decided case as open: a block-form table as another's row type "**is permitted today** (§11), and depth one already answers it — its own arrays stay INLINE there", with the size consequence named and the follow-on narrowed to making them out-of-line.
- **D15** — §13.6 now reads "the layout and the fill are shaped by (§19.1)". I re-walked all 37 §19.x references; every one resolves to the right subsection.
- **D16** — §14 is rewritten into present-tense rejected alternatives ("The block form: the shapes it is not, and why"), and the content survives intact — the circular wire argument, the near miss, the per-field stride trigger and its zero-declared-strides evidence are all still priced. "was first designed", "were considered", "overturning" are gone.

### Notes, all carried

§19.1: "**The agreement is with the ARITHMETIC, not with one frame** … they land every array at the same offset for ANY counts — the table above is one instance of that, chosen to be legible rather than measured." The nine-row table is unchanged and still reproduces from the rule alone (192 … 572,672, `BlockBytes` 577,472, `BlockMaxBytes` 7,879,488). §19.3 gained the `Pack`/`Size`-set-the-managed-layout clause and "**C++'s `sizeof( bool ) == 1` is ASSERTED, not assumed**". "It's tables all the way down" is down to two placements — §13.6 where it governs and §19's close where it lands.

---

### One residual, and it does not block

§11 still narrates: "What is NOT refused, and it is worth saying **because the earlier shape of this feature refused it**". Same class as D16, in the one place the sweep did not reach. The observation is worth keeping — it is genuinely useful to say that a block-form table nests and is pointed at like any other — it just needs to stand on its own footing rather than on a draft nobody will ever read: *"What is NOT refused, and it is worth saying because a reader coming from the block's base requirement will expect it to be:"* or simply drop the clause. Half a line, and fine to fold into the next commit that touches §11.

---

**LAND.** The ids are machine-computed and verified against the compiler's own function, the layout reproduces from the text alone, the wire is provably the ordinary wire, and every one of D9–D16 is carried — several of them further than the finding asked. This is the page to build from.

Issue #292, opened 2026-09-02
The BUILD VERSION: one generated digest of the wire protocol id and the layout ids of the tables as they cook — the key cooked assets are addressed by


gafferongames:
Owner, 2026-09-02 (Glenn, live), verbatim: "so the hash of the asset and the cooked version is the asset the runtime searches for." / "asset hash, cooked version hash." / "(as in, build version, schema table protocol id effectively...)" / "so this is a new concept. some sort of 'build version' that depends both on wire protocol id, and protocol id for current set of tables, as they would cook."

**The concept.** A unit's BUILD VERSION is one digest generated by the compiler: the type wire's protocol id (SPEC.md §3) combined with the cook ids of the unit's tables — each table's layout id (SPEC-TABLES.md §7: packed-layout facts + this build's sizeof/offsetof, keyed by wire ids) — so it moves when the wire moves OR when any cookable table's layout moves, and stays when neither does. It is the second half of the cook key: a cooked artifact is addressed by (asset hash, build version); the runtime searches for exactly that pair; a distributed build cache produces and serves by it. The estate precedent: space's connect-time id already combines the unit protocol id and the delta id with FNV (space CLAUDE.md, "two first-class ids").

**Spec first.** SPEC-TABLES.md §7 (cooking) states the key; a new short section (or SPEC.md §3.3) defines the build version: what it digests, in what order, that it is printed by `schema build-version <unit>` and emitted as a constant (`BuildVersion` / `BUILD_VERSION`) beside `ProtocolId` in every backend; that it never enters the projection (it is derived from it); independence tests (a table edit moves the build version and not the protocol id; a type edit moves both). Then the emitter, all backends, with goldens.

Gate item for schema#248, after the block form (#291) lands; queued with the wire v2 emitter (#251), which its cooked form depends on.

gafferongames:
Owner rulings, same minute (verbatim): "Any cook binary divergence = new build version." / "call it 'build version'" / "now, the trick is, protocol id remains as schema and types only" — (1) the NAME is 'build version'; (2) the DEFINITION: any divergence in the bytes a cook would produce is a new build version — it digests every fact a cook depends on (the tables' layout ids, and the type wire's protocol id since types nest in tables and the build's ABI is part of the cook); (3) the PROTOCOL ID stays the type wire's id and nothing else — tables never enter it (§10's independence holds: a table edit moves the build version, never the protocol id; a type edit moves both).

gafferongames:
[...]

PR #297, opened 2026-09-02
SPEC: the build version — one digest of the wire and the tables as they cook (#292, spec-first)

gafferongames:
Spec only — no compiler code. Closes the spec half of #292.

`SPEC-TABLES.md` gains **§20 The build version**; §7's cook-key sentence now names it, §10 states the pair from the tables side, §13.7 records the owner's verbatims, §15 gains two follow-ons. `SPEC.md` §3.2 states the boundary from the type wire's side. `USAGE.md` teaches it.

## The definition, in one line

**Any divergence in the bytes a cook would produce is a new build version** — one digest over three groups of facts, and the grouping is the definition:

1. **the type wire** — the unit's protocol id (types nest in tables by value);
2. **the layout** — every table's layout id (§7), which already carries the packed-layout facts keyed by wire id *and* this build's `sizeof`/`offsetof`;
3. **the meaning** — what a wire load PUTS in those slots: specified defaults, enum variant order and names, flags bit positions, union arm order and names.

Group 3 is the finding. A cook is produced from a builder a load filled, so **a specified default changed re-cooks the same asset into different bytes while moving no offset and no layout id** — same asset hash, same key, and the cache serves an artifact whose values are last week's with nothing anywhere to refuse it. It is §4.1's silent class arriving at the cook. A build version that digests layout only has this hole.

## Decisions, each labelled — these are the ones open for your word

- **D1 — it lives in `SPEC-TABLES.md` §20, not `SPEC.md` §3.3.** It depends on table layout ids, and putting a table-dependent concept in the language document is what §10 and the scope sentence keep apart. `SPEC.md` §3.2 instead states the boundary from its own side and points here. Appending §20 also leaves every frozen section number alone.
- **D2 — two instruments, one per group.** The meaning projection is **SHA-256, low 64 bits, over canonical text** — exactly §3.1's instrument, for exactly its reason: many small facts a walk could forget, so what it depends on must be printable and diffable. The **fold is `fnv1a64`** over the concatenated ids, 8 bytes big-endian each: its inputs are already digests, its input list is the unit's table list, and it must be a **constant expression in every backend's own language with no library**. A streaming 64-bit fold is one line everywhere; SHA-256 is not. (This also matches space's precedent — the connect id folds the unit protocol id and the delta id with FNV.)
- **D3 — TWO constants, because half the facts belong to the build.** A layout id mixes *this build's* `sizeof`/`offsetof` (§7), so the build version inherits that nature and is settled at the **target language's** compile time. `SchemaBuildVersion` is the half the compiler owns — printable, golden-pinnable, diffable, and what `schema build-version` prints, labelled. `BuildVersion` folds the full layout ids and the byte order and is the only one that keys a cook. **This is the biggest call in the PR**; the alternative is §15's *compiler-settled build version* below.
- **D4 — the build's BYTE ORDER rides in it**, one byte. A cook is produced in its target's byte order (§7), so two otherwise identical builds produce different cook bytes — and a key that can address two different artifacts is a defect in the key, whatever the cook's magic catches downstream.
- **D5 — the block digest (§19.3) is EXCLUDED, deliberately.** This diverges from the dispatch reading, so it is called out. A block is not a cook and §7 already rules the two must not share an id; folding it in re-admits exactly that sharing through the cache key and invalidates every cooked artifact in the unit for an edit no cook's bytes depend on. It also buys no discrimination — every fact the block digest sees is a by-value layout fact of the marked table or one of its row types, and each of those contributes its own cook layout id. Consequence, stated as a test: **adding `| block` to an already-fixed table moves nothing**, which is §2.7's "adds a form, removes nothing" made checkable.
- **D6 — the layout id ADDRESSES nothing and the build version REFUSES nothing.** Two ids, two jobs: the build version is unit-wide and coarse because a cache key must never address two artifacts; the layout id stays per-table and narrow because a guard should not invalidate a format nothing touched (§7's "a mesh catalogue cooked last year opens against this year's build"). §7 is otherwise unchanged. The cooked header keeps the layout id and does not gain the build version.
- **D7 — a table RENAME moves it; a `was` field rename does not.** §7 states the second as an obligation ("a `was` rename must not invalidate every cooked file in existence") and §20.3 honours it.
- **D8 — the `json` key (§16.4) does not move it.** A cook is produced beside the WIRE, which is the format of record; a JSON document is an earlier pipeline stage whose output is a wire file with a hash of its own.
- **D9 — baseline-only facts do not move it**, and the baseline's rendering version does not either.
- **D10 — a unit with no tables still has one**, deliberately not equal to its protocol id, so nobody can substitute one for the other.
- **D11 — meaning lines carry their ORDINAL and are then sorted**, so a reorder is a changed line and file layout reaches nothing.

## Named follow-ons added (§15) — the two roads not taken here

- **The meaning facts inside the per-file layout id.** §20.5 states the residual plainly: the layout id is a layout check and was never a meaning check, so a cook copied by hand out of an old tree opens against a build whose layout did not move. Inside the pipeline the key moved and the artifact was re-cooked, so nothing can produce that file; closing it per FILE re-prices every existing layout id and wants your word on which facts belong to a guard rather than to a key.
- **A compiler-settled build version** — lifting §19.3's model (the compiler computes the layout, both backends assert it, a disagreement is a build error) to the cook. That collapses D3's two constants into one, printable and pinnable in full, and turns ABI drift from a runtime refusal into a loud build failure. It commits the compiler's layout model to every cookable closure and not just to blocks and rows, which is the decision.

## Held by test (§20.7), for the emitter pass

Independence both directions; the meaning group's negative controls (a default change moves the build version and NO layout id; likewise an enum reorder, an enum rename, a flags reorder, a union arm reorder); the exclusions each with the edit that proves it; goldens over the corpus beside SPEC §7.2 gate 2; and the cross-language leg — two backends' `BuildVersion` are EQUAL when their ABIs agree, which is the equality a C++ tool cooking for a C# runtime already rests on and which nothing else asserts at unit scope.

## Verification

The §20 fold was implemented **twice, independently, from the page's text alone** (throwaway Go, deleted) and run over the `tabledemo` corpus — real protocol id `0x12cf1828bc593ac7`, real table list and real meaning facts read out of `tables/examples/tables.baseline`, stand-in layout ids (the real ones need a C++ build). The two readings agree on every number, and every discrimination behaves as §20 says:

```
tables 20   meaning facts 48
meaning digest      A 0x7bd6e7bbf7ead622   B 0x7bd6e7bbf7ead622
SchemaBuildVersion  A 0x33c56d0b36cb3eea   B 0x33c56d0b36cb3eea
BuildVersion        A 0xc3b7350e1b5be951   B 0xc3b7350e1b5be951   AGREE

a specified default changed          MOVED
an enum variant renamed              MOVED
two enum variants swapped            MOVED
a table renamed / a table added      MOVED
a flags variant renamed in place     held
declaration order reversed           held
```

`make check` is green.

## Two implementation lags the emitter pass will meet

Not touched here — both are the implementation trailing the spec, not spec changes:

- `internal/codegen/cpptable/pointers.go` emits `<T>LayoutId` as **`uint32_t`**, while §7 states the layout id is 64 bits and sized like a digest.
- Only variable-length tables emit a `LayoutId` today (`compiler/tables_test.go` asserts a value-only table has none). §20.1 group 2 needs one per table, which §7's "the cook of a FIXED root table is the same idea with nothing in it" already implies.

Draft, not for merge — the decisions above want your word first.

### DIFF
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -4581,16 +4581,16 @@ out-of-line array's element and pitch, and every row field's offset, size and
 kind — the `block` and `slot` lines of §20.2. A cook and a block share the id
 and differ in their MAGIC, which is where a form's identity belongs (§7).
 
-**A declared MAXIMUM is not a line in the projection**, and for the block's
-own facts that is exactly right: a maximum sizes the storage and moves the
-`offset_of`s written into the instance, and it moves no offset a consumer
-reads AT, because a consumer takes every `offset_of` from the instance
-(§19.2). A block's `slot` lines therefore do not move when one is raised — a
-triple is sixteen bytes whatever the maximum is. **But the table's own
-by-value `record` lines DO move**, because raising a maximum grows the inline
-array's storage and every later field's offset with it, and those lines are
-what a COOK of that table is opened against. So the build version moves, and
-`BlockOpen` refuses the edit like any other (§19.4).
+**A declared MAXIMUM moves NO `slot` line**, and for the block's own facts
+that is exactly right: a maximum sizes the storage and moves the `offset_of`s
+written into the instance, and it moves no offset a consumer reads AT, because
+a consumer takes every `offset_of` from the instance (§19.2). A triple is
+sixteen bytes whatever the maximum is. **But it is a `bound=` on the table's
+own `field` line, and the by-value `record` lines move with it**, because
+raising a maximum grows the inline array's storage and every later field's
+offset — and those lines are what a COOK of that table is opened against. So
+the build version moves, and `BlockOpen` refuses the edit like any other
+(§19.4).
 
 **What the id sees, and what it does not.** It sees layout and it sees
 meaning — the facts that decide what a load puts in a slot are §20's group 3.
@@ -4724,9 +4724,10 @@ and never the protocol id; a type edit moves both.**
 Tooling cooks before any game binary exists, so Y has to be knowable from the
 schema alone. The compiler owns every fact in it, including the layout: it
 computes each record's layout from its own C ABI model — the model §19.3
-already states and both backends already assert against — and emits the id as
-one constant. A build whose compiler lays a record out differently fails to
-BUILD, loudly, naming the type and the field (§20.3); it never reaches a cook.
+states and both backends MUST assert against (§20.3) — and emits the id as one
+constant. A build whose compiler lays a record out differently is meant to
+fail to BUILD, loudly, naming the type and the field; those asserts are owed
+and not yet emitted, which §20.3 states in full.
 
 **Backend status: specified, unimplemented.** No backend emits `BuildVersion`,
 `schema build-version` does not exist, and `schema cook-check` (§7) does not
@@ -5061,16 +5062,15 @@ wrong fails to build instead of degrading.
   integer it rides as, a union arm's payload swapped. Each keeps the kind, the
   size, the offset and the wire id and changes what the bytes MEAN, and each
   moves a `type=`, `enum=`, `union=`, `key=` or `payload=` token;
-
-- **a declared MAXIMUM raised or lowered.** A maximum is not itself a line in
-  the projection, but the inline storage it sizes is: the array field's `size`
-  grows, every later field's `offset` moves and the record's `sizeof` moves
-  with them. It could not be excluded even if a block would rather it were —
-  a build that read another's cook under a larger declared maximum would read
-  past the region — and under one same-build entry point per form there is
-  nothing for the exclusion to buy: `BlockOpen` refuses this edit exactly as
-  it refuses an appended field (§19.4), and both sides regenerate. The
-  block's own `slot` lines still do not move, because a triple is sixteen
+- **a declared MAXIMUM raised or lowered.** It is a `bound=` on the field
+  line, and the inline storage it sizes moves with it: the array field's
+  `size` grows, every later field's `offset` moves and the record's `sizeof`
+  moves with them. It could not be excluded even if a block would rather it
+  were — a build that read another's cook under a larger declared maximum
+  would read past the region — and under one same-build entry point per form
+  there is nothing for the exclusion to buy: `BlockOpen` refuses this edit
+  exactly as it refuses an appended field (§19.4), and both sides regenerate.
+  The block's own `slot` lines still do not move, because a triple is sixteen
   bytes whatever the maximum is;
 
 - **a specified default changed, added or removed**, and **a declared range
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -70,6 +70,42 @@ one language and read every frame in another — so it is the hottest path
 tables have, and the fastest-correct mission bites hardest there. §12.1 is
 the gate that decides whether it earns its place.
 
+**THREE FAST PATHS, and one deliberately slower one** — the owner's framing:
+*"notice how we now have fast paths for three things? types, cooked tables,
+and fast write/read tables like render data."* Each is fast for its own
+reason, and the reasons do not transfer:
+
+- **Types (SPEC.md)** — same build on both ends, so the bytes ARE the struct
+  under a hardcoded bitpacked wire and one protocol id. Nothing is negotiated
+  because nothing may differ.
+- **Cooked tables (§7)** — the work is done OFFLINE and once: tooling builds,
+  cooks, and endian-fixes for a known target, and the game memory-maps the
+  result and points at it. The reader does no parse because the writer
+  already did every part of it.
+- **The block form (§19)** — both sides GENERATED against one layout the
+  compiler settled, so the producer goes wide with nothing to synchronise and
+  the consumer indexes rows at a pitch it reads. Nothing is discovered at
+  frame time because everything was decided at build time.
+
+**And the tolerant wire (§3) is slower on purpose**, because it is the one
+form that CROSSES VERSIONS: ids, kinds, lengths and a read report are what
+let any reader read any data, years apart, and every one of them is a byte
+the three fast paths do not spend. That is the trade this document exists to
+make, and it is why the fast paths are accelerators BESIDE the wire rather
+than replacements for it.
+
+**And one path is generic on purpose and may allocate** — the reflection
+surface (§8): the field walk, the text form (§16), the packer (§17), a
+viewer. In the owner's words: *"we also have the flexibility of the
+reflection based stuff that is more generic and does allocations if we want
+it for tooling, editor, whatever where perf and allocations aren't such a
+big deal."* It buys REACH — walk any declaration by name without knowing its
+type — and it pays in speed and in allocation, which is the right trade
+where a tool or an editor is the caller and the wrong one anywhere the
+three fast paths run. **That is the ladder whole**: three fast paths, one
+slower path that crosses versions, one generic path that trades speed for
+reach.
+
 ## What allocates, and what never does
 
 - **A type never allocates**, in any backend, on any path. This rung does
@@ -83,12 +119,22 @@ the gate that decides whether it earns its place.
   allocated on read and write — and a backend with a native union (C++)
   allocates nothing for the same declaration.
 - **The variable-length class allocates by nature**, and assuming
-  otherwise is foolish: the arena on build, the region on load. C++ keeps
-  the caller-owned form — `LoadMeasure` sizes the region and the caller
-  supplies it, `Builder` grows storage the caller owns — which is
-  allocation with the caller holding the pointer, not its absence. Other
-  backends allocate inside their runtime and say so. No port contorts
-  itself toward zero allocation for a variable-length table.
+  otherwise is foolish: the arena on build, the region on load. It allocates
+  through a **caller-provided allocator with malloc semantics** — C and C++
+  take the allocator; other languages take their platform's default — and it
+  allocates **at BUILD time, in bulk, never per record on the fill path**: a
+  worker carves its nodes out of a slab the arena already holds (§6.4), so
+  going wide costs no allocation and no lock per node. C++ keeps the
+  caller-owned form — `LoadMeasure` sizes the region and the caller supplies
+  it, `Builder` grows storage the caller owns — which is allocation with the
+  caller holding the pointer, not its absence. Other backends allocate inside
+  their runtime and say so. No port contorts itself toward zero allocation
+  for a variable-length table.
+- **EVERY READ PATH ALLOCATES NOTHING**, in every class, on every form. A
+  load fills caller-owned storage, a region is caller-owned, `Open` and
+  `BlockOpen` point at bytes the caller already has, and the reflection walk
+  reads in place. The allocation in this document is a BUILDING cost, and
+  building is TOOLING's path — the game points at the cook (§7).
 
 **Backend status: C++, and C# for the FIXED class.** C++ carries both classes;
 C# carries the fixed class (§6.1) — optionals, enum-keyed arrays and all — and
@@ -1604,24 +1650,45 @@ SceneCook( builder, buffer, data, attribution );   // write it
 const Scene * scene = SceneOpen( bytes, size );    // point at it, or NULL
 ```
 
-**The pipeline.** `schema pack` writes the WIRE, and the wire is the
-format of record (§17). A cook is produced beside it, next to the build
-that will read it — on the machine or on a distributed cook farm — for
-exactly the layouts that build knows. A game loads its own cooked assets
-and never a foreign one; tooling reads and writes the generic wire,
-because the generic form is the one that carries every version. And if
-load time does not demand the accelerator, the game just uses the generic
-table: cooking is a choice made per asset, never a requirement of the
+**The pipeline, in the owner's words**: *"the optimized path is still
+available, it is tooling does the build, then cooks to the rad binary format.
+and the game just points at it and works."* — *"(plus endian fixups)"* —
+*"this is the way."* `schema pack` writes the WIRE, and the wire is the format
+of record (§17). A cook is produced beside it, next to the build that will
+read it — on the machine or on a distributed cook farm — for exactly the
+layouts that build knows. **TOOLING BUILDS; THE GAME POINTS.** A game loads
+its own cooked assets and never a foreign one; tooling reads and writes the
+generic wire, because the generic form is the one that carries every version.
+And if load time does not demand the accelerator, the game just uses the
+generic table: cooking is a choice made per asset, never a requirement of the
 format.
 
+**The endian fix-up is part of the COOK, and it stays.** A cook is produced in
+the byte order of the build it is cooked for, so the fixing happens where the
+target is known — offline, once, on the writing side — and never on the
+reading side, which is what makes `Open` a match and a point rather than a
+pass over the region (below). The byte order is a fact of the build version
+(§20.1), so a cook for a foreign order is not this build's file and refuses.
+
 **A cooked artifact is CONTENT-ADDRESSED by a pair — the hash of its source
-asset, and the cooked-version hash, which is this build's table protocol id
-and on this page is the LAYOUT ID (below)** — and that pair is the key the
-runtime searches for, the key a distributed build cache produces under and
-serves from. It is why the cooking side is a build cost rather than a
-runtime one: the work happens offline, once per (asset, layout), and the
-game does a lookup. That is the fact the performance ladder cites when it
-calls the wire and the cook read-hot and write-cold.
+asset, and the unit's BUILD VERSION (§20)** — and that pair is the tuple the
+runtime searches for, the tuple a distributed build cache produces under and
+serves from. It is why the cooking side is a build cost rather than a runtime
+one: the work happens offline, once per (asset, build version), and the game
+does a lookup. That is the fact the performance ladder cites when it calls the
+wire and the cook read-hot and write-cold.
+
+**The ASSET HASH is the hash of the WIRE FILE the cook was produced from.**
+The wire is the format of record and a cook is produced beside one (below,
+§17), so naming the wire file is what makes the pair well defined: a
+pipeline that ran a text tree through `schema pack` and then cooked in one
+step still keys on the bytes the cook actually read, and an edit upstream of
+those bytes reaches the key through them.
+
+**The build version both ADDRESSES a cooked artifact and REFUSES one**: it is
+what the store is indexed by, and it is what `Open` checks out of the header.
+There is no second VERSION id — the design has two in total, the protocol id
+for the type wire and this one for everything cooked or blocked (§20).
 
 The example pair is the whole rule. A huge data file naming every mesh in
 the game, or every texture — that is a cook. A configuration file small
@@ -1629,18 +1696,20 @@ enough that the cost of loading it does not matter is the wire, and stays
 the wire, and keeps the flexibility that comes with it.
 
 - **The header build-locks it**: a magic (which is also the byte-order
-  check), a LAYOUT ID, and the lengths of the two parts below. The
-  reserved words are reserved: a non-zero one means a writer used a form
-  this build does not understand, and `Open` refuses rather than ignoring
-  it.
+  check), the unit's BUILD VERSION (§20), and the lengths of the two parts
+  below. The reserved words are reserved: a non-zero one means a writer used
+  a form this build does not understand, and `Open` refuses rather than
+  ignoring it.
 
   A cooked file never crosses builds, so most of the header's shape is
   the implementation's business. **Three widths are not**, because each
   one decides something semantic:
   - **The magic is read BYTEWISE, before anything else**, since it is
     what establishes the byte order every other header field is written
-    in.
-  - **The LAYOUT ID is 64 bits.** Under the rule below a matching id
+    in. It is also what separates a cook from a BLOCK (§19): the two
+    accelerators carry the same build version and different magics, and a
+    form's identity belongs in its magic rather than in a second digest.
+  - **The BUILD VERSION is 64 bits.** Under the rule below a matching id
     means `Open` checks nothing further, so the id is the sole guard
     between a runtime and a foreign region; it is sized like a digest,
     not like a version counter.
@@ -1649,57 +1718,64 @@ the wire, and keeps the flexibility that comes with it.
     ceiling §3.1 just removed — on exactly the huge mesh or texture
     catalogue a cook exists for.
 
-  The layout id is a compile-time digest mixing the schema's
-  packed-layout facts with this build's own `sizeof` AND `offsetof` for
-  every type and field in the closure, so schema drift and ABI drift both
-  refuse — `sizeof` alone cannot see a member that moved inside an
-  unchanged total, and the packed form is read by offset. It is the
-  region form's twin of the protocol id.
+  The build version is a compiler-settled digest over every fact a cook's
+  bytes depend on: the type wire's protocol id, every record's layout as the
+  compiler's own C ABI model computes it, and the facts that decide what a
+  load PUTS in those slots (§20). It KEYS A FIELD BY ITS WIRE ID, not its
+  source name — the identity `was` preserves (§5), because a `was` rename
+  moves no byte and must not invalidate every cooked file in existence.
+  Reordering two same-shaped fields does move bytes, and it moves which id
+  sits at which offset, so it still refuses. Both directions are held by test.
 
-  **It keys a field by its WIRE ID, not its source name.** That is the
-  identity `was` preserves (§5), and a `was` rename moves no byte — so it
-  must not invalidate every cooked file in existence. Reordering two
-  same-shaped fields does move bytes, and it moves which id sits at which
-  offset, so it still refuses. Both directions are held by test.
+  **ABI drift is a BUILD ERROR, not a refusal here.** The compiler computes
+  the layout and both backends assert it, so a compiler that lays a record
+  out differently fails to build, naming the type and the field (§20.3). That
+  is what lets the id be settled from the schema alone, which is what tooling
+  needs: a cook is produced before any game binary exists.
 
-  A format that rarely changes therefore keeps its id across builds by
-  construction: the id follows wire ids and offsets, so a layout nothing
-  touched does not move, and a mesh catalogue cooked last year opens
-  against this year's build if nothing in its closure changed.
+  **The id is UNIT-WIDE, so a cook outlives only what the unit does.** A mesh
+  catalogue is re-cooked when any table in its unit moves, not only when its
+  own closure does. That is the model rather than a cost — the work is
+  offline and a build cache exists to absorb it (§20.6) — and it is the price
+  of two ids in the whole design instead of three.
+
 - **A cooked file is TWO PARTS, and the header locates both.** The DATA
   comes first, at the region's alignment: it is `Lock`'s region written
   verbatim, the root at its base, and it is what the runtime points at.
   The ATTRIBUTION follows it: the node directory of §6.3, which says
-  where every node starts and what type it is. Nothing that reads the
-  structure touches the attribution, so a shipped build need not carry it
-  at all — the header records its length as zero and the file is just
-  data.
+  where every node starts and what type it is. **Nothing that READS the
+  structure touches it** — it is written beside the data for the TOOL
+  (`schema cook-check`, below — **not built**), so a build that ships no
+  tooling need not carry it at all: the header records its length as zero and
+  the file is just data.
 - **`Open` checks the header and points, and this is the WHOLE check**,
-  because nothing else is checked at all: the magic, the byte order it
-  establishes, the layout id, the two part lengths against the `size` the
-  caller passed — a truncated file refuses — and the alignment of the
-  base. On a match the bytes ARE what this build wrote, in this build's
-  layout and this build's byte order, so there is nothing to validate and
-  nothing to fix up: `Open` returns the root. That is the runtime path
-  and it is the default. On any failure it returns NULL and the caller
-  falls back to a wire load, which is the path that carries every
-  version.
-- **`OpenValidated` is the other entry point, and it is named rather than
-  implied.** It takes a cooked file whose provenance the caller does not
-  trust, or one a tool is diagnosing, and checks the data against the
-  attribution before returning anything. There is no SILENT bypass
-  anywhere: a caller either gets the layout-id guarantee, or asks for the
-  check by name. A file with no attribution part cannot be checked, so
-  `OpenValidated` refuses it and says which part is missing. **A cook
-  that arrived from somewhere else is where it earns its keep**: the
-  pipeline above contemplates a distributed cook farm, and a file that
-  crossed a machine boundary carries a matching layout id and nothing
-  else — a caller that wants more than the id's word takes the check by
-  name.
+  because nothing else is checked at all: **the magic, the byte order it
+  establishes, the build version, every RESERVED word zero, the two part
+  lengths against the `size` the caller passed — a truncated file refuses —
+  and the alignment of the base.** That is the enumeration, and it is stated
+  once: §11 and §20.6 cite it rather than repeat it. On a match the bytes ARE
+  what this build wrote, in this build's layout and this build's byte order,
+  so there is nothing to validate and nothing to fix up: `Open` returns the
+  root. That is the runtime path and there is no other. On any failure it
+  returns NULL and the caller falls back to a wire load, which is the path
+  that carries every version.
+
+- **`Open` is the RUNTIME's only entry point.** There is no second one: a
+  build either wrote a file or it did not, and the build version is what says
+  which. A cook is an accelerator for a build's own assets, and a runtime
+  that wanted to reason about a foreign one has already left the fast path.
+- **Validating an untrusted cook is a TOOL, not a runtime surface**:
+  `schema cook-check` (**not built**), over the same reflection descriptors
+  (§8) the runtime already carries, checking the DATA against the ATTRIBUTION.
+  That is where the case lives — a file whose provenance a person doubts, or
+  one a tool is diagnosing — and it is a person's decision to run it, not a
+  parameter on a load. A file with no attribution part cannot be checked, so
+  the command refuses it and says which part is missing. The ATTRIBUTION is
+  written beside the data for exactly this reader.
 - **The check is a SCAN of the attribution, not a traversal of the
   graph.** Two passes, in order:
   1. **The directory itself**, linearly and with no state: it lies inside
-     the file, every type id names a table this build has, the
+     the file, every type id names a table the unit has, the
      materialized offsets ascend, each is aligned for its own type, and
      each node's storage fits before the next entry. A sentinel entry
      (§6.3) refuses here — a cooked file is an accelerator and cannot
@@ -1721,6 +1797,7 @@ the wire, and keeps the flexibility that comes with it.
   the root can reach. The cost is `O(R + P log N)` — linear in the
   region, plus one search per pointer slot — with no allocation and no
   per-node state, and it terminates on every input.
+
 - **Pack order and checking are INDEPENDENT.** The region is packed in
   pre-order (§6.3) because that is the order the wire numbers nodes in
   and one order is simpler than two, but nothing in the check reads that
@@ -1737,22 +1814,23 @@ the wire, and keeps the flexibility that comes with it.
   A fixed-size table is one struct (§6.1), so its cooked form is the
   struct's bytes behind the header: memcpy it, or point at it where it
   lies. There is no region, no node table and no attribution, and the
-  layout id is the whole of the check.
+  build version is the whole of the check.
 - **Every refusal is loud and the fallback is a real wire load**: wrong
-  magic or byte order, a layout id this build did not produce, a
-  truncated file, an unaligned base — and, under `OpenValidated`, an
-  attribution part that is missing, leaves the file, carries a sentinel
-  entry, names a type this build does not have, does not ascend, or
-  overlaps a node with the next; a reference that leaves the region, that
-  the directory does not name, or that it names as another type; a
-  misaligned reference; a buffer outside its node's extent; a count
-  companion outside its declared bound.
+  magic or byte order, a build version this build did not produce, a
+  truncated file, a non-zero reserved word, an unaligned base. **And `schema cook-check`'s refusals are
+  the tool's, not the runtime's**: an attribution part that is missing,
+  leaves the file, carries a sentinel entry, names a type the unit does not
+  have, does not ascend, or overlaps a node with the next; a reference that
+  leaves the region, that the directory does not name, or that it names as
+  another type; a misaligned reference; a buffer outside its node's extent;
+  a count companion outside its declared bound.
+
 - **Alignment.** The header pads the data part to the region's alignment,
   so a base the allocator or `mmap` gave you is already aligned; `mmap`
   gives page alignment for free.
 - **Endianness is part of the COOK, not of `Open`.** A cook is produced
   in the byte order of the build it is cooked for — the cook knows its
-  target — so a matching layout id already means a matching byte order
+  target — so a matching build version already means a matching byte order
   and `Open` never fixes anything up. Cooking for a foreign target is
   where a byte swap would live if one is ever wanted (§15); a cooked file
   whose recorded order is not this build's is simply not this build's
@@ -1764,13 +1842,16 @@ for it. Here the tolerant wire stays the format of record and the cooked
 form is a build-locked accelerator beside it, produced only where load
 time asks for one. The two-form split is the design.
 
-**A COOK and a BLOCK are different accelerators and must not share a layout
-id.** Both are build-locked projections of one declaration, but they lay it
-out differently — a cook writes the by-value form verbatim, a block writes
-the projection with its rows out of line (§2.7) — so one id covering both
-would let a runtime accept one where the other was written. Each carries its
-own id, computed over its own facts, and neither is checked against the
-other's.
+**A COOK and a BLOCK are different accelerators and a runtime must never
+accept one where the other was written.** Both are build-locked projections
+of one declaration, but they lay it out differently — a cook writes the
+by-value form verbatim, a block writes the projection with its rows out of
+line (§2.7). **What separates them is the MAGIC**, which each form has its
+own of and which is read bytewise before anything else. They share the BUILD
+VERSION, because a build version answers "which build?" and not "which
+form?", and the projection carries both forms' facts (§20.2): a block's
+`slot` lines sit beside its record's `field` lines, so an edit to either
+moves the one id.
 
 **And prior art gets one MEASURED sentence, from the case §19 exists for.**
 The render data this document's second gate is held to (§12) was built with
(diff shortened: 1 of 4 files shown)


gafferongames:
## Cold read — §20, adversarial, no part in writing it

Read whole: CLAUDE.md, SPEC.md §3, SPEC-TABLES.md §2.7/§3/§4/§4.1/§5/§7/§10/§16/§17/§18/§19, the diff, #292 with every comment, #248. The fold was implemented from the page's text and run over `tables/examples/` (throwaway Go, deleted). **All three published numbers reproduce exactly** — and reproducing them is what surfaced three of the four defects below.

The rendering that produces `0x7bd6e7bbf7ead622` — recovered by searching 78,400 candidates against the published target, because the page states none:

```
default <Table>.<field> <value>        # e.g. "default WeaponConfig.damage 21.0"
enum  <Enum>  <ordinal, 0-based> <name>
flags <Flags> <bit>                    # NO NAME on the line
union <Union> <ordinal, 0-based> <arm>
```
sorted byte-wise, `\n`-joined, one trailing `\n`, no form token in the text; SHA-256 low 64 big-endian. Fold: `fnv1a64` over ASCII `schema-build-version 1` ++ BE64(protocol id) ++ BE64(meaning digest) ++ BE64(each layout id, tables sorted by name) ++ `0x01` for `BuildVersion`.

```
meaning digest      0x7bd6e7bbf7ead622   match
SchemaBuildVersion  0x33c56d0b36cb3eea   match
BuildVersion        0xc3b7350e1b5be951   match
```

---

## DEFECTS

**D-1. The meaning group is not closed. A tightened numeric range diverges a cook's bytes and moves nothing.**

§3 of this page, on the neutral wire: *"a ranged integer rides as its storage-width integer kind... The bounds and resolutions stay on the DECLARATION side, where they validate and clamp on load (§4) — they never change what the bytes look like."* §4: *"Out-of-range value (bounds tightened since the writer): clamped to the reader's declared bounds, counted."*

So a declared range is a **load-time clamp**. Tighten one and a wire load puts a different value in the slot — while the offset does not move, the size does not move, the storage kind does not move, the wire id does not move (§4: *"the bound is not part of identity"*), and none of §20.1's four meaning kinds moves. Same asset, same key, different cook bytes. This is exactly the failure §20.1 group 3 was written to close, one construct over, and it defeats the invariant verbatim.

Three instances, all legal in a table body today:
- a numeric field's declared `[min, max]` tightened;
- a compressed float's declared range narrowed (resolution too, if a load quantizes — the page does not say whether it does, which is its own gap);
- `bits(N)` narrowed **within one storage width** — `bits(12)` to `bits(10)` both ride as the same unsigned kind and store at the same width, but §8 states the implied range `[0, 2^N − 1]` is carried in the descriptor and *"the codec has always clamped to it (§4)"*. The implied range moves; nothing else does.

Nothing else catches it either: §18.1's baseline records `name, id, kind, element kind, array shape, capacity, presence, default, was` — **no range line at all** — so `schema check` passes it in silence too.

Fix: group 3 gains a fifth kind — every table field's **effective declared range** (and a compressed float's resolution), EVALUATED, on the same terms as a default. It is the same silent-value class and it belongs in the same list. §18.1 wants the same line for the same reason, but that is a separate PR.

**D-2. The flags rules are internally contradictory, and flags should not be in group 3 at all.**

§20.3 says the build version MOVES on *"a `flags` variant inserted, removed or reordered"* and does NOT move on *"a `flags` variant renamed in place"*. **No function of a flags declaration can do both.** Take `Perks { Shielded, Cloaked, Turbo }` and apply three renames in place: `Shielded`→`Tmp`, `Cloaked`→`Shielded`, `Tmp`→`Cloaked`. Each step is a rename in place and moves nothing by rule; the endpoint is the declaration reordered, which must move by rule. Contradiction.

The reference implementation resolves it by dropping the name, and my run confirms the consequence:

```
flags variants REORDERED (Shielded <-> Cloaked)   held  <-- does not move
flags variant RENAMED in place                    held
```

So §20.3's move list and §20.7's negative control (*"Same shape for... a flags variant reordered"*) are both **false against the only implementation that produces the published digest**. The PR's verification table tested `a flags variant renamed in place → held` and never tested the reorder — the untested row is the broken one.

The invariant picks the winner, and it is not the move list. A flags value rides as its raw mask and a load copies it verbatim; a reorder changes **no cook byte**, only what the bits MEAN. That is §4.1's silent class, whose answer this document already gives: discipline (append at the end) plus §18's baseline, which REFUSES a flags insert/reorder/rename at compile time. Putting it in a cache key is over-refusal on a fact no cook's bytes depend on.

Fix: **drop flags from group 3** and strike "flags variant inserted, removed or reordered" from §20.3, keeping the rename exclusion. Then §20.1's "exactly four kinds" becomes three, and the closure argument gets stronger rather than weaker — enum, union and default all genuinely change a stored value; flags does not. One caveat to settle first: the page never states whether a table load MASKS a flags value to the declared width. If it does, then the flags **width** (not order) is a real meaning fact and belongs; the order still does not. Say which.

**D-3. The meaning projection has no line grammar and no example. The page is not implementable.**

SPEC.md §3.1 points at `ir.WireProjection` and a live `schema projection`. §18.1 prints a full worked baseline. §20 gives the reader neither, and `--facts` (§20.6) is described without a single sample line. Every one of these is a guess a second implementer must make, and each changes the digest:

- the line's shape and prefixes (`default `/`enum `/`flags `/`union `), the separator, `Table.field` vs `Table field`;
- **whether a field is identified by NAME or by WIRE ID** (see D-4);
- whether only fields WITH a default get a line, or every field gets one (the sparse reading is what makes 48; the dense reading gives ~100);
- ordinal base — 0 or 1;
- the join, and whether a trailing newline is present;
- what bytes the `BuildVersionForm` token contributes to an `fnv1a64` fold whose other inputs are declared as "each id eight bytes, big-endian" (it is ASCII `schema-build-version 1`, which no reader could derive);
- the meaning digest of an EMPTY set (§20.6 turns on it) — SHA-256 over `""` or over `"\n"`;
- the SCOPE of group 3 (see R-6).

"Implemented twice and agreed" is not evidence the page is implementable — both readings were the author's, and by the PR's own account the facts were read out of `tables.baseline`, a format §20 never names. **Fix: §20.1 gains the line grammar and §20.6 gains a `--facts` sample, exactly as §18.1 carries one.** The recovered rendering above is a serviceable starting text.

**D-4. As specified-and-implemented, a `was` rename MOVES the build version.**

The default line keys the field by NAME. `WeaponConfig.speed` (already `was=velocity`, id `0x2e46`) renamed again under `was`:

```
a field with a default renamed under `was`   MOVED
0xc3b7350e1b5be951 -> 0x9204b4e39773a3b3
```

That violates §20.3's exclusion, D7, §20.7's test, and §7's stated obligation — *"a `was` rename moves no byte — so it must not invalidate every cooked file in existence."* §7 keys its layout id by wire id for precisely this reason and says so. §20.1 says nothing about which identity a meaning line uses.

Fix: state that a meaning line identifies a field by its **wire id** (with its table), as §7 does. Then `was` holds and a rename without `was` still moves (the id changes) — redundantly with group 2, which is fine.

---

## RISKS, with recommendation

**R-1 (D5). The decision is right; its first reason is not.** §20.3 argues that folding the block digest in *"would re-admit exactly that sharing through the cache key."* It would not. §7's rule is about the id **in the cooked file's header**, which is what lets a runtime accept one form where the other was written; the build version is a cache key and mixing the block digest into it gives no runtime any new thing to accept. Strike that clause.

The exclusion is still correct, on the two sound reasons: no cook's bytes depend on any block-only fact, and the block digest's facts are a subset of the cook layout facts. I tried to construct the stale-serve case you asked about and **it cannot be built on this page**: a `| block` table is an ordinary fixed table whose cook writes the by-value form verbatim (§7), a block is never persisted (§19.1 leaves its tail unspecified), and the one fact the block digest excludes — the declared maximum — is squarely inside the cook layout id (§20.3, "an array's bound changed"). D5 holds. Two wording fixes: strike the clause above, and note that a row element may be a declared `type` rather than a table (§2.7), so it contributes through the marked table's own closure (§7) and not through group 2's "every table."

**R-2 (D6). The coarse key deletes §7's mesh-catalogue property in practice, and the page still advertises it.** §20.5 keeps *"a mesh catalogue cooked last year opens against this year's build"* as the layout id's virtue. `Open` will indeed accept that file — but under D6 the cache will never hand it to anyone, because a single unrelated table gaining a field moves the unit-wide key and every artifact in the cache is re-cooked. For the motivating case (*"a huge data file naming every mesh in the game, or every texture"*) that is the whole cost of the design. §20 prices it as *"a key that is too coarse costs a re-cook"*, which understates it. Recommend: say it in the case's own terms in §20.5, and bank a follow-on for a narrower cook key (per-table or per-closure) beside §15's existing "declared compatible set", which addresses `Open` and not the key.

The trace itself is sound. Cache key `(asset hash, build version)` → file → `Open` checks magic, byte order, layout id, part lengths, alignment. A build at version B folds every table's full layout id into B, so a hit under the right key implies matching layout ids and `Open` cannot refuse it — **except on an `fnv1a64` collision or a corrupt file, where it returns NULL and the caller falls back to a wire load.** Both the implication and the collision caveat are the reader's obvious question and neither is stated; one sentence in §20.5 closes it.

**R-3 (D3). The split is necessary — keep it — but the number that keys the cache is not obtainable by any tool.** The alternative is real: §19.3's model would collapse the two constants into one printable, pinnable number and turn ABI drift from a silent runtime refusal (`Open` returns NULL, wire fallback, a performance cliff nobody sees) into a loud build error. `cpp_native` is not the blocker one would expect — SPEC.md guarantees layout identity by derivation and already asserts it. The real price is that it commits the compiler's C-ABI model to **every cookable closure on every target**, where §19.3 commits it only to blocks and rows; a platform the model gets wrong becomes a hard build failure instead of a graceful degrade. Under #248 — nine languages to port, two backends with tables today — that is premature. Recommend: keep D3, keep the §15 follow-on, and revisit after the second backend's cook lands, which is exactly what §20.7's cross-language leg would then be measuring.

One hole to close now, though: §20.6 forbids the only value a tool can print from keying a cache (*"no cache may use it"*), and `BuildVersion` is a constexpr that exists only after compiling. A cook farm running the target binary is fine; a **dispatcher** deciding which farm may serve which artifact — the case §20.2 names — cannot get the number at all. Say how it does (a generated entry point, a symbol, a `--abi` mode), or the "distributed build cache produces under and serves from" sentence has no mechanism behind it.

**R-4. The cook FORMAT's own version is not a fact of the build version.** The region's pack order (§6.3), the node-directory encoding, the header shape — change any of them in the compiler or runtime and a cook's bytes diverge with the build version unmoved. §7 explicitly contemplates the pack order changing (*"the layout can change without silently weakening it"*). No correctness bug follows (an old file is internally consistent and still opens), but the invariant as written is literally violated. Recommend a **cook-form token** folded in beside `BuildVersionForm` — it also covers the header-width migration in R-10 — or narrow the invariant's wording to "any divergence in what a load produces". State which reading is meant; the section opens on the byte reading.

**R-5 (D8). The `json` exclusion rests on a fact §7 does not state.** §20.3's argument is that a JSON document *"is an earlier stage whose output is a wire file with a hash of its own"* — true only if the cook's asset hash is the **wire file's** hash. §7 says *"the hash of its source asset"*, and §17's `schema pack` turns a JSON tree into wire, so a pipeline keying on the JSON tree and running pack→cook as one step has a stale-hit hole on a `json` key rename. Cheap fix, and it belongs in §7 where the hash is defined: the asset hash is the hash of the wire file the cook was produced from.

**R-6. Group 3's scope is unstated.** §20.6 asserts a table-less unit has *"the meaning digest of an empty set"*, which is only true if group 3 is scoped to the table closure — but §20.1 says "every `enum`", unqualified, and a unit with no tables still has enums. The scope must be the **transitive** table closure, including enums reached only through a nested `type`: a type's enum field stores an ordinal into a cooked region, and the protocol id does not cover it (SPEC.md §3.1 excludes variant names, and an enum reorder moves neither `max` nor storage bits). One sentence.

**R-7. §20.7's union-arm negative control does not isolate group 3.** SPEC.md §3.1 already includes *"union variant order"* in the protocol id, so a union arm reorder moves the build version through group 1 and the test would pass with group 3's union fact deleted. Use an arm **rename** (which the protocol id excludes) as the control, and note the overlap.

**R-8. Group 2's facts do not name KIND, but §20.3 relies on them.** §20.3 says the version moves on a field *"retyped"*. §7 defines the layout id over packed-layout facts, `sizeof`/`offsetof` and wire ids; §20.1 restates it as "keyed by WIRE ID". A `uint32` changed to `float32` moves no offset, no size and no wire id — yet the load kind-mismatches and the slot lands at its default, so the cook's bytes change. §19.3's block digest names kind explicitly (*"every row field's offset, size and kind"*); §7's does not. Add the word to §7 or to §20.1 group 2.

**R-9. Guards are absent from both lists.** §4.1 goes out of its way to place a guard added/removed adjacent to the silent class. It correctly moves nothing here (a load finds a field by id whatever branch encloses it), but §20.3's exclusion list should say so, since a reader who just read §4.1 will look for it.

**R-10. The implementation lags are undercounted — three, not two.** Beside the 32-bit `LayoutId` and the variable-tables-only emission, `internal/codegen/cpptable/arena.go:268` writes `TableCookedHeaderWrite( uint8_t *, uint32_t layout_id, uint32_t bytes )` — a **32-bit part length**, and only one of the two §7 requires. §7 states 64 bits with its reason (*"a 32-bit header field would reimpose at cook time the ceiling §3.1 just removed"*), and the corpus already carries 70,000-byte fields. Also: §16.1 and the top of the page carry backend-status paragraphs; **§20 carries none**, so the three lags live only in a PR body. Recommend a status line in §20 on §16.1's model, and an issue so the emitter pass inherits all three.

**R-11. Two verbatims gained a period inside the quotes** — §13.7's *"call it 'build version'."* and *"...same build version."* Neither has one in #292. Everything else in §13.7 matches the issue and its five comments word for word, including the ordering.

**R-12. The published run did not exercise D3.** The stand-in made a table's schema layout id and its full layout id the same value, so `BuildVersion` is `SchemaBuildVersion`'s input plus one byte. §20.7's last test — a `sizeof` perturbation moves one and not the other — is untested by construction, which is worth knowing given D3 is the biggest call in the PR.

---

## VERIFIED GOOD

- **The finding is real and it is the section's reason to exist.** Group 3 catches a class a layout-only digest cannot see, and the default case is exactly as described.
- **The independence pair reads correctly everywhere it appears** — §20 opening, §10's new paragraph, §20.7's first test and SPEC.md §3.2 all say *a table edit moves the build version and never the protocol id; a type edit moves both*, in those words. SPEC.md's frozen `table=false message=false round=nearest` projection tokens are untouched, `schema id` is stated unchanged, and no table fact reaches the wire-shape projection in either direction.
- **The connect gate matches #292 exactly**, including the part that is easy to lose: cooked assets are LOCAL on both sides and nothing about a cook crosses a connection.
- **D4** (byte order) is right and its reason is right.
- **D5** holds; I could not construct the stale-serve case (see R-1 for the two wording fixes).
- **D10, D11** verified against the implementation: declaration order reversed → held; a table renamed, added or removed → moved.
- **D2's instrument split is well argued** — SHA-256 over text where a walk could forget a fact, a constant-expression fold where the inputs are already digests.
- **§20.5 is honest about its residual** rather than claiming a lock the design does not have. That matters against the owner's *"can only be loaded by a binary with that same build version"*: what §20 delivers is addressing plus a layout guard, and the meaning class remains reachable by a hand-copied file. The page says so.
- **Numbering and register**: §20 appended, §13.7 appended, frozen sections untouched, present tense throughout, no decision narration in the spec text. USAGE's "Specified, not yet implemented" note follows the block form's precedent at line 1025.

---

## VERDICT: **LAND-AFTER-FIXES**

The design is sound and the finding that produced it is the right finding. Four defects should be fixed before this is the page a ruling rests on — D-1 and D-2 change what §20 digests, D-3 and D-4 change whether two people can implement it and get the same number. None of them touches D3, D5 or D6, so the three decisions are ready for a ruling as they stand, with R-1's clause struck, R-2's price stated and R-3's dispatcher question answered.

*Numbers reproduced with throwaway Go over `tables/examples/tables.baseline`, deleted. Branch not touched.*

gafferongames:
## Delta — all four defects fixed, the rulings applied, `fab27f7`

Thank you for reproducing the numbers; three of the four defects only exist because you did.

### D-1 — ranges join group 3

Confirmed exactly as written: §4 clamps a load to the **reader's** declared bounds, and §4 also says *"the bound is not part of identity"* — so a tightened `min`/`max`, a narrowed compressed-float range or resolution, and a `bits(N)` narrowed within one storage width each change what a load stores while the offset, the size, the kind and the wire id all hold. Group 3 gains `range` and `step` lines.

And the closure argument is no longer a claim. §20.1 now carries a **table assigning every declaration-side fact** — package, type facts, table name, field name, `was`, kind, width, order, array shape and bound, capacity, `?T`/`*T`, arm payload, `cpp_native`, default, range, enum, union, flags, guard, `json`, `const`, type tag, `| block`, comments/layout/order, baseline facts — to **wire / layout / meaning / none-with-reason**. A meaning fact is defined as one that changes a stored value while moving no offset, no size and no wire id, and the table is what shows there is no fifth.

### D-2 — flags leave group 3 entirely

Your composition argument is decisive, and the rule it exposes is the right one. Settling the caveat you flagged: **the page states no mask-to-width rule and §4's report has no counter for a bit outside the declared set**, so a load copies the mask verbatim and neither the order nor the declared width is a meaning fact. The field's storage **width** is a layout fact and group 2 already has it. §20.4 and §20.8 now test **the reorder holding** — the untested row was indeed the broken one.

### D-3 — line grammar and a worked example

§20.2 states the five line shapes, the single-space separator, zero-based decimal ordinals, the byte-wise sort, one `\n` per line, the empty set as the empty string, sparse `default`/`range`/`step` lines, and both form tokens' literal ASCII. Your recovered rendering was right on every point it covered; where it guessed I have now made the page say it.

The worked example is a four-field table plus an enum, with the protocol id and both layout-id halves given as literals, and it publishes **every intermediate digest** rather than only the last:

```
default ShipConfig.15a9 21.0        MeaningDigest       0x31ab36593dc6a96c
default ShipConfig.2e46 500.0       SchemaCookId        0x516f523e28b710d3
default ShipConfig.d272 Silver      CookId              0xb603cac4f732e73e
enum Grade 0 Bronze                 SchemaBuildVersion  0x2339502ee7f36cc9
enum Grade 1 Silver                 BuildVersion        0xcaa52d95241145c8
enum Grade 2 Gold
range ShipConfig.7c9d 0 100         empty unit          0x68bfee44b000750d
```

§20.8 makes that example a golden, digest by digest, so a port reproduces intermediates and not just the end.

### D-4 — meaning lines key on the wire id

Four lowercase hex digits, after the declaring record's name and a `.`, never the source name — §7's rule one document on. Re-verified: a `was` rename **holds**.

---

### Rulings

**D3 kept, dispatcher hole closed.** A cross-platform dispatcher addresses by **`(asset hash, SchemaCookId of the root table, TARGET)`** — a named identifier for the language, platform and ABI facet, the application's name and not schema's — every part of it compiler-settled and available pre-compile. The **loading binary verifies the full `CookId`**, which the cache records beside the artifact; schema defines the id, and recording and checking it is the obligation on the cache. Your R-3 reasoning is now in §15's follow-on verbatim, including that the revisit point is after the second backend's cook lands.

**D5 kept, first reason struck.** You are right that mixing the block digest into a cache key gives no runtime anything new to accept. The two sound reasons stand alone now: no cook byte depends on a block-only fact (a declared maximum is squarely inside the cook layout id), and folding it in would invalidate artifacts for nothing. Your second wording fix landed too — a row element may be a declared `type`, contributing through the marked table's own closure rather than as a table of its own.

**D6 REFINED — my decision, flagged for your ruling.** R-2 is right that a unit-wide key deletes §7's mesh-catalogue property in practice, and understating it as "costs a re-cook" was wrong. So the key got narrower rather than the sentence honest:

- **`CookId(T)`, per table** — `fnv1a64` over the token `schema-cook-id 1`, the protocol id, T's meaning digest, T's name + `0x00`, T's layout id, and the byte order. Closure-scoped, so it moves when the catalogue's own closure moves and not when someone else's table does. **This is the cache key's second half.**
- **`BuildVersion`, per unit** — the fold over every table's cook id. What a binary prints, what "same build?" compares, and it moves whenever any cook id does.

§20.6 also states what you asked for: a hit under the right key **implies** `Open` cannot refuse — every fact it checks is folded into the id — with the two exceptions named, an `fnv1a64` collision and a corrupt or truncated file, both landing on the wire fallback.

### Also fixed, from the risk list

- **R-4** — the cook token doubles as the **cook form's** version: bump `schema-cook-id N` when the region's pack order, the node directory's encoding or the header's shape changes. The invariant now holds on its byte reading.
- **R-5** — §7 states the asset hash is the hash of the **wire file** the cook was produced from, which is where it belongs and what makes D8's argument load-bearing.
- **R-6** — group 3's scope is the **transitive table closure**, enum keys of `[E]T` included. Better citation than mine or yours: §3 says *"an arm's body is a table body whether the arm names a declared `type` or a `table`"*, so a nested type's fields resolve by wire id and its enums by name hash.
- **R-7** — the union control is a **rename**, with the §3.1 overlap noted.
- **R-8** — **kind** is now stated as a layout-id fact in §7 itself, with the `uint32`→`float32` case as its reason.
- **R-9** — guards are in the exclusion list, with §4.1's round-trip distinction.
- **R-10** — §20 carries a **backend-status paragraph** on §16.1's model naming all three lags, the third being `arena.go`'s 32-bit part length against §7's 64-bit rule and its "one of the two" problem.
- **R-11** — both periods removed. §13.7 is verbatim.
- **R-12** — the fixture now makes the schema and full layout digests **differ by construction**, so the perturbation test is real, and §20.8 says so as an obligation.

### Re-verification (throwaway Go, deleted)

Corpus `tabledemo`, revised rules — per-table cook ids, wire-id keys, flags dropped, ranges added, schema and ABI layout halves distinct:

```
SchemaBuildVersion 0x9dd927a252ba7f2b   BuildVersion 0x512ae19912dcddc6

a specified default changed                    MOVED
a declared range tightened (new)               MOVED
an enum variant renamed / two swapped          MOVED
a union arm renamed                            MOVED
a table renamed                                MOVED
a FLAGS variant reordered                      held
a FLAGS variant renamed in place               held
a `was` rename                                 held
declaration order reversed                     held

D3 perturbation: SchemaBuildVersion HELD, BuildVersion MOVED
```

`make check` and `go test ./compiler/... ./ir/...` green. Still draft — D6's refinement is a decision, not a fix.

gafferongames:
## Revised to the ruling — two ids, one constant, `e0b012f`

"yes on protocol id and build version." §20 is restated from the tuple sentence down.

### One constant, one instrument

`BuildVersion` is **the low 64 bits of SHA-256 over one canonical COOK PROJECTION** — exactly how the protocol id is taken over the wire-shape projection, and for the same reason. The `fnv1a64` fold went out with the second constant: it existed only so a target's compiler could fold an ABI term at *its* compile time, and nothing folds any more. One id, one instrument, one text a person can diff.

The projection carries the wire (the protocol id), the **layout** of every record in the table closure by wire id and kind, the **meaning** facts (defaults, ranges, enum and union vocabularies), and the byte order. The grammar is stated in full — record/field/block/slot/enum/union lines, the token order, the sort, the indentation — and the worked example publishes its **whole text** beside its digest:

```
schema-build-version 1
protocol 0123456789abcdef
byteorder little
record ShipConfig sizeof=12 alignof=4
    field 15a9 kind=10 offset=0 size=4 default=21.0
    field 2e46 kind=10 offset=4 size=4 default=500.0
    field 7c9d kind=6 offset=8 size=1 min=0 max=100
    field d272 kind=7 offset=10 size=2 default=Silver
enum Grade
    variant 0 Bronze
    variant 1 Silver
    variant 2 Gold
                                    ->  BuildVersion 0x122277610e69c72e
                                        (empty unit  0x49947af3382f914e)
```

Every number in it is derivable from the page: the wire ids from §5's fold, the offsets from §19.3's C ABI rule.

### Compiler-settled, and what that costs

The compiler computes the layout and both backends assert it — §19.3's model, now unit-wide. **ABI drift is a BUILD ERROR naming the type and the field**, not a silent `Open` returning NULL and a wire fallback nobody sees. That is what lets Y exist before a binary does, which is what tooling needs. The price is stated rather than hidden: the compiler's layout model is now committed to every cookable closure, so a platform it gets wrong fails to build instead of degrading.

### What retired

- **The per-table layout id, gone from the design.** §7's header carries the build version; `Open` is magic, byte order, build version, the two lengths, alignment — match and point. §19's prologue carries the same id (`build_version`, not `layout_id`). §11 drops `LayoutId` and `BlockLayoutId`; the suffix claim is 34.
- **What still separates a cook from a block is the MAGIC**, read bytewise before anything else — which is where a form's identity belongs, and better than a second digest was.
- **The per-root cook id, gone.** The tuple is `(asset hash, build version)` and nothing finer; a new Y is a new cook and the store absorbs it.
- **§18 untouched**, in both directions.

### One behaviour changes, and it is a decision I want your eye on

**A raised declared MAXIMUM now moves the build version.** The brief said to keep it excluded, and I could not: a maximum is no line in the projection, but the inline storage it sizes is — the array field's `size`, every later `offset`, the record's `sizeof` — and those lines are what a **cook** of that table is opened against. A build reading another's cook under a larger declared maximum reads past the region. §19.3's own rule ("a fact both sides must AGREE on is in the digest; a fact a consumer READS is not") put the maximum outside a *block-only* digest; merging the digests moves it to the other side of that rule.

The consequence is stated in §19.4: its third edit moves from `BlockOpen` to `BlockOpenCompatible`, so the default path absorbs none of the three and the named path absorbs all three — which is the shape §14's "no silent bypass" asks for anyway. The block's own `slot` lines still do not move, because a triple is sixteen bytes whatever the maximum is.

### Also folded, where they belong

- **§7's pipeline, in your words**: *"the optimized path is still available, it is tooling does the build, then cooks to the rad binary format. and the game just points at it and works."* / *"(plus endian fixups)"* / *"this is the way."* The endian fix-up is part of the cook and stays — §15 defers only cooking for a **foreign** target, and names #303's emulated big-endian leg as what holds it.
- **The opening ladder states itself whole**: *"notice how we now have fast paths for three things? types, cooked tables, and fast write/read tables like render data"* — each with its own reason (types: same build, the bytes are the struct; cooked: the work done offline, the game points; the block: both sides generated against one layout, the writer goes wide) — beside the tolerant wire, slower because it alone crosses versions, and the reflection surface, generic and allocating *"for tooling, editor, whatever where perf and allocations aren't such a big deal."*
- **The allocation rule, in full**: types and fixed tables never allocate on any path; a variable table's builder allocates through a **caller-provided allocator with malloc semantics** (C/C++; other languages per platform default), at build time and in bulk, never per record on the fill path; and **every read path allocates nothing**. Building is tooling's path — the game points at the cook.
- **Gate 2's bar** is same speed, or not significantly slower.
- **§13.7** carries the new verbatims, including the id count.

### Verification

Throwaway Go, deleted. Every discrimination behaves under the single fold:

```
a specified default changed / a declared range tightened     MOVED
an enum variant renamed / two swapped                        MOVED
a field's KIND changed with its width unmoved                MOVED
a field's OFFSET moved with sizeof unmoved                   MOVED
the target byte order changed                                MOVED
a record renamed                                             MOVED
a `was` rename (wire id unmoved)                             held
```

`make check`, `go build ./...` and `go test ./compiler/... ./ir/...` green.

gafferongames:
## Both subtractions applied — `17446b3`

### `BlockOpenCompatible` is gone

**A block is same-build, and §19.4 is restated from that.** Both sides are generated from one declaration by one compiler run, so a consumer older than its producer is half a shipped build, not a case to absorb — and a refusal is the honest answer to it. One entry point: `BlockOpen` checks the magic, the byte order it establishes, the build version, the extent, the alignment, and each array's `offset_of`. A match points; anything else refuses and both sides regenerate.

The three named edits collapse into one rule rather than three cases — a field appended to the table, a field appended to a row, a maximum raised: each moves the build version, each refuses. **Which settles the maximum question cleanly.** It moves the id, it has to (a build reading another's cook under a larger maximum reads past the region), and with no tolerant path left there is nothing for excluding it to buy. §20.4 now says that instead of flagging a behaviour change.

§14 records the dropped alternative with the reason it could not have worked anyway: a single number cannot be verified against a PREFIX of the facts that produced it, so an append-tolerant digest was never available. §15's stride-headroom follow-on loses the absorption it leaned on and gains the refusal.

### `OpenValidated` is gone

**The runtime has one `Open`** — magic, byte order, build version, the two lengths, alignment; match and point. Validating a cook whose provenance a person doubts is **`schema cook-check`**, a tool over the same reflection descriptors (§8) the runtime already carries, and §7's two-pass attribution scan moves there whole, unchanged: the directory linearly and with no state, then every node in directory order, following no reference and decoding no value, `O(R + P log N)`.

**The attribution part is written beside the data for that reader** — stated as present-state, as asked — so a build that ships no tooling still records its length as zero and the file is just data. §11's refusal row splits accordingly: the runtime's four refusals stay the runtime's, and the attribution's are named as the tool's.

### Sweep

`BlockOpenCompatible` and `OpenValidated` appear nowhere in the tree. §11's generated-name claim is 32 suffixes. USAGE's block section says a schema edit is a regenerate, and its cooked-form section loses a stale layout-id-era paragraph — *"`Open` validates before it points — a bounds walk over the reference graph"* — which was describing a check the runtime has not had since #251; it now states the header match plus the tool.

`make check`, `go build ./...` and `go test ./compiler/... ./ir/...` green. Head is `17446b3`.

gafferongames:
## LAND-AFTER-FIXES

Cold read of `17446b3` from the page alone. The invariant, the two-id story, §20.5 and §13.7 are right. Two defects break the section's own promise; three are honesty items.

### D-1 — the `field` line names no REFERENT, so a cook diverges under an unmoved id (line 4253)

`field <id> kind=<n> offset=<n> size=<n>[ elem=][ stride=][ default=][ min=][ max=][ step=]`

§20.1's closure table promises carriage no token provides: *"a union arm's payload type | layout + wire"*, *"`?T` presence companion | layout"*, *"`cpp_native` | layout"* — and nothing carries a nested record's NAME, an enum's name, a keyed array's KEY, or the array class. `arm <ordinal> <name>` has no payload either. Three edits out of `tables/examples`, each moving **zero lines**:

- `Buff{multiplier f32}` and `Debuff{amount i32}` are both `sizeof=4 alignof=4` and both stay in the closure as `union Effect`'s arms. Retype any `Buff` field to `Debuff`: same kind=13, size, offset, wire id. The nested body then decodes by `0x39cc`, not `0x32e0`.
- `PackConfig.thresholds [Difficulty]int32` → `[Team]int32`: kind=16, elem=4, four slots, same offsets, both enums stay in the closure. Slots ride by the KEY's variant-name hashes (§3.2).
- `LoadoutConfig.grade Grade` → `uint16`: an enum rides as kind `7` (§3) and Grade stays in the closure via `grades`. The slot then stores the raw name hash, not the ordinal.

§18.1's baseline records exactly these — `type= enum= union= flags= key= array= bound= optional=` — and says why: *"A field that names a declaration records WHICH KIND of declaration it names."* **Fix:** mirror that token set onto the `field` line; add `payload=` to `arm`. (`cpp_native` is the opposite error: SPEC.md guarantees layout identity by derivation, so it belongs in `none`.)

### D-2 — the worked example, which §20.8 pins as a golden, is not derivable from the page (4288–4323)

`enum Grade { Bronze, Silver, Gold }        // storage uint16` → `field d272 kind=7 offset=10 size=2`

Storage is not declarable. SPEC.md §4.2: *"max derives as the count… **Enum storage derives from the enum's own max** — the smallest unsigned integer that fits."* max = 3 → **uint8**. The tree agrees (`generated/cpp/Enums.h:14`: a three-variant enum → `enum class Team : uint8_t`). The line derives as `field d272 kind=7 offset=9 size=1`, and the digest becomes `0x9362ec310fbc3078`. **Fix:** widen the enum past a byte, or correct the line and republish the number.

Second guess in the same block: `variant 0 Bronze`, but Bronze's stored ordinal is 1 (`None = 0`, omitted). §20.2's *"ordinals are ZERO-BASED"* makes the printed ordinal a positional index, not the value §20.1 says group 3 captures. Say which, and say `None` is omitted.

### D-3 — "both backends already assert against" is false twice (line 4130)

Line 146 of this page says the block form is *"specified and unimplemented."* The tree emits **no** `sizeof`/`offsetof` `static_assert` anywhere: `internal/codegen/` has one layout assert (`cpptable.go:452`, `is_standard_layout`), and `offsetof` appears only to POPULATE descriptors from the target compiler's answers — the opposite direction to §20.3. C# has no blittable struct at all: `cstable.go:546-548` states *"A C# field has no offsetof and a C# class has no meaningful sizeof."* §20's three emitter obligations therefore miss the largest: rewriting the C# fixed class as blittable structs with generated padding. **Fix:** strike "already"; add both to the status. On 32-bit System V `alignof(uint64_t)` is 4, so §20.3's model disagrees with the ABI — and with no asserts today, two builds share a version and cook different bytes. *"It never reaches a cook"* is not true of this tree.

### D-4 — the RESERVED WORDS drop out of `Open` (§20.6, §11:2153, §7:1751)

§7's header bullet: *"a non-zero one means a writer used a form this build does not understand, and `Open` refuses rather than ignoring it."* Its next bullet calls the five-item list *"the WHOLE check, because nothing else is checked at all"*, and §20.6 and §11 enumerate without it. **Fix:** one enumeration, in all three.

### D-5 — `schema cook-check` does not exist and carries no status note

Present tense in §7 (×3), §11 and USAGE:1146 (*"check it with the tool"*). `cmd/schema/main.go` has check/id/projection/tables-baseline/fmt/generate/pack/unpack. `schema build-version` is flagged; this is not.

### Reproduced (throwaway Go, deleted)

- Worked-example text verbatim off the page → SHA-256 low 64 bits, big-endian = **`0x122277610e69c72e`** ✓
- Empty unit, three header lines, same protocol id → **`0x49947af3382f914e`** ✓
- Wire ids from §5 alone: `damage 15a9`, `velocity 2e46`, `armor 7c9d`, `grade d272` ✓
- Derived from the page's *rules* rather than its text → `0x9362ec310fbc3078` ✗ (D-2)

Guesses forced: the empty unit's protocol id (a zero id gives `0xc49dabcb839e0c8a`); whether a plain `uintN` has an effective range (I read: no); whether `bits(N)` renders `min=0 max=2^N−1` (yes, untested); and what a **variable-length** table's `record sizeof/alignof` is — never stated, and the corpus is pointered.

### Residues

- **2180, "The nine block spellings"** — seven remain; nine counted `BlockOpenCompatible` and `BlockLayoutId`. The 32-suffix list itself is right.
- **`internal/check/check.go:2665`** still reserves the `LayoutId` suffix §11 dropped, and none of the seven block spellings — a new divergence this PR creates, unmentioned in §20's status.
- **§13.7's first verbatim drops a period**: #292 has *"…the asset the runtime searches for."* Everything else there is word-for-word.
- **"TWO ids in this design and no others"** (4118, §7:1740) — §6.3's node directory carries a per-node `type id (u64)`. Scope it to unit-wide version ids.
- `layout id`, `LayoutId`, `BlockLayoutId`, `BlockOpenCompatible`, `OpenValidated`, cook id: **zero** across all four pages. Clean.

*Branch not touched.*

gafferongames:
## Delta-read fixes, rebased onto main — `dc684e3`, MERGEABLE

Thank you for reproducing from the *rules* and not only the text — that is what caught D-2, and D-1 is the better catch.

### D-1 — the `field` line now names its REFERENT

You are right that §20.1 promised carriage no token provided, and all three of your corpus edits moved zero lines. The line now carries §18.1's own token set, for §18.1's stated reason:

```
field <id> kind=<n> offset=<n> size=<n>
    [ type=<Name> | enum=<Name> | union=<Name> ]
    [ elem=<n> ] [ array=fixed|bounded|keyed ] [ bound=<n> ] [ key=<Name> ]
    [ stride=<n> ] [ optional=true ]
    [ default=<v> ] [ min=<v> ] [ max=<v> ] [ step=<v> ]
```

and `arm <stored tag> <name> payload=<Name>`. `Buff`→`Debuff` moves `type=`; `[Difficulty]`→`[Team]` moves `key=`; `Grade`→`uint16` moves `enum=`.

**§20.1's table is re-cast as a CARRIAGE table** — every fact now names the token that carries it, or sits in `none` with its reason, and the rule is stated as the rule: *a fact with no token is a promise the digest does not keep*. Two rows moved to `none` under it, both yours: `cpp_native` (SPEC.md §6.1 guarantees layout identity by derivation) and — the one I had to decide — **a flags field's REFERENT**. There is deliberately no `flags=` token: a slot holds a raw mask, a load copies it verbatim, so swapping one flags declaration for a same-width other changes no cook byte. It is stated as deliberate and §20.8 carries the negative control.

Three referent controls added to §20.8, each a same-shape swap every other fact survives.

### D-2 — the worked example is derived, not asserted

`Grade`'s storage is **`uint8`**: `None = 0` implicit, three variants from 1, `max` derives as the count, *"enum storage derives from the enum's own max"* (SPEC.md §4.2). `generated/cpp/Enums.h` agrees. The annotation is gone and the field is `offset=9 size=1`.

And you are right about the ordinals: **the number is the STORED VALUE**, since that is what group 3 captures, so they run `1 2 3` and `None` is never listed — same for union arms. The example now derives every number step by step (ids from §5, kinds from §3's closed set, storage from §4.2, layout from the C ABI rule, variants as stored values), and the digest is **`0x7402a36de22d9728`**. The empty unit is unchanged at `0x49947af3382f914e`, with its protocol id now named explicitly.

Your three forced guesses are answered on the page: a plain integer with no declared bound carries **no** `min=`/`max=`; `bits(N)` carries `min=0 max=<2^N−1>`; and a variable-length table is a record like any other — its node is a struct with a `sizeof` and `alignof`, and a pointer field is a kind `17` slot whose `type=` names the pointee.

### D-3 — "already assert" struck, and the status reordered

You are right twice over. §20.3 now says the asserts are **owed**, that the C++ backend emits one `is_standard_layout` assert and no `sizeof`/`offsetof` assertion (and that the `offsetof` it does emit populates descriptors from the target compiler's answers — the opposite direction), that the C# fixed class is not blittable at all, and that **until both land the guarantee is specified and not held: two builds whose ABIs differ would share a build version and cook different bytes**. Your 32-bit System V `alignof(uint64_t)` point is on the page as the reason the model is asserted rather than assumed.

The status list is five obligations, largest first, with **C# blittable storage** at the top.

### D-4 and D-5

§7's `Open` enumeration is the one place, now including **every reserved word zero**, and §11 and §20.6 cite it instead of repeating a partial list. `schema cook-check` carries a not-built note in all five places (three in §7, §11, USAGE), and §20's status names it beside `schema build-version`.

### Residues

Seven block spellings, and here is the list I counted: `Block  BlockStorage  BlockBegin  BlockBytes  BlockMaxBytes  BlockOpen  Counts`. §13.7's first verbatim regains its period. **"Two ids"** is scoped to unit-wide VERSION ids, with §6.3's per-node type id named as the identity it is. `internal/check/check.go:2665` left alone — another worker owns it this hour.

### Rebase

Onto `917fddb` (#304 no-marker, #299 the view, #303 the big-endian leg, #308). Resolutions: #304's marker removal kept throughout and my §20/§19 marker wording swept to match; #304's residual `BlockOpenCompatible` in §19.4, §19.5 and the on-the-side emission rule resolved in favour of this branch's removal; #299's new §13.7 (the view) kept and mine renumbered to **§13.8**, with §13.6's pointer updated. §15's cross-endian follow-on already names #303's leg.

`make check`, `go build ./...`, `go test ./compiler/... ./ir/...` green. `mergeable: MERGEABLE`. Head `dc684e3`. Still draft — not merging.

gafferongames:
## LAND — after three one-line strikes, none of which touches the digest

Delta read of `dc684e3`. Only what differs.

**(1) The number now derives.** I wrote the projection from the rules before reading the page's block — ids from §5, kinds from §3, `uint8` from SPEC.md §4.2, offsets 0/4/8/9 and `sizeof=12 alignof=4` from the C ABI rule, `enum=Grade` in the stated token order, variants as stored values `1 2 3`, `None` unlisted. It matches the page **byte for byte**, and hashes to **`0x7402a36de22d9728`**. Empty unit **`0x49947af3382f914e`** ✓. Every guess from my last read is now answered on the page.

**(2) Carriage holds.** `Buff`→`Debuff` moves `type=`; `[Difficulty]`→`[Team]` moves `key=`; `Grade`→`uint16` moves `enum=`; a `?T` companion moves `optional=true` and the offsets. **The flags `none` row is sound, and I tried to break it**: SPEC.md §4.2 makes flags storage `uint64` in *every* target, so "same-width" is automatic, and SPEC.md's defaults list — bool, integer, float, enum, 128-bit, `fixed` — excludes flags, so no flags field can carry a specified default whose evaluated mask a reorder could move. The slot is a raw u64 copied verbatim. No cook byte moves. `cpp_native` in `none` is sound too, on §6.1's derivation plus the standing standard-layout assert.

**(3) §20.3 is honest and accurate against the tree** — zero `static_assert( sizeof`/`offsetof` in `internal/codegen/`, zero `StructLayout` in `cstable/`. The five-item status is right, largest first. `schema cook-check` carries a not-built note in all five places ✓, and §11 now cites §7's six checks including the reserved word ✓.

### The three strikes

1. **`SPEC-TABLES.md:4726`** — §20's OPENING still reads *"the model §19.3 already states and both backends already assert against"*. That is the exact D-3 sentence; it was fixed in §20.3 and survived here, three lines above the status list that contradicts it. Strike "already assert against" → "must assert against".
2. **`USECASES.md` is unswept** (it arrived with the rebase, so it was clean at my last read): `:16` *"one exact layout id"*; `:97` *"The layout id is a 64-bit digest…"*; `:103`+`:110` `OpenValidated` as *"the other entry point"* and as **proof** that *"round-trip[s] over the pointer unit under `make`"* — it is in no `.go`/`.h`/`.cs`; `:133` `BlockOpenCompatible`.
3. **`:4584` (§19.3) and §20.4** both still say a declared MAXIMUM *"is not a line in the projection"*. `bound=<n>` now carries it. Both conclusions survive; both premises are stale.

Residues otherwise clean: "nine block spellings" gone, §13.8's verbatim has its period, no `gains \`| block\``.

*Branch not touched.*

gafferongames:
## Three strikes applied, ready for review — `8f44e4c`, CI green

**(1) The surviving "already assert".** You are right that it was the exact D-3 sentence, three lines above the status list that contradicts it. §20's opening now reads *"the model §19.3 states and both backends MUST assert against (§20.3)"*, and the sentence after it says the asserts are owed and not yet emitted rather than claiming a build failure the tree cannot produce.

**(2) USECASES.md swept.** It arrived with the rebase, so it had my last revision's whole vocabulary still in it:

- the forms table and entry 4's contract now say **one exact build version**, and entry 4 describes what it digests;
- `Open` is stated as the runtime's only entry point with its six checks, reserved word included;
- `OpenValidated` is gone, replaced by `schema cook-check` named as a tool **and as not built**;
- **the proof claim is cut back to what `make` actually runs.** You were right that `OpenValidated` is in no source file; I checked what remains and `CookMeasure`, `Cook` and `Open` are genuinely exercised over the pointer unit (`test/tables/main.cpp`, `test/tables/cook_endian_main.cpp`), so the claim now names those three and nothing else;
- entry 5 loses `BlockOpenCompatible` for the one same-build `BlockOpen`, and its proof line now names the missing layout asserts as the largest gap instead of leaving the contract paragraph to imply they exist.

**(3) The maximum's premise.** Both places opened on "a declared maximum is not a line in the projection" — stale the moment `bound=` landed. §19.3 now reads *"A declared MAXIMUM moves NO `slot` line"* and §20.4 *"It is a `bound=` on the field line"*. Both conclusions are unchanged and now rest on the right fact: a triple is sixteen bytes whatever the maximum is, and the by-value `record` lines move, which is what a cook is opened against.

Thank you for trying to break the flags `none` row rather than taking my word for it — the `uint64`-in-every-target point and the defaults-list point are both stronger arguments for it than the one on the page, and neither had occurred to me.

`make check`, `go build ./...`, `go test ./compiler/... ./ir/...` green locally. CI: all nine checks **SUCCESS** (inline-gate skipped). `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`, out of draft. Head **`8f44e4c`**. Not merging — yours to land.

PR #299, opened 2026-09-02
SPEC: the view — one file per unit carrying every declaration, on the side, for editors and debug builds (#157, spec-first)

gafferongames:
Spec-only, per #157's spec-first step: SPEC-TABLES.md §8, SPEC.md §4.2 and §6.1, USAGE.md. No compiler code, no corpus change — `| view` and a `--views all` request are stated as refused-by-name today, the way `| block` is.

## Placement decision: §8 rewritten, not a new section

§8 becomes **"Reflection: the view"** and today's descriptor text becomes **§8.1 The descriptors**, unchanged in substance. Every existing `§8` citation — §16's walk, §19.2's reflective read, §11, and a dozen comments in `internal/codegen/*` — still lands on the section that carries the columns, so nothing has to be re-cited. A separate §20 would have put the view a document away from the vocabulary it *is*: the owner's "(but with a view on that type?)" asks for the same surface, not a second one.

## What the page now says

- **§8.2 The type view.** Three askers: `| view` on the declaration, a table reaching the type (as today), or a build viewing everything. The view IS §8.1's `TableTypeInfo` — one vocabulary, one walker — so a printer written against tables walks a viewed type without being told. A field with **no table-wire kind** (`fixed`, `ufixed`, the 128-bit family) is **described at kind 0**, not refused: `examples128` has 19 such fields, and `--views all` over that unit must produce a listing rather than a diagnostic.
- **§8.3 The unit registry.** `UnitView()` returning constant data: `types` (those with a view), `tables` (always), `enums`, `flags`, `unions`, `constants`. Enum rows are the named values in order beside `max`; a flags row's value is a bit index with id 0; a union lists tag 0 then its arms with payload name always and payload descriptor where the payload is viewed; a constant carries name, declared storage and folded value. Every entry names its schema file; order is files-by-basename then declaration order, which is what makes §8.7 a byte comparison.
- **§8.4 Who asks.** `--views declared` (default) / `all` / `none`, carried as a generate option — the `compiler.Options` contract already says a target that cannot honor a key errors rather than falling back, which is exactly the refusal this needs. §2.2's zero-cost gate is restated: a unit that asks for nothing emits no view file, not an empty one.
- **§8.5 On the side.** One `<Package>View.h`/`<Package>View.cpp` per unit (`<Package>View.cs` for C#), at the **leaf** of the include graph, pulling data headers and no wire header. Data in the `.cpp` per §13.5. Descriptors a table needs stay in the table header — a table's descriptor points at its nested types', so a game that loads tables must keep linking them.
- **§8.6 What it does not carry**, **§8.7 Held by test** (the editor gate, the corpus pin, the red-gate check, the zero-cost gate).
- **§11** gains the view's refusals, and claims `TableType` plus `UnitView`/`UnitViewInfo`/`ViewType`/`ViewVocabulary`/`ViewVariant`/`ViewConstant` unconditionally — a generate flag must never change what the language accepts. **§14** records the four models weighed; **§15** the follow-ons.
- **SPEC.md** §4.2: the `| view` attribute, type declarations only, excluded from the wire-shape projection (no `ProtocolId` moves, exactly as a type tag does not). §6.1: the `--views` request and the per-unit view file, C++ and C#.
- **USAGE.md**: "Inspecting a whole build: the view" — an editor-shaped walk over `UnitView()`, printing constants, vocabularies and every declaration's properties.

## Decisions taken (each open to the owner)

1. **§8 rewritten in place** rather than a new section — citation compatibility and one vocabulary. 
2. **Entry point `UnitView()`**, in the unit's own namespace beside `ProtocolId`; no package baked into the identifier.
3. **File named `<Package>View.*`, per unit.** A fixed name would collide: the bench corpus generates **two units into one directory** (`generated/bench/cpp`), which the Makefile calls out.
4. **A viewed type's accessor is `<Name>TableType()`** — already in §11's claimed 36 suffixes, so the view adds no new per-declaration spelling. The `Table*` names are kept even in a table-free unit: renaming them forks the walker.
5. **`| view` marks types only.** Enums, flags, unions and constants are catalogued unconditionally wherever a registry exists — they cost names and values, a type costs a descriptor table. §11 refuses the marker on the free kinds rather than accepting one that selects nothing.
6. **`| doc` is NOT introduced.** Doc comments are already deferred with a pinned design (SPEC §4.1, §9 q5); when they land the view carries them as a column. Listed in §15 rather than invented here.
7. **Type TAGS are carried** in the registry (data, no meaning attached) — `Vec3 | vec3` exists in the corpus and an editor wants it.
8. **The editor gate lives in §8.7, not as a third §12 gate.** §12's two gates are expressiveness gates; this one is reflection completeness. §12 is untouched.
9. **Three `--views` modes.** `none` exists so a game build pays nothing for markers a tools build wanted — the "same declarations serve both" property.
10. **The corpus pin is the compiler's IR listing**, with each backend's generated program byte-comparing against it, so C++ and C# are one view the way they are one wire. The spec explicitly does not ask for a new CLI verb.

## Open for the owner

- The `Table*` spellings in a table-free unit (`TableFieldInfo` describing a packet-only type) — kept for one-vocabulary reasons; aliases are cheap if he dislikes the read.
- Whether the default should be `declared` or `none` — I chose `declared` so a marker works without a flag; `none` as default would make the marker inert until asked for.
- Whether registry entries should carry the source file basename at all (three extra pointers) — kept because "everything that was in the schema built" includes where it was declared.
- Build identity in the registry (package + protocol id only today) — deliberately silent, since the build-version work is in flight elsewhere.

Not touched: SPEC §3.3/§7 (#292 in flight), §12, the block-form sections beyond citations, and anything in `internal/`.

Refs #157

### DIFF
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -1988,6 +1988,29 @@ were never checked for collisions (§5's refusal is scoped to the closure, as
 one id and a text-form key for a text form that does not exist. A type
 inside a closure carries both as it always did.
 
+**The same rule governs EVERY id column the view carries, for the same
+reason.** §5's refusals are scoped to the closure throughout: the variant
+refusal reaches a vocabulary through a closure member's field, and through
+the KEY of a closure member's keyed array, so a declaration no table closure
+reaches has ids nothing ever checked. So **an enum, a flags or a union no
+table closure reaches carries `id = 0` on every variant row** — the
+registry's `ViewVariant.id` (§8.3) and a descriptor's `variant_id()`
+function (§8.1) alike — and **an enum-keyed array whose KEY enum no table
+closure reaches carries `key_id( slot )` of `0` in every slot**. `0` is the
+reserved id no declared name folds to (§5), and it already spells "no
+table-wire identity here" in the two places the columns state it: a flags
+bit's id, and slot 0 of a keyed array. A vocabulary a closure reaches
+carries its checked ids exactly as it did before.
+
+**The compiler is unchanged, and is right to be.** Inside a closure it
+refuses a collision by name; outside one it accepts the unit, because the
+packet wire identifies a variant by its ORDINAL and carries no name hash
+(§5) — an enum nothing in a table reaches keeps every spelling it ever had,
+and widening the refusal unit-wide would reject a packet-only schema over an
+identity its wire does not have. The consequence is the rule above, stated
+so a tool never has to know it: **outside a table closure an id is not
+unique**, and the view hands out none.
+
 **The vocabulary keeps its `Table` spellings** — `TableFieldInfo`,
 `TableTypeInfo`, `<Name>TableType()` — in a unit that declares no table at
 all — and every unit now has a view, so every unit spells them. They are one
@@ -2018,10 +2041,13 @@ struct ViewVariant          // one enum variant, one flags bit, one union arm
 {
     uint64_t value;         // an enum's value, a flags BIT INDEX, a union's tag
     const char * name;
-    uint16_t id;            // the table-wire id (§5); 0 for a flags bit
+    uint16_t id;            // the table-wire id (§5); 0 for a flags bit, and 0
+                            // throughout a vocabulary no table closure reaches
+                            // (§8.2)
     const char * payload_name;      // a union arm's payload TYPE name, else NULL
-    const TableTypeInfo * payload;  // that payload's descriptor where it has a
-                                    // view (§8.2), else NULL
+    const TableTypeInfo * payload;  // that payload's descriptor — never NULL for
+                                    // a declared payload (§8.2); NULL on tag 0
+                                    // and on an enum or flags row
 };
 
 struct ViewVocabulary       // an enum, a flags or a union declaration
@@ -2034,7 +2060,7 @@ struct ViewVocabulary       // an enum, a flags or a union declaration
     const ViewVariant * variants;
 };
 
-struct ViewType             // a type with a view, or a table
+struct ViewType             // one declaration: a type, or a table
 {
     const char * name;
     const char * file;
@@ -2070,9 +2096,11 @@ const UnitViewInfo * UnitView();
   registry.
 - **An ENUM lists its NAMED values in order, beside its `max`.** Row 0 is
   `None`, the reserved id (§5), then the variants in declared order with the
-  wire id each rides under. A value inside `[0, max]` that no row names is
-  `| max = K` headroom (SPEC §4.2), and a tool that shows such a value
-  unnamed is showing the truth rather than inventing a name for it.
+  wire id each rides under — `0` throughout where no table closure reaches
+  the enum, because nothing checked those ids (§8.2). A value inside
+  `[0, max]` that no row names is `| max = K` headroom (SPEC §4.2), and a
+  tool that shows such a value unnamed is showing the truth rather than
+  inventing a name for it.
 - **A FLAGS row's `value` is a BIT INDEX and its `id` is 0**, because a
   mask's variants ride by position and have no per-variant wire id (§4) —
   the same rule §8.1's columns state, in data rather than in functions.
@@ -2220,23 +2248,38 @@ applied to reflection). The scope is per backend because acceptance already
 is: C# refuses a pointered unit by name (§11), and a gate cannot ask a
 backend for a listing of a unit it declines to emit at all. Completeness is
 the count the pin carries: every declaration, every field of every type,
-every variant of every enum, flags and union, and every constant, in the
-schema's own order. How the compiler produces its half is the
-implementation's business — a test that walks the checked unit is enough,
-and nothing in this section asks for a new command.
+every variant of every enum, flags and union, and every constant, each set
+in DECLARATION-NAME order (§8.3) — a variant list keeps its declared order,
+which is the one order that is a property of the declaration itself. How the
+compiler produces its half is the implementation's business — a test that
+walks the checked unit is enough, and nothing in this section asks for a new
+command.
 
 **A gate that cannot go red proves nothing.** Dropping one declaration, one
 property, one variant or one constant from an emitter's registry turns the
 corpus gate red, and that is established by doing it rather than assumed.
 
 **The INDEPENDENCE gate, and it is what the whole section rests on.** The
-view adds a file and moves nothing else: for every corpus unit, every
-generated file other than `<Package>View.*` is byte-identical to what the
-same unit emitted before the view existed, and `schema id` reports the same
-protocol id. That is what proves reflection costs the type wire's generated
-code not one byte — §10's independence, claimed for a table edit, holding
-for a view. Beside it, the containment gate of §8.4: not one of the six
-registry symbols appears in any generated file but the view pair.
+view adds a file and moves nothing else: every generated file other than
+`<Package>View.*` is byte-identical to what the same unit emitted before the
+view existed, and `schema id` reports the same protocol id. That is what
+proves reflection costs the type wire's generated code not one byte — §10's
+independence, claimed for a table edit, holding for a view.
+
+**Its left-hand side is the RECORDED SOURCE GOLDENS**, because with nothing
+selecting the view there is no second generate to compare against: "before
+the view existed" is a snapshot, and the repo keeps one — the source goldens
+pin every backend's generated text byte for byte, with the protocol id
+beside them, so the gate reduces to "landing the view moves nothing in the
+recorded goldens", which is mechanical and already red-capable. Today those
+goldens cover the `examples` and `examples128` units, and **extending them
+to the two TABLE corpora is a named follow-on (§15)** — that is the half
+where movement is plausible, since §8.5 draws the boundary between the table
+header and the view file straight through `<Base>Table.h`, and until the
+snapshot exists the gate has no referent for those units.
+
+Beside it, the containment gate of §8.4: not one of the six registry symbols
+appears in any generated file but the view pair.
 
 ## 9. Relocatable, and built wide
 
@@ -2925,8 +2968,8 @@ Owner rulings, 2026-09-02, in the order given.
 - **The question that opened it, and both halves are yes**: "We should
   revisit the 'view' concept. Does a type have a view generated for it,
   alongside it, and a table has the view built in?" A table's descriptors
-  are built in (§8.1); a type's view is generated alongside it when asked
-  (§8.2). That is the section's whole shape.
+  are built in (§8.1); a type's view is generated alongside it, for every
+  type, always (§8.2). That is the section's whole shape.
 - **What has to be walkable, beyond fields**: "I would find it valuable to
   be able to walk enums, constants at runtime via some reflection." Constants
   had no runtime surface at all before §8.3, and enums had one only through
@@ -3380,6 +3423,14 @@ pre-empted here.
   per table: walk the registry, print or compare every declaration, with the
   instance-level walk of §8.1 underneath it. The listing §8.7's gate already
   produces is the dump half of it in embryo.
+- **THE SOURCE GOLDENS EXTENDED TO THE TABLE CORPORA.** §8.7's independence
+  gate compares against a recorded snapshot of generated text, and the
+  snapshot covers the `examples` and `examples128` units only — nothing
+  records what the two table corpora emit, so for exactly the units whose
+  table headers the view's boundary runs through, the gate has nothing to
+  compare. It is the same harness pointed at two more directories, and it
+  belongs with the emitters rather than with this page, because a snapshot
+  taken before the code that could move it is a snapshot of nothing.
 - **The view in a ported backend** — C++ and C# carry it; every other
   backend emits no view file until it emits the same registry against the
   same pin (§8.7). Nothing is refused meanwhile, because nothing in a schema
--- a/SPEC-TABLES.md
+++ b/SPEC-TABLES.md
@@ -104,6 +104,15 @@ variable-length table's block form is (§19). C++ and C# take it together,
 because the form is an ABI between two languages and one language alone
 cannot hold the gate it exists for (§12.1).
 
+**The VIEW's type half and unit registry (§8.2–§8.7) are specified and
+unimplemented.** What ships today is §8.1: a table's descriptors, built in,
+and the same descriptors for every type a table reaches. No backend emits a
+view file yet, so a declaration carrying `| view` and a request for
+`--views all` are refused by name with §8 cited, never accepted and ignored;
+the default over a unit that marks nothing asks for no view file and is
+unaffected. C++ and C# take it together, because the gate it exists for is
+one listing both backends reproduce.
+
 ## 1. Purpose
 
 - **Users build their own formats.** A table declaration plus nesting is a
@@ -1767,7 +1776,26 @@ builder with a serialization point in it, whatever the read side costs. A
 cook does not answer it either — a cook is produced from a builder by a
 single-threaded `Lock` (§6.2), which is exactly the shape that lost.
 
-## 8. Reflection
+## 8. Reflection: the view
+
+**One vocabulary, one walker, three things it describes.** A TABLE carries
+its descriptors BUILT IN, always — they are what the table runtime, the text
+form (§16) and the block form's read side (§19.2) are written against. A
+TYPE gets the SAME descriptors generated ALONGSIDE it when something asks:
+a `| view` marker on the declaration (§8.2), or a table reaching it. And a
+UNIT gets a REGISTRY (§8.3) — one entry point listing every declaration in
+the build, so a tool that knows the unit and nothing else enumerates the
+types, the tables, the enums, the flags, the unions and the constants, and
+walks each one's properties with no schema files on hand.
+
+The registry and the type views are emitted ON THE SIDE (§8.5): one
+generated file per unit that an editor or a debug build links and a game
+build never compiles. Which of them exists is decided by WHO ASKS (§8.4) —
+a build flag and a declaration marker, never a change to what a table
+carries. Nothing here rides a wire: `| view` moves no protocol id, no
+generated wire byte and no baseline row (§10, §18).
+
+### 8.1 The descriptors
 
 For every table in the unit's closure the generated header carries static
 descriptors: field name, type name, wire id and kind, storage offset and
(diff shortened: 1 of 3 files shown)


gafferongames:
## Cold read — #299 (the view)

Fresh clone, `gh pr checkout 299`, read against `main`'s §8/§2.2/§11/§13.5/§16/§19.2 and every comment on #157. Every claim below is either quoted from the tree or reproduced with a scratch Go program over the checked IR and the pinned `bin/schema`. I had no part in writing the page.

---

## DEFECTS

### D1. `| view` collides with the type-tag namespace, and the status paragraph is not present-state

The page's own status line says a declaration carrying `| view` is *"refused by name with §8 cited, never accepted and ignored"*. It is accepted today, silently:

```
type ShipState | view
{
    health int16 | min = 0, max = 1000
}
```
`schema check` → exit 0. Walking the IR: `tagged declarations: [ShipState | [view]]`.

`| view` parses as a **type TAG named `view`** — the type attribute section already takes bare identifiers as tags (SPEC §4.2), which is the same slot the marker is being added to. `| block` had no such problem: it sits on a `table`, where tags are not accepted.

Two things follow, and both need saying on the page:
- the status paragraph is false as written (a spec-first page may say "unimplemented", not "refused" when the compiler accepts and ignores it);
- §8.3's `ViewType` carries `tags`, so `view` would be simultaneously a marker and a listed tag. The page must reserve `view` out of the tag namespace — and that is a **language change** (a type tagged `view` becomes refused), stated as one, or a non-colliding spelling must be chosen.

### D2. The 128-bit family does not answer kind 0 under "the same rules"

§8.2: *"`fixed`/`ufixed` and the 128-bit family have no table kind ... `kind` is `0`"*, and the view *"IS §8.1's surface ... filled by the same rules."* The rule that fills that column is `ir.TableScalarKind`, and its `TInt` arm dispatches on width 8/16/32 with **`default:` catching 64 and 128 alike**. Run over `examples128`:

```
~ 128-bit field WideProbe.entity_id uint128 -> TableScalarKind answers u64 (NOT 0)
~ 128-bit field WideProbe.energy    int128  -> TableScalarKind answers i64 (NOT 0)
~ ... flux, bias, seed, LudicrousState.keys, target_id, DegenerateProbe.locked_wide
```

Eight fields, every one silently described as a 64-bit scalar sitting at a 16-byte field's offset. That is the worst failure available to the editor this section exists for.

It also means **the "19 such fields" number is not the number of fields with no table-wire kind.** My walk reproduces exactly 19 — and they are the `fixed`/`ufixed` fields alone (5 FixedProbe, 6 UnsignedProbe, 1 DegenerateProbe, 3 FixedVec, 4 FixedQuat). With the 128-bit family the count is **27**. Either the sentence's rule is wrong or its count is; both are load-bearing for what `--views all` over that unit must produce.

### D3. Kind 0 is DESCRIBE, not EDIT — and the page reads as sufficiency

You asked whether the kind-0 rule lets an editor *display and edit* such a field. **Display of the declaration: yes. Read or write of the value: no.** What the descriptor hands a generic walker for `reach fixed(112, 16) | min = -1000000, max = 1000000` is: `name`, `type_name` as the string `"fixed(112, 16)"`, `offset`, `elem_size`, `has_range`, and two `double`s. There is no numeric shape column — no signedness, no `I`, no `F`, no storage width — so nothing can decode those 16 bytes without **parsing the type_name string**, which is a schema-file dependency wearing a disguise.

And the bounds do not survive in general: `range_min`/`range_max` are `double`, with the emitter's own note *"int64 ranges beyond 2^53 lose precision here"*. `examples128`'s particular bounds happen to be exactly representable; a `uint128 | min = 0, max = 340282366920938463463374607431768211455` is not, and rounds **up past the true max** — an editor clamping to `range_max` accepts an illegal value.

§8.2 says the field is *"described at kind 0"* and *"the range columns carry the declared bounds"* and stops. Fix in one sentence: a kind-0 field is describable and **not** generically readable or writable, and either carry the numeric shape (signed / int_bits / frac_bits / width) or point at §15's packet-layout follow-on as the thing that closes it. §15 already half-admits this (*"the no-table-wire-kind fields of §8.2 are exactly the ones it would fill in"*) — §8.2 should not read as complete while §15 says it isn't.

### D4. The name claim is incomplete — by the PR's own rule

§11's new paragraph widens **`TableType`** to every declaration, *"so a name that is legal without views and refused with them would make a GENERATE FLAG change what the language accepts, which no flag may do."* Correct rule; the application misses most of its own surface.

**Per declaration, the descriptor emission spells three names, not one.** Grepping the two table emitters for per-declaration `Table*` format strings returns exactly:

```
%sTableFields   (inline const TableFieldInfo <Name>TableFields[] = ...)
%sTableInfo     (inline const TableTypeInfo  <Name>TableInfo    = ...)
%sTableType     (inline const TableTypeInfo * <Name>TableType();)
```

All three are in §11's 36, and all three are claimed today **only for closure members** (`check.go`: `if c.tableClosure[name] { c.addTableSymbols(...) }`). Widening one of the three is not enough.

**Worse, the unit-level primitives are conditional too.** `check.go`:

```go
if len(c.tables) > 0 {
    // ... claimed only when a unit declares a table, so table-free units
    // keep their whole namespace.
    for _, gen := range tablenames.Claimed() { add(gen, ...) }
}
```

`tablenames` holds `TableTypeInfo`, `TableFieldInfo`, `TableReport`, `TableWriter`, `TableReader`, `TableKeyed`, `TableRef`, … — and §8.2 requires a **table-free** unit's view file to define those primitives. So today a packet-only unit may legally declare `type TableFieldInfo`, and `--views all` makes its generated code not compile: a generate flag changing what the language accepts, the exact outcome the new paragraph forbids.

`TableKeyed` is not hypothetical here — enum-keyed arrays occur in **type** bodies: `tables/examples` has `ScoreBoard.per_team [Team]`.

Fix: widen the per-declaration triple **and** the `tablenames` unit-level set to every unit unconditionally, and say so in §11 in one paragraph rather than one name.

### D5. The `--views` refusal contradicts the contract it cites

The PR body: *"the `compiler.Options` contract already says a target that cannot honor a key errors rather than falling back, which is exactly the refusal this needs."* The contract, in full (`compiler/generate.go`):

> A generator **ignores keys it does not know**, and an absent or empty value means that target's default; a value it knows but cannot honor is an error, never a silent fallback.

Only the second half was quoted. The seven backends with no view emitter do not *know* `views` — under the contract as written they **ignore it silently**, which is precisely the accepted-and-ignored outcome §11's new bullet forbids (*"refused with §8 cited and never emitted with the registry missing"*). The refusal requires every backend to learn the key, or the Options contract to change. The page must state which, and cite the contract accurately.

### D6. `| view` is not closure-transitive, and the page's own reason says it must be

§8.2 gives the reason a table's closure is viewed: *"a walk must be able to enter a nested type, so every type in a table's closure has had a view all along."* That reason is not applied to the marker. Under `--views declared`:

```
type ShipState | view { position Vec3 }
```

`ShipState` lands in `types`; `Vec3` gets nothing; the descriptor's nested-type column is NULL. USAGE's own printer —

```cpp
if ( field.table ) printf( "  -> %s", field.table->name ); // recurse
```

— dead-ends at `position` and says nothing about why. §8.3 says *"`types` holds exactly the types with a view"*, which makes that the specified behaviour rather than an oversight.

Either the marker views its closure (and §8.3's sentence takes the closure with it), or the page states plainly that a `declared` view is shallow and a walker meets NULLs inside marked types. Silence here is the guess.

---

## RISKS + recommendation

- **R1 — §8.7's corpus gate over-claims.** *"For every unit in the corpus ... every backend's program byte-compares against it, so C++ and C# are one view the way they are one wire."* The C# backend **refuses `tables/pointers` by name** (§11's variable-length refusal), and `Makefile:308` asserts that refusal as a gate. Scope the sentence to the units each backend carries.
- **R2 — the zero-cost gate's two halves disagree.** §8.4: *"not one symbol of the registry appears in any other generated file. The build fails if **a spelling of the view** surfaces in a unit that asked for none."* By §8.2 the view's vocabulary **is** the `Table*` spellings, so the second half read literally reds every table unit built with `--views none`. §8.7 says it better ("not one registry symbol"). Name the seven: `UnitView`, `UnitViewInfo`, `ViewType`, `ViewVocabulary`, `ViewVariant`, `ViewConstant` — and nothing else.
- **R3 — independence is asserted, never gated.** §8.4's gate is one-directional: it proves `none` emits nothing, never that `all` leaves everything else alone. The check, and it is one line in §8.7: generate each corpus unit three times (`--views none|declared|all`) into three trees; **every file other than `<Package>View.*` byte-identical across all three**, `schema id` unchanged, and no `<Package>View.*` in the `none` tree. That is the only thing that proves the type wire's generated code does not move by a byte, which is the property the whole section rests on.
- **R4 — `<Package>View.h`'s spelling has no home, and can collide with a schema file.** C++ spells the package **verbatim** (`namespace tabledemo`); only C# capitalizes it (`capitalize(unit.Package)`). "the package name spelled as that target spells it in a file name" names a rule C++ does not have — say `capitalize(package)`. Separately: a schema file named `TabledemoView.schema` in `package tabledemo` generates `TabledemoView.h` and collides with the view header. Same latent class as `FooTable.schema` vs `<Base>Table.h`, unrefused today — but the view is where a new fixed-shape file name enters, and one §11 line closes it.
- **R5 — the `id` column is unsafe on a viewed packet-only type.** Reproduced: `type Probe | view { aapy int32  aaym int32 }` checks clean; the same two fields inside a table closure are refused — *"fields aapy and aaym collide on table-wire id 0xe032 — rename one"*. So a viewed type's descriptors can carry two fields under one id, and a tool keyed by id conflates them. Say the `id` column is 0/meaningless where the field has no table wire, or extend the refusal. Quieter member of the same class: `json` is non-NULL on a viewed type that §8.2 says has no text form — a walker reasonably reads that as "there is a text form here".
- **R6 — registry ORDER, and my ruling on open item 3.** Files-by-basename then declaration order means a **file rename or a declaration moved between files reorders the entire listing**, while SPEC §3.2 holds that *"File layout, declaration order, comments, whitespace ... move nothing"* for the id and the wire. §8.3 claims the listing is *"stable across builds and diffable between them"*; under a rename it is not, and the `file` column churns beside it. **Recommendation: keep the `file` column, drop it from the order.** Sort each set by declaration NAME. It is stable under a rename and a move, it is still a byte comparison for §8.7, grouping-by-file stays one pass over the column, and the owner's *"everything that was in the schema built"* keeps its "where".
- **R7 — dangling citation.** §8.3 says *"exactly as §8.1's `TableRef` is"*. `TableRef` appears **nowhere else in the spec** — it exists only in the C# emitter. A reader cannot follow it.
- **R8 — USAGE.** *"That adds one file to the unit — `TabledemoView.h` and `TabledemoView.cpp` ... one pair for the whole unit."* One file, then two.
- **R9 — no §13 entry, and the sibling PR has one.** #157 carries seven owner verbatims. §13.6 "The block form, ruled" landed yesterday on exactly this pattern, and the sibling spec-first PR of the same day (#297) adds **§13.7 "The build version, ruled" — "Owner rulings, 2026-09-02, in the order given."** #299 adds §14 and §15 and no §13. The verbatims that produced this section — *"Does a type have a view generated for it, alongside it, and a table has the view built in?"*, *"walk enums, constants at runtime"*, *"the set of types" / "the set of tables"*, *"walk each table set of properties" / "(but with a view on that type?)"*, *"inspect everything that was in the schema built"*, *"This can be on the side if you want"* — belong in a §13 subsection. (Also: #297 touches §11 and §15 too, so one of the two takes a small rebase.)

### The other three open items

1. **`Table*` spellings in a table-free unit** — **keep them.** One vocabulary is the whole design; aliases fork the walker every tool is written against. But D4 is the price of keeping them and has to be paid in the same page.
2. **Default `declared` vs `none`** — **`declared`.** A marker inert until a flag is passed is the accepted-and-ignored shape §11 refuses for `| stride` (*"an inert attribute is a lie about the declaration"*). Put that reason on the page; it currently argues only convenience.
3. **Build identity** — silence is right while #297 is in flight, but **name it in §15** as a follow-on so the two pages meet rather than both staying quiet.

---

## VERIFIED GOOD

- **The 19 is right for what it counts.** An independent walk over the checked IR reproduces exactly those 19 `fixed`/`ufixed` fields, one for one. (D2 is about what the sentence *around* the number claims.)
- **Kind 0 is genuinely free.** `ir.TableKind*` runs 1..16; nothing spells 0. The reserved value is real.
- **`| view` moves no id.** `schema projection` over a type with and without the tag is byte-identical, and tags do not enter the projection at all — SPEC §4.2's *"exactly as a type tag does not"* is exact.
- **Two units into one directory is a real layout, and decision 3 is right.** `Makefile:566-568` generates `package bench` and `package realworld` into `generated/bench/cpp`. A fixed file name collides there; `<Package>View.*` does not.
- **The corpus is clean for every new claim.** No declaration named `UnitView`/`UnitViewInfo`/`ViewType`/`ViewVocabulary`/`ViewVariant`/`ViewConstant`, and none ending `TableType`/`TableInfo`/`TableFields`, anywhere in `examples`, `examples128`, `tables`, `bench`.
- **The leaf claim holds.** `<Base>Table.h` includes `<Base>.h` and other `<X>Table.h` and no wire header, so a view consumer really does inherit no serialize runtime.
- **USAGE's C++ walk compiles against the real vocabulary** — `name`, `type_name`, `is_array`, `array_bound`, `optional`, `has_range`, `range_min`, `range_max`, `table` all exist on `TableFieldInfo`, and the unqualified spellings follow the doc's existing convention.
- **Zero cost under `--views none` is genuinely zero**, including for an enum-heavy unit: no view file exists, so "catalogued unconditionally" costs nothing. The decision-5 asymmetry (names and values vs a descriptor table) prices correctly.
- **Placement.** §8 kept its number, so all ten `SPEC-TABLES.md §8` citations under `internal/`, `ir/`, `compiler/`, `test/` still land. §14's four-model shape matches the existing register exactly, and §8.7 states its own negative control (*"established by doing it rather than assumed"*) — the estate's law, not a gesture at it.
- **Decision 6 (no `| doc`)** matches SPEC §4.1's pinned design and §9 q5's deferral; §8.6 and §15 say the same thing without contradicting each other.
- **§3.3, §7, §12 untouched**, as the body says.

---

## VERDICT: **LAND-AFTER-FIXES**

The shape is right and the reasoning is the owner's own — one vocabulary, on the side, paid for by whoever asks. The blockers are D1 (the marker collides with the tag namespace and is accepted today), D2 (the 128-bit rule contradicts the rule it invokes, and the count is a different set), D4 (a generate flag changes what the language accepts, against the page's own rule), D5 (the cited contract says the opposite for unknown keys), and D6 (the marker is shallow where the page's reasoning demands a closure). D3 is a one-sentence honesty fix and should ride with them. R3's three-tree byte comparison is worth adding before the emitters start, because it is the gate the independence claim currently lacks.

gafferongames:
## Delta — cold read addressed, one pass, same branch (`44a4a9c`)

Every defect was reproduced here before the page moved; the two numbers below are my own walk, not a restatement.

### Blockers

**D1 — `| view` collided with the tag namespace, and was accepted today.** Reproduced: `type ShipState | view { … }` → `schema check` exit 0, IR carries the tag. Fixed as a **language change**, stated as one: SPEC §4.2's Type tags section gains the rule (a claimed valueless attribute spelling cannot also be a tag), the `| view` bullet states the reservation and that it is a change, and **§4.11 gains the row** — a type tagged `view` is refused by name, id-neutral because tags never enter the projection. SPEC-TABLES' status paragraph now names **the two front-end changes that precede any emitter** (this reservation, and D4's claims), so "refused by name" is a statement about a front end that is being asked for, not about one that silently accepts.

**D2 — the 128-bit family.** Confirmed exactly: `TableScalarKind`'s `TInt` arm dispatches 8/16/32 and lets `default:` take 64 **and** 128, so `WideProbe.entity_id` answers kind 9 and `WideProbe.energy` kind 5. My count: **19 fixed/ufixed + 8 128-bit = 27**, matching yours. §8.2 now states the no-table-wire-kind set as `fixed`/`ufixed` **and** `int128`/`uint128`, says 27, and adds the paragraph the page owed: the shared kind function names 128 explicitly rather than defaulting it, and that correction is part of the view's work because the view is what makes the path reachable.

**D3 — describe, not edit.** §8.2 now says it in the open: name, declared spelling as text, offset, size, and `double` bounds — enough to LIST, not to decode; no numeric-shape column, so nothing reads those bytes without parsing `type_name`; and the `double` bounds round a 128-bit maximum **up**, the one direction an editor must not clamp to. §15 carries **the typed descriptor form** as the follow-on (signedness / I / F / width, bounds that survive the value, and whether the same columns serve the packet wire), with the page explicitly claiming no sufficiency.

**D4 — the claim was one name of a surface.** §11's paragraph is rewritten to claim, unconditionally in every unit: the **three per-declaration spellings** (`TableFields`, `TableInfo`, `TableType`), and the **unit-level table-runtime names** — `TableTypeInfo`, `TableFieldInfo` at the head, plus `TableKeyed` (with `ScoreBoard.per_team [Team]` named as the reason a keyed array reaches a `type` body), `TableRef`, `TableReport`, `TableWriter`, `TableReader`, `TableEnumId`, `TableEnumValue` and the rest of that one registry. Stated as front-end law, with the reason: a generate flag may not turn a legal unit's generated code into code that does not compile.

**D5 — the contract.** Quoted in full on the page now, both halves: a generator **ignores keys it does not know**, which is exactly why `views` may not be left unknown to the seven — so **every backend knows the key and either honors it or refuses it by name**. SPEC §6.1 and §8.4 both say it that way.

**D6 — the marker is closure-transitive.** §8.2 states it on the page's own reason (a walk that enters a nested type needs that type's descriptor), §8.3's `types` sentence takes the closure with it, and USAGE says marking `ShipState` views `Vec3`, so the `// recurse` line no longer dead-ends.

### Risks

- **R1** corpus gate scoped: each backend byte-compares the pin **for the units that backend accepts**, with C#'s pointered-unit refusal named as the reason.
- **R2** the zero-cost gate names the six — `UnitView`, `UnitViewInfo`, `ViewType`, `ViewVocabulary`, `ViewVariant`, `ViewConstant` — and nothing else, restated as §2.2's machinery-vs-columns distinction so a table unit built with `--views none` cannot red on its own descriptors.
- **R3** the **independence gate** is in §8.7 as you constructed it: each corpus unit generated three times (`none|declared|all`), every file other than `<Package>View.*` byte-identical across all three, `schema id` unchanged, no view file in the `none` tree.
- **R4** `capitalize(package)` + `View`, with the C++ home spelled out (file name capitalized, `namespace tabledemo` unchanged), and a §11 line refusing a schema file whose generated name would be the view file's.
- **R5** a field of a type **no table closure reaches** carries `id == 0` and a **NULL** text key — those two columns describe a wire such a type does not have, and its ids were never collision-checked.
- **R6** ruling taken: each set **ordered by declaration name**, `file` column kept, with SPEC §3.2's "file layout moves nothing" as the reason on the page.
- **R7** the dangling `§8.1's TableRef` is gone — replaced by the reason itself (no initialization order across a partial class's files).
- **R8** "one pair of files … one pair for the whole unit however many schema files it has".
- **R9** **§13.7 "The view, ruled"** added with all seven verbatims in the order given, following §13.6's pattern. If #297 lands first this renumbers to §13.8 — happy to take that rebase.

### The other three

Rulings accepted and now argued on the page rather than in the body: `Table*` spellings **kept** (D4 is the price and is paid in the same page); default **`declared`**, with the reason stated as §11's own — a marker inert until a flag is passed is the inert attribute `| stride` is refused for; build identity **silent, named in §15** so the two pages meet.

`make check` clean over the whole corpus.

gafferongames:
## Owner ruling folded in — the marker and the flag are gone (`a610635`)

Subtraction pass, item 1: **no `| view`, no `--views`.** The view file on the side carries everything — every type, every table, every enum/flags/union, every constant — for every unit; a consumer that compiles it pays for it, one that does not pays nothing.

**Removed:** SPEC §4.2's `view` attribute *and* the tag-namespace reservation that existed only to make the marker parse (D1 dissolves — there is no spelling to collide); SPEC §4.11's reserved row; SPEC §6.1's `--views declared|all|none` and the Options-contract paragraph (D5 dissolves — no key to know or ignore); §8.4's three modes; §11's five view refusals; §14's marker-and-flag reasoning; closure-transitivity (D6 moot — every type is viewed, so a union arm's payload descriptor is never NULL and the recursion never dead-ends); the three-tree independence gate.

**Replaced with:** §8.2 *"The type view: every type, on the side"*; §8.4 *"What it costs: the include rule"* — placement answers the cost question the flag was answering, and answers it **per consumer rather than per generate**, so one unit's output serves the editor and the game at once; §11's view bullet is now one sentence saying the view refuses nothing a schema can write, because nothing in a schema can ask for it wrongly.

**The name claims got stronger, not weaker.** They were unconditional because a flag must not change what the language accepts; now they are unconditional because **every unit emits a view file that defines the descriptor primitives**, so a table-free unit may never declare `TableFieldInfo`, `TableTypeInfo`, `TableKeyed`, `TableRef`, … or the per-declaration `TableFields`/`TableInfo`/`TableType` triple. D4 stands as written and reads better.

**The independence gate reduces to what it is now about:** every generated file other than `<Package>View.*` byte-identical to what the unit emitted before the view existed, `schema id` unchanged — plus §8.4's containment gate (not one of the six registry symbols outside the view pair). R3's substance kept, its three-tree machinery dropped with the flag that needed it.

**Everything else from the cold read stands:** kind 0 for `fixed`/`ufixed` **and** the 128-bit family with the truthful dispatch note (8 fields answering a 64-bit kind today; 27 in the set); kind 0 **describes and does not decode**, with the typed descriptor form in §15; order by declaration name with the `file` column kept; corpus gate scoped to the units each backend accepts; `id == 0` and NULL text key on a type no closure reaches; `capitalize(package)` with its C++ home and the §11 file-name collision line; §13.7 with the seven verbatims — now with the subtraction ruling recorded beside them.

§14's rejected-model list keeps the marker design as **model 2**, with the reason on the page: it existed to answer "what does this build pay for", and placement answers that by itself, so the attribute, the flag, the closure rule, the per-kind refusals and the reserved word went out together. USAGE carries the *"Specified, not yet implemented"* banner the block form uses.

`make check` clean; CI green on the previous head and re-running on this one.

gafferongames:
## Cold read — delta on `44a4a9c` + `a610635`

Fresh clone, re-checked out, everything below re-run from scratch against the new head. The subtraction is the right call and it dissolves more than the three defects it was aimed at: with no marker and no flag there is no tag-namespace collision, no contract to misquote, no closure rule to get wrong, and no "did the flag move anything" gate to design. The page got shorter in the places that were carrying the most risk.

**Reproduced, from the page as it now reads:**

- **The registry is complete over all four corpora.** I rebuilt `UnitView()` from §8.3 against the checked IR and walked every declaration and every property:

| unit | types | tables | enums | flags | unions | consts | fields walked |
|---|---|---|---|---|---|---|---|
| `example` | 70 | 0 | 6 | 2 | 3 | 22 | 223 |
| `ludicrous` | 7 | 0 | 1 | 0 | 0 | 1 | 34 |
| `tabledemo` | 4 | 16 | 6 | 1 | 1 | 3 | 75 |
| `graphdemo` | 1 | 11 | 1 | 0 | 0 | 1 | 39 |

Nothing is missing: every `ir.Decl` kind plus `Tables` maps onto a row, **no union arm lacks a payload descriptor, and no nested column is NULL** in any of the four — which is the property §8.2's "every type, no subset" was bought for, and it holds.

- **27 kind-0 fields in `examples128`, exactly.** The corrected set enumerates as 19 `fixed`/`ufixed` + 8 of the 128-bit family — `WideProbe.{entity_id,energy,flux,bias,seed}`, `LudicrousState.{keys,target_id}`, `DegenerateProbe.locked_wide`. The number and the set now agree, and §8.2's dispatch note ("lets the widths it does not name fall to the 64-bit answer ... eight fields ... describing a 16-byte field as an 8-byte one") is precisely what `ir.TableScalarKind` does today.
- **`id == 0` / `json == NULL` outside a closure:** fires on 223 fields in `example` and 34 in `ludicrous`, on none in the two table corpora (every type there is a closure member). The rule is well-formed and non-trivial.
- **The two new claims are correctly stated as not-yet-front-end.** Today: a table-free unit declaring `type TableFieldInfo`, `type TableKeyed` and `type UnitView` checks **exit 0**; a schema file named `R299dView.schema` in `package r299d` checks **exit 0**. The status paragraph's "one front-end change comes BEFORE any emitter" is doing real work and is the right sequencing.
- **§13.7's verbatims: all eleven match #157 exactly.** Extracted every quoted string from §13.7 and grepped each against the issue body — 11/11 `OK`, and they cover every owner quote on the issue with none invented.
- **No `view` tag / `--views` flag residue.** The only surviving mentions are the two deliberate historical ones (§13.7's subtraction bullet, §14's rejected model 2). `view` is not reserved anywhere, correctly. USAGE's banner matches the block form's own precedent at line 1021 word for word.

---

## DEFECTS

### D-a. The unchecked-hash hazard is closed for fields and left open for variants — and for keys

§8.2 now says, correctly: a field of a type no closure reaches carries `id == 0`, *"its field ids were never checked for collisions (§5's refusal is scoped to the closure)"*. That reasoning applies unchanged to **three more id-bearing columns**, and the page closes only one. `checkTableVariantIdentity` is scoped to `closureNames` exactly as the field check is. Reproduced:

```
enum Grade { Aaes, Aaol }        // both fold to table-wire id 0x4111

type Holder  { g Grade }         -> check exit 0          (packet-only)
table Holder { g Grade }         -> refused: "enum Grade: variants Aaes and Aaol
                                    collide on table-wire id 0x4111 ... rename one"

type Board { per_team [Team]int32 }   // Team = { Aaes, Aaol }  -> check exit 0
```

So in the registry:
- **`ViewVariant.id`** — an enum or union no table closure reaches hands a tool two variant rows under one id. In `examples` this is every vocabulary in the unit: 6 enums and 3 unions, zero tables, so **every variant row there carries an unchecked id**.
- **`variant_id()` and `key_id()`** in §8.1's columns, for the same declarations, reached through a viewed type's descriptor rather than through the registry.

The fix is the sentence §8.2 already wrote, extended: a vocabulary no table closure reaches carries `id = 0` on its variant rows — which is what a flags row already does, for the same "there is no wire here" reason — or the §5 refusal widens. Leaving it is the one place the page still hands a tool a number it must not trust.

### D-b. Four residues of the subtraction, two of them contradictions inside the normative text

1. **§8.3's struct block still describes the old subset model**, twenty lines above the bullet that corrects it:
   ```cpp
   const TableTypeInfo * payload;  // that payload's descriptor where it has a
                                   // view (§8.2), else NULL
   ```
   against *"The descriptor is never NULL for a declared payload: every type in the unit has a view (§8.2), so an arm is walkable to its bottom."* The struct block is the half an implementer types in.
2. **`struct ViewType  // a type with a view, or a table`** — there is no longer a type without one.
3. **§8.7 still pins the old order.** *"...and every constant, **in the schema's own order**"* — against §8.3's *"Each set is ordered by DECLARATION NAME, and deliberately not by where the declaration lives."* The pin is a byte comparison, so the two sentences specify two different gates.
4. **§13.7's first bullet contradicts the ruling the last bullet records**: *"a type's view is generated alongside it **when asked** (§8.2)"* — nothing asks any more. It reads oddly precisely because it is the bullet that opens the section the subtraction closed.

Plus two in SPEC.md that are pure leftovers of removing `view` from a three-item list — both lines are otherwise identical to `main`, and reverting `,` to `and` restores it exactly:
- the `TypeDecl` grammar comment: *"qualifiers = the type TAG, the cpp_* pair"*
- §4.2: *"type declarations take a tag, the `cpp_native`/`cpp_include` pair (below);"*

---

## RISK

**The independence gate is well-specified and its left-hand side is thinner than it sounds.** §8.7: *"every generated file other than `<Package>View.*` is byte-identical to what the same unit emitted before the view existed, and `schema id` reports the same protocol id."* With no flag, "before the view existed" can only mean a recorded snapshot — and the repo has one: `internal/goldens` pins generated source byte-for-byte plus `id.txt`, so for `examples` and `examples128` the gate reduces to *"landing the view moves nothing in `testdata/golden/`"*, which is already mechanical and already red-capable.

It has **no referent for `tables/examples` and `tables/pointers`** — `goldens_test.go` covers `corpusDir` and `corpus128Dir` only, and no other test compares table-unit source. That is the half where movement is plausible: §8.5 splits `<Name>TableType()` between the table header (types a table reaches) and the view file (types none does), so the boundary the view draws runs straight through `<Base>Table.h`. Recommend extending the source goldens to the two table corpora — same harness, two more directories — either in this page as part of the gate or as the landing PR's first commit. Worth naming now because the gate's wording implies a comparison that, for those units, nothing performs.

Minor, non-blocking: **§13.7's subtraction bullet is the only one in §13 carrying no owner words.** Every other bullet there, and every bullet of §13.1–§13.6, quotes. If the ruling has a phrasing on record it belongs beside them; if it does not, the bullet is fine as prose and this is an observation, not a defect.

---

## VERIFIED GOOD

- **The subtraction is sound and the page argues it on its own terms.** §8.4's *"per CONSUMER rather than per generate: one unit's output serves the editor and the game at once, with no second generate and no second tree to keep in step"* is a better property than the flag had, and §14's model 2 records what went out and why in one place. The "completeness is what an editor needs, and a subset is what a marker sells" paragraph is the argument, stated once.
- **D2 fully closed.** The no-wire-kind set is named in full, the count is the set, and the dispatch correction is stated as a correction that *belongs with the view* — the honest framing, since the view is what makes the path reachable.
- **D3 fully closed and then some.** §8.2's "Kind 0 DESCRIBES; it does not decode" says the `double` bound *rounds UP, "the one direction an editor must never clamp to"*, disclaims sufficiency explicitly, and §15's typed-descriptor-form follow-on names all three things it would need. USAGE carries the same limitation in the reader's own words. Nothing over-claims.
- **D4 fully closed, and wider than I asked.** §11 claims all three per-declaration spellings (not the one), and widens the unit-level table-runtime set — `TableTypeInfo`, `TableFieldInfo`, `TableKeyed`, `TableRef`, `TableReport`, `TableWriter`, `TableReader`, `TableEnumId`, `TableEnumValue` — to every unit, with the `ScoreBoard` keyed-array-in-a-type case cited as the reason `TableKeyed` is not hypothetical. That is the whole class.
- **R1** scoped correctly, with the reason on the page (*"a gate cannot ask a backend for a listing of a unit it declines to emit at all"*) rather than as a bare exception.
- **R4** answered on both halves: `capitalize(package)` given a C++ home (*"it is a FILE name, and generated file names are basename-shaped in every target"*, `namespace tabledemo` unchanged inside), and the `<capitalize(package)>View.schema` file-name collision refused in §11 with the reason it lives there.
- **R5** answered for fields (D-a is its unanswered half). **R6** answered exactly as recommended — name order, `file` column kept, and the argument tied to SPEC §3.2 (*"a listing that reordered when a file was renamed ... would be the one artifact in the tree that did"*). **R7** gone: the `TableRef` citation is replaced by the actual C# initialization-order reason. **R8** gone.
- **§11's view bullet is the right shape**: a construct that refuses nothing, stated as a property with the reason, not as an omission — and it counts its own additions correctly (two name claims, one file-name collision).
- **§8.5's leaf/no-wire-header claim still holds** against the emitters: `<Base>Table.h` includes `<Base>.h` and sibling table headers, no wire header.
- **USAGE** is consistent with the new design throughout — no flag in the invocation, "you pay for it by compiling it", the name-order sentence, and the kind-0 limitation stated to the reader rather than left in the spec.
- **§3.3, §7, §12 untouched**; §8 keeps its number, so all ten `SPEC-TABLES.md §8` code citations still land.

---

## VERDICT: **LAND-AFTER-FIXES**

The design is settled and every blocker from the first read is gone. What remains is one substantive sentence (D-a: variant and key ids outside a closure, the symmetric half of a rule the page already gets right for fields) and four residues, two of which are outright contradictions between a struct block and the bullet under it — which on a spec-first page is the failure that actually reaches an implementer. None of it is design work; all of it is one line each. The independence gate's missing left-hand side for the table corpora is worth a decision now even if the goldens extension lands with the emitters rather than here.

PR #305, opened 2026-09-02
Keyed arrays: iteration over valid slots; Slot<Key>() removed (#289 #290)

### LINKED ISSUE #289: Dogfood finding: iteration over an enum-keyed array — every call site wants `for ( auto & ship : root.ships )` instead of `for ( int i = 1; …) root.ships[(ShipType) i]`

### LINKED ISSUE #290: Dogfood finding: §2.4's "prefer Slot<Key>(), which is most places" is ZERO places in a data-driven engine — reword the guidance

gafferongames:
Closes #289. Closes #290.

**Spec and code in one PR, deliberately.** The spec change is one paragraph of §2.4 and this PR is its implementation — splitting them would have put a claim on the page that nothing yet emitted.

## What the dogfood said

Two findings from space#443, one construct:

- **#290**: `Slot<Key>()` was used **zero** times. Keys in a data-driven engine are runtime values — an enum field read out of a config, a key a tool hands in — so a compile-time accessor serves a case that does not occur, and §2.4's "prefer it, which is most places" was wrong about the primary consumer. Every access went through `operator[]( E )`, whose assert `NDEBUG` compiles out; the page never said so plainly.
- **#289**: every consumer of a keyed array wants the whole array, and spelled it `for ( int i = 1; i < (int) NumShipTypes; i++ ) … m_live.ships[(ShipType) i]` — the slot-0 rule re-derived by hand at every site (five of them in one file of the dogfood).

## What lands

**§2.4 rewritten.** `Slot<Key>()` is REMOVED, not demoted — a subtraction, on the measurement. `operator[]( E )` is *the* accessor, its assert is a debug guard, and the page states outright that a shipped build carries no check on a keyed index. **ITERATION is the safety**: the keyed type iterates its valid slots, `1 .. E.Max`, yielding the key beside the element, so slot 0 never appears at a call site and no consumer spells a lower bound, a cast or an `E.Max` of its own.

```cpp
for ( auto [ ship_type, ship ] : fleet.ships )   // slot 0 is not in the range
{
    ship.health *= 2.0f;                         // the element is a reference
}
```

**§11**'s keyed refusal restated on the same terms. **§8**'s key columns and **§16.2**'s `[E]T` row need nothing — iteration is a language surface, not a wire or descriptor fact. **No new claimed spelling**: `begin`/`end` and the C# enumerator are members reached through their owner, and the nested `Entry`/`Iterator` types claim nothing at unit level, so `internal/tablenames` is untouched.

**C++ and C# together.** C++ yields a nested `Entry { E key; T & element; }` by value — a proxy, so the structured binding names the reference and iterating fills an array as well as reading it; const iteration yields const elements. `auto & [ ... ]` will not bind to the proxy and that refusal is by design; the spelling is `auto [ ... ]`, or `auto && [ ... ]`. The iterators carry the `iterator_traits` typedefs, so `std::distance` works.

C# gets a struct enumerator over the same range with `Deconstruct`, so `foreach (var (key, element) in keyed)` reads the same way and allocates nothing. **Two things vary by port, and §2.4 now says both**: its key is the slot index — the currency C#'s indexer already takes, the language having no non-boxing generic enum conversion — so a site wanting the key as its enum still writes the cast there; and its `Entry.Element` is a value, so a class element (a nested table) is the live instance while a scalar or enum element is a copy — **in C#, iteration reads those and the indexer writes them**. A ref-yielding C# enumerator is a follow-on, not this construct's shape.

The storage type is still one member — trivially copyable, standard-layout, relocatable (§9).

**USAGE** shows the range-for, not the integer loop.

## Held by test

- Both legs iterate **every keyed array their backend can generate** and see slot 0 in none of them: the space-shaped `Pack` root (one record per ship type, one threshold per difficulty), the nested `Keyed` config, and BOTH generations of the evolution unit — V1's three over a four-variant key, V2's four over a five-variant one. The pointered `Depot` is C++'s alone, because C# refuses a pointered unit by name. Plus writing through the iteration, const iteration, `std::distance` over the traits typedefs, and a value filled by iteration riding and reading back by name.
- **The `None` assert on `operator[]` stays, and is now tested rather than claimed in a comment**: a forked child indexes `None`'s slot and must die on the assert. (C#'s `ArgumentOutOfRangeException` on the same index was already tested.)
- **Negative control** — `make tables-keyed-iteration-negative-control`: the EMITTER's `begin()` is sabotaged to slot 0 through `go build -overlay` (no tracked file written), the whole corpus is regenerated from the sabotaged compiler, and **`test/tables/main.cpp` itself must go red — 120 failures**. It is the control this change needs, because an untouched slot 0 holds the same declared defaults every other untouched slot does: a walk that visited it would look identical to a walk that did not.

## Gates

`make test` green end to end (nine languages, the goldens, both tables legs including the ASan twin), plus the zero-cost gate, the C# standalone and pointer-refusal gates, the generic-walk gate, both existing negative controls and the new one; `make check` and `go test ./...` green. No wire change, no descriptor change, no baseline move.


gafferongames:
**LAND-AFTER-FIXES.** The mechanism is right and genuinely zero-cost. Every defect below is a page or a claim; no emitted code needs to change.

### Defects, ranked

**1. §2.4: "The element is a REFERENCE to the slot, so iterating is how a whole array is filled as well as read."** The port paragraph qualifies only the *key currency*, never the element. C#'s `Entry.Element` is a `readonly T` copy, so for scalar, enum and struct slots iteration **reads and cannot write**. Reproduced: C++ `for ( auto [t,q] : fleet.quotas ) q = t*100` leaves `quotas[Bomber] == 300`; the identical C# loop leaves `Quotas[(int)Bomber] == 0`. Written down only in a `test/cs-tables/src/Program.cs` comment, not the page. *Fix:* one clause on §2.4's port sentence and USAGE's C# paragraph: where a port's element is a value, iteration reads and the indexer writes.

**2. §2.4: "and neither does a lower bound, a cast, or an `E.Max` a consumer had to spell for itself."** The cast half is C++-only. C# yields `int`, so any site wanting the key as its enum still writes `(ShipType) key` — #289's cast moved from the index to the key, not removed. *Fix:* say it where the port sentence names the index currency.

**3. PR body: "Both legs iterate every keyed array in the corpus."** The C# leg does not: `grep -c PackConfig test/cs-tables/src/Program.cs` → **0**, so `Pack.ships`/`Pack.thresholds` and `V2.Bank` are never iterated there; C++ omits V1's three. *Fix:* add the loops or restate as "every element kind".

**4. `Makefile:331-345` — the negative control validates a surrogate, not the guarded test.** It builds `keyed_iteration_negative_main.cpp`, a re-implementation of the sweep, which **exits 0 under the sabotage and 1 without it**. So it proves the sabotage is observable by a fixture written for it, not that `test_keyed_iteration` reddens. I patched the real headers' `begin()` to slot 0 and rebuilt `build/schema_test_tables`: **97 failures** — the guarantee *is* held, by a path the control never exercises. *Fix:* compile `test/tables/main.cpp` against the sabotaged include dir, require non-zero exit, drop the surrogate. The comment "the fixture must go red" also inverts that fixture's actual exit code.

**5. `auto &` — #289's own words and #290's sketch — is a hard compile error, and no page says so.** `for ( auto & [ ship_type, ship ] : fleet.ships )` → clang: *"non-const lvalue reference to type 'Entry' cannot bind to a temporary of type 'Entry'"*, with a note naming `begin`. Loud and correct — but the spelling that opened the issue does not compile. *Fix:* a parenthetical in §2.4 — the entry is a proxy: `auto [ key, element ]` or `auto &&`; `auto &` will not bind.

**6, and residues.** `Iterator` has no `iterator_traits` typedefs and no `operator==`, so `std::distance( k.begin(), k.end() )` fails on a missing `difference_type` — range-`for` only; fine if deliberate, but say so. `test/cs-tables/src/Program.cs:1186` still reads "C# expresses neither a compile-time key nor…" — no port has one now. `grep -rn 'Slot<'` is otherwise clean — every hit is `TableSlot<T>`, the unrelated arena handle. The C++ assert lost its sentence with `Slot<Key>()`: it prints only the expression while C# throws the full message; `assert( key != E::None && "slot 0 is None's…" )` restores parity.

### Reproduced

Fresh 4-variant unit (one keyed nested table, one keyed `int32`) generated through the public `compiler` API, built with the repo's own `-std=c++17 -Wall -Wextra -Werror -ffp-contract=off`.

- `sizeof(TableKeyed<T,E,5>) == 5*sizeof(T)` for both; `is_trivially_copyable`, `is_standard_layout`, `is_trivially_destructible`, `!is_polymorphic`; `sizeof(Fleet)` == the two summed. **4 entries == E.Max**, ascending from 1, slot 0 never; writes read back by `operator[]`; const instance yields `const T &`. Clean at `-std=c++20`/`c++23`.
- **Zero cost:** at `-O2` the range-`for` and the hand-written `for (i=1; i<Max+1; i++)` emit **byte-identical arm64** — `ldur q0,[x0,#44]; addv.4s; fmov; ret`. `nm -gu` on the object is empty.
- **NDEBUG true:** `-DNDEBUG -O2`, `ships[ShipType::None].health = 7.0f` runs and writes slot 0; without it, SIGABRT.
- **C#:** 4 entries, never 0; class-element mutation visible via the indexer; value-element mutation lost (see 1); allocation over 1000 `foreach` passes = **0 bytes**.
- **Name claims hold:** `table Entry`/`Iterator`/`ConstIterator`/`Enumerator` beside two keyed arrays compile in both languages.
- Green: `go test ./...`, `schema_test_tables`, `test/cs-tables`, the new control (per 4).

gafferongames:
All six addressed, on `abc4c7a` (rebased onto `3c81b34`, then `6d88ef1`). No emitted C++ semantics changed — the read's soundness findings still hold.

**1. The element is a reference in C++, a value in C#.** §2.4 now states the two things that vary by port rather than qualifying only the key: the KEY's currency, and **whether the ELEMENT is a reference**. C#'s `Entry.Element` is a `readonly T`, so a class element is the live instance and a scalar or enum element is a copy — **iteration reads those and the indexer writes them**. In §2.4, in USAGE's C# paragraph, and in the C# emitter's own comment. The C# emitter is *not* changed to ref-yield: KISS, and a follow-on only if space's port needs it.

**2. "No cast" is C++-only.** Said where the port sentence names the index currency: a C# site that wants the key as its enum still writes the cast, so #289's cast moved from the index to the key there rather than vanishing.

**3. Coverage claim — extended, not narrowed.** C++ gains V1's `bank`/`tokens`/`ranks`; C# gains `PackConfig`'s `ships` and `thresholds`, V1's three, and `V2.Bank`. Every keyed array either backend can generate is now walked; the pointered `Depot` stays C++'s alone because C# refuses a pointered unit by name. The PR body says that instead of "the corpus".

**4. The control walks the real path.** It sabotages the EMITTER's `begin()` through `go build -overlay` — the big-endian control's rule, no tracked file written — regenerates the whole corpus from the sabotaged compiler, builds `test/tables/main.cpp` itself and requires it to fail: **120 failures**, and the target also refuses a sabotage that patched nothing and a redness that is not a `CHECK`. The surrogate `keyed_iteration_negative_main.cpp` is deleted. Same target name.

**5. `auto &` is a compile error by design.** §2.4 and USAGE both say so now: the entry is a proxy handed out by value, the spelling is `auto [ key, element ]`, and `auto && [ ... ]` binds if a reference form is wanted.

**6, and the residues.** The iterators carry the five `iterator_traits` typedefs (`forward_iterator_tag`) and `operator==`; `std::distance` over both the mutable and const iterators is held by test. `test/cs-tables/src/Program.cs:1186` corrected — no port has a compile-time key now. And the assert regained the sentence it lost with `Slot<Key>()`: `assert( key != E::None && "slot 0 is None's and is never valid…" )`, so C++ prints what C# throws.

`make test` green end to end, `go vet`/`go test ./...` green.

PR #309, opened 2026-09-02
The block form: C++ block builder and C# reflective read, golden-locked across the boundary (#288 #287)

### LINKED ISSUE #288: Expressiveness gate 2 — THE RENDER DATA: using tables or types, the per-frame render data must be implementable end to end (C++ producer, C# consumer pointing at it, no hand mirror)

gafferongames:
The block form, both backends, golden-locked across the boundary. Against `main` (49bbdc7), rebased. Closes #288's implementation half and #287.

Every fixed table has a block form, generated on both sides and emitted **on the side**, and the two-language gate is green: a C++ producer writes a block, a C# consumer opens those very bytes and reads every field of every row — twice, through the generated blittable struct and through the block descriptors.

## The layout numbers

The compiler's layout model reproduces §19.1's worked table to the digit, in the model, in the generated C++, and in the running program:

| | |
|---|---|
| array starts | **192 / 320 / 26752 / 84352 / 92992 / 95872 / 495872 / 559872 / 572672** |
| used extent | **577,472** |
| storage from the declared maxima | **7,879,488** (the page's 7.51 MiB) |
| projection | **176** — the 24-byte prologue, a `uint64` and nine triples |

`tables/block/Render.schema` is the dogfood: the nine-array render frame with space's own maxima, and `RenderShip` carries two `bool`s beside its scalars — the case where C#'s two layout models disagree (§19.3). `tables/block/Padded.schema` declares the opposite on purpose: field orders that leave holes, plus every inline storage class the form leaves where it is.

## The build version is §20's number

`ir/buildversion.go` implements §20.2's canonical **cook projection** — the three header lines, records sorted by name with fields in declaration order carrying `kind=`/`offset=`/`size=` and the referent tokens (`type=`/`enum=`/`union=`/`key=`/`payload=`), a `block` line with `slot` lines, then the enums and unions with STORED VALUES from 1 — hashed SHA-256, low 64 bits big-endian.

It reproduces the page's own two numbers exactly, and the projection **text** is pinned character for character beside each, so a port reproduces the text and not only the number:

| | |
|---|---|
| §20.2's worked example | **`0x7402a36de22d9728`** |
| the same unit with no table | **`0x49947af3382f914e`** |
| the block corpus (`tables/block`) | **`0x92b8887e8391446a`** |

`schema build-version [--facts] <unit>` lands (§20.7). A Makefile gate holds the CLI, the C++ constant and the C# constant to **one** number — a tuple a store is indexed by cannot mean two things — and `TestEmittedBuildVersionIsTheSpecsNumber` generates the page's example under the page's protocol id and requires the emitted literal to be `0x7402a36de22d9728` on both backends. `--facts` output is pinned as a golden per unit under `testdata/golden/build-version/`.

§20.8's batteries land in `ir/buildversion_test.go`: the independence pair both directions, the meaning group's seven controls, the layout group's three, the four **referent** same-shape swaps a digest without the name tokens passes in silence, and the seven exclusions (`was`, a flags variant reordered, a flags variant renamed, a flags field's referent swapped, a guard removed, a json key changed, a comment).

**One reconciliation, reported rather than assumed.** §20.2's prose says "every record with a block form is followed by its PROJECTION"; its worked example — which §20.8 makes a golden — carries **no** `block` line for `ShipConfig`, a fixed table with no bounded array of structs. The number is the contract, so a `block` line is emitted for a record whose block form MOVES something, and a table with no out-of-line array contributes none. One line to change if a reader disagrees.

## The prologue, exactly

Three `uint64`s, and §19.1 is updated in this PR to say so with the reason:

| offset | word | what it is for |
|---|---|---|
| 0 | `magic` | what **refuses** a foreign byte order: read bytewise it is this build's constant, or that constant byte-reversed, or not a block |
| 8 | `build_version` | the unit's id (§20) |
| 16 | `byte_order` | what **records** which order wrote it — `1` little, `2` big — so a refusal can name the order and a dump can read it |

The build version can do neither job: §20 digests `byteorder` as a **generation input**, `little` for every target schema generates for today, so two builds of one schema for two byte orders emit the SAME id. The CI log for the big-endian leg is the evidence — both builds print `build version 0x92b8887e8391446a`, one with `byte order 1` and one with `byte order 2`.

The page said two words; the implementation taught it three, so **the page changes here, in the same PR, with the reason on it** — §19.1's prologue paragraph, its worked table restated at 176 (152, 168 and 176 all round to 192, so not one layout number moves), and §11 claiming `byte_order` beside `magic` and `build_version`.

## The big-endian leg carries the block form (#306)

`test/tables/block_endian_main.cpp` joins `make tables-big-endian` on the cook driver's own model — `write` / `accept` / `refuse`, built both ways, with a file between them. Green on s390x under qemu in CI:

```
./build/schema_test_block_endian write build/block-host.bin
  wrote build/block-host.bin: 1984 bytes, build version 0x92b8887e8391446a, byte order 1
qemu-s390x ./build/schema_test_block_endian_be write build/block-target.bin
  wrote build/block-target.bin: 1984 bytes, build version 0x92b8887e8391446a, byte order 2
qemu-s390x ./build/schema_test_block_endian_be accept build/block-target.bin      -> accept OK
qemu-s390x ./build/schema_test_block_endian_be refuse build/block-host.bin        -> refuse OK (this build's order is 2, the file's is 1)
./build/schema_test_block_endian accept build/block-host.bin                      -> accept OK
./build/schema_test_block_endian refuse build/block-target.bin                    -> refuse OK (this build's order is 1, the file's is 2)
```

The refusal case proves three things and not just the first: `BlockOpen` returned false rather than pointing at rows it cannot read; the file's magic read the way THIS build reads is this build's magic **byte-reversed**, so byte order is what refused it and not some other mismatch; and the recorded word names the other order. Then it lays out and opens a block of its own, because a clean refusal leaves the reader able to carry on.

## Gate 2 — the two measured numbers, on the Air

**This bench is a MacBook Air, not the box.** The box run is a later sitting under the bench rules. Each half runs its golden gate first and REFUSES to bench on a mismatch; arms are interleaved in one sitting, medians of 15, both halves over the same representative frame.

| number | hand-written | generated | ratio |
|---|---|---|---|
| **per-frame C++ WRITE** | 22.71 / 18.36 / 16.70 µs | 21.05 / 16.99 / 15.48 µs | **0.927 / 0.925 / 0.927** |
| **per-frame C# READ** (Release) | 13.822 / 13.527 / 13.588 µs | 13.546 / 13.649 / 13.606 µs | **0.980 / 1.009 / 1.001** |

Absolutes move with the Air's mood; the paired ratio does not. **Both clear the bar in the owner's words — the same speed, or not significantly slower.** The C++ write is ~7% *faster* than the hand-written scatter; the C# read is indistinguishable from the hand mirror. Both C# arms read through a contiguous `ReadOnlySpan<Row>` reinterpret at pitch == sizeof, with nothing per row.

**An instrument defect, reported because a harness is code and rots too.** The first C++ sitting read 18x, which was entirely the *hand* arm being eliminated: its offsets are compile-time constants, every frame's stores were redundant with the last one's, and the compiler removed the loop. The fix is a per-frame seed written into one field of every row, plus a volatile sink.

## The refuser, and how it detects

- **source-level** — `make tables-block-fill-refuser` extracts the generated fill path between its own markers from every `<Base>Block.h` and refuses `malloc|calloc|realloc|operator new|new |std::mutex|std::lock_guard|std::unique_lock|std::atomic|std::thread|pthread_|__atomic`. It also refuses if the extracted region is implausibly small, so the markers breaking cannot pass as the property holding;
- **runtime** — `test/tables/block_main.cpp` replaces global `operator new`/`delete` with counters and asserts the count does not move across `Begin`, the accessors and the fill. It measures the fill SERIALLY on purpose: the parallelism is the caller's loop and `std::thread` allocates on its own account.

Beside them: four workers over disjoint index ranges of all nine arrays, byte-identical to a serial fill, under the sanitized twin as well.

## The negative controls, pasted

**(a) the fill refuser, red under an injected lock and allocation:**
```
block fill refuser: the generated fill path allocates nothing, locks nothing, takes no atomic
block fill refuser negative control: the gate goes red on an injected lock and allocation
```

**(b) the projection's offset token is load-bearing** — `TestBuildVersionOffsetTermIsLoadBearing` reorders two same-size, same-alignment row fields and asserts BOTH that the real build version moves AND that a **positional-versus-keyed** sabotage — the projection with offsets stripped and its lines sorted, i.e. a digest that kept the facts and lost the positions — does *not*.

**(c) the C# padding fields, deleted:**
```
System.InvalidOperationException: schema block layout: Blockdemo.Block.PaddedRow.Value sits at 1
in this runtime and 8 in the schema the C++ side asserts — the two sides disagree about the bytes
(SPEC-TABLES.md §19.3)
```

## The gates

`tables-block` (C++ leg + **sanitized twin** + C# consumer), `tables-block-zero-cost` (no `<Base>Table.h`/`.cpp`/`.cs` carries one symbol of the form — `test/cs-tables` excludes `*Block.cs` from every glob, and that exclusion *is* the property), `tables-block-build-version`, `tables-block-fill-refuser` and its negative control, `tables-block-padding-negative-control`, `tables-block-gate2` (deliberately not in `make test`: a correctness suite whose verdict depends on the machine's mood is not one), and the block half of `tables-big-endian`.

## Decisions the page did not make, labelled as mine

- **The C# blittable records live in the nested namespace `<Pkg>.Block`.** The unit's own namespace already holds a sealed *class* of each of those names. A declaration named `Block` is refused by the C# backend, by name.
- **A block-form table's projection is `<Pkg>.Block.<Table>Projection`**, separate from the row record of the same name: a table can be both a block root and another block's row. In C++ it is nested as `<Table>Block::Projection`, so C++ claims no new file-scope name.
- **C# spells the entry point `<Table>Block.Open(out block, pointer, bytes)`**, a static on the claimed `<Table>Block` type.
- **The block DESCRIPTORS live in the Block files, not as columns on `TableFieldInfo`** — which is what keeps the Table headers byte-identical. They carry more than the page asked: each out-of-line array's element column carries that ROW's own layout, and a nested record's its own, so a walker DESCENDS through the graph with no name per row type. A C# `TableTypeInfo` has no offsets to give it.
- **A table whose closure carries a UNION has no block form**, and says so in its Block header: §19.3 pins the C# side to Sequential with generated padding, which cannot overlay arms.
- **Begin touches the prologue and the triples and nothing else** — not the projection's own scalars, which are the producer's exactly as the rows are.
- **`<Table><Field>Span` returns the contiguous view and `<Table><Field>` the pitch-driven one.** In C++ the latter converts implicitly to `T *`, so §19.1's `RenderShip * ships = RenderFrameShips( block );` reads exactly as the page writes it.
- **`OpenValidated` and the block spellings joined the claimed suffix set** (`Block`, `BlockStorage`, `BlockBegin`, `BlockBytes`, `BlockMaxBytes`, `BlockOpen`, `Counts`), plus two per out-of-line array for the row accessors.
- **The generated C# is `unsafe`**, and a project that compiles a Block file sets `AllowUnsafeBlocks`. Everything emitted is C# 9, for Unity.
- **C# has no block PRODUCER** — no `Begin`, no `Counts`, no `BlockStorage`. The page's C# half is the reflective read and the typed fast path; a producer is a follow-on.

## Not built here, named

- **`BuildVersion` beside `ProtocolId` in every backend** (§20.7) — it is emitted where the block form needs it, in the Block files. The other eight backends emit no Block file at all, and §20's own status list owns the rest.
- **space's render path** does not move here (#444).

## Verification

`make check`, `go test ./...`, `make shape-gate` (22 exemptions, both gate-2 files on the ledger) green. Full CI green on the branch head — [run 33616546345](https://github.com/mas-bandwidth/schema/actions/runs/33616546345): `test (ubuntu-latest)`, `test (macos-latest)` (both run the whole block battery), **`big-endian`** (which now carries the block driver), `windows`, `lint`, `shape-gate`, `vuln`, and all six inline gates.


gafferongames:
**CI green on the branch head** — [run 33613782000](https://github.com/mas-bandwidth/schema/actions/runs/33613782000), triggered with `gh workflow run ci.yml --ref block-form` because the `pull_request` leg did not re-fire on the last two pushes: `test (ubuntu-latest)`, `test (macos-latest)` (both run the whole block battery), `big-endian`, `windows`, `shape-gate`, `lint`, `vuln` and all six inline gates. `cpp-lock` passed on the earlier `pull_request` run and is not part of a dispatch.

Three CI-only corrections since the body above was written, all in this branch:

- **GCC's `-Wclass-memaccess` was right.** The C++ leg compared a filled row against a twin with `memcmp`, and zeroed the twin with `memset` — which GCC refuses on a struct carrying member initialisers. Value-initialising the twin does not answer it either: a generated table struct is an *aggregate*, so `= {}` initialises the members and leaves the PADDING unspecified, and a whole-row `memcmp` would be comparing bytes nobody defines. The comparison is field by field now, which is what the fill actually wrote; the block's padding is §19.1's business and the golden pins it separately. Clang never said a word about either half — a real cross-compiler catch.
- `gocritic`: the meaning projection's range chain is a `switch`.
- `modernize`: `max()` for the array-start alignment, `SplitSeq` in the test.

gafferongames:
**LAND-AFTER-FIXES.** The form is right, the layout contract holds across the boundary under an independent consumer, and the zero-cost property is real. One consumer-side safety hole and a cluster of page defects stand in the way.

Cold read, fresh clone, `block-form` @151f832, read-only. Everything below was run.

## What I reproduced

- **A C# consumer written from §19 alone** — my own `Sequential, Pack=1, Size=N` structs and derived offsets, no peek at `RenderBlock.cs` — opens the C++ producer's 1984-byte block and agrees with the generated side on all 16 checks: projection 176, `RenderShip` 88, triple 16, ships `320/4/88`, both `ReadOnlySpan` reinterprets over identical bytes, row digest `e49b9781bbbb0032` both ways, every field of every row.
- **Forgeries: 9 of 10 refused**, ASan+UBSan `-fno-sanitize-recover=all`, exact-size `posix_memalign` buffers so a redzone sits at the region edge. No report, no crash, no read past the region.
- **Build version reproduced independently** (Go, SHA-256 low 64 big-endian over the `--facts` text): `0x92b8887e8391446a` (block), `0x2500934549f29bb7` (examples), and from §20.2's printed projection typed by hand, `0x7402a36de22d9728` and `0x49947af3382f914e`. Grammar holds.
- **Fill path**: allocator called exactly once, in `Create`, never in `Begin`; zero global allocations across Begin+accessors+fill; 4 workers over disjoint ranges byte-identical to serial. Refuser, its control, the padding control, zero-cost and build-version gates green.
- **The zero-cost gate as §19 states it**, which the tree does not run: I built `origin/main`'s compiler (no Block emitter) and diffed its Table output against this branch's. **30 `*Table.h/.cpp/.cs` across examples, pointers and block: byte-identical, 0 differ.** The property is real; the shipped gate is a symbol grep, so §19/§19.5 should say so or the Makefile should do this.

## Defects, ranked

1. **`BlockOpen` accepts a count past the declared maximum — both backends.** Neither `Open` reads a `Max` constant. Reproduced: `ships.count = 5000` against `ShipsMax = 4096` opens in C++ and C#, `ShipsSpan.Length == 5000`. `Begin` refuses this on the producer side; the consumer does not. Memory-safe, but a consumer sizing anything by the maximum overflows. *Fix: `if ( count > <Field>Max ) return false;` per array, and add it to §19.2's "WHOLE check".*
2. **§20's status text is falsified by this PR and untouched by it.** `:4813`, `:4816` ("No backend emits `BuildVersion`, `schema build-version` does not exist"), `:4826` ("Neither backend asserts the compiler's layout model"), `:5111` ("Those asserts … are NOT in the tree today"). All four now false. *Fix: rewrite the status to what remains owed.*
3. **The prologue's shape is outside the digest for a unit with no out-of-line array.** A one-table unit gets `0x7bfee932381dd3aa` and a `ConfigBlock::Projection` of 32 bytes that is entirely the 3-word prologue — and no `block` line in its projection. This PR changed the prologue 2→3 words and left `ir/buildversion.go:35 BuildVersionForm = 1`. Two builds across that edit share an id and write incompatible blocks. *Fix: make the block form's shape a form-version fact and bump it.*
4. **§19's worked examples do not compile against what ships** — the peeks my from-the-page consumer needed. `:4569` spells `RenderFrame.BlockOpen(…, pointer, bytes)`; it is `RenderFrameBlock.Open(out …, IntPtr, long)`, and its bare `RenderShip` is a sealed *class* in that namespace, so the page binds the wrong type. `:4462` goes from `RenderFrameBlockStorage storage;` straight into `Begin`; the caller-provided allocator (`storage.Create(…)`) is an owner ruling §19 never states, so the page makes `Begin` return false with no reason. And **§19 never gives the magic's value.** USAGE was updated here; §19 was not.
5. **§19.5's two negative controls are absent, and one is inert.** Nothing perturbs the compiler's layout model to prove both backends go red. And "perturb one row type's pitch constant on one side only" cannot go red in C#: `ShipsStride` is emitted and never asserted (one grep hit, the declaration). *Fix: assert each `Stride` constant against `Unsafe.SizeOf` in `Verify`.*
6. **§19.5's race claim is not held by the sanitizer chosen.** The twin is `address,undefined`; there is no ThreadSanitizer in the tree. ASan finds no races, and byte-identity between a serial and a deterministic 4-worker fill finds none either. *Fix: a TSan build, or drop the sentence.*
7. **The fill refuser's plausibility floor is aggregate.** 2033 lines extracted, 811 of them `RenderBlock.h`'s; its markers could break entirely and the gate still passes at 1222. *Fix: floor per file.*
8. **§11's suffix list is short of the code.** `:2618` says **32** and lists 32; `check.go` claims 34 — `OpenValidated` (added here) and `LayoutId` unlisted. Reproduced: `type FrameOpenValidated` is refused by a rule the page does not state. *Fix: add both, 32 → 34.*
9. **§20.2's prose is wrong, not the golden.** `:5004` "Every record with a block form is followed by its PROJECTION" plus §19's "every FIXED table has one" demands a `block` line for `ShipConfig`; the page's own projection carries none, and I reproduced its digest from the printed text exactly. The number is the contract, so the reconciliation is right. *Fix: "every record whose block form lays an array out of line".*
10. Nits: §19.3 promises a BUILD ERROR on both sides; C# throws at first use — say so. `cpptable/block.go:417` emits "PROLOGUE of two uint64s" into every generated header (there are three), same stale comment in `ir/blocklayout.go:401` and `check.go:1755`. §19.1's "the worked layout below is unchanged" narrates a change a cold reader cannot resolve.

## Residues

`Open` trusts the caller's `bytes`: overstating your buffer opens (verified). That is §19.2 as written, but it is where "the WHOLE check" leans on the caller. The triple's `count` is `uint32` and the C++ accessors cast to `int32_t` — unreachable (a high-bit count needs a 137 GB extent), but present. C# unit primitives land in the alphabetically first Block file, so `RenderBlock.cs` will not compile alone; C++ guards them into every header. No fixture for the §11 prologue-name refusal, which §19.5 requires with a control — by hand it refuses and a rename clears it. Gate 2 is fair on both halves (golden gate first and refusing, interleaved, medians of 15, per-frame seed into every row, sink, Release on C#), but the C++ sink takes only the returned extent, so nothing proves the row stores survived — the shape of the 18x artifact found once already.

gafferongames:
**All eleven addressed, CI green.** Head `8c7c0c8`, rebased on `main` (f55e711 — #311's audit edits, including §19.1 taking the three-word prologue paragraph). Full CI: [run 33621908428](https://github.com/mas-bandwidth/schema/actions/runs/33621908428) — `test` on both OSes, `big-endian`, `windows`, `lint`, `shape-gate`, `vuln`, all six inline gates.

Thank you for the read. The tenth forgery was the one that mattered and I would not have found it: `Begin` refuses a count past the maximum and `Open` did not, so the two ends of one contract disagreed.

## The dogfood's two, first, because they blocked tonight's Unity test

Both from one root cause — the C# generator emitted per DECLARING FILE and skipped any file with no `table`:

- **the shared runtime went to the PROTOCOL ID's home**, which in a real unit is the constants file, so its Block.cs was never written and every reference to `Schema.TableBlockMagic`, `Schema.BuildVersion`, `TableBlockTriple`, `TableBlockRows<T>`, `TableBlockInfo`, `TableBlockLayout` was undefined;
- **a blittable record went to its declaring file's Block.cs**, which is never written when that file declares only `type`s.

Fixed at the root: the runtime and **every** blittable record of the unit are emitted once, into the **block home** — the first file that declares a table *with* a block form, which by construction gets a Block.cs. C# has no include guard and one assembly sees every file, so "emitted once, anywhere it exists" is the whole requirement; C++ takes the other road for its own reason (a consumer may include one header alone), and §19.2 now states the rule and both reasons.

`tables/blockhome` is a corpus unit with exactly that shape — `Constants.schema` sorts first and declares no table, `Data.schema` declares only `type`s and one of them is a block row's field, `Frame.schema` declares the table. `test/cs-block` compiles it, which *is* the gate, and `make tables-block-home-negative-control` puts the defect back through `go build -overlay` and requires the compile to fail on an undefined name.

## The nine

1. **`BlockOpen` refuses a count past the declared maximum, both backends.** Your ten forgeries ship as a battery on both legs — **10 of 10 refused**, under the ASan/UBSan twin as well, and the C# side runs the same ten against the same bytes. Forgery 9 (a count whose rows leave the extent) is kept *under* the maximum on purpose so the extent check is what refuses it and not row 10's.
2. **The prologue's shape is in the digest for every unit.** §20.2's header gains `block prologue=magic:8,build_version:8,byte_order:8`, unconditionally — named and widthed rather than counted, so it moves when the shape moves and there is no counter to forget. Re-derived: **`0xc211ce2f3414aa7c`** and **`0xe2eeb510ec9621cb`**, page text and every corpus golden re-pinned. And §20.2's prose is fixed your way — the golden is right: a `block` line is emitted for a record whose form *lays an array out of line*.
3. **§20's status at all four sites** now says what is true and what remains owed: the id and its projection are live; the asserts exist for the block closure and nowhere else; the C# check throws at first use rather than failing the build; the cook has neither constant nor `cook-check`.
4. **§19's examples compile against what ships**: `RenderFrameBlock.Open( out …, IntPtr, long )`, the rows are structs in the unit's `.Block` namespace and never the sealed classes of the same names, the allocator is on the page with its signature and "exactly once, at build time", and the magic's value is stated.
5. **Both named negative controls ship**, and the C# `Stride` constants are asserted against `Unsafe.SizeOf` so the first is not inert. A pitch perturbed on one side turns the C# check red; a field's offset moved in the **compiler's** model turns both backends red.
6. **A ThreadSanitizer leg**, under a second, beside the other two — and its own control. That control was flaky first time out: overlapping *ranges* is a real race but whether TSan observes one depends on scheduling, and it fired here and not on CI's macOS runner. It now has every worker write **one row**, after a start barrier, twenty thousand times. Same defect, made certain. §19.5 names TSan as the leg that holds the claim.
7. **The fill refuser's floor is per file and anchored** to the `BlockBegin` functions that file declares, with markers balanced. Verified: blanking `RenderBlock.h`'s markers now goes red.
8. **`OpenValidated` is out.** The two lists are reconciled mechanically at 26 + 7, and §11's "the checker does not claim them yet" is gone — it does. The two-per-array accessor rule is on the page now too.
9. **The zero-cost gate does what you did.** 38 Table sources pinned from the **pre-block compiler at f55e711** and byte-compared, so the identity is measured against a build that could not emit a Block file at all. The symbol grep stays beside it.

Nits: §19.3 says plainly that C# throws at first use; the stale "PROLOGUE of two uint64s" in three files; §19.1's narration removed. Your residue about gate 2's sink: it now digests four rows of two sections, so a dead-store elimination cannot hide behind an extent that did not move.

Two page notes from the dogfood: **§19.1 — `Begin` does not clamp, it refuses** (a producer that gathered more rows than a maximum clamps at the call site; nothing downstream will), and **USAGE — `Open`'s 64-byte alignment binds the consumer**, and a pinned managed `byte[]` does not satisfy it.

And a build defect this work tripped over: `GO_SOURCES` globbed `cmd internal` only, so an edit to `ir/` or `compiler/` did **not** rebuild `bin/schema` and every gate below then measured a stale compiler.

## Numbers

| | |
|---|---|
| build version, `tables/block` | `0xb34bc386659d9873` |
| build version, `tables/blockhome` | `0x831c4abc9b71036d` |
| gate 2, C++ write (3 sittings) | **0.893 / 0.899 / 0.896** |
| gate 2, C# read, Release (3 sittings) | **0.998 / 0.998 / 1.000** |

The dogfood's own measurement of the generated form against its hand walk lands in the same place: **0.85**.

Still not merged.

gafferongames:
**LAND.** All eleven verified at 8c7c0c8; nothing I raised is still open, and the two fixes with the most room to go wrong went right. Only the delta follows.

- **Forgeries 10/10 refused** — C++ under ASan+UBSan `-fno-sanitize-recover=all`, C# through the generated `Open`. Count > max is refused symmetrically with `Begin` on both backends, and it does not over-refuse: a block claiming exactly `ShipsMax` = 4096 still opens, `ShipsSpan.Length == 4096`.
- **The prologue is in the digest.** Re-derived §20.2's numbers from the page's own text with the new fourth header line: `0xc211ce2f3414aa7c` and `0xe2eeb510ec9621cb`, both exact. The hole is closed where I found it — the one-table unit with no out-of-line array now moves on a prologue edit: `0xf38b669eadf50b28` → `0x20db9e4259092ee3` dropping `byte_order`, → `0xfdf64cdef27bf3d2` narrowing it to 4.
- **Block home**: two units built to break §19.2. One puts the home in `C_First` with a *second* block table in `D_Second`, a row type shared by both, and a row type declared in `E_Late` — a file sorting *after* the home; the other's only block table is in the last file, behind a constants file and a types file. Both generate, compile and run under `TreatWarningsAsErrors` + `AllowUnsafeBlocks`: every record lands in the home, `D_SecondBlock.cs` references them without redeclaring.
- **Race control red 3/3**, deterministically — write-write on one row after a start barrier, not a scheduling-dependent range overlap. Clean TSan leg green 3/3.
- **Zero-cost pins**: f55e711 is on `main` with no `ir/blocklayout.go` and no `cpptable/block.go`. I built its compiler and regenerated the *current* schemas: **38/38 pinned Table sources byte-identical, 0 missing.** The pins are the pre-block compiler's output, not this branch's.

One departure from the report, and the better call: `OpenValidated` was dropped from `tableGeneratedVerbs` rather than added to §11. No emitter spells it anywhere, so the claim guarded nothing; §11's 26 + 7 now equals the checker's 33 exactly, and the page states that it must.

Also green here: fill refuser (floor now per file) and its control, padding control, build-version gate at `0xb34bc386659d9873`, the C++/ASan/C# legs, `go test`.

Issue #310, opened 2026-09-02
check: the `Root` suffix claim outlives what it guarded — no emitter spells `<X>Root`


gafferongames:
`tableGeneratedVerbs` (internal/check/check.go) claims `<X>Root` for every closure member, and no backend emits that name: the builder's accessor is the member `GetRoot` (internal/codegen/cpptable/pointers.go), renamed so `table Root` — SPEC-TABLES.md §11's own canonical example — stays legal. Drop the claim, with SPEC-TABLES.md §11's list moving in the same change so the page and the checker never disagree. Found by the audit on #248 (A7).

PR #311, opened 2026-09-02
SPEC: the audit's edits — honesty notes, the third class, refused pointer arrays, counts, the stride cut, `*bytes` to follow-on (#248)

gafferongames:
The audit on #248, applied to the pages. One commit per item; nothing deployed, no emitter touched (one code COMMENT, a stale count).

1. **§7 + §6.3 honesty notes** — §7 gains a `Backend status` paragraph: the C++ backend emits wire v1's cook, where `Open` matches the header and then WALKS the region behind a monotonic high-water mark (the traversal §14 rejects by name), the header carries one 32-bit part length, there is no attribution part and no node directory, and `schema cook-check` is not built; the emitter moves to §7 with #251. Beside it the owner's scale requirement is stated as a requirement — open O(1) in file size, the byte order settled at cook time, pages touched only as used — with the 1 GB gate marked **not built**. §6.3 gains its own note: no backend writes the node directory.
2. **§4.1's "there is no third"** — the silent class is **three**: default, flags, and a field's REFERENT dropped or swapped for one that cannot stand in (both sides kind `7`, nothing to report), which is the class §18 exists for. §18's opening follows.
3. **Arrays of table pointers** — the compiler refuses them; the page said legal in four places. All four now say REFUSED, and §15 gains the follow-on the diagnostic cites, carrying its two spellings verbatim ("declare a bounded array of tables by value, or a pointer to a table that holds the array").
4. **"Renaming a `type` field is free"** — false: SPEC.md §3.1 puts field NAMES in the projection, so a `type` rename moves the protocol id and buys a lockstep redeploy. Fixed at both sites (SPEC-TABLES §5, SPEC.md §4.2). The two "what moves the id" lists (SPEC.md §3.1's consequences, FAQ) were audited against §3.1 and are correct as they stand.
5. **The counts** — the checker's per-member set, `tableGeneratedVerbs` in `internal/check/check.go`, is **26**, counted on main:

   ```
   Measure  MeasureBody  Save  SaveBody  Load  LoadBody
   LoadMeasure  LoadMeasureBody  LoadBuilder  TableType  Builder
   At  Root  Emplace  Pack  PackMeasure  OpenWalk
   Cook  CookMeasure  Open  LayoutId  TableFields  TableInfo
   FromJson  ToJson  ToJsonMeasure
   ```

   §11's printed list is now that list exactly (it had gained the seven block spellings and lost `LayoutId`), §8's "27" and §11's "36" both become 26, and `internal/tablenames`'s "23 name-first suffixes" comment with them. The block form's seven — `Block BlockStorage BlockBegin BlockBytes BlockMaxBytes BlockOpen Counts` — are stated separately as law the checker does not claim yet, owed with the form (#287) and owed BEFORE its emitters.
6. **`Root`** — the checker still reserves it, so the page keeps it rather than freeing a name ahead of the compiler; it is named on the page as a claim that guards nothing (no emitter spells `<X>Root`; the accessor is the member `GetRoot`) and filed as **#310**. `LayoutId` is named the same way, retiring with #251.
7. **`section`** — not live vocabulary since §13.6 collapsed the header into the table. The surviving uses are the optional's, and they read "group" now (SPEC-TABLES §2.1, §3, §12; USAGE). §14's rejected `section` CONSTRUCT and §13.6's verbatim keep the word, as the names of what they name.
8. **§2.5 `*bytes`** — cut to a §15 follow-on citing #259, keeping the reasoning in the owner's words ("yes, i like byte buffer as primitive" / "since it can be nulled." / "instantly, it turns a table into variable"). The spelling does not parse, so every statement of it as live law goes with it: §2's overview bullet, §2.1's pointer target, §2.2's mode rule, §3's framing family and presence rule, §3.1's pointer-edge paragraph and its 4 GiB example, §4.1's and §18.2's edit lists, §6.3's buffer extents, §7's and §11's cook-check refusals, §8.1's used-length column, §16.2's two rows, SPEC.md's EBNF production, and USAGE's Byte buffers section.
9. **The declared stride is CUT** ("OK got it. We cut it.") — the §15 headroom follow-on is gone, §11's `| stride` refusal-by-name is gone (an unknown attribute is refused as unknown), §19's pitch bullet loses "in this version": the pitch is `sizeof` rounded to alignment, derived, always, and it rides in the triple. §13.6 records the ruling with the hard requirement it is held to ("I mean as long as we can make it work with the fast blitted C# stuff." / "that's the hard requirement. so if we don't need stride anymore then c'est la vie."), and §14's decision reads as a cut rather than a deferral.
10. **USAGE's cook section** — reads against §7 as it stands: the whole enumeration `Open` checks, open time flat in the file's size, the tool for an untrusted file, and a fixed table's own cook (it said value-only tables get no `Cook`/`Open`, which contradicted §7 and §11). A one-paragraph note says what the emitter does today. USAGE:1424's "`--lang cpp` only today" becomes C++ and C#; README:112 already reads "(C++ and C# today)" from #308 and is untouched.
11. **`space#443`** — a private cross-repo number on public pages, replaced by the finding itself: 803 values byte-identical, every injected bit flip refused rather than read as data. Three sites in USECASES and, for the same defect, §12's gate-1 paragraph.
12. **One table for the three frames** — §4.1 gains it: per edit, what the read report says, what the baseline does, and whether the build version moves. §18's opening and §20.1 cite it instead of restating their own lists.
13. Nothing owed — #297 landed the compiler-settled build version and left no "after a second backend" wording anywhere.
14. **§20.1's flags `none` row** — the stronger basis is on the page: SPEC.md §4.2 makes flags storage `uint64` in every target and a flags field carries no specified default, so a slot is a raw `u64` copied verbatim and no declaration-side flags fact can reach a cook's bytes.
15. **§13 gains the 2026-09-02 verbatims** — the aim at the head of the section ("We aim to build the best cross-language data type system for games." / "This includes save games, tooling, cooking to runtime efficient structures and so on."), the cook in §13.4 ("We do cook. We just aren't using it in space game right now." / "We can keep the gigabyte scale stuff for v2, but think about it as we work now."), and the stride cut in §13.6 with item 9. The aim is also USECASES.md's first sentence, quoted.

**Gates.** `make check` green; `go build ./...` and `go test ./...` green. Every §-citation on every page resolves in the document it names — 906 of them, by a Go scratch checker written for this pass and deleted after. No `previously|used to|no longer|was ` in any added line outside §13/§14.

**Judgments made rather than followed**, each one flagged for the owner:

- **Item 5, the number.** The brief said make every page number equal the checker's. The page and the checker do not describe one set: the checker claims 26 and the page claimed 32, the difference being the seven block spellings the checker does not claim and `LayoutId`, which it does. Making the page say 26 flat would have deleted law that ought to be there — a name free today must not become a collision the day the block form lands. So the printed set is the checker's 26 exactly, every count on the page equals it, and the block form's seven are stated separately as owed with #287 rather than counted as claimed.
- **Item 6, `Root`.** The checker reserves it, so the page keeps it and #310 carries the removal, page and checker moving together — the brief's own fallback.
- **Item 8, the §2.5 number.** §2.5's law is gone and its heading is not: renumbering §2.6 and §2.7 would move 25 citations on four pages and every reference outside the tree. The slot holds three sentences saying the spelling does not parse and naming §15.
- **Item 7, the replacement word.** The brief said "array" / "the triple"; both fit the block-form senses, and the surviving uses are the optional's — an optional GROUP, which is the page's own word for a guarded group two paragraphs away.
- **Item 11, the fourth site.** §12's gate-1 paragraph cites `space#443` too, and the audit (C10) named it. Fixed with the three in USECASES so no private issue number is left on a public page.
- **The `tablenames` comment.** A code comment carrying the stale count 23 and citing §11 — a two-word fix inside the doc pass rather than a fifth stale number left behind.
- **Three follow-through commits** beyond the fifteen items: item 8's blast radius across the page, one present-state phrasing fix the gate caught, and one reflow pass over the paragraphs the edits disturbed.


gafferongames:
**LAND-AFTER-FIXES.** All fifteen edits are present and, as described, correct against the tree. Four defects — three of reach, one of overreach; all one-commit fixes, and two are on USAGE, the page a user reads.

## Defects, ranked

1. **USAGE.md:1265 — item 2 stopped at SPEC-TABLES.** "**Exactly two edits are different**" survives, and USAGE's own refuse list sixty lines down (`:1332`) names the third and says why it is silent — "*both ride as kind 7*". The page contradicts itself in one section. *Fix: "Exactly three", plus the referent clause, mirroring §4.1.*
2. **USAGE.md:1190 — item 10 replaced a sentence true of the tree with one false of it.** "*A FIXED table cooks too… memcpy it or point at it where it lies*" is §7's law; the emitter gives a fixed table **no `Cook`, no `Open`, no `LayoutId`** — in a fixed-only unit and a pointered one alike (`Config` gets only `ConfigOpenWalk`). The status note beside it names the walk, the 32-bit length and the missing attribution, not this. §7's four-way status (`SPEC-TABLES.md:1692`) has the same hole. *Fix: a fifth clause in both notes — no fixed table has a cook today.*
3. **SPEC-TABLES.md:1381 — one cell of item 12's table is wrong.** "a declaration or field RENAMED with no `was`, on a `type`" → baseline "not covered". A `type` in a table closure IS projected; renaming it warns, live: `warning: tables.baseline: table Vec: no longer in the closure under that name; Vector carries 2 of its 2 identities…`; renaming its field passes in silence. *Fix: "warns when the `type` is in a table closure (§18.3)".*
4. **SPEC-TABLES.md:2688 — item 5 fixed the list and left the paragraph.** §11 still claims `<Table><Field>` and `<Table><Field>Span` per out-of-line array as law; `grep -n Span internal/check/check.go` returns nothing. The class item 5 closed for the seven. *Fix: same "owed with #287" sentence.*

Minor: §11's "`LayoutId` … (§7's status)" cites a status that never mentions an id; §20's does (obligation 3).

## Against #309

- **Prologue — the page is wrong.** §19.1 says "*a generated PROLOGUE of two `uint64`s*", worked at "*the 16-byte prologue*" / 168. #309 emits three — `magic`@0, `build_version`@8, `byte_order`@16 (`block.go:419-421`, asserted at `:484-486`) — projection 176. Ruling 5 kept byte order a prologue field; §19.1's own "*the prologue is free in this shape*" already absorbs 176→192.
- **§11's prologue claim — the page is wrong, and it is a live hole.** Page refuses `magic` or `build_version`; #309 refuses those **and `byte_order`** (`check.go:1765`). A user may declare `byte_order` on a fixed table today.
- **The seven — agree, spelling for spelling.** But #309 also adds **`OpenValidated`** to `tableGeneratedVerbs`; no emitter spells it and no page mentions it. **The code is wrong** — the `Root`/#310 defect this PR files, against #309's own rule that "a claim nothing needs takes a name away for free".
- Per-array accessors, and pitch derived / no `| stride`: agree. Live: `table T | stride` does not parse; a field `| stride = 64` → *unknown attribute "stride" … (SPEC §4.2)*, which is what §11 now says.

## Verified, with numbers

`internal/check/check.go:2661` holds **26**; §11's block is that list character for character, and §8:2048, §11:2609, §11:2661 and `tablenames.go:91` all equal it — no 23/27/32/36 anywhere. Item 1: `Open`→`OpenWalk` (`pointers.go:803`), one 32-bit length (`:764`), no attribution or directory in `cpptable`/`cstable`, no `cook-check` in `cmd/` or `internal/`. Item 3: `[..8]*Node` and `[8]*Node` both refused live, the diagnostic matching §15's quote verbatim. Item 14: flags storage is `uint64` in every target (SPEC.md:456), and a flags default is refused live. Item 11: `grep -rn 'space#'` over every `.md` — empty. Citations: a Go scratch resolved **940** `§N.M` references across the five pages to a heading in the document each names — **0 misses**. Present state: added lines outside §13/§14 matching `\b(previously|used to|no longer|was|now)\b` → 4 hits, every one the `was` ATTRIBUTE in backticks. The aim is verbatim on both pages, "build" not "make", both sentences; the stride cut and "*c'est la vie*" match #248 word for word; the scale verbatims match #251. `make check` and `go test ./...` green here.

## Residues

- **"We do cook. We just aren't using it in space game right now." (§13.4) is on no comment in this repo** — I searched every issue and PR comment since 2026-08-01. Every other verbatim in the pass resolves to one I can name.
- §2.3:398 still reads "*one framing with three spellings*" (USAGE:801 likewise) while §3:702 now says two — audit B3, outside the fifteen.
- **Item 8's stub: sound.** Renumbering would move every §2.6/§2.7 citation on four pages and outside the tree, and a signpost where a reader goes looking for `*bytes` is worth more than tidy numbering. The heading should say what the slot now says, though, so it stops advertising a construct as a section.

gafferongames:
All four defects fixed, the minor with them, and the prologue now matches #309. One commit: `71eb1fc`, rebased on `main` at `49bbdc7`.

1. **USAGE:1265** — "Exactly **three** edits are different", with the third named where the reader meets it: a field's referent dropped or swapped for one that cannot stand in, riding under the same kind either way, citing §4.1. The refuse list sixty lines down now agrees with the sentence above it.
2. **USAGE:1190 and §7's note** — both say it. USAGE: *"the emitter gives `Cook`, `CookMeasure` and `Open` to VARIABLE-LENGTH tables only, and only in a unit that declares one, so a fixed table has no cook surface today"*; §7's status goes from four ways to **five**, with the same clause. Verified as you described: `schema generate --lang cpp tables/examples` emits no `Cook`, no `Open`, no `LayoutId` at all, and in `tables/pointers` the fixed members get `<X>OpenWalk` and nothing else.
3. **§4.1's cell** — split in two, and each half is what the tool actually prints. A DECLARATION renamed **warns** when a table closure reaches it, naming what carries its contents on and how many identities that candidate carries (§18.3) — reproduced live: *"table Vec: no longer in the closure under that name; Vector carries 2 of its 2 identities, so it is judged as the rename it looks like"*. A `type`'s FIELD renamed passes the baseline in silence (exit 0), and the read report counts it `unknown` on the table wire, whose field id is the name's hash — so the row also says it moves the protocol id, which is item 4's fact.
4. **§11's per-array accessors** — moved into the owed list: *"and this claim is owed with the block form too, on the same terms as the seven above (schema#287): the checker makes it today for neither."*
5. **Minor** — `LayoutId` now cites §20's status, obligation 3 by name (a 32-bit id where the build version is 64), which is the status that mentions an id.

**Against #309**, taking your sentences rather than writing my own, so the merge is textual:

- §19.1's prologue paragraph is #309's verbatim — three `uint64`s, `magic` / `build_version` / `byte_order` (`1` little, `2` big), the "what each of the three does" paragraph, and the "free in this shape either way" note; the worked layout is 176 / 24-byte prologue, and "152, 168 and 176 all round to 192".
- §11 claims all three names, in #309's phrasing.
- One thing I added on top, because your own defect 4 rules it: **the checker claims none of the three today** — verified live, a table declaring `magic`, `build_version` and `byte_order` checks clean on main — so the sentence says the refusal lands with the form and its emitters (#287), the same shape as the seven block spellings and the per-array accessors. If #309 lands first the sentence is one line to drop.
- Not taken, because they are #309's to land and false of main: the ladder's "the BLOCK FORM is live in C++ and C#", §19.2's C# surface, and USAGE's block section. This branch still describes main.

Noted and not acted on, since both are outside the fifteen and outside your four:

- **`OpenValidated` in #309's `tableGeneratedVerbs`** — agreed, and it is the `Root` defect exactly: a claim nothing emits, taking a name for free. schema#310 is filed for `Root`; this belongs in the same change or in #309 before it lands.
- **§2.3's "one framing with three spellings"** (and USAGE:801) against §3's two — audit B3, unclaimed.
- **"We do cook. We just aren't using it in space game right now."** — you are right that it is on no comment in this repo. It came to me in the dispatch brief's addendum as an owner verbatim of the 2026-09-02 evening, alongside the scale sentence that does resolve (#251). If it cannot be sourced, §13.4's bullet should lose the quotation marks and keep the fact.
- **The §2.5 heading** — agreed; it reads `### 2.5 Byte buffers` now, and its three sentences say the slot is a signpost. Renaming it further is a word, not a decision, and I left it where a reader looking for `*bytes` will still find it.

`make check`, `go test ./...` and the citation gate (912 references, 0 misses) green here.

PR #316, opened 2026-09-02
CI: the windows job never compiles C++ — the estate's Visual C++ hard requirement is ungated

### LINKED ISSUE #286: CI: the windows job never compiles C++ — the estate's Visual C++ hard requirement is ungated
Found by the #283/#285 worker while reporting: the `windows` CI job runs `go vet` and `go build` only; it never compiles a generated header or the tables/JSON legs with MSVC. The estate's standing rule (Visual C++ is a hard requirement: emulate extensions, never require them; serialize.h holds the reference pair) is therefore not gated by CI on any PR to date. Add an MSVC leg that compiles the generated C++ corpus (packet headers, the tables leg incl. `<Base>Table.cpp`, the JSON batteries) with `/W4 /WX` (or the repo's chosen flags) on a windows runner; negative control: introduce a GCC-only extension in a generated header in a throwaway and show the leg red. Until it lands, every C++ claim of Windows coverage in PR bodies is unverified — note it in CLAUDE.md's CI description so nobody repeats the mistake.

gafferongames:
Draft while the leg is measured against the two-minute law. Closes #286 when it lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -400,11 +400,30 @@ jobs:
         run: go run golang.org/x/vuln/cmd/govulncheck@latest ./...
 
   # The compiler that generates C++ for a Windows-targeting game engine was
-  # never itself run on Windows. `make test` needs the six sibling runtime
-  # clones and a POSIX toolchain, so vet + build is the cheap first step the
-  # issue prescribes; growing this into a real test leg is a follow-up.
+  # never itself run on Windows, and until the MSVC leg below nothing in this
+  # file compiled a single line of what it EMITS with cl (issue #286). vet +
+  # build prove the COMPILER runs there; the MSVC leg proves its OUTPUT
+  # builds there, which is what the estate's standing rule actually asks for:
+  # Visual C++ is a hard requirement, emulate extensions, never require them.
+  #
+  # The leg is the TABLES corpus plus every packet header it emits beside it,
+  # not `make test` — `make test` needs the six sibling runtime clones, the
+  # three self-contained targets' toolchains and a POSIX toolchain, and none
+  # of that is what this job is for. It needs one sibling (serialize, header
+  # only, and only the packet headers include it) and cl.
+  #
+  # THE PIN. The runner image is NAMED rather than `-latest`, the argument
+  # the big-endian leg above makes: which extensions cl accepts is exactly a
+  # toolchain-version fact, so the image it comes from has to be a decision
+  # rather than whatever `-latest` moved to overnight. The step after the
+  # compile records what the pin resolved to.
+  #
+  # Its NEGATIVE CONTROL is the point of the leg, not decoration: a GNU
+  # statement expression is planted in one generated header, and the leg must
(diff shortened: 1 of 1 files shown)


gafferongames:
Closing without merging, per the measurement below — the leg is correct and the corpus is not ready for it.

**The leg works.** Its negative control is RED as required: a GNU statement expression planted in one generated header, which every POSIX leg in this file accepts silently under `-Wall -Wextra -Werror`, is refused by cl with C2059/C2143. So a green from this leg would mean something.

**The corpus does not compile under MSVC**, and that is the finding #286 was looking for. cl 19.51.36256 x64 (VS 2026, `windows-2025` image), `/W4 /WX /std:c++17 /permissive- /EHsc`, one TU at a time:

| translation unit | result | cl time |
|---|---|---|
| `jsonkeys/JsonKeysTable.cpp` | pass | 1.0 s |
| `v1/V1Table.cpp` | **C4456 → C2220**, `V1Table.h(922)`: declaration of `element_id` hides previous local declaration (see `V1Table.h(915)`) | 5.3 s |
| `v1/_AllHeaders.cpp` | same C4456 | 0.6 s |
| `block/RenderTable.cpp` | **fatal error C1060: compiler is out of heap space**, `RenderTable.h(3900)` | **116 s**, then dies |
| the negative control | red, as required (C2059) | 0.3 s |

**Two defects, neither of them a CI problem.**

1. **C4456, shadowed local.** `element_id` is declared twice in one scope in the emitted table reader. gcc and clang do not warn on this without `-Wshadow`, which is why every POSIX leg is green on it. It lives in `internal/codegen/cpptable` — outside `bench/LOCK` — so it is fixable.
2. **C1060, out of heap space on `tables/block`.** `blockdemo::RenderFrame` is 7 879 320 bytes (arrays of 32000 / 20000 / 8192 / 4096 inline records). Clang compiles that same TU in **0.15 s**; cl spends 116 s and exhausts its heap. The suspects are in `internal/codegen/cpptable/codecs.go`: `sizeof( RenderFrame{}.lasers[0] )` (codecs.go:1188-1195) materialises a prvalue of a 7.9 MB aggregate in an unevaluated operand, and `+[]( void * p ) { new ( p ) RenderFrame{}; }` (codecs.go:1106) value-initialises one. `sizeof( decltype( RenderFrame::lasers[0] ) )` needs no object at all.

**And the two-minute law is not close.** `block/RenderTable.cpp` alone costs 116 s — the whole budget, on one of the corpus's 33 generated `.cpp` files, and it fails at the end of it. Job setup through the start of cl is 47 s (checkout, setup-go, vet, build, one serialize clone, generation). A first attempt with `/MP` across the whole corpus ran 3 m 13 s and finished no unit.

So the leg cannot land until those two are fixed, and adding it with `block` excluded would be a green light over the one unit that actually breaks. The branch `msvc-leg` is parked with the working leg on it — vcvars via vswhere, per-unit `/I`, an `_AllHeaders.cpp` per unit so the packet headers and the all-variable unit are built too, and the negative control — so whoever fixes the emitter has the gate ready to turn on.

Issue #133, opened 2026-08-25
Split the CI gate: fast PR leg for iteration, full matrix certifies on main


rowan-claude:
Ruled by Glenn, 2026-08-25: "I think #2 split the gate makes sense." His calibration, verbatim: "It's reasonably slow, and for good reason. but for iteration, should be faster." His size class: "it's a mini-durr".

The split: the PR-triggered workflow trims to a fast iteration gate (Go tests plus one backend's conformance on one platform, about 2 minutes); the FULL five-language certification matrix keeps running on push to main, which it already does today as the post-merge run. So the change is one workflow edit removing matrix legs from the PR trigger, not new machinery. Iteration instruments and certification instruments, separated per the standing doctrine.

Measured basis (2026-08-25, the six-feature campaign): full matrix 8m20s to 10m35s per run, roughly half of every feature PR's open-to-merge wall time. The fast leg discriminates nearly all real failures; cross-language divergence that only a matrix catches still fails loudly on main minutes later.

Sequencing: after the six-feature campaign merges (do not change CI under its in-flight PRs). Not covered by the release hold, which governs releases, not CI config.

rowan-claude:
Enacted and measured, 2026-08-25: the split landed as PR #135 (fast PR leg; certification fans out on push-to-main + nightly 09:17 UTC + workflow_dispatch) and PR #138 (PR conformance on ubuntu only, immune to macOS pool contention). Measured result: PR gate 8:20-10:43 before, 1:16-1:45 after (final shape 1:20); certification 5:41-7:44, never cancelled (the cancel-in-progress hazard on main was found and fixed in the same work). The 1-2 minute law is met with margin. Closing with receipts.

PR #319, opened 2026-09-02
CI: an msvc leg on the certification side — cl /W4 /WX over the generated corpus (#286)

### LINKED ISSUE #286: CI: the windows job never compiles C++ — the estate's Visual C++ hard requirement is ungated

gafferongames:
Follows #317, which fixed the two emitter defects that made this leg impossible. The corpus now compiles clean under cl; this wires the gate that keeps it that way.

## Where it runs, and why there

`if: github.event_name != 'pull_request'` — **push to main, the nightly schedule, and `workflow_dispatch`**. Same condition and the same reasoning as the inline-budget gates above it, and it is a measurement, not a preference.

Whole job, four runs: **2:27, 2:28, 2:26** — and **one at 5:14**, on input byte-identical to a 2:27. One run in four at 2.1x the median is a flake in the PR gate, which is worse than a slow job. At 5:14 it was also the longest job in that gate (test-ubuntu 3:23, big-endian 1:36). On push-to-main it costs nothing anybody waits for, and a Windows regression fails loudly minutes after merge — the bargain #133's gate split already struck for the inline gates.

## A correction to #317's numbers

#317's body said `block/RenderTable.cpp` was ~92% of the cost. **That was inferred from `/MP`'s scheduling order and it is wrong.** Compiling one TU at a time says so ([run 33640076155](https://github.com/mas-bandwidth/schema/actions/runs/33640076155)):

| translation unit | cl (two runs) | clang, this bench |
|---|---|---|
| `block/PaddedBlock.cpp` | 23.0 s / 22.0 s | 0.18 s |
| `block/PaddedTable.cpp` | 19.1 s / 19.4 s | 0.14 s |
| `block/RenderBlock.cpp` | 19.2 s / 19.4 s | 0.13 s |
| `block/RenderTable.cpp` | 19.2 s / 19.3 s | 0.15 s |
| `block/_AllHeaders.cpp` | 0.2 s | — |
| the other nine units, 39 TUs | 9.1 s total | — |
| **corpus** | **89.7 s** | |

The block unit is 80.7 s of 89.7 s, spread almost evenly across **four** translation units — not one pathological function, and reproducible to within a second across two runs, which also says the 5:14 outlier was the whole runner rather than anything in the tree. Every one of them includes `RenderTable.h` (`PaddedTable.h` and `RenderBlock.h` include it directly, `PaddedBlock.h` through `PaddedTable.h`), and that header defines `blockdemo::RenderFrame`, a **7 879 320-byte aggregate**. cl charges ~20 s per translation unit for having that type in scope. clang charges ~0.15 s. Roughly **130x, uniform**.

The mechanism was right; the shape of the number was not, and the instrument that corrected it is the one this job now carries.

## What the job is

- **Named `msvc`**, separate from `windows`. `windows` keeps riding every PR with vet + build (the compiler runs on Windows); `msvc` proves what it **emits** builds there.
- **Pinned image, stated**: `windows-2025`, Visual Studio 2026, cl 19.51.36256 — the argument the big-endian leg makes, since which extensions cl accepts is a toolchain-version fact. The compile step prints what the pin resolved to on the day.
- **Flags**: `/W4 /WX /std:c++17 /permissive- /EHsc`.
- **Corpus**: the ten tables units plus every packet header emitted beside them. Each unit gets an `_AllHeaders.cpp` covering the headers no `.cpp` covers — the packet headers, and `P2.schema`, whose tables are all variable-length so it emits no `.cpp` at all. A header that *has* a `.cpp` is left out of it; on the block unit, compiling it twice cost more than the whole rest of the corpus.
- **One `cl` per translation unit, each with a clock line. No `/MP`.** Parallel TUs are faster and unmeasurable. Dropping `/MP` cost 89.7 s against 81 s and bought the correction above; any future argument about this job's cost gets settled from its log.
- **Negative control in its own step, asserted.** A GNU statement expression planted in one generated header of a copied tree — every POSIX leg in this file takes it silently under `-Wall -Wextra -Werror`. cl must refuse it; if cl accepts, the step goes red saying why that matters. The control's *failure* is the step's success, so the step ends on an explicit `exit /b 0` — `cmd` would otherwise hand the job the exit code of the compile that was supposed to fail (that exact bug ate one measurement run).

## Documentation

`CONTRIBUTING.md`'s "The gates a change has to pass" gains the present-state line: which two gates are not in the PR leg, where they do run, and the measured reason for `msvc` — plus the consequence for a contributor, which is the point of writing it down. **A change to the C++ emitters wants `gh workflow run ci.yml --ref <branch>` before merging, because a green pull request has not compiled anything with cl.** The `ci.yml` header comment records the same split.

## CI on this PR

- **The `pull_request` run does not exercise this leg** — that is the design. `msvc` shows as **skipped**. Everything else green, proving the workflow still parses and nothing else moved.
- **The leg itself ran by `workflow_dispatch` on this branch, twice, both green:** [run 33640076155](https://github.com/mas-bandwidth/schema/actions/runs/33640076155) (`msvc` 2:28) and [run 33641197840](https://github.com/mas-bandwidth/schema/actions/runs/33641197840) (`msvc` 2:26, on the current head). Whole corpus clean under `/W4 /WX`, negative control red as required in both (`cl refused the planted extension`), per-TU timings in the logs. The tables above are from those runs.

Closes #286.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,16 @@ name: CI
 # A cross-compiler inline regression a PR cannot see still fails loudly on
 # main minutes after merge, and nightly re-certifies a quiet main.
 #
+# The `msvc` leg sits on the certification side for the same reason and by the
+# same measurement (#286, #317): the job cost 2:27, 2:28 and 2:26 across three
+# runs and 5:14 on a fourth, with byte-identical input — the longest job in the
+# PR gate at its slow end, and a flake at any end. Its job comment carries the
+# per-translation-unit numbers and where they come from.
+#
 # For a full run on a branch before merging (e.g. a change near the hot
-# paths): gh workflow run ci.yml --ref <branch>.
+# paths, or anything that moves the C++ emitters): gh workflow run ci.yml
+# --ref <branch>. That is also the only way to exercise the msvc leg from a
+# pull request, since it does not run on the pull_request event.
 
 on:
   push:
@@ -400,11 +408,10 @@ jobs:
       - name: govulncheck
         run: go run golang.org/x/vuln/cmd/govulncheck@latest ./...
 
-  # The compiler that generates C++ for a Windows-targeting game engine was
-  # never itself run on Windows. `make test` needs the six sibling runtime
-  # clones, the three self-contained targets' toolchains and a POSIX
-  # toolchain, so vet + build is the cheap first step the issue prescribes;
-  # growing this into a real test leg is a follow-up.
+  # The compiler that generates C++ for a Windows-targeting game engine, run on
+  # Windows: vet + build, cheap enough to ride every pull request. What it
+  # EMITS is compiled with cl by the `msvc` job below, which runs on the
+  # certification side.
   windows:
     runs-on: windows-latest
     timeout-minutes: 15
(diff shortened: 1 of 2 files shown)


gafferongames:
Dispatch runs that actually exercise the leg (the `pull_request` run cannot — `msvc` skips there by design):

- [run 33640076155](https://github.com/mas-bandwidth/schema/actions/runs/33640076155) — `msvc` green, 2:28
- [run 33641197840](https://github.com/mas-bandwidth/schema/actions/runs/33641197840) — `msvc` green, 2:26, on the current head

Both: whole corpus clean under `cl /W4 /WX /std:c++17 /permissive- /EHsc` (19.51.36256, `windows-2025`), negative control red as required, per-TU clock lines in the log. The head commit since then is comment-only in `ci.yml` — verified with `git diff <dispatched> HEAD -- .github/workflows/ci.yml` filtered to non-comment lines, which is empty.

One note on the `pull_request` run: `big-endian` was cancelled at the 15-minute timeout on its first attempt, inside the `apt-get` step that installs the s390x cross toolchain and qemu — before it compiled anything. Re-run, green in 1:30. Runner flake, not the tree.

PR #321, opened 2026-09-02
wire v2 (1/3): schema cook, cook-check, uncook — the cooked form from §7, and §7 completed to the byte

gafferongames:
The first of the wire v2 sprint's three. `schema cook`, `schema cook-check` and `schema uncook` produce, validate and read back §7's cooked form, in both byte orders — and §7 is completed to the byte, so the C++ worker can implement `Open` from the page without reading any of this Go.

Rebased on 112efbb (#322). Do not merge.

## The commands

```
schema cook       --root <Table> --in <file.bin|tree-dir> --out <file.cook>
                  [--byte-order little|big] [--attribution <file>] [--tolerate] [--verbose]
schema cook-check [--root <Table>] [--attribution <file>] [--verbose] <file.cook>
schema uncook     --root <Table> --in <file.cook> --out <file.bin> [--attribution <file>]
```

`--in` takes a wire file or the directory tree `schema pack` reads, so one command covers *tooling does the build, then cooks to the rad binary format*.

## The owner's ruling, applied

**The region reference slot is EIGHT bytes** (owner, 2026-09-03 on #251: *"move now"*), over the two facts already on the page: *"u64 now, why fuck around"* for wire v2's widths, and *"Assume we have say, 100mbs or many gigabytes of data in Assets.bin at some point."* for the scale. A self-relative delta cannot reach further than the width it is stored in, so a four-byte slot bounded **one region** at 2 GiB — a ceiling a mesh or texture catalogue is exactly the thing to meet.

Recorded in §13.4 with the verbatims. The reach refusal and §15's three widening roads are deleted as moot. §6.3 states the cost in one place: eight bytes a reference in every variable-length record, needed or not, and every affected unit's build version moves so every cooked file is re-cooked — pre-release, a regeneration. A value-only table pays nothing: no pointer, no slot.

**It moved a latent bug beside it.** The emitted `TableRef` was `uint32_t` with a comment claiming region deltas are *always positive* — which §6.3 refutes for every shared node's back-reference, and which its own deref already contradicted by sign-extending. It is now `int64_t`, signed, with the comment corrected.

**Goldens: only the pointered unit moved** — `tables-pointers`' build version and its three generated headers. The value-only corpus, the block corpus and **every wire golden** are byte-identical; a region encoding touches no wire. Regenerated, never hand-merged.

**`test/tables/main.cpp` pinned the slot at four and went red** — the gate doing its job. Re-pinned at eight and strengthened while open: alignment, null in both encodings, a negative delta as an ordinary value, and the pointer field's offset against the alignment a cook writes at. The C++ battery passes, Lock/Save/Load over the pointered corpus included.

## What had to be built underneath

§3.1's flat node table had no writer anywhere — spec ahead of every backend — and a cook is a region, so the wire engine got one first: kind `17`, the reserved `0xFFFF` repeating field with its record-never-straddles rule, the depth-first pre-order numbering, the cycle refusal, the authoritative record scan, and every one of §3.1's read failures as one of §4's events. Then `internal/tablecook` on top, laying the region out at §20.3's C ABI model through a new `ir.RecordLayout` — the same model the block asserts and the build version's `record` lines come from; a second walk would be a second ABI.

One implementation gap fell out: `skip` and `KindWidth` did not know kind `17`, so a fixed-width skip §3 already specifies was framing damage. Fixed.

## Gates, with numbers

| gate | result |
|---|---|
| `wire → cook → wire` byte-identical, both orders | 6 pointered graphs × 2 orders + every pinned pack tree × 2 orders, green |
| header stamps `schema build-version` | equal; any other id refuses |
| big-endian byte-swaps every scalar | same length, differs; every header word the other's reversal (bar the order word, whose VALUE differs); `Scene.version` reads `03 00 00 00` / `00 00 00 03` at the offset the layout model gives; the file opens in its own order; directories entry-for-entry equal |
| hostile battery over `cook-check` | **25** directed forgeries + 400 seeded mutations over 12 fixtures; bar is REFUSE, OR ACCEPT AND BE WHOLE; nothing panics |
| its negative control | pass one removed via `go test -overlay` → red on the bar |
| the C++ tables battery | passes on the regenerated pointered headers |
| scale fixtures, re-measured at 8 bytes | 100 MB gen **0.5 s** / check **1.5 s** (3,276,795 nodes); **1 GB gen 0.61 s / check 15.2 s (33,554,427 nodes, 1.5 GB file)** |
| CLI round trip | tree→cook→wire and wire→big-endian cook with split attribution→wire, both `cmp`-identical |
| cross-endian, on s390x under qemu | the same command on a big-endian host produces **byte-identical** cooks in both orders, and either host checks either file |

`go test ./...`, `go vet ./...`, `golangci-lint run ./...` and `gofmt` clean. CI runs the 1 MB and 100 MB legs inside the two-minute rule; the gigabyte is `make tables-cook-scale-1gb`, by hand. **The open-cost gate itself is the RUNTIME's, not this tool's** — `Open` is what must be O(1), and a tool that walks a directory measures its own scan. These files are the C++ worker's inputs.

## Every page addition

**§7.1 The file, byte for byte** (new) — the header's eight `u64` words and offsets; the magic `0x4B4F4F434D484353` (`SCHMCOOK`, §19.1's `SCHMABLK` shape); what the magic does versus the `byte_order` word; `alignment` floored at 8; the data offset DERIVED as `align_up(64, alignment)`; `data_length` rounded to the alignment; `attribution_length` as `16 × nodes` with the count derived from it; the whole-file size check; the reserved words; the attribution part's framing, node extents and separability.

**§7.2 The region, byte for byte** (new) — nodes in §3.1's pre-order at their own alignment; §20.3's model named as the only one; a table of every declaration's STORAGE PIECES; an enum slot holds the ordinal, not the wire hash; an enum-keyed array is positional in a region and by name on the wire; a union is tag plus overlaid arms with only the set arm written; a counted array writes all N slots with the tail value-initialized; `*T` is eight bytes at eight and null is zero; and **every byte no field covers is zero**, because content addressing makes two cooks of one wire one artifact.

**§7.3 A cook, worked to the byte** (new) — a four-table example in §20.2's register, with layouts, numbering, offsets, deltas, directory and a full annotated hex dump of all **200** bytes. Taken from real output, re-derived after the widening.

**§7.4 What the check reads, and what it does not** (new) — pass one's refusals in full and why pass two then needs no bounds check of its own; pass two reads exactly three things (reference slots, count companions, union tags) and nothing else.

**§7.5 Held by test** (new) — the gates above as law.

**§7 body** — the fixed-root bullet (decision 1); the backend-status line separates the tool from the emitters.

**§6.3** — null in a region is a delta of zero, with the argument; the slot is eight bytes and signed, so one region reaches everything; the cost stated.

**§13.4** — the owner's ruling with its verbatims.

**§3.1** — what a reader that CAN name `0xFFFF` counts (nothing, and why).

**§15** — the three widening roads deleted; moot.

**§20** — status item 4 moves to "the tool carries it, the generated `Open` is owed".

**USAGE.md** — the three commands worked, `--attribution`, the status lines.

## Decisions, labelled

Places §7 was silent or in tension, where I chose and said so on the page.

1. **A FIXED root's cook carries a directory of one entry** — accepted by the coordinator, kept. §7 said "no attribution" for a fixed root; it is now ONE REGION OF ONE NODE, because a fixed root's count companions are exactly as forgeable in one struct as in a graph and with no attribution part `cook-check` could check *nothing*. 16 bytes, one shape instead of two.
2. **`cook-check` reads union TAGS.** §7 says the scan reads no field value. A tag is not a payload: it is the discriminant saying which arm's storage is live, and a scan that skipped it would leave every reference and companion inside an arm unchecked. Read, bounds-checked, and said so in §7.4.
3. **The magic's value** — `SCHMCOOK` in little-endian store order, beside §19.1's `SCHMABLK`, which §19.1 says a consumer written from the page needs as a constant.
4. **The header's shape** — §7 fixes three widths; the other five words, the 64-byte size and the derived data offset are mine. The `byte_order` word and its two values are lifted from §19.1.
5. **The region's alignment is floored at 8**, or a region of byte-only records aligns at 1 and the attribution part needs its own padding rule.
6. **`--attribution <file>` writes the part beside the cook** rather than inside it, and `cook-check`/`uncook` take the same flag to rejoin. §6.3 licenses placing the parts apart. A split moves only the length word, and the CLI gate proves the rejoin exact.
7. **A counted array's tail slots hold the value-initialized element**, matching `T x[N] = {}` and what a wire load leaves, so a cook of a loaded wire and a cook of a built structure agree.
8. **Cook refuses a wire it cannot read whole** — "refuse on mismatch" where a region would take a not-materialized sentinel. Softer report events print and exit nonzero, as `pack` does, with `--tolerate`.

## Open, for the owner — not invented

- **Whether a cook should record the asset hash.** §7 makes the store's key `(asset hash, build version)` but says the pair is what a store is indexed BY. Nothing in the header carries the asset hash, and stamping one would be inventing a field. A reserved word is there if it should.
- **Whether `--byte-order big` should also move the build version.** §20.1 digests `byteorder` as a generation input but the compiler emits `little` unconditionally, so today two orders share an id and the magic plus order word do the separating — which is exactly what §19.1 rules for a block. Consistent, and worth a look while the cook is new.
- **What `Open` does with a cook carrying no attribution part.** Nothing: the runtime never touches it. Stated, but worth confirming it is not meant to refuse one.

## Not in this PR

The generated `Open` on any backend (worker 2's C++, worker 3's C#), and the emitter step of #251 — kind `17` in `cpptable/pointers.go`, the resident node directory, retiring `<Name>LayoutId` and the v1 depth cap. The four claimed spellings stay claimed.


gafferongames:
Three open items, decided consistent with the block form (Rowan; reversible on a word): (1) no asset hash in the header — the (asset hash, build version) key lives in the store, the file is data; (2) `--byte-order big` does not move the build version — one build version per build, the byte-order word in the header distinguishes the files, a foreign order is refused by magic; (3) `Open` ignores a missing attribution part — it is the tool's, the runtime never reads it. Landing; the C++ Open is written from this page next and files any page defect it meets.

Issue #320, opened 2026-09-02
MSVC charges ~20 s per TU that has a multi-megabyte fixed table in scope (130x clang) — find what in the emitted header cl pays for


gafferongames:
Measured on the nightly msvc leg (#319): every TU that includes `RenderTable.h` (7,879,320-byte `RenderFrame`) costs cl 19.51 ~19–23 s at /W4 /WX; clang 0.13–0.18 s. Uniform and reproducible to within a second across four TUs. #317 removed the aggregate prvalue in sizeof (which was the out-of-heap failure); this residual is something else cl evaluates per TU for a huge aggregate — suspects: default member initializers / value-initialisation paths, constexpr layout asserts over the struct, the descriptor tables, or the `reset` placement-new lambda. A game with a large fixed table pays this on every MSVC build, which is the iteration loop (owner's law: the loop rate bounds productivity). Bisect by stripping emitted sections from a copy of the header under cl on the leg (workflow_dispatch), find the term, fix the class in the emitter.

gafferongames:
Owner ruling, 2026-09-03, verbatim: "Yes, this sounds like something we should try to avoid. Remember the 1-2 minute rule for CI that runs on each PR." — the fix's bar: the emitted header no longer costs cl ~20 s per TU, and the msvc job fits the per-PR rule (under two minutes, measured on CI) so it moves from nightly into the PR gate.

gafferongames:
## The report: it is all front end, and the term is `T{}` over the whole aggregate

Measured on run [33688601075](https://github.com/mas-bandwidth/schema/actions/runs/33688601075) (branch `msvc-per-tu-cost`, cl 19.51.36231 / VS 2026 on `windows-2025`), with temporary probe steps on the `msvc` leg.

### 1. Baseline, this run — the four block TUs reproduce #319

| TU | cl |
|---|---|
| `block/PaddedBlock.cpp` | 22.55 s |
| `block/PaddedTable.cpp` | 20.44 s |
| `block/RenderBlock.cpp` | 20.20 s |
| `block/RenderTable.cpp` | 20.10 s |
| `block/_AllHeaders.cpp` | 0.21 s |
| the other 39 TUs, together | 9.17 s |

95 s for the corpus step; 83.3 s of it is those four.

### 2. `/Bt+ /d2cgsummary` — front end vs back end

A bare TU (`#include "RenderTable.h"` + one `int`), same flags:

```
time(c1xx.dll)=19.85205s        <- front end
[...]

PR #338, opened 2026-09-03
tables in Go: the tolerant wire, JSON, reflection, the block form, the cook (preview) — mirrored from C++, harness green

gafferongames:
The Go tables port: the tolerant wire, the text form, the reflection
descriptors, the block form and the cook. Mirrored from `internal/codegen/cpptable`
— the reference — and following `internal/codegen/cstable` wherever a managed
language had already answered the same question.

## The gate: the conformance matrix

Rebased onto main after #340 (the Rust port). `make conformance`, FOUR legs,
all twelve surfaces:

```
surface       cpp            cs             go             rust
--------------------------------------------------------------------------
wire          pass 18/18     pass 18/18     pass 18/18     pass 18/18
report        pass 8/8       pass 8/8       pass 8/8       pass 8/8
json-read     pass 18/18     pass 18/18     pass 18/18     pass 18/18
json-write    pass 18/18     pass 18/18     pass 18/18     pass 18/18
json-hostile  pass 66/66     pass 66/66     pass 66/66     pass 66/66
cook          pass 6/6       pass 6/6       pass 6/6       pass 6/6
cook-foreign  pass 6/6       pass 6/6       pass 6/6       pass 6/6
block         pass 2/2       pass 2/2       pass 2/2       pass 2/2
block-foreign pass 2/2       pass 2/2       pass 2/2       pass 2/2
block-dump    pass 2/2       pass 2/2       pass 2/2       pass 2/2
forgery       pass 11/11     pass 11/11     pass 11/11     pass 11/11
cook-forgery  pass 111/111   pass 111/111   pass 111/111   pass 111/111
```

Registering the leg is one line in `test/conformance/drivers.txt` plus
`test/conformance/go/`, exactly as `test/conformance/README.md` says. The Go
leg is ONE binary answering every surface in process, which is why its `cook`
costs one exec rather than five.

**`cook-foreign` and `block-foreign` are new in this round, for all FOUR
legs** — the Rust driver listed ten surfaces where the contract now carries
twelve, so this branch registers them there too — and they exist because the blind read was right that a refusal nobody
exercises is a rule on a page. Each takes the corpus's own fixture and reverses
the eight bytes of its magic word, so the file is foreign to WHOEVER reads it,
and the expected verdict is `refuse` on every leg on every host. That is what
lets the big-endian leg report them GREEN rather than absent — the surfaces are
byte-order-neutral by construction, so they are the two accelerator surfaces a
big-endian reader CAN answer.

**And the same matrix on a BIG-ENDIAN target.** `make conformance-big-endian`
cross-builds the driver for s390x and runs it under the emulator the C++
big-endian legs already pin, so the wire's own claim — every field id, every
length and every scalar rides little-endian whatever the host is — is a gate
rather than a sentence:

```
surface       go-be
--------------------------
wire          pass 18/18
report        pass 8/8
json-read     pass 18/18
json-write    pass 18/18
json-hostile  pass 66/66
cook-foreign  pass 6/6
block-foreign pass 2/2
cook          absent
block         absent
block-dump    absent
forgery       absent
cook-forgery  absent
```

The remaining accelerator surfaces are ABSENT rather than failing, and
`test/conformance/go/driver-be` says why: a block and a cook are produced in the
byte order of the build that wrote them (§19.1, §7), so a big-endian reader is
CORRECT to refuse this corpus's native fixtures and has no byte-order-neutral
verdict to give about their CONTENTS. What it does have is the refusal itself,
and `cook-foreign`/`block-foreign` are exactly that surface. The C++ leg's own
cross-endian gate proves the refusal from the other side.

The pinned s390x toolchain is CI's, not a system binary (`BE_CXX`/`BE_RUN`,
Makefile) — the row above is CI's.

## What is emitted

Four sources per unit file, all four `go/format`-clean:

| file | what |
|---|---|
| `<Base>Table.go` | storage structs, `Reset`/`Measure`/`Save`/`Load`, the enum identity pair, the §8 descriptors |
| `<Base>Block.go` | §19's block form: blittable rows at the C ABI, `<T>BlockOpen`, the row accessors, the §8.1 descriptors with their row-walk columns |
| `<Base>Cook.go` | §7's cooked form: `<T>Open`, `<T>At`, the pointed records, the descriptors |
| `<Home>TableJson.go` | §16's text form: ONE generic walk over the descriptors, once per unit |

A table-free unit's Go output is byte-identical with and without the backend,
and a test holds it (`TestTablesMoveNoGeneratedGoByte`).

## The scope, and what is named rather than missing

The FIXED class on the wire, and both accelerators — which is C#'s scope, for
C#'s reason and with C#'s refusal. A unit whose closure declares a pointer gets
no `<Base>Table.go` at all, and every source it does emit opens with a banner
naming each refused table and the follow-on. The refusal is of the WIRE and of
nothing else: a pointered unit's `<Base>Block.go` and `<Base>Cook.go` ARE
emitted, and `<Root>Open` opens its cooked assets in full — which is what lets
§7's "a root is any table, and every table gets one" hold in Go too.

The VARIABLE class on the wire — the arena, the builder, the region, the
node-table codec (kind 17) — is the named follow-on, §11's own entry now
covering both ports.

## The blocking finding, and what closed it

The blind read found 48 unexported package-scope identifiers the emitter mints
— `tableJson*`, `tableBlock*`, `tableCook*` — that §11 did not claim. **In Go
unexported is not private**: a package is ONE namespace, so a schema declaring
`const tableJsonMaxDepth = 5` beside a table produced generated code that would
not compile, with no diagnostic naming why.

**Two things closed it, and the first is a reduction rather than a list.** The
three families were DERIVED — one variable per block record, per cook record
and per union field, so the claimed set grew with the schema and could not be
written down. They are now three FIXED slices — `tableBlockRecords`,
`tableCookRecords`, `tableUnionArms` — indexed by ordinal, so 48 derived names
became a bounded family a page can name and a test can check.

The second is registration: all 51 lowercase names are in
`internal/tablenames`, `internal/check` refuses a user declaration that spells
one with a diagnostic in the house shape (it says the Go half puts unexported
names at package scope, and cites §11), the scan in `compiler/tables_test.go`
is `[Tt]able` so it SEES the lowercase family, and the reader's own repro is in
the diagnostics corpus with its negative control — the same three spellings
must compile fine in a table-free unit.

## The six lesser findings

- **The fuzzer could not claim a short extent or an unaligned base.** Both arms
  are in it now; the second is what a caller placing a block inside its own
  buffer actually does wrong.
- **§16.5's per-backend walker negative control now exists for Go**
  (`make conformance-negative-control-go-walk`): one field offset off in the
  generated walk, through `go build -overlay`, must turn `json-read` red and
  nothing else.
- **The hand-ordered `…Max()` mirror is gone** from the fuzzer — §19.2 says it
  died, and the fuzzer was the last thing keeping it alive.
- **The soak states its identification rather than assuming it.** It samples
  every allocation (`MemProfileRate = 1`), diffs the profile across the steady
  phase, and classifies each site as the RUNTIME's or the PORT's. Finding this
  cost me a real bug in my own instrument, below.
- **`TableKeyed` guards both ends now** — `None` and past `E.Max` — and both
  panics are tested, with the boundary and the placement beside them.
- **The text form's allocation gate exists**, both directions. The PR claimed
  it before this round and only the wire path was measured.

## The rebase onto #340

The Rust port landed first and touched the same shared surfaces this branch
does. Everywhere the two met, BOTH languages are kept rather than one winning:

- `test/conformance/drivers.txt` and `bench/tables/legs.txt` carry four legs;
  `bench/SHAPE-GATE.allow` carries both ports' callsites (31 ledgered files).
- `internal/tablenames` carries a `Go` bit beside `Rust`, 25 shared entries
  marked for both, and Go's own eight `TableCookStorage` members — which only
  Go flattens into the package namespace, where C# and Rust scope them inside a
  type.
- `compiler/builtin.go`'s refusal now names all four backends, and
  `TestTablelessTargetsRefuseTables` sweeps the five that remain (c, dart,
  elixir, java, js) rather than the six it swept when Go was one of them.
- `docs/SPEC-TABLES.md`'s backend status, block-form paragraph and keyed-array
  section name all four; the keyed accessor's per-language KEY spelling gains
  Go's `int(ShipTypeBomber)` beside Rust's `ShipType::BOMBER.0 as u64`; §11's
  block-home rule keeps Go's own addition (the three descriptor SLICES).
- `docs/USAGE.md` keeps both ports' save/load and text-form examples in full.
- **The generated bench tree is REGENERATED, not merged.** #348 gave the
  corpus's `nonce` a declared default of 1, which moves the unit's build
  version and three lines of its codec. The `corpus_id` is unchanged
  (`2f7567e1e25ba918`), so the bench boards still key to the same corpus.

## Three things found on the way, not on the list

- **A 4.8 MB Mach-O executable was tracked at the repository root**, arrived in
  #345, and my own earlier commit carried a rebuilt copy of it — 4.8 MB of pure
  churn in this diff. Nothing builds or runs it: the Makefile builds the same
  program as `build/conformance-harness` with an explicit `-o`. Deleted, and
  `.gitignore` now names it so `go build ./test/conformance/harness` from the
  root cannot reach a commit again.
- **`test/conformance/README.md`'s budget projection was wrong by six times.**
  "Nine languages each starting a runtime per surface lands near 20 s" did not
  follow from its own table even before this round — the C# leg alone was 7.4 s
  there. A managed leg costs about 0.9 s per process start, so nine is about
  150 s: past the two-minute rule on its own. The page now says what the budget
  actually turns on, which is the SHAPE of each leg — a factor of twenty
  between the two shapes measured here — and re-measures the table for four
  legs and twelve surfaces.
- **`docs/USAGE.md` claimed six allocation gates where four existed**, naming
  `FromJson`, `ToJsonMeasure` and `ToJson` among them. The page was right about
  what the port should hold, so the gates were written rather than the claim
  trimmed; both directions measure zero.

## What Go spells differently, and why

Each is at the site in the emitted source as well as here.

- **An enum's wire identity is a METHOD** — `value.TableEnumId()`,
  `(&value).TableEnumValue(id)`. Go has no overloading, so a free pair would
  have to mint a per-enum unit-level name §11 does not claim; a method claims
  nothing at package scope.
- **An enum-keyed array IS the plain `[E.Max]T`** the schema means. Go has
  neither operator overloading nor a generic array extent, so there is no
  wrapper type; the shift and BOTH refusals live in one `TableKeyed` helper,
  so no call site spells either.
- **The claimed verbs stay FREE FUNCTIONS**, because Go has them: `<X>Open`,
  `<X>At`, `<X>Measure`, `<X>Save`, `<X>Load`. What is a member is only what §11
  already leaves a language free to make one — a block's row accessors and
  per-array constants, a cook's `Type`, `Root` and the three §7.1 facts.
- **The blittable records are a SECOND set of structs** under §11's claimed
  `<Name>Row` suffix, for the reason the C# port's are: the wire's storage
  follows the Go PACKET emitter (`string(N)` is `[N]byte`) and the C ABI record
  is `char[N+1]`. One declaration cannot be two types.
- **The layout contract is a generated `init()` that REFUSES**, naming the
  record, the field and both numbers. Go could state most of it as a constant
  expression but not with a message a person can act on.
- **A cyclic descriptor graph is linked in an `init()`**, because Go refuses an
  initialization cycle among package-level variables — a closure in an
  initializer included — and a cooked graph is cyclic by design.
- **No locale.** `strconv` is locale-free, so the decimal point the C++ text
  walk crosses on the way out and back in does not exist here.

## The instruments

**Allocation is MEASURED, not asserted.** C++ holds "the codecs allocate
nothing" by construction and its gate is a grep; Go has a garbage collector, so
`testing.AllocsPerRun` is the gate and zero is the number — `Load`, `Measure`,
`Save`, the wire round trip, and now `FromJson`, `ToJsonMeasure` and `ToJson`.
The gate carries its own negative control (an escaping buffer, an escaping
sub-reader), because `AllocsPerRun` reads zero for an allocation the compiler
proved does not escape and a gate written without one measures nothing.

**The soak** reads and writes the whole 18-case corpus in a loop — wire both
ways, text both ways, every byte compared — and watches
`runtime.MemStats.Mallocs`, which must not scale with the loop.

**369,372 passes of the 18-case corpus in one hour — about 20 million round
trips — allocated ZERO objects.** Not two, as the previous round reported: the
identification bug below is what produced that two, and with the profile read
correctly there is nothing to identify, because `Mallocs` does not move at all
over the hour. The bound is a RATE (`iterations/10000 + 8`, so 44 here) and it
was never approached. The EXACT zero per operation is held next door by the
per-operation gates.

It found a real defect on its first run, and one no byte comparison could have:
the text walk rebuilt a union field's arms table on every call, ten objects per
`ToJson` of an instance carrying five unions.

**And it found a defect in ITSELF this round, which is why the identification
is now printed rather than assumed.** The first version snapshotted the
allocation profile BEFORE the GC cycle the profile needs, so the setup's own
objects — the scratch buffer, the corpus — were attributed to the steady phase:
it reported "4 of the 4 allocation sites are the port's own" while `Mallocs`
said the loop had allocated ZERO. It now reads the profile only when `Mallocs`
actually moved, flushes three cycles on both sides, and carries
`TestSoakIdentifierCanGoRed` so the classifier itself has a control.

**The forgery fuzzer** over both accelerators' `Open`, with the C++ fuzzer's own
contract — REFUSE, or OPEN and be WHOLE. The oracle re-derives every bound from
the descriptors and from the triples the instance carries, never from `Open`'s
arithmetic, and reads every row through a bounds-checked view of the extent —
Go's bounds checks are on in every configuration, which is what the sanitized
twin does for C++. It claims a SHORT EXTENT and an UNALIGNED BASE as well as
mutating bytes. It runs plain and under `-race`. Its two negative controls
remove the extent check and the declared-maximum check from the EMITTER through
`go build -overlay` and each must turn it red on that check and no other.

**The harness's own negative control**: one byte off in one wire answer turns
the matrix red on that surface and leaves every other green.

## The bench, and a lever that did not transfer

Registered in `bench/tables/legs.txt` (#339): one line plus
`bench/tables/go/leg`, and a runner ported from the C++ reference — shape-blind
at the one call site the ledger names, gated on the golden and the 64-variant
round trip before any clock starts.

Re-paired against the post-#350 C++ arm. A PAIRING CHECK on the Air, seven
interleaved rounds, board and CSV at
`bench/tables/results/2026-09-03-pairing-go2-arm64-macbook.*`. It certifies
nothing — the box sitting is the certification.

| lang | path | M msg/s (median) | ratio to C++ |
|---|---|---:|---:|
| cpp | write | 1.454 | 1.00 |
| cs | write | 0.246 | 0.17 |
| **go** | **write** | **0.294** | **0.20** |
| cpp | round_trip | 0.493 | 1.00 |
| cs | round_trip | 0.155 | 0.32 |
| **go** | **round_trip** | **0.186** | **0.38** |

**The ratio moved from 0.82/0.78 on a denominator alone.** #350 forced the
generated C++ bodies and the writer primitives inline and took the C++ write
arm from 0.322 to 1.454 M msg/s; nothing about the Go leg moved, and Go still
leads C# on both rows.

**Whether the same lever exists in Go is measured, not guessed.** It does not
exist as a flag: Go already inlines the writer's puts (`-gcflags=-m` reports
`inlining call to (*TableWriter).Put16` at every callsite), and
`-gcflags=all=-l=4`, which inlines the nested bodies too, moves the leg by
nothing outside this laptop's noise. **The round-trip #350 removed SURVIVES
inlining in Go** — an inlined `TableStatSaveBody` still emits, per put:

```
MOVD 24(R0), R2      // load w.Offset
ADD  $1, R2, R2
MOVD R2, 24(R0)      // store it back
MOVD 24(R0), R2      // and load it again for the next put
LDP  (R0), (R4, R5)  // reloading w.Buffer's pointer and length too
```

because Go has no `restrict` and does not separate a `[]byte`'s payload from
the struct naming it, so the store through `Buffer` is assumed to alias the
writer.

**It exists as a generated shape, and it is worth 1.42x** — measured over the
mechanism in isolation (92 fields of `tag16, kind8, value32`, cursor in the
struct against cursor in locals, output asserted byte-identical): 846.7 ns
against 595.3 ns per record. §15 names the pass — hoist the cursor, merge the
per-field bounds checks, write back once. **It is not taken here**, because a
gate lands in its simplest form and this is a codec restructure with its own
goldens to re-prove. It is ruled with a number rather than an intention.

So the gap decomposes: about 1.4x of it is a Go codegen pass this port has not
done, and the rest is the language's bounds discipline paid where the format's
contract asks for it.

## The budget

`make conformance` is ~21 s for FOUR legs over twelve surfaces, inside the
two-minute rule with most of it to spare. **The Go leg alone is 0.79 s** —
against 0.81 s for rust, 0.92 s for cpp and 16.5 s for cs, which is where the
budget actually goes: 804 of the 1,072 cases cost about two and a half seconds
between them and the remaining 268 cost sixteen. One native exec per surface and no more, because its `cook` and
`cook-forgery` are answered in the same binary as everything else rather than
delegated.

## The pages

`docs/SPEC-TABLES.md`'s backend status names Go beside C# for the fixed class
and the block form as live in three languages; §11 reconciles the claims — Go
widens `TableCookHeaderBytes` from scoped to claimed, adds the eight
`TableCookStorage` members and the 51-name lowercase family, all registered in
`internal/tablenames` with `TestTableRuntimeNamesAreClaimedGo` holding the list
honest in both directions. §15 gains the cursor-hoist pass above.
`docs/USAGE.md` gains the Go save/load and text-form examples.

**The six page defects from the read are fixed**, `FORMAT.md:132` included: the
emission order is now stated exactly — a node's leaf lines are emitted
depth-first, so a parent's remaining leaf lines FOLLOW its descendants' node
blocks rather than preceding them, and `#present` is emitted after the value,
with an absent `?T` still emitting its value line.
`test/conformance/README.md` and `FORMAT.md` both state the two foreign
surfaces and their expected `refuse`.


gafferongames:
**LAND-AFTER-FIXES** — one real defect against a §11 promise, cheap to close; everything else I could measure is green. Written from the pages alone, `go doc` for spellings, no emitter until the adversarial half.

### Defects, ranked

**1. The registry does not claim the Go runtime's LOWERCASE names, so a legal schema generates Go that does not compile.** `package probe / const tableJsonMaxDepth = 5 / table Thing { n int32 }` →
`./ProbeTableJson.go:33:2: tableJsonMaxDepth redeclared in this block / ./Probe.go:21:7: other declaration`.
48 identifiers are exposed corpus-wide (`tableJson*` 42, `tableBlock*` 3, `tableCook*` 3); I reproduced build failures on four. The scan is `compiler/tables_test.go:673`, `\b(?:Table[A-Za-z0-9_]*|BuildVersion)\b`, copied from the C# leg at `:582` where every runtime name is PascalCase — Go is the first backend with an unexported package-scope runtime family, so the regex is blind to exactly what this PR adds. **Fix:** `[Tt]able` at `:673`, then register the family.

**2. `test/go-tables/fuzz_test.go:223` — the fuzzer can never claim a SHORT extent.** `*claim = r.IntN(len*2)` is clamped back up to `len(image)`, so `:220`'s "larger or smaller than the bytes it has" names a case the code cannot produce; truncation is left to the fixed 11-row battery, and `place()` already handles it. **Fix:** drop the clamp.

**3. Same file, `:131` — `place()` always 64-aligns**, so the manifest's `<pointer>` column is never fuzzed. **Fix:** jitter the base.

**4. §16.5's per-backend walker control is one backend short.** `conformance-negative-control-go` (Makefile:2645) flips an output byte in the DRIVER — the cpp shape. §16.5 asks for the walker's offset arithmetic sabotaged in the EMITTER, red on `json-read` and green elsewhere, as the `-cs` control is. Go has real offsets, so that shape ports directly.

**5. `fuzz_test.go:102` keeps a hand-ordered mirror** — `b.CamerasMax(), b.ShipsMax(), …` paired positionally with the descriptor's out-of-line fields, which is the mirror §19.2 says died. **Fix:** key them by field name.

**6. `soak_test.go:203`'s "an hour moves it TWICE … a forced collection or a stack move" is ASSUMED** (the "or" is the tell). I identified it: 150 s / 3,049,187 passes at `MemProfileRate = 1` moved `Mallocs` by 12, all attributed to `runtime.allocm` / `gcBgMarkWorker` / `gcStart.func4`. True, and still not established by the code.

**7. `TableKeyed` (gotable.go:777) guards None but not `> E.Max`.** Under `-gcflags=all=-B` the None panic survives and `TableKeyed(slots[:3], 4)` reads past the array silently. §2.4's argument for the unconditional compare is symmetric. **Fix:** `if key <= 0 || key > len(slots)`. And §2.4:504's "in every port" is weaker in Go than the page admits: the storage IS the bare `[E.Max]T`, so `slots[0]` never reaches `TableKeyed` at all.

### Page defects

- **SPEC-TABLES.md:2036, :2092, :5642** spell the C++ and C# cook/block surfaces "because a consumer written from this page needs the spelling", and never spell Go's. I needed `go doc`.
- **USAGE.md:805** promises "the same columns C++ carries", but its C++ example (:1704) writes `field.table->name` where Go has `Table func() *TableTypeInfo`; `TableBlockInfo`/`TableCookInfo` appear on no page at all.
- **FORMAT.md:132** "Nodes are numbered in visit order … one line each": the pins do NOT emit node blocks consecutively — a parent's remaining leaf lines follow its descendants (Scene.dump's root `tree`…`meta` land at line 493, after node 125). I wrote what the page says and got a reordered dump on 3 of 6 cooks. **:141–142** likewise never says `#present` follows the value lines, nor that an absent `?T` emits its value anyway.
- **§19 never states a block row's enum slot holds the ORDINAL at the enum's own width** — §7.2 (:2399) says it for the cook only, and `Kind` is the WIRE kind. I read `team` as u16 and got 256 for 0.

**(g) the five s390x absences: right verdict, thin vocabulary.** `cook`'s expectation is the node DUMP, `block`'s a hard-coded `open\n`, and both forgery batteries carry `open` rows a foreign-order reader must refuse — no neutral answer exists, so `driver-be` is correct. But the leg cross-builds with both readers in it (I built the s390x ELF), so "ABSENT = missing FEATURE" mislabels "reader present, no neutral expectation". A `cook-foreign`/`block-foreign` surface expecting `refuse` closes it; today only C++'s `make tables-big-endian` gates that refusal.

### Reproduced

- **8/8 wire goldens**: my own descriptor walker (not `ToJson`) matches every JSON golden semantically; `Measure`+`Save` byte-identical (398/2/46/152/210133/414/234/2 bytes); every report silent.
- **8/8 accelerator fixtures**: my own block ROW and cook NODE dumps, descriptors alone, byte-identical to all eight pins (450/69 lines; 123/126/126/102/128/123 nodes), once I found the FORMAT.md gap above.
- **Matrix**: ten surfaces green on three legs. Go alone 260/260, median 1.03 s (README: 1.07 s). All three: 24.2 s vs the README's 10.5 s — four other workers building here; #320 untouched.
- **Claimed names**: registry-for-Go and emitted `Table*`/`BuildVersion` corpus-wide are the same 33, both directions, zero drift — the hole is the lowercase family the scan cannot see.
- **Allocations**: 0/run for `Load` (report and nil), `Measure`, `Save`, `FromJson`, `ToJsonMeasure`, `ToJson` over all 8 goldens, and 0/run for every descriptor accessor including `Reset`. My generic walk allocates 161/run — mine, and licensed.
- **Soak** 3 min: 11,534 passes × 18 cases, 0 objects (bound 9). **Fuzz** `SEED=20260903 N=5000000`: 170 s green, both emitter-sabotage controls fire. `go vet` and `-race` clean everywhere; staticcheck not pinned here.
- **None refusal** panics under `-gcflags=all=-B`, `-ldflags=-s -w`, `-trimpath`, `-l=4`. Iteration keys 1..3 = Red/Blue/Green; `KeyName(0)="None"`, `KeyId(0)=0`; no storage index maps to None.
- **Bench**, cpp+go paired once on this Air: write **0.80** (PR 0.82), round_trip **0.73** (PR 0.78), same `corpus_id 2f7567e1e25ba918`. Absolutes ~33% under the PR's and go round_trip spread 60% — the ordering reproduces; nothing here certifies.

PR #340, opened 2026-09-03
tables in rust: the tolerant wire, JSON, reflection, the block form, the cook (preview) — mirrored from C++, harness green

gafferongames:
The Rust table backend (`internal/codegen/rusttable`), mirrored from
`internal/codegen/cpptable` the way `internal/codegen/cstable` is, registered
as the third leg of the tables conformance harness, of the block forgery
fuzzer's negative controls, and of the tables bench.

**The blind read's four defects and five page defects are fixed on the branch.**

## The matrix

```
TABLES CONFORMANCE — surface x language

surface       cpp           cs            rust
--------------------------------------------------------
wire          pass 18/18    pass 18/18    pass 18/18
report        pass 8/8      pass 8/8      pass 8/8
json-read     pass 18/18    pass 18/18    pass 18/18
json-write    pass 18/18    pass 18/18    pass 18/18
json-hostile  pass 66/66    pass 66/66    pass 66/66
cook          pass 6/6      pass 6/6      pass 6/6
block         pass 2/2      pass 2/2      pass 2/2
block-dump    pass 2/2      pass 2/2      pass 2/2
forgery       pass 11/11    pass 11/11    pass 11/11
cook-forgery  pass 111/111  pass 111/111  pass 111/111
```

All ten surfaces, 260 cases. `cook-foreign` / `block-foreign` are **not
registered**: they are not in the contract on main yet.

## The four defects

**1. The text writer allocated 74 times per `ToJson` — fixed, and the number
is now a gate.** The float formatter built its digits with `format!`. It
formats through a **64-byte stack sink** — the C++ walker's own `char[64]`,
and provably enough rather than generously chosen: style f is used only when
the decimal exponent is in `[-4, precision)`, so the integer part is at most
18 digits and the fraction at most 20. `core::fmt` writes a float through its
own stack buffers, so a non-allocating sink makes the whole formatter
allocation-free; a write past it REFUSES, as the C++ walker refuses an
`snprintf` that would not fit. **json-write stayed 18/18 — byte-identical
output**, which is the proof you asked for.

```
$ make tables-rust-alloc-audit
instance                  allocations
root_full                           0
…                                   0     (all 18)
alloc audit: load, measure, save, from_json, to_json_measure and to_json
             — 0 allocations, every instance
```

**2. The soak could not go red on it — fixed.** It gated on LIVE-BYTE drift,
a **leak** instrument: a path that allocates and frees the same bytes every
iteration reads `+0` there forever, which is exactly how 74 allocations per
`ToJson` sat under a green one-hour soak. It now gates on the allocation
**count**, per iteration, through an audit closure that boxes the value and
sizes the buffer *before* the counter is snapshotted — so the count measures
the codec and not the driver. `make tables-rust-alloc-negative-control` puts
one allocation per iteration inside the measured region and **both gates go
red on it**.

**3. Nothing could be left uncompiled — fixed.** The two accelerators are
**cargo features, both on by default**. `--no-default-features` compiles
**16,109 of tabledemo's 23,290 lines** instead of all of them; `--features
cook` and `--features block` take one without the other;
`make tables-rust-features` builds all four combinations on all ten units.

**The feature gate was not a formality: its first run found two real
couplings**, each a fact belonging to *neither* accelerator that had been
sitting inside one. The unit's **BUILD VERSION** (§20 answers "which build?",
not "which form?", and both forms compare against it) and the blittable
**`<Name>Row` records** (a cooked record IS the blittable row, §7.2/§19.3, so
a block-only build wants the same structs a cook-only build does). Both now
have their own always-compiled modules.

**4. `unsafe` — the page now states the truth.** There are **four** sites, not
three. The fourth is §8.1's own descriptor surface: the `reset: fn(*mut u8)`
hook takes a raw pointer because a generic walker holds no type to spell, and
a union field's four columns reach a Rust enum's payload the same way — and
every table carries them always, in the wire module, behind no feature. The
generated table **wire** remains safe Rust throughout. USAGE lists all four,
says which two the features switch off, and gives the line count.

## The five page defects

- **§8.1's block positions** were described as columns *on the table field
  descriptor*. No backend does that — C++ and Rust both emit a separate
  `TableBlockFieldInfo` reached from the block's own type descriptor. Stated
  correctly now. (Pre-existing.)
- **§8.1 never said how to tell `bytes(N)` from `[..N]uint8`.** Both carry
  kind u8 with the array column set — the wire framing is identical, which is
  the point — and the **type name** is the discriminator by design, `bytes`
  being a keyword no declaration can claim. Written down.
- **§8's "one vocabulary, one walker" is one walker PER UNIT**, and in a
  language whose units are separate namespaces that is a real limit: a tool
  walking two units holds two nominally distinct `TableFieldInfo` types and
  writes its walk once per unit. Named, with the trade a shared descriptor
  library would make.
- **USAGE:1840 contradicted USAGE:2124.** Now a cross-reference, plus the
  allocation claim it earns. **§2.4** admitted only C#'s `int` cast; every
  port indexes by the KEY and what varies is the key's *type* at the call
  site, so Rust's `u64` is there beside C++'s enum and C#'s int.
- **§11's "claims are UNCONDITIONAL — every unit, tables or not" is not in
  force.** The paragraph says so now and names the deadline: before the first
  view file is emitted, not after. (Pre-existing.)

## The gates

| gate | result |
|---|---|
| conformance matrix | green, all ten surfaces, 260 cases |
| `tables-rust-alloc-audit` | **0 allocations**, every path, every instance |
| `tables-rust-alloc-negative-control` | one allocation reds **both** gates |
| `tables-rust-features` | wire-only, cook-only, block-only, everything |
| `tables-rust-walk` | one table runtime, the same bytes in every unit |
| `tables-rust-clippy` | every unit builds under a consumer's clippy |
| `tables-rust-fuzz` | 409,746 mutants, none read outside the extent |
| `tables-rust-fuzz-miri` | 51,940 mutants under miri — by-hand, 478 s |
| `tables-block-fuzz-{extent,maximum}-negative-control` | red on the Rust leg |
| `tables-rust-names-negative-control` | red when the §11 claim is removed |
| `tables-rust-big-endian` | every layout const assert evaluated for s390x |
| `tables-rust-soak` | 404,848 iterations, drift +0 |

## The bench

**PAIRING CHECK, not a sitting.** Board:
`bench/tables/results/2026-09-03-rust-inline-arm64-macbook.md`, against the
inlined C++ arm (#350):

| lang | path | M msg/s | ratio to C++ |
|---|---|---:|---:|
| cpp | write | 2.4107 | 1.00 |
| **rust** | **write** | **2.0148** | **0.84** |
| cpp | round_trip | 0.8226 | 1.00 |
| **rust** | **round_trip** | **0.6211** | **0.76** |

Derived read: rust/cpp **0.72**. Applying #350's force-inline rule to Rust was
measured against a banked prediction that **MISSED** — predicted write
+10–60%, measured **+103%**, past the 2x line I had set as the refutation. The
board says so. The remaining read gap is filed for a line-level profile on a
quiet box, not chased here.

## Two more deviations, both named in the package doc

- **The read report is not inside the reader** — a Rust reader holding
  `&mut TableReport` could not hand out a sub-reader over its own buffer.
- **The cooked and blocked records are their own `<Name>Row` structs**, as in
  C#: SPEC §6.1's Rust column spells `string(N)` as `[u8; N]` where the layout
  model spells `char[N + 1]`. `#[repr(C)]` IS the layout, so no generated
  padding — const asserts hold it at compile time.

## What is refused, by name

The **variable class's wire surface**, on exactly the C# terms (§11); both
accelerators emitted, so a pointered unit's cooked assets load. A **union in a
cooked record**, for the reason the block form has always refused one.

DRAFT: not for merge.


gafferongames:
**LAND-AFTER-FIXES.**

Blind read at `2177626`: an independent Rust consumer written from the pages alone, over the crates' public API only. **142 checks, 0 failures.** All 18 wire goldens load with a silent report; my own walker over `<name>_table_type()` reproduces the JSON goldens semantically; `measure == save ==` golden bytes on each; a short buffer refuses. Both block images open and my reflective row walk reproduces the pinned dumps' values; all six cook fixtures open and walk, every count companion in bound; every §7/§19.2 refusal I could derive fires (truncation, reserved word, alignment 0 and 3, byte-reversed magic, unaligned base). Keyed: iteration yields `1..=E.Max`, key 1 lands in slot 0, key 0 and `E.Max+1` panic. `make conformance` green on three legs. **The port invents no contracts** — its three descriptor tables mirror `V1Table.h`/`V1Block.h` field for field.

## Defects

**1. The text WRITER allocates — `table_runtime.rs:942,944,958,961`** (`format!`/`to_string()` in the float formatter; byte-identical in all ten units). Counting global allocator: `root_config_to_json` **74 allocations / 759 bytes per call**, `to_json_measure` the same again. Every other path measured **0**: wire load, measure+save, keyed and union load, `from_json`, `BlockOpen`+row sweep, `SceneCook::open`+chain walk. Against USAGE.md:1840-42 ("allocates nothing beyond the instance you passed in"). **Fix:** format through a stack `[u8; TABLE_JSON_MAX_NUMBER]` sink — the constant (512) is already emitted and C# already does this.

**2. The soak cannot go red on (1) — `test/conformance/rust/src/main.rs:1124-28`** gates on `drift != 0` over LIVE bytes, a LEAK instrument. My 75 s run: **`drift +0` while the soak's own line printed 5,957,932 allocations in 60 s** (~1,072 per pass over 18 instances). It counts `TOTAL_ALLOCATIONS` and does not gate on it. **Fix:** snapshot that counter after the warm pass and fail if it moves — it goes red today.

**3. The `unsafe` claim is wrong on both halves — USAGE.md:2124-31.**
- A fourth site: **71 `unsafe` lines across `*_table.rs` outside the text form's entry points** — every `<name>_table_reset_at` (§8.1's `reset: fn(*mut u8)` hook) and every union field's four descriptor columns (`read_tag`/`clear`/`select`/`payload`). §8.1 says every table carries those always.
- **Nothing can be left uncompiled.** `lib.rs` declares `mod <base>_block; mod <base>_cook; mod block_runtime; mod cook_runtime;` unconditionally; no `[features]`, no `cfg(` anywhere in the ten crates. A consumer that opens no block and reads no cook still compiles **7,767 of tabledemo's 23,032 lines (33%)**. §19's "the Block files cost nothing unless they are included" has no Rust analogue. **Fix:** state it as a Rust divergence, or gate the accelerator modules behind default-on cargo features. (216 `pub unsafe fn` corpus-wide, all with `# Safety` — that half holds.)

**4. §8's "one vocabulary, one walker" is one walker PER UNIT here.** Each crate declares its own `TableTypeInfo`/`TableFieldInfo`, so a tool walking two units holds two nominally distinct types; I had to emit mine through a macro per crate. Worth naming beside the two divergences already listed.

## Page defects

- **SPEC-TABLES.md:2928-45** puts the block projection offsets and row-walk columns *on the table field descriptor*. No backend does that — C++ (`V1Block.h:169`) and Rust both emit a separate `TableBlockFieldInfo`, so a reader of §8.1 looks for `projection_offset` on `TableFieldInfo` in vain. Pre-existing.
- **§8.1 never says how to tell `bytes(N)` from `[..N]uint8`** — both carry `kind: 6, is_array: true`; only `type_name` separates them, and §16.2 needs base64 for one and an array for the other. I had to guess.
- **USAGE.md:1840** ("the one place the generated table code is `unsafe`") contradicts USAGE.md:2124-31. **§2.4** admits only C#'s `int` cast; Rust indexes by `u64` (USAGE:793).
- **SPEC-TABLES.md:3603-06** ("claims are UNCONDITIONAL — every unit, tables or not") is not in force: a table-free unit accepts `const TABLE_COOK_MAGIC` and `type TableReport`. Pre-existing.

## Reproduced, with numbers

- **Fuzzer, fresh seed `90903172026`, N=6M: 20,764,746 mutants, 5,195,393 opened, none read outside the extent** (4:12). The oracle is genuinely independent — `walk_opened` re-derives every bound from the descriptors and the instance's triples, then reads every row byte; both emitter negative controls turn it red. Miri leg clean: 51,940 mutants, 1,710 opened, nothing read outside the extent (18:05).
- **Names:** all three spellings refuse, as do `TableReport`/`TableKeyed`/`TableBlockTriple`/`table_bits_to_float`/`BuildVersion`; the negative control goes red naming `build_version`.
- **File order:** a first-sorting file leaves all three runtime files byte-identical (banner included — it names the first *table-declaring* file) and the crate still builds. The packet backend still relocates `PROTOCOL_ID`/`Error`/`Result`; the table homes are clean.
- **Big-endian:** `tables-rust-big-endian` runs (s390x installed) and passes; the cross-endian *refusal* is gated by `cook_magic_reversed`, `cook_magic_block`, `cook_byte_order_0/2/3`, `block_magic_swapped` and `block_byte_order`. clippy green at stable 1.97.1.
- **Bench, CONTENDED window** (two workers at 100%, load 5.4, spreads 12.5–35.8% against the board's 0.0%): cpp 1.533/0.618, rust 0.998/0.310, cs 0.200/0.124 M msg/s → **0.65 / 0.50 / 0.43**. Neither confirms nor refutes 0.84/0.76/0.72; it re-confirms the board's own caveat.
- **Profile of the read arm** (`sample`, 20 s, same 64 variants): **9,554 of 9,676 samples (98.7%) in `table_mixed_load` itself**; the only named callees are `_platform_memset`/`__bzero` (1.1%) and `_platform_memmove` (0.2%). **The top term is the decode body** — a flat 8,424-byte function opening with an inlined full-record prefill of the 1,352-byte `TableMixed` (§8.1's `<Name>Reset`, which `BenchTableTable.h:1369` also calls, so not an asymmetry). That rules out the cheap answers: no allocation, no bounds-check helper, no indirect dispatch, no memcpy storm. I could not split prefill from decode on this box — five interleaved rounds gave read 0.455 / load-only 0.463 / internal-prefill-removed 0.474 M op/s inside a 0.33–0.49 spread. The remaining ~1.4x is **distributed inside the flat decode body**: the next instrument is line- or PMU-level on a quiet box, not another sampling profile.

PR #344, opened 2026-09-03
tables: a declared bound AT the storage limits emits no clamp (#342)

### LINKED ISSUE #342: cpptable: a declared bound at the storage limits emits a tautological clamp (g++ -Wtype-limits reds it)
Found by the tables bench corpus (#330 / PR #339) on the way in.

A table field declared with a bound spanning its whole storage width —

    nonce uint64 | min = 0, max = 18446744073709551615

— makes `internal/codegen/cpptable` emit, in `<Root>LoadBody`:

    if ( decoded_v < 0ull ) { decoded_v = 0ull; r.report->clamped++; }
    else if ( decoded_v > 18446744073709551615ull ) { decoded_v = 18446744073709551615ull; r.report->clamped++; }

Both comparisons are tautological. **g++ rejects them** under the repo's own
flags (`-Wall -Wextra -Werror` → `-Wtype-limits`); Apple clang does not warn,
so a tree carrying such a field reads clean on macOS and reds on CI:

    error: comparison of unsigned expression in '< 0' is always false [-Werror=type-limits]

**The rule the emitter needs is one it already has.** It ALREADY elides the
`bits(N)` width clamp when N equals the storage width — `ack_bits bits(32)`
generates no clamp at all — so the same "the check cannot fire" test applied to
a declared `min`/`max` closes this. Both ends need it: a `max` at the storage
maximum is as tautological as a `min` at the storage minimum, and the signed
case (`int64 | min = -9223372036854775808`) is the same shape.

Scope note: this is the C++ table emitter. Whether the C# emitter has a twin is
unchecked — C# has no `-Wtype-limits` equivalent that would have surfaced it,
which is a reason to look rather than a reason not to.

The bench corpus does NOT carry the spelling, so nothing is blocked: it uses a
bound that is live at both ends and says why at the field
(`bench/corpus/BenchTable.schema`). Leaving the corpus red would have been the
other way to keep this visible; the comment plus this issue is the trade.

gafferongames:
Closes #342.

## The defect

A table field bounded at its storage width made both table emitters write a comparison with no false case:

```cpp
if ( decoded_v < 0ull ) { decoded_v = 0ull; r.report->clamped++; }
else if ( decoded_v > 18446744073709551615ull ) { ... }
```

The issue named gcc's half. **Both compilers red it, and they split the work:**

| | diagnostic | catches |
|---|---|---|
| gcc | `-Wtype-limits` (implied by `-Wextra`) | the unsigned half — `decoded_v < 0ull` |
| clang | `-Wtautological-type-limit-compare` (implied by `-Wextra`) | the signed half — `decoded_v < -128` |

Neither catches the other's. So the failure mode is worse than "clean on macOS, red on CI": it is *whichever half you happened to spell*. Both halves are in the corpus case below, so the gate holds on either platform.

## The fix

The rule the emitter already had. It drops a `bits(N)` width clamp when N is the storage width; `tableClampEnds` now applies the same "this check cannot fire" test to a declared `min`/`max`, **at each end independently** — signed and unsigned, every integer width, scalars and array elements alike. Four emission shapes come out of it: no clamp, upper only, lower only, both.

`docs/SPEC-TABLES.md` §4 is untouched and needs no sentence: an elided end is one that could never have clamped or counted. Read reports do not move, and neither do bytes — clamping is read-side, and a clamp that cannot fire writes nothing either way.

Mirrored into the **C# emitter**, which had the identical twin. C# raises no diagnostic for a comparison that cannot fire; the shape mirrors anyway, because one table codec in two languages is the point.

`bits(N)` cannot declare `min`/`max` (the checker refuses it — §4.6 applies them to integer fields), so the bits half of the rule is the width clamp that was already there. `RangedWidths` pins it beside the new one.

## The corpus case — `tables/examples/Ranges.schema`

Every integer width four ways: both bounds at the limits, each limit alone, one value off each end. Plus an array element at a full span, and the six `bits(N)` widths that do and do not fill their storage.

## Gates

- **`make tables-clamp-limits`** — the generated header compiles under `-std=c++17 -Wall -Wextra -Werror -Wtype-limits`. (`-Wtype-limits` is named explicitly even though `-Wextra` implies it on gcc, so the flag that matters is in the recipe and not one level of implication away.)
- **`make tables-clamp-limits-negative-control`** — puts the pre-fix emission back through `go build -overlay` (the two comparisons deciding which ends are live become non-strict, which is always true given the checker already refuses a bound outside its storage), regenerates, and **requires the build to go red on a comparison that cannot fire**. Locally, on Apple clang: **19 of them**. It also asserts the sabotage leaves the *rest* of the corpus byte-identical — no other declared bound in the tree sits on a storage limit, which is exactly why this survived until the tables bench corpus met it.
- Both wired into `make test`.

## Wire identity

**No wire golden moves** — `testdata/wire/` is untouched by this branch, and structurally so: the change is read-side only, and the clamps it removes are the ones that could never have fired. `./build/schema_test_tables` byte-compares the pinned instances before it passes.

The only movement in existing goldens is **one line per C++ Table header** — `BuildVersion`, because the unit gained three tables — plus the `tables-examples` build-version pin and the baseline. `generated/` is byte-unmoved (checked `generated/bench/tables/{cpp,cs}` explicitly: `BenchTable.schema` deliberately carries no bound at a storage limit).

## Green here

`go test ./...`, `gofmt`, `go vet`; `schema check tables/examples`; `build/schema_test_tables`; `test/cs-tables` (dotnet); `make tables-zero-cost`, `tables-block-zero-cost` (41 Table sources byte-identical), `tables-cs-json-walk`, `tables-cs-standalone`, `tables-cs-refuses-pointers`, `conformance`, `tables-json-walk`, `tables-json-negative-control`, `tables-cook-cli`, `shape-gate`. The gcc half of the diagnostic is CI's to prove — there is no gcc on this bench.

## Three judgments worth a look

1. **The corpus file is named `Ranges.schema`, not `Bounds.schema`, for one incidental reason:** the C# table runtime is emitted once per unit into `ir.ProtocolIdHome` — the unit's *first* file. A file sorting before `Guarded` relocates 2149 lines of shared runtime into it. That is correct, designed behavior and the goldens absorb it fine, but it is not what this PR is about, so the name keeps the review diff to the change. Flagging it rather than burying it: if you would rather the corpus file be named for what it holds and the home move ride along, say so and I will rename.

2. **`u*_high` and `u*_inside` carry `= 1`, and they have to.** Without it `internal/tabletext`'s round-trip test goes red: a field whose declared range excludes zero has an implicit zero default that is *out of its own range*, so the value's default form does not survive ToJson → FromJson (8 clamp events). The checker validates a **declared** default against the range but does not require one — so `x uint8 | min = 1, max = 255` is accepted today and its zero form is unreadable-without-clamping. That looks like a real latent gap (a diagnostic: "a range excluding zero needs a declared default"), and it is a separate change. Want it filed?

3. **The helper is duplicated across `cpptable` and `cstable` rather than shared.** That follows the local convention — `tableKindWidth`, `tableIntLit`/`csIntLit`, `tableFieldTypeName` are all already per-package — and it avoids a new export in `ir`, which is a semver commitment. The alternative reading is that this is emission law two copies of which is two behaviors. I took the convention; happy to hoist it if you read it the other way.


gafferongames:
CI green, all legs. The gcc half is now proven rather than predicted — from the ubuntu `make test` log:

```
g++ -std=c++17 -Wall -Wextra -Werror -Wtype-limits -ffp-contract=off -Ibuild/tables-generated/examples -fsyntax-only build/clamp-limits-tu.cpp
clamp-at-storage-limits gate: the bounded corpus carries no comparison that cannot fire
negative control: the storage-limit clamps back turn the build red — 27 comparisons that cannot fire
```

**27 on gcc against 19 on Apple clang** — the 8 are the unsigned `< 0` rows clang stays silent on, which is the issue's own half. Each compiler reds a strict superset of nothing: the corpus needs both spellings and now carries both.

Legs: test (ubuntu-latest, gcc) pass 11m3s · msvc pass · windows pass · big-endian (s390x g++-13) pass · shape-gate, lint, vuln, cpp-lock, caa pass. macOS clang is not a PR leg by the repo's own budget (it runs on push to main); its half ran on this bench — `build/schema_test_tables`, `test/cs-tables`, `conformance`, and both clamp gates, all green under Apple clang.

PR #352, opened 2026-09-03
tables in c: the tolerant wire, JSON, reflection, the block form, the cook (preview) — mirrored from C++, harness green

gafferongames:
| soak | **704,512 iterations over 18 cases in 3,601 s, live allocation 1,293,504 bytes at the first iteration and 1,293,504 at the end — FLAT.** The golden gate re-saves every exact case byte for byte before the clock; `make test` runs the twenty-second one |
| forgery fuzzer | 100,000+ rounds over five fixtures under ASan/UBSan; it WALKS what it opens, aims half its damage at the words the checks read, biases values into the legal-but-escaping band, and fuzzes the caller's extent and base too |
| the fuzzer's own control | deleting the block reader's extent guards turns it RED as a heap over-read (`make tables-c-fuzz-negative-control`) |
The tables layer in C: the tolerant wire, the read report, the reflection
descriptors, the text form, the block form and the cook — mirrored from
`internal/codegen/cpptable`, with the conformance harness as the gate.

## The matrix

`make conformance`, all five legs, twelve surfaces — `cook-foreign` and
`block-foreign` included, which #338 added to the contract while this branch was
out:

```
surface       cpp            cs             go             rust           c
-----------------------------------------------------------------------------------------
wire          pass 18/18     pass 18/18     pass 18/18     pass 18/18     pass 18/18
report        pass 8/8       pass 8/8       pass 8/8       pass 8/8       pass 8/8
json-read     pass 18/18     pass 18/18     pass 18/18     pass 18/18     pass 18/18
json-write    pass 18/18     pass 18/18     pass 18/18     pass 18/18     pass 18/18
json-hostile  pass 66/66     pass 66/66     pass 66/66     pass 66/66     pass 66/66
cook          pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6
cook-foreign  pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6
block         pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
block-foreign pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
block-dump    pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
forgery       pass 11/11     pass 11/11     pass 11/11     pass 11/11     pass 11/11
cook-forgery  pass 111/111   pass 111/111   pass 111/111   pass 111/111   pass 111/111
```

The C leg's row is the last column: **twelve of twelve, 268 cases.**

The two foreign rows reverse the file's MAGIC word rather than swapping the
host, which makes the file foreign to whoever reads it and is the only shape the
expectation can take without depending on where the leg runs. They are also the
only rows whose expected answer is a REFUSAL, so a driver that never made the
file foreign would pass them by accident everywhere —
`make conformance-negative-control-c-foreign` neuters the swap and requires both
rows red while `cook` and `block` stay green.

Registering the leg was one line in `test/conformance/drivers.txt` plus one
driver, exactly as the contract says. The C leg alone runs in **0.45 s** (median
of three, arm64 macOS) — inside the run-to-run variation of the whole run — so
the two-minute rule is untouched. The harness README's budget table is
re-measured as one sitting so the new row is comparable to the rows beside it.

## The bench

`bench/tables/legs.txt` gains `c bench/tables/c/leg`. **Read the label first:**
this is a PAIRING CHECK on a MacBook Air with sibling port workers on it, and
the results file says of its own numbers that it *"certifies nothing, and no
number here may be quoted as the tables layer's performance"*. The board and
both raw CSVs are committed at
`bench/tables/results/2026-09-03-pairing-c-arm64-macbook.{md,csv}`. Seven rounds,
cpp and c INTERLEAVED within each round so both legs see the same window,
median of seven:

| row | cpp | c | ratio |
|---|---|---|---|
| write | 2,405,517 msg/s | 2,418,263 msg/s | **1.005x** |
| round_trip | 819,897 msg/s | 851,435 msg/s | **1.038x** |

Spreads 0.8–2.6%. **C is at parity with C++** on the tolerant wire, marginally
ahead on both rows — which is the ladder's "same speed, or not significantly
slower", answered.

### #350's lever, measured in C

Paired before/after, identical source with the qualifier defined away in the
BEFORE arm, seven rounds interleaved, median of seven:

| row | before | after | ratio |
|---|---|---|---|
| write | 469,413 | 2,332,375 | **4.97x** |
| round_trip | 334,800 | 830,851 | **2.48x** |

Banked before measuring: *"write 2x to 5x, round trip 1.5x to 2.5x; under 1.2x
on write would be a refutation."* Both landed inside, and within a few percent
of C++'s own 5.31x / 2.40x — the mechanism transfers whole, because it is the
same mechanism: the cursor lives in the caller's `TableWriter` and a `uint8_t *`
store may alias it, so an out-of-line put reloads and re-stores it. The
qualifier stops at the fixed class for the reference's reason: a pointered body
reaches its pointee through the depth-carrying form, and a recursive
`always_inline` does not compile under gcc.

## What the port had to spell differently, and why

C is not C++ with different punctuation, and four mechanisms have no direct
translation. Each is the C form of one C++ mechanism rather than a new idea, and
each is stated in the package doc:

- **No overload set for enum identity.** C++ finds `TableEnumId` by overload
  resolution. C emits the switch AT THE USE SITE — the same switch, inline,
  resolving no call — which claims not one name per enum.
- **The descriptor's vocabulary columns are DATA.** C++ spells an enum's names
  and wire ids as captureless lambdas; C has none, and a named function per enum
  would claim a name per enum. The same facts ride as a `TableVariantInfo` table
  indexed the way `enum_max` bounds.
- **The keyed array is a plain array.** C's storage IS `T slots[E_MAX]`; what
  C++ puts in `operator[]` lives in `TableKeyedAt` over `TableKeyedSlot` —
  assert plus abort, in every build, held under `-DNDEBUG` by
  `make tables-c-keyed-none-refusal-ndebug` with its own negative control.
- **The header/source split is wider.** C has no inline variables, so the
  descriptors and the text form's walk are DEFINED in `<Base>Table.c` and only
  declared in the header. A TU that includes the header for the wire pays for
  neither.

**Every external the backend emits carries the package** —
`schema_<package>_<type>_<what>_` — because C has no namespace and two units
whose type names collide have to LINK together, which is what the conformance
driver itself does with `tblv1`'s `Cfg` and `tblv2`'s `Cfg`. The name-first
surface §11 states is `static` in the header and forwards to them, so a call
site reads the same in C as in C++.

## A defect class C has and the other two do not

**The preprocessor.** A schema's constants, enum variants and flag masks are all
`#define`s in the C target, and the generated sources define macros beside them.
A collision there is not a redeclaration error — it is a SILENT REWRITE: the
generator's `#ifndef` sees the user's definition standing, skips its own, and
every later use expands to something else with nothing in the build saying so.

The C target now reserves `SCHEMA_`/`schema_`, every macro this backend defines
moved under it, and `internal/check` refuses a declaration that spells one of
the macros the generated C actually defines. **The set is enumerated, not a bare
prefix test**: `enum SchemaKind` folds to `SCHEMA_KIND_ALPHA`, which collides
with nothing, and refusing it would take a name from every schema for free —
`TestCReservedMacroNeighborsAccepted` is that half.

This also closes an EXISTING exposure in the packet emitter — `enum Schema
{ Unused }` folds to `SCHEMA_UNUSED`, which is that emitter's own macro — so the
refusal is unconditional rather than table-only. Five repro cases join
`internal/check`'s break-the-language suite, and
`compiler.TestCGeneratorMacrosAreOwned` holds the two halves together by naming
every declaration with one prefix and looking at what is left: a scan that
recognises no spelling and no family.

The split between the two registries is deliberate. `internal/tablenames` holds
the RUNTIME IDENTIFIERS — the `Table*`, `table_*` and `kTable*` families this
backend defines — because that is what it is for and what the §11 claim reads.
The MACRO set lives in `internal/check` instead: it spans both C emitters, the
packet one included, and half of it predates tables entirely, so hanging it off
a table registry would have made a table-free unit's exposure invisible.

## §11's claims

`internal/tablenames` gains a `C` backend bit and the C backend's own spellings.
The list is longest for this backend, and for one reason: C has no namespace to
put a runtime in and no nested scope to hide one in, so a spelling the other two
can scope away is a unit-level name here. Everything a consumer never types
carries the `schema_<package>_` prefix instead and claims nothing.

`compiler.TestCTableRuntimeNamesAreClaimed` is the C# scan's twin over the C
output and asserts both directions. `internal/check`'s `tableGeneratedVerbs`
gains seven name-first spellings C needs and C++ does not, because C++ puts them
on a class and a member function claims nothing: `BuilderInit`,
`BuilderShutdown`, `BuilderLock`, `BuilderRoot`, `BlockStorageCreate`,
`BlockStorageDestroy`, `BlockType`.

## One thing the conformance corpus does not reach

Every `instance` in the manifest is FIXED, because the harness's wire goldens
are — so the whole VARIABLE class (the arena, the builder, `Lock`'s pack walk,
`LoadMeasure`, the region sink, the self-relative deref) has no conformance case
in any language. I wrote the port's pointer half and would have shipped it
unexercised, which is not a thing to notice after merge.

`test/c-tables/variable_main.c` is the gate: build a five-node chain through the
arena plus a pointed-at fixed table and a by-value nested variable table, Lock,
and prove **the mutable form and the locked region write the same bytes**; then
size a region from the wire's framing alone, load into it, and walk it back out
through the region's own self-relative derefs. It runs plain and under ASan +
UBSan, because the arena, the pack walk and the region sink are the three places
this backend does pointer arithmetic on caller-owned memory.

## The negative control that changed a gate's design

I wrote the forgery fuzzer to OPEN forged files and assert it never crashes,
then asked what would make it go red. Deleting the block reader's
rows-inside-the-extent check: **it stayed green.** An `Open` validates and
points — it reads no row — so a removed guard produces a wrong `open` and no
symptom at all. The fuzzer was proving the reader does not crash on bytes it
never looks at.

Three things came out of chasing that:

1. **The fuzzer now WALKS what it opened** — every row of every array, every
   node of the graph, which is what a consumer does and where a guard that
   stopped guarding reads past the caller's buffer.
2. **The walk immediately found real UB in my own driver**: the cook walk did
   `slot + delta` with an attacker-controlled delta and range-checked the
   result. A forged delta near 2^63 overflows the pointer, so the check was
   examining a value the standard never promised. It is integer arithmetic,
   bounded before it becomes a pointer, now.
3. **The damage is aimed.** A block image is 2,816 bytes and the words an Open
   reads are the first hundred-odd of them, so uniformly random damage lands in
   a row essentially every time; and a uniformly random 64-bit value is never a
   legal-but-escaping count. Half the offsets now go to the prologue region and
   half the values are small or near the file's own.

And one thing worth recording about the READER: deleting the extent check alone
leaves it correct, because the padding check catches the same forgery as a side
effect. The control names both lines for that reason — a control that deletes a
redundant guard is a control that proves nothing.

## The gates

| gate | how |
|---|---|
| conformance, twelve surfaces | `make conformance` — green above, with a negative control per half: one field index off in the C walk turns `json-read` red alone, and a driver that never swaps the magic turns the two foreign rows red alone |
| the same, under ASan + UBSan | `build/conformance-c-asan`, all ten green |
| negative control | one field index off in the C walk turns `json-read` RED, and `json-write` and `wire` stay green (`make conformance-negative-control-c`) |
| zero-cost | `make tables-c-zero-cost` — no pointer machinery in a value-only unit's header |
| block zero-cost | the C Table sources carry no block symbol, and the pinned Table files rose from 38 to 68, byte-identical to their pins |
| one generic walk | `make tables-c-json-walk` — byte-identical in 19 generated `.c` files |
| forgery fuzzer | 100,000+ rounds over five fixtures under ASan/UBSan, extent and base fuzzed too; refuses to pass if nothing refused |
| the VARIABLE class | `make tables-c-variable`, plain and sanitized |
| soak | `make tables-c-soak` — allocations FLAT; the golden gate re-saves every exact case byte for byte before the clock |
| keyed None refusal | held under `-DNDEBUG`, with a negative control that deletes the abort |
| big-endian | `make tables-c-big-endian`, riding `tables-big-endian` — **passing in CI on s390x under qemu** |
| the documented surface | the USAGE C example compiles and runs against the corpus — a doc example that does not build is a lie |
| `make test` | the C leg rides the tables chain |

## Big-endian

`tables-c-big-endian` PASSES in CI on s390x under qemu — the one leg I could not
run on this Mac, because the pinned cross toolchain is Linux-only. It is the
soak binary cross-compiled and run for zero seconds: the golden gate re-saves
every exact case and byte-compares, which is exactly the assertion the tolerant
wire's little-endian-by-construction rule makes. CI's step gains `BE_CC` beside
the `BE_CXX` it already named.

## What CI caught that this bench did not

The forgery fuzzer read `build/cook-open/*.cook` without declaring them: they
happened to be lying around here and were not on a clean runner. A fuzzer whose
subject happens to exist is a fuzzer that passes by not running, so the fixtures
are now a declared prerequisite and the target was re-verified after deleting
them.

## The diff's shape

About 6,200 lines of Go emitter, 2,000 lines of C driver and gates, and the rest
is PINNED GENERATED C under `testdata/golden/tables/{examples,block,pointers}-c`
— the same header/source pair the C++ and C# tables goldens already commit, on
the same terms and re-pinned by the same `make update-goldens`.

## The blind read's fixes

All four code defects the review named, plus the pages:

1. **`generated/bench/tables/c/BenchTable{Block,Table}.h` were stale** against
   the guard rename — `make generated-current` failed and I had not run it.
   Regenerated; both headers move.
2. **`#define BuildVersion` is gone.** An object-like macro carrying a common
   PascalCase identifier rewrites it everywhere in the CONSUMER's translation
   unit — `struct AppBuild { uint64_t BuildVersion; };` after the include stops
   compiling — and no front end can refuse that, because the collision is with
   the consumer's own code and not with a schema declaration. It is
   `SCHEMA_<PKG>_BUILD_VERSION_VALUE` now, under the reserved prefix, and
   `internal/check` reserves that spelling too. It stays a macro rather than a
   `static const` for one reason worth stating: the block descriptors are
   constant data that name it in their initialisers, and a `static const` is
   not a constant expression in C. The object-like `Table*` constants the
   review named in the same breath — `TableCookMagic`, `TableCookByteOrder`,
   `TableCookMaxAlign`, `kTableCookHeaderBytes`, `TableBlockMagic`,
   `TableBlockByteOrder` — ARE now `static const` and no longer macros at all.
   `kTable*` stay macros because they are array bounds, and `TableKeyedAt`
   because it must yield an lvalue.
3. **The soak was a leak instrument printing an allocation claim.** It now
   counts allocator CALLS over the measured loop by interposing
   `malloc`/`calloc`/`realloc`/`free` and forwarding to the platform's real
   allocator by a spelling that does not recurse (`__libc_*` on glibc, the
   default zone on macOS), and requires **zero**. Measured: 0 over 3,072
   iterations, which is the review's own number. `make
   tables-c-soak-negative-control` plants one matched malloc/free pair per
   iteration and demonstrates the point exactly: the drift gate stays silent
   and the call count refuses at 36,864.
4. **The force-inline comment claimed a case that cannot happen.** Two units'
   headers in one translation unit is dozens of redefinitions, not a macro
   collision. The comment now states the real rule — the names carry the
   package because two units LINK into one program, and they do not share a TU
   and cannot — in both the emitted text and the emitter.

Pages: §2.1's pointer slot is **eight** bytes, not four (§6.3's number, and
`sizeof(TableRef)`). §11's suffix count reconciles to the checker's **40** with
the C backend's seven listed. The ladder no longer says "C and C++ take the
allocator" for the variable class — neither does; the arena and `Lock` call
`calloc`/`malloc` by name, the caller owns the LOAD region, and the
caller-provided triple is the BLOCK form's, where it is real. §6.2 and USAGE
gain the C variable-length surface — `TableSink`, `TableCtx`, `<T>Emplace`,
`<T>At`, the builder functions and the members that stand in for C++'s
`AsConst`/`Region`/`RegionBytes`/`Locked` — and both examples were compiled
verbatim before they were written down. USAGE's C section now reads whole
instead of splitting C#'s in two. Three stale statements my landing falsified
are corrected (`SPEC-TABLES.md` on the pointered unit, `bench/tables/README.md`
on C having no tables backend).

**Not fixed, reported.** C refuses `None` and not a key past `E.Max`. That
mirrors C++, which is the point of the port, but the five backends do not agree
with each other, so here is the state of each rather than a phrase:

| backend | refuses `None` | refuses a key past `E.Max` | the line |
|---|---|---|---|
| C++ | yes | **no** | `if ( key == E::None )` — `internal/codegen/cpptable/cpptable.go:205`, in `TableKeyed<T,E>::RefuseNone`, `assert` then `abort` |
| C | yes | **no** | `if ( key == 0 )` — `internal/codegen/ctable/ctable.go:215`, in `TableKeyedSlot`, `assert` then `abort` |
| Go | yes | **yes** | `if key <= 0 || key > len(slots) {` — `internal/codegen/gotable/gotable.go:820`, one `panic` for both |
| C# | yes | yes, but not from the emitted code | `if (key == 0)` — `internal/codegen/cstable/cstable.go:648`, `throw ArgumentOutOfRangeException`; past `E.Max` is the CLR bounds check on `Slots[key - 1]`, an `IndexOutOfRangeException` |
| Rust | yes | yes, but not from the emitted code | `if key == 0 {` — `internal/codegen/rusttable/runtime.go:135`, `panic!`; past `E.Max` is the slice index panic where the returned `usize` is used |

So: **Go is the only backend whose accessor refuses both.** C# and Rust refuse
both in effect, but the upper half is their runtime's bounds check rather than
anything the emitter wrote — the message a caller sees is the language's, not
the schema's, and it names neither the key enum nor `E.Max`. C++ and C refuse
`None` only: `slots[key - 1]` past the end is a read of memory that is not a
slot, in every build.

Closing it in C alone would make C and its reference disagree about a
documented refusal, so I have not. It belongs with §2.4's "no
undefined-behaviour path in any build", which reads wider than the None case,
and the fix is one line in each of C++ and C. Same for the
shared-node-packed-twice finding, which the review confirms is the reference's
behaviour too.
## Pages

SPEC.md §6.1's C row gains the table pair, the wider split, the
package-carrying externals and the reserved prefix. USAGE gains a C section —
raw material; the design seat writes the prose. SPEC-TABLES.md's backend-status
lines gain C where the port now qualifies. CLAUDE.md records the preprocessor
rule as a learning.

## Not in this PR, named

- **§3.1's flat node table (kind 17) is not on the wire in any backend.** The
  constant lives in `ir/buildversion.go` and the reference rides a pointer as a
  kind-13 nested body — framing identical to a by-value nesting, which is what
  makes the three framings one framing. This port mirrors that. If §3.1's node
  table is meant to be live, it is a wire change for every backend at once and
  not something a port invents.
- The variable class's TEXT FORM (§16.1, schema#275) — absent by absence, loud:
  a pointered table gets no `FromJson`, and the header says so.

## The CI budget

The C leg's per-build cost is trimmed three ways, none of which weakens a gate:
two seconds of the soak ride `make test` instead of twenty (what the Go leg
already does — the gate is the allocator CALL COUNT per iteration and it reads
the same at either length); the fuzzer runs 25,000 rounds there and the full
100,000 under `make tables-c` (the enumerated passes cover the boundaries
whatever N is); and the two conformance negative controls compile at `-O0`,
since a control is built to be run once and thrown away and the optimiser was
most of a twenty-eight-unit build. Together about a minute.

That work predates the rebase: #372's split had already moved `make test` off
the pull-request path, so the timeout raise it came with is dropped and the
trims are kept on their own merits.

## The C spelling (the Java blind read's rule)

A backend's table half uses the identifier convention its PACKET half already
uses in the same language. C was the only backend where the two disagreed:
`generated/c/` writes `write_arms_agree` and `schema_utf8_valid_`, and the table
half was writing `GunnerConfigLoad`, `TableWriterPut32` and `TableKeyedAt` —
C++'s casing in a C file.

So §11's `<Name>Load` is `<name>_load` here, as it already is in Rust and as it
already is `<Name>Load` in Go, C# and C++. §11 names the SUFFIX SET a
declaration may not collide with; each backend spells that set in its own
language, and the claim is untouched because the checker refuses the
DECLARATION, not the emitted identifier.

89 per-type verb sites and 86 runtime functions moved to snake_case, six runtime
constants with them, `TableKeyedAt` to `SCHEMA_TABLE_KEYED_AT` and
`<X>BlockMaxBytes` to `SCHEMA_<PKG>_<NAME>_BLOCK_MAX_BYTES` — no unprefixed
macro survives in this backend. Types stay PascalCase, which is what the packet
half does. The registry SPLITS where a name is shared: the same runtime value is
one identifier in Rust and a different one here, so both are registered and both
are claimed for every target.

Two gates went vacuous under the rename and are repaired rather than deleted:
the force-inline test's forbidden half now matches the C spellings, and a fresh
negative control forces a variable-class body to carry the qualifier and
requires that half to go red. The rule is stated at the emitter site, in SPEC
§6.1's C row and in §11 — on the page, because it is every backend's rule and
not C's.

## The rebase

Rebased onto `b16227f2` — main with Java (#356) and the CI split (#372) on it.
Thirty-four commits replayed; every conflict resolved as a UNION — the C row
sits beside Java's and Go's and Rust's in `drivers.txt`, `legs.txt`,
`tablenames`, the budget table, every backend-status line and §11's lists, and
replaces nothing. `USECASES.md` stays deleted (#365).

`ci.yml`'s per-leg conformance matrix gains its C entry, written exactly as the
cpp and go entries are, with `build/conformance-c` as its build target and a
line saying the leg names no sibling runtime. `tables-c-release` is derived by
`certify.yml` from the Makefile, so no other workflow moved.

The harness budget table is re-measured as ONE sitting across all six legs,
because a table whose rows come from different sittings can say the whole costs
less than one of its parts. The laptop was busy for it and the entry says so.

Head: `f5ca043`. The two-minute gate is green on every check, `conformance (c)`
included.

Draft, not for merge.


gafferongames:
**LAND-AFTER-FIXES**

Blind read: a C consumer of `tables/examples` and `tables/pointers` written from the pages alone, touching the generated code only through its declarations.

**Reproduced.** All 8 tabledemo wire goldens load with an all-zero report and re-save byte-identical. My own reflection walker's JSON is *semantically identical* to all 8 conformance goldens — the descriptors are complete. Consumer + all 10 generated `.c` build under `-std=c99 -Wall -Wextra -Werror -Wshadow -Wconversion -pedantic` (stricter than the repo's C flags) and run clean under `-fsanitize=address,undefined -fno-sanitize-recover=all`. Variable class: 1 calloc (4,194,304 B) at `BuilderInit`, **0 for 106 `Emplace`**, 1 malloc + 1 free at `Lock`, live back to 0; `Load` allocated nothing; locked and loaded regions walk identically and re-save reproduces the 3,492-byte wire. Cook: `SceneOpen` on a 6,112 B fixture, 122 nodes walked; magic / build_version / byte_order / short length each refused. Block: **1 allocator call at `BlockStorageCreate` (959 B), 0 at `Begin` and per row**; `BlockOpen` accepts a memcpy'd copy, refuses flipped magic/build_version/short extent; `Begin` at 4 rows refuses naming `reserves` 4 > 3. Two units with colliding `Cfg` link into one program. Fuzzer walks what it opens: 3,800,000 rounds, seed 1788421999, 5m55s under ASan/UBSan — 1,914,083 opened, 1,885,917 refused, no crash, no over-read; control reproduced as a heap over-read. Conformance: C leg alone **260/260 in 0.44 s user / 2.3 s wall**; full matrix green. Keyed `None` under `-DNDEBUG -O2` **dies (SIGABRT)**. Checker refuses `BuildVersion`/`TableKeyedAt`/`kTableAlign`/`CfgBuilderInit`; `enum SchemaKind` passes; all 134 runtime names the emitter spells are in the 161-name registry.

**Defects**

1. `generated/bench/tables/c/BenchTable{Block,Table}.h` — **committed tree is STALE**: still the pre-rename guards `BENCHTABLE_SCHEMA_*`; the emitter now writes `SCHEMA_BENCHTABLE_*` (54 lines each). `make generated-current` fails. Fix: `make bench-tables && git add generated/bench/tables/c`.
2. `examples-c/TablesTable.h:380` (every `<Base>Table.h`/`Block.h`) — `#define BuildVersion 0x…ull`: an unprefixed PascalCase object-like macro. It rewrites the identifier everywhere in the consumer's TU — `struct AppBuild { uint64_t BuildVersion; };` after the include fails to compile — and the front end cannot refuse it, because it is not a schema declaration. C++ scopes it in a namespace. Fix: spell it `SCHEMA_<PKG>_BUILD_VERSION_VALUE` (the guard name exists) or a `static const uint64_t`; same shape for `TableCookMagic`, `TableKeyedAt`, `kTable*`.
3. `test/c-tables/soak_main.c:225-246` — the soak is a **leak** instrument, not an allocation one: it samples `live_bytes()` after iteration 1 and at the end, fails only on drift, then prints "the read and write paths allocate nothing". A malloc/free pair per iteration is invisible to it. Counting allocator over one iteration of the same paths (root_full, profile_elide, keyed_config, wide_blob, archive): **malloc=0 calloc=0 realloc=0 free=0**. Add the call counter and require zero after warm-up — the gate is free today.
4. `TablesTable.h:~44` — the comment justifies the package-scoped inline macro with "a consumer compiles several packages' Table.h files in one translation unit". That is dozens of redefinitions (`TableReport`, `TableUnionArmInfo`, …), and SPEC.md:1396 says two units cannot share a TU. Fix: state the real reason.
5. The C builder drops four C++ accessors (`AsConst`, `Region`, `RegionBytes`, `Locked`) and both builder overloads of `Measure`/`Save`. Post-`Lock`, `SceneBuilderRoot` is NULL (correct), so a C caller must write `(const Scene *) builder.region` plus a zeroed `TableCtx` — struct-member reach-in documented nowhere.
6. `TableKeyedSlot` refuses `None` but not `> E.Max`: key 9 on `[Team]` (Max 3) read out of bounds and returned 1065353216. Mirrors C++, so a reference-level gap — but §2.4's "NO undefined-behaviour path here in any build" reads wider than the None case.

**Page defects**

- `SPEC-TABLES.md:313` — a "**four-byte** relocatable reference"; §6.3:1672,1699 and §7.2:2419 say eight, and `sizeof(TableRef)` / `elem_size` are 8.
- `SPEC-TABLES.md:3515-3517` — "the count and the spellings here equal `tableGeneratedVerbs` exactly". Page lists 33, checker holds **40**. Unstated: `BuilderInit BuilderShutdown BuilderLock BuilderRoot BlockStorageCreate BlockStorageDestroy BlockType`.
- `SPEC-TABLES.md:135-137` — "C and C++ take the allocator" for the variable class. Neither does: the arena hardwires `calloc`/`free`, `Lock` hardwires `malloc`. The caller-provided triple is the **block** form's (§19.1), where it is real.
- `SPEC-TABLES.md:1654-1661` and `1114-1118` — "a shared node is packed ONCE" / "one node … in a builder and in a region alike". Measured: an aliased node is packed **twice** by `Lock` (region 240 B vs 208 B), identically in C and C++. §3.1's honesty note (869-885) disclaims only the wire.
- Stale: `SPEC-TABLES.md:874-878` "the only one that accepts a pointered unit"; `USAGE.md:1659` "cpp and cs today"; `bench/tables/README.md:174-175` "C has no tables backend".
- `SPEC-TABLES.md:2939-2941` — the descriptors' **node type id** is claimed and emitted by no backend.
- Nothing says how a walker separates `bytes(N)` from `[..N]uint8` — identical columns; only `type_name == "bytes"` does, and §16.2 needs base64 for one.
- §6.2/§6.3 and USAGE's *Pointers* give **only C++**. The entire C variable surface (`TableRef`, `TableSink{region,worker}`, `TableCtx{arena}`, `<T>Emplace`, `<T>At`, `<Root>Builder*`) is on no page; `USAGE.md:766-796` also interleaves C and C# storage paragraphs under no heading.

**Bench.** Two paired sittings on a loaded Air: C/C++ write 1.112 then **1.018**, round_trip 1.156 then **1.048** (C++ round_trip spread 17.7% then 7.9%). Direction agrees with 1.005/1.038 and the quieter sitting sits within noise of it — no refutation. `results/2026-09-03-pairing-arm64-macbook.md` labels Air numbers a pairing check that "certifies nothing"; the body quotes 1.005/1.038 unlabelled, with no CSV committed.

**§3.1 kind 17.** Confirmed: a C pointer field rides `id, kind=0x0d, L, body` (`aa 79 0d 09 …`), no node table, exactly as C++ does; the page's honesty note is honest. One consequence worth a line: `schema cook` on a backend-written pointered wire reports `kind_mismatch 5` and refuses without `--tolerate` (schema#251, pre-existing).

PR #353, opened 2026-09-03
tables in JavaScript: the tolerant wire, JSON, reflection, the block and cook readers — mirrored from C++, harness green

gafferongames:
The fixed class on the tolerant wire (§3) with the read report (§4), the text form
(§16) as one generic walk over the descriptors (§8), and the READING side of both
accelerators — the block form (§19) and the cook (§7). Mirrored from C++, with C#
as the worked second port; no contract invented.

**The conformance row is green on all twelve surfaces**, at the same counts the
six existing legs answer:

```
surface       cpp            cs             go             js             rust           java           c
-----------------------------------------------------------------------------------------------------------------------
wire          pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16
report        pass 12/12     pass 12/12     pass 12/12     pass 12/12     pass 12/12     pass 12/12     pass 12/12
json-read     pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16
json-write    pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16     pass 16/16
json-hostile  pass 66/66     pass 66/66     pass 66/66     pass 66/66     pass 66/66     pass 66/66     pass 66/66
cook          pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6
cook-foreign  pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6
block         pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
block-foreign pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
block-dump    pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
forgery       pass 11/11     pass 11/11     pass 11/11     pass 11/11     pass 11/11     pass 11/11     pass 11/11
cook-forgery  pass 111/111   pass 111/111   pass 111/111   pass 111/111   pass 111/111   pass 111/111   pass 111/111
```

It runs as its own shard in the split PR gate (#372) — `- lang: js`, targets
`build/conformance-harness build/tables-generated-js/.stamp`, setup-node, and no
sibling clone because generated JavaScript imports none — and answers all twelve
surfaces in **3.2 s**, so the two-minute rule (#320) has the whole of its budget
left. Its expensive half is `tables-js-release`, which certify.yml derives from
the Makefile with no edit of its own: the fuzz oracle at ten times the PR scale,
the allocation gate at seven times the iterations, and ten minutes of soak.

## The reading tier, stated

JavaScript has no struct layout — no offsetof, no sizeof, no way to place a
field — so the two accelerators are READ and never produced: every field is read
at the offset the compiler settled, through a `DataView` with an explicit
little-endian flag at every call. The tolerant wire is the one form this backend
both writes and reads, because that wire is a byte stream and not a layout.

The VARIABLE class on the wire — the arena, the builder, the region and the flat
node table (kind 17) — is a named follow-on, refused by name exactly as C#
refuses it, with both accelerators emitted all the same. **The tolerant reader
still SKIPS a node index**, which is what #376 asked of every fixed backend the
day it moved the pointered wire: a P2 message read by the P1 or P3 codec is one
unknown id, one kind mismatch, and a usable value out the other side.

## What allocates — and the three CI reds it took to name it

A flat heap is a LEAK instrument and nothing more: an allocation made and
collected every iteration leaves it exactly as flat as no allocation at all. So
`make tables-js-alloc` measures **bytes allocated per iteration, per path**, and
gates the paths that must be zero. Its negative control puts one extra
allocation per iteration in and requires them to go red.

```
                              bytes  ceiling  sample  instrument
  KeyedConfig Load              0.0   <= 8    1/4096    0.74   no 64-bit field in its closure
  KeyedConfig Measure           0.0   <= 8    1/4096    0.78
  KeyedConfig Save              0.0   <= 8    1/4096    0.65
  RootConfig Load              67.1   <= 512  1/3641    0.55   one BigInt per 64-bit field it carries
  RootConfig Measure            0.0   <= 8    1/4096    0.78
  RenderFrame ships walk        0.0   <= 8    1/4096    0.86   a block row read at its offset
  Scene head deref              0.0   reported 1/4096   0.82   a cook reference followed
  RootConfig ToJson         16041.9   reported 1/64    10.16   the generic path, licensed
  RootConfig FromJson       12161.2   reported 1/64    10.16
```

**This gate went red on CI three times, and my first two answers were wrong.**
I called it instrument noise, then relaxed the ceiling to 64 bytes with a "four
different machines" argument. Both were inverted: four runners on one pinned
node is ONE runtime read four times, and a steady fifteen bytes is what a small
allocation looks like, not what noise looks like. The estate's law is that an
unexpected red freezes the work and gets debugged to solid ground.

**The cause: a double that crosses a JavaScript call boundary is a HEAP NUMBER**
— sixteen bytes, per call — unless V8 inlines the callee. `w.putF32(x)` and
`r.getF32()` were exactly that boundary, once per float field. Whether V8 inlines
them depends on the size of the body around the call, so the same helper
allocated in a big table's codec and not in a small one's, and the same build
allocated in one process and not the next. That is the estate's law that a
generated codec must not depend on the compiler's inlining budget, arriving in
a JIT.

The evidence, in the order it was taken:

- Reproduced under the CI-pinned runtime (`dist/node-v20.20.2-darwin-arm64`),
  where the developer's node 26 read zero: Load 15.6, Save 15.5, Measure 0.0.
  The empty-loop control on that runtime reads 0.00.
- Cross-checked off the heap-delta instrument entirely: `--trace-gc` counts 39
  scavenges over five million iterations against 1 for the empty control.
  Calibrated against a known object — empty 0.00, empty+1 object 34.27, the read
  15.02, read+1 object 44.53. (The inspector's sampling profiler reported 27 KB
  over three million calls and named nothing: it was the instrument at fault
  here, which is why "it is the instrument" needed a control first.)
- **The delta histogram named it.** In a 15.1 window there are 73 positive
  samples at the sample interval, each 65,608 bytes over 4,096 iterations —
  16.02 bytes an iteration, evenly spread. One sixteen-byte object per call,
  every call; not one heap step misattributed to a body.
- Bisected to the field: `TeamConfigSaveBody`, which writes ONE float, reads
  0.00 in every run. `HullConfigSaveBody`, which writes TWO, reads 15.56 and
  sometimes 30.25 — one box and two.
- `--trace-opt`/`--trace-deopt` are IDENTICAL between a 0.0 run and a 15.1 run.
  The bodies are optimized in both; what differs is which callees were inlined,
  which no trace prints and which the histogram does.

**The fix**: the generated body reads and writes a float through the view itself,
so no double crosses a call on a codec path at all. The four float primitives are
gone from the shared runtime with nothing left calling them. Five consecutive
runs of the whole gate on the pinned node now read identically; under the old
emission the same gate read 15.1 on a randomly varying path in three runs of four.

**And the gate refuses to certify off the pin.** The zero-allocation claim is a
property of OPTIMIZED code — the same source reads 1192 bytes per iteration under
`--no-opt` — so it can only be made on a named runtime. The leg holds
`PinnedNodeMajor` and refuses any other major unless `SCHEMA_JS_ALLOC_ANY_NODE=1`;
the Makefile routes every JavaScript gate through `NODE`, pinned to the version
CI runs and unpacked in `dist/` beside the Dart, Java and BEAM toolchains. The
harness also measures ONE PROCESS PER PATH: nine bodies through one call site
made it megamorphic, and a number taken there is a number about the harness.

§19.5 states the rule for every port that follows — a float field is read and
written by the generated body, never across a call — and states what a port
cannot close this way: its own public accessor returning a float to a caller.

## The casing rule, stated once

A consumer sees ONE unit, not two halves of one, so every identifier this
backend emits is spelled the way the JavaScript PACKET emitter spells the same
kind of thing. Types, functions and constants were already UpperCamelCase; the
DATA MEMBERS were not, so a reader wrote `value.Gunner` on one line and
`info.fields[0].name` on the next, out of the same generated file. They are
`Name`, `Kind`, `Fields`, `ArrayBound`, `GetChild` now, and so are the reader's
and writer's state (`Offset`, `View`, `Bytes`, `Limit`, `Report`, `Overflow`)
and the block and cook handles (`Bytes`, `View`, `Region`, `Length`, `Used`).
METHODS stay lowerCamelCase — the only methods generated packet code calls are
the stream's, and a reader's `get32` sits in exactly that position — and so do
locals and IIFE-private bindings, which nothing outside can name. What the
SCHEMA spells keeps the schema's spelling. The rule is in the emitter's package
doc, beside the storage rule it follows from.

One case-only bug came out of that pass and CI caught it: two `".json"` file
names had been swept up with the descriptor member, which a case-insensitive
macOS filesystem cannot see and the Linux conformance shard can.

## The instruments, and what each one found

- **The fuzzer oracle** over both accelerators' readers: a forged block or cook
  either refuses, or opens and is walked in full through its descriptors without
  one read leaving the bytes the caller passed. In a language where a read past
  a view THROWS, "refuse, or open and be whole" and "no exception escapes a
  reader" are the same sentence. Two negative controls remove a bound each and
  require it to go red.
- **Accessors against descriptors**, both accelerators: every field of every row
  read both ways and compared, with a sabotaged-emitter control for each.
- **The text differential against a THIRD implementation** — the compiler's own
  Go engine — which found a `%g` tie-break the pinned goldens could not: C rounds
  half-to-even and `toExponential` rounds by magnitude, so `-266744.625` rendered
  `-266744.63` against C's `-266744.62`, and both candidates round-trip to the
  same float32, so no shortest-precision search steps past it. 199,817/199,817
  agree with C `snprintf` now.
- **WHAT ALLOCATES, as a rate** — above. Before the float boundary it had already
  found two real allocations: a block row accessor reading its `offset_of` with
  `getBigUint64` (247 B/iter) and the cook's `At` reading its delta the same way
  (37 B/iter). Both are composed from two 32-bit reads now.
- **The other byte order, refused twice over**: the harness's `cook-foreign` and
  `block-foreign` surfaces hold the MAGIC half as data across every leg; the leg
  keeps the ORDER WORD half, which no surface holds — a file whose magic is
  intact and whose order word was forged back is exactly what a reader leaning on
  one check would open.
- **Zero cost for a unit with no tables**: the packet output is byte-identical
  with and without the table backend, and a unit that declares tables adds
  exactly the three table modules.

## The soak, and the board

**One hour, 393,968,000 round trips, gated on the allocation COUNT and not on
heap drift.** A flat heap is a leak instrument and nothing more, so the soak
re-measures the same paths after the hour and fails if a rate GREW:

```
                       before      after
  KeyedConfig Load        0.0        0.0     zero-floor
  KeyedConfig Measure     0.0        0.0
  KeyedConfig Save        0.0        0.0
  RootConfig Load        67.3       67.3     the BigInts it must have
  RenderFrame walk        0.0        0.0
  Scene head deref        0.0        0.0
  heap 6.60 MiB -> 6.98 MiB across 98,002 samples (5.7%)
```

**The bench board** (`bench/tables/results/2026-09-03-pairing-js2-arm64-macbook.md`)
is a PAIRING CHECK on a MacBook Air and certifies nothing. Seven interleaved
rounds, seven legs:

| path | cpp | c | rust | java | js | go | cs |
|---|---:|---:|---:|---:|---:|---:|---:|
| write | 1.000x | 0.932x | 0.871x | 0.655x | **0.157x** | 0.142x | 0.141x |
| round_trip | 1.000x | 0.990x | 0.810x | 0.549x | **0.248x** | 0.269x | 0.258x |

JavaScript is the fastest of the three managed legs on write and a shade behind
them on round-trip, which is the shape the allocation gate predicts: the write
ratio is one runtime against another, and the round-trip ratio carries the one
BigInt per 64-bit field READ, on a corpus with six of them per record. An
earlier sitting the same day is **discarded whole** — three rows over §2.3's 40%
invalid line, because a soak and five other builds were on the box — and named
in the board's prose rather than averaged in.

## Not in this PR, by name

- **The VARIABLE class on the wire** — refused by name, both accelerators
  emitted all the same, exactly as C# refuses it.
- **A per-BYTE claim.** §12.1's block form is where the per-frame JavaScript
  path lives, and this leg's corpus is the tolerant wire.
- **An `inline` verdict.** Every §4 verdict pass is ahead-of-time disassembly
  and a JIT has no build-time artifact to gate; the js bench rows carry
  `inline=unknown` by construction, which is why they never enter a ratio table.

**Do not merge** — this is the reading tier's first port and the coordinator's
call.


gafferongames:
## Blind read — the JavaScript tables port

**Verdict: LAND-AFTER-FIXES.** The bytes are the reference's on every surface I could reach with my own consumer, the refusals fire, the instruments are real measurements with controls that go red. Two contract misses, one measured claim the page overstates, and a handful of page gaps stand between this and the port the pages describe.

### Defects, ranked

1. **`TableKeyed` refuses `None` and nothing else.** `internal/codegen/jstable/jstable.go:743-756` (`refuseNone`). §2.4: "REFUSES EVERY KEY THAT NAMES NO SLOT — `None` below the storage and anything past `E.Max` above it … THE GUARD IS SYMMETRIC … in every port, AT BOTH ENDS." Repro on `tables/examples` `KeyedConfig`: `cfg.Teams.get(4)` → `undefined`; `get(-1)` → `undefined`; `cfg.Teams.set(4, t)` → `Slots.length` becomes 4, `for..of` then yields key 4 (a key no variant names; §2.4 says iteration offers valid slots only), and `KeyedConfigSave` walks a constant 3 and silently drops the slot (Measure 2, Save 2, Load back silent). Silent loss, no throw. Fix: one unsigned compare in `get` and `set` — `((key - 1) >>> 0) >= this.Slots.length` → `RangeError`. The leg (`test/js-tables/main.mjs` `checkKeyedSurface`) tests only `None`; add the upper end and its negative control.

2. **A block module re-exports the table rows but not the `type` rows its own accessors hand offsets into.** `internal/codegen/jstable/block.go:213-227` re-exports `g.imports[home+"Block"]`, which holds only what the module's own code referenced. `RenderBlock.js:23` exports `RenderShipRow` etc. and not `RenderVector3Row` / `RenderQuaternionRow`; `RenderShipRow.PositionAt(at)` returns an offset only `RenderVector3Row.X(view, at)` can read, and that name is `undefined` from the module USAGE tells a consumer to import (`import { RenderFrameBlock, RenderShipRow } from "./RenderBlock.js"`). Same in `FrameBlock.js:23` (`ArmorConfigRow`, `ArmorPlateRow`, `GunnerSettingsRow`, `FiringGroupRow` missing). The banner says "one module, one whole surface"; it is not. Repro: `B.RenderVector3Row.X(...)` → `TypeError: Cannot read properties of undefined`. The cook side (`GraphCook.js:43`) already re-exports its whole closure; do the same here.

3. **The allocation floor as stated is not what the documented per-frame loop pays.** USAGE, JavaScript section (~l.734): "A field declared `int64`, `uint64`, `bits(N > 32)` or a `flags` mask costs one BigInt per read … that is the whole of the floor", with the hoisted `LoadBody` → `SaveBody` loop right under it. Independent instrument (perf_hooks GC observer + `--trace-gc` scavenge counts at a fixed `--min/max-semi-space-size=1`, 2M iterations after 400k warm-up, one path per process, pinned node 20.20.2; a planted `{a: i}` per iteration adds +75 scavenges ≈ +39 B/iter on every path, so the instrument goes red):
   - `KeyedConfig` LoadBody+Measure+SaveBody: 4 scavenges = idle (4). **Zero. The claim holds for the no-64-bit class.**
   - `RootConfig` LoadBody alone: 172 (~90 B/iter — the licensed BigInts). Measure alone 3. SaveBody alone 3 (≈ 0).
   - `RootConfig` LoadBody + SaveBody in ONE body — the USAGE loop: **454 (~238 B/iter)**, 2.6× Load alone, with Save alone at zero. Same on `ProfileConfig`: load 172, save-after-one-load 0, load+save 448. Isolated: `x = view.getBigInt64(); writer.put64(x)` = 150 scavenges against `put64(old)` = 57 — a freshly read BigInt is boxed again on its way into `put64` (TurboFan keeps it unboxed and rematerialises at the call — a hypothesis; the numbers are not). `--no-turbo-inlining` changes nothing; a 40× inlining budget gives 141 / 377 — still 2.7×.
   - The port's gate (`tables-js-alloc`) measures Load, Measure and Save as three separate bodies (`main.mjs` `allocationPaths`), and the soak's steady phase runs the allocating entry points, so neither can see the combined body. The gate IS a measurement with a planted control — that half is right. Fix: add the hoisted Load→Save loop as a gated path and put the number on the page: a 64-bit field on the wire costs about two and a half BigInt-sized allocations per round trip, not one.
   - Where a BigInt IS allocated per record: wire `Load` of every `int64`/`uint64`/`bits(>32)`/`flags` field (~30 B each); block `Row.Flags` / `Version` accessors (139 B per 5-ship walk, ~28 B/row — `RenderShip.flags` on the hottest path §12.1 has, every frame; a `flags` with ≤ 32 variants could be a Number, or a `FlagsHas(view, at, bit)` accessor beside it); cook `At` composes two u32 reads and is zero (2M derefs, 4 scavenges) — but `main.mjs:1122` still labels it "following one reads a BigInt (§6.3)" and leaves it ungated (`null` ceiling). Gate it at the zero floor; the label is stale.

4. **`Open(readFileSync(path))` returns `null` for a block, with no reason.** Node pools Buffers under 4 KiB: `block_render.bin` arrives at `byteOffset 568`, `RenderFrameBlock.Open` refuses `byteOffset % 64` and hands back `null`. The cook fixtures are over 4 KiB and open. The manifest's `cook_lead_N` rows make "an unaligned base refuses" the contract, so this is a page defect first: USAGE's block example must say `bytes` is a `Uint8Array` whose `byteOffset` is a multiple of 64 (copy a pooled Buffer), or the first thing every Node programmer writes fails silently.

5. **`FromJson(value, "…a string…", report)` returns `false` with `Malformed` set** instead of refusing the type. A string is the native JSON currency; reading it as malformed text is the wrong answer. Accept a string (encode at the boundary) or throw a `TypeError` naming the argument.

6. **History in code comments** (doctrine: present state only; git carries history): `jstable.go:106` ("allocated per row and per edge until the allocation gate measured it"), `:908` ("were four statements of it"), `:994` ("where a float field's sixteen bytes came from") — the last two land in every generated `<Package>Table.js` (`TabledemoTable.js:161`, `:247`); `Makefile:75`, `:629`, `:697`; `test/js-tables/main.mjs:1026`, `:1232`.

### Page defects

- SPEC-TABLES §7 backend status (l.2165-2175) names C++ `<Root>Open`, C# `<Root>Cook.Open`, Java `<Root>Cook.open`; JavaScript's `<Root>Cook.Open` (and Go's `<Root>Open`) are missing. §7's per-backend spelling table (l.2400) has C++/C#/Go columns and no JS column, though the JS shape differs from all three (`XCook.At(cook, slot)` → `-1` / `-2`).
- §11 (l.3766) states the variable-class wire refusal and the accelerator emission for "the C#, Go and Java backends"; JavaScript refuses the same way (no `<Base>Table.js`, banner in every Block/Cook module) and is not named.
- §16.1 backend status (l.5122): "the FIXED class, in C++, C#, Rust, Java and C" — JavaScript (json-read/json-write 16/16, hostile 66/66, the differential) and Go are missing.
- "BIG-ENDIAN, per backend" (l.283-300) has a paragraph each for C++, Rust, Go and Java and none for JavaScript. No JS big-endian leg exists (CI's `big-endian` job runs C++ and Go); the port's answer is Java's — the reader's order is the DataView's explicit `true` — and its cross-endian gate is the two foreign surfaces plus the leg's order-word check. Say so there.
- USAGE JavaScript section: the floor sentence (defect 3); the block example's alignment requirement (defect 4); l.1899 "the C++, C# and C block backends emit `BuildVersion` … and the JavaScript one emits it to compare against" — Go, Rust and Java emit it too.
- The casing rule (types/functions/constants/data members UpperCamelCase, methods and locals lowerCamelCase) lives only in `jstable.go:37-63`; a consumer meets `report.Unknown` beside `reader.reset()` and `teams.get()` with no page saying why.

### Feel — as the JavaScript programmer

Not the surface I would have written; most of the distance is C++ transliterated, and the reasons are in the emitter rather than on the page. Idiomatic alternatives, per site:

- **Out-params and sentinels.** `XLoad(value, bytes, report) → boolean`; `XSave(value, buffer) → size | -1` (two meanings); `XToJson(value, buffer) → -1`; `XCook.At → -1 | -2`; `Open → null` with no reason. The JS shape: return the value (`load(bytes, report) → value`, `save(value) → Uint8Array` with `saveInto(value, buffer) → bytesWritten` beside it), throw `RangeError` / `TypeError` for a CALLER error (short buffer, unnameable enum, a string where bytes were expected), and keep the report for DATA tolerance — which is what the report is for. `Open → null` is right for a frame loop; a `why` on the refusal or an `open` / `tryOpen` pair would separate "wrong build" from "you handed me a pooled Buffer".
- **Naming.** UpperCamelCase properties (`ship.Health`, `report.Unknown`, `cook.Region`) are C#'s convention; JS is camelCase. The port's argument — the packet emitter already does this, one unit one spelling — is a real one, and belongs on the page. `TableEnumIdGrade` / `TableEnumValueGrade` is Go's spelling.
- **Text.** JSON in and out as byte buffers with a separate Measure call is the C++ API verbatim. The idiomatic pair is `toJson(value): string` / `fromJson(value, text: string | Uint8Array, report)`, with the byte form beside it for zero-copy.
- **Iteration.** `for (const [key, e] of fleet.Teams)` is right and native. A bounded array as `Weapons` + `WeaponsCount` and a string as `Name` + `NameLength` are C structs; an iterator over the live prefix and a `name()` decode helper would cost the codec path nothing.
- **Rows as `(view, at)` accessor objects** (`RenderShipRow.ObjectId(block.View, at)`) is the honest shape for a language with no structs — the one place the C++-ness is forced, and the page says so. Fine.
- **ESM only, zero dependencies, `DataView` with an explicit little-endian flag everywhere, caller-owned `Uint8Array`s in and out, BigInt for 64-bit and Number below, one frozen runtime home per unit re-exported everywhere** — all right, and stated. `TableReport` not cleared by `Load` mirrors C++ (a reused report accumulates); a JS caller expects `report.reset()`.

### Reproduced

- **My own consumer from the pages** (`build/read/consumer.mjs`, gone with the worktree): **580 checks, 3 failures — all three the keyed guard above.** Covered: the 16 fixed wire instances (Load silent, Measure = golden length, Save = golden bytes, buffer one short → -1, hoisted `LoadBody`/`SaveBody` pair = same bytes); the 16 JSON goldens both ways, with and without the trailing newline; 12 report cases (+3 absent for `tblp2`, as Go/C#); 66 hostile trees; my own descriptor walks writing both block row dumps and all six cook node dumps byte-identical to the pins; all 122 forgeries (11 block, 111 cook) through my own patcher; both foreign surfaces; unaligned, truncated and null bases; hand-made §3/§4 bytes (unknown id, kind mismatch, repeated id last-wins with `Duplicate` 0, unknown kind 18 → malformed, string past capacity clamps and counts, positional body under a keyed field → kind mismatch, keyed key 0 → malformed with the parent continuing and `Load` true — checked against the C++ reference compiled from the same unit: `ok=1 malformed=1`); save-side refusals (unnameable enum in a field and in a fixed array, count and length past their bounds).
- `./build/conformance-harness run --only js`: all twelve surfaces pass; `wire 16/16 +4a`, `report 12/12 +3a`; 4.1 s wall.
- The port's own instruments in this worktree on the pinned node: standalone, generic-walk and refusal gates OK; leg OK; fuzz with a fresh seed `0x5eed20260903`, N=20000 → 12412 refused / 7588 opened, no exception escaped; accessor and slot negative controls red; text differential 60/60 both ways, its negative control red on 19/60; `conformance-negative-control-js` red on `json-read` alone; alloc gate green (Keyed 0.0 / RootConfig Load 67.1 / ships walk 0.0 / deref 0.0), its negative control red at 299.3 B; soak 60 s green.
- Claimed names: 20 runtime spellings (`TableReport` … `TableCookMaxAlign`, `BuildVersion`, `TableEnumIdGrade`, `TableEnumValueGrade`) each refused in a schema under `--lang js`; `XOpenWalk` accepted — the negative control.
- Bench: `bench/tables/results/2026-09-03-pairing-js2-arm64-macbook.md` is labeled "PAIRING CHECK, not a sitting". Correct.
- Big-endian: CI `big-endian` passes (C++ and Go); there is no JS leg; the cross-endian refusal is held by `cook-foreign` / `block-foreign` (harness, 8/8) and the leg's order-word check.

gafferongames:
Every item in the read, what it became (all on the rebased branch, on top of #369):

**Defects**

1. **Keyed guard** — `TableKeyed.get`/`set` refuse every key that names no slot: one unsigned compare, `((key - 1) >>> 0) >= Slots.length`, plus an integer check, throwing `RangeError` that names the range. The leg holds None, `E.Max + 1`, `-1`, a fraction and `undefined` on both `get` and `set`, and that a refused `set` grows nothing; `make tables-js-keyed-negative-control` puts a None-only guard back through `go build -overlay` and requires the leg to go red on `E.Max + 1`.
2. **One module, one whole surface** — every `<Base>Block.js` re-exports the block home's whole surface (the runtime and every `<Name>Row`, nested `type` rows included), and every `<Base>Cook.js` the cook home's plus the block-home records its closure names. `RenderBlock.js` now exports `RenderVector3Row`/`RenderQuaternionRow`; the leg reads `PositionAt`'s offset through the re-exported row and holds it to the home's.
3. **Allocation** — the +1 BigInt on the hoisted Load→Save loop was `BigInt.asUintN(64, v)` inside `put64`: measured at a fixed 1 MiB semi-space, `save` alone read 283 scavenges/2M and `load+save` 451 against `load` 171; with the BigInt handed straight to `setBigUint64` (which wraps mod 2^64 by spec), `save` reads 4 (= idle) and `load+save` 171 (= load). The gate now measures the loop the page documents: `KeyedConfig Load+Save` 0.0 B/iter (gated ≤ 8), `RootConfig Save` 0.0 (gated ≤ 8), `RootConfig Load+Save` 67.4 against `Load` 67.1 (gated ≤ 512); cook `At` composes two u32 reads and is gated at the zero floor (0.6 B/iter, label corrected); the ships walk now includes a `FlagsHas` read and stays 0.0; the whole-mask `Flags` read is stated as the licensed BigInt and gated at a ceiling (223.9 B per 5-row walk ≤ 320). A `flags` field's row accessor gains `<Member>Has(view, at, bit)` — one u32 read, no BigInt — beside `Flags`; a 64-bit row field stays a BigInt and the page says so. The soak's steady phase runs the hoisted pair. USAGE states the floor as measured: zero for the no-64-bit class and for every write, one BigInt per 64-bit field READ, 67 B/iter on the corpus root's loop.
4. **Open on a pooled Buffer** — `Open` answers `null` for one thing only: bytes that are not this build's. A view at an unaligned `byteOffset` throws `RangeError` naming the offset, the alignment and the fix (`new Uint8Array(bytes)`); no `Uint8Array` at all throws `TypeError`. USAGE's block example starts with the copy and says why. The manifest's unaligned rows still read "refuse" (the driver maps a caller-error throw to the verdict), and the fuzzer classifies the throw on the leads it misaligns on purpose.
5. **FromJson given a string** — a string is now the text form's currency: `XFromJson(text, report, value?)` takes a string or a `Uint8Array` and hands the value back; anything else is a `TypeError`, never `Malformed`. `XToJson(value)` answers a string; `ToJsonMeasure` is gone from the JS surface.
6. **History out of comments** — `jstable.go` (the three sites, two of which landed in every generated `<Package>Table.js`), the Makefile's three, `main.mjs`'s two: present state only.

**Page defects** — §7 status names the Go `<Root>Open` and the JavaScript `<Root>Cook.Open`; the per-backend spelling table has a JavaScript column; §11's variable-class bullet names JavaScript and its banner; §16.1's status names Go and JavaScript; the big-endian section has a JavaScript paragraph (the reader's order, no JS leg, the two foreign surfaces plus the leg's order-word check); USAGE's floor sentence, block example and `BuildVersion` sentence are corrected; the casing rule and the storage rule are on both pages, with the reason (the packet emitter's classes are the closure's).

**Feel** — applied where the wire and semantics stay the reference's, stated at the site where they cannot: `XLoad(bytes, report, value?)` returns the value (a fresh one, or the caller's overlaid in place — the zero-alloc half); `XSave(value)` returns a `Uint8Array` of exactly Measure's size and `XSaveInto(value, buffer)` the bytes written; no sentinels anywhere — a buffer too small, a count or length past its bound, an enum value or union tag no variant names, a float the text form cannot spell are the CALLER's errors and throw `RangeError` naming the field, while what the DATA did stays the report; cook `At` answers the offset or `null`, and a delta that leaves the region throws naming the cook as corrupt; the enum identity pair answers `undefined` for no identity; `TableReport.reset()` exists; the descriptor's `reset` is `Reset` (a member holding a function, like `GetRaw`). Kept and stated: UpperCamel data members and lowerCamel methods, and `Uint8Array`+`Length` / `Array`+`Count` storage — both are the packet emitter's, and a table's closure decodes into the classes that emitter already wrote, so a second spelling one declaration over would be the worse surface. `SaveInto` is claimed in §11 and the checker (49 suffixes).

Local: every JS gate, the keyed control, the alloc control, the differential and its control, fuzz at 100k, the harness js leg (12 surfaces, cook-write absent), a 30 s soak and the bench smoke are green on the rebased tree; the generated bench JS is regenerated.

gafferongames:
Two more from the register audit: `make tables-js-fuzz-negative-control` builds the emitter with BOTH block extent bounds removed (rows past the body, and the padding check) through `go build -overlay`, regenerates, and runs the fuzzer at the pinned seed — red at mutant 32 on a read that escaped the buffer. Each bound alone leaves 20000 mutants green at that seed (the other check catches what the removed one missed), which is why both come out; the Makefile states the measurement. It sits in `make test` beside the fuzzer. And the `Scene head deref` allocation row carries the zero-floor ceiling (`<= 8`, measured 0.6 B/iter): the delta is two u32 reads.

gafferongames:
## Confirming read: LAND

Head f6c25611, own worktree, pinned node v20.20.2, every item run rather than read.

1. **Keyed guard** — own consumer on `KeyedConfig` (`Team.Max` = 3): `Teams.get`/`set` throw `RangeError` naming the range for 4, 0, -1, 1.5 and `undefined`; `Slots.length` stays 3 after a refused `set`; iteration yields keys 1..3 only. `make tables-js-keyed-negative-control`: a None-only guard through `go build -overlay` turns the leg red on "accepted E.Max + 1 as a key".
2. **Whole surface per module** — `RenderBlock.js` exports `RenderVector3Row`/`RenderQuaternionRow`, the same objects as `BlockdemoBlock.js`'s; `RenderVector3Row.X(block.View, ShipsAt(block,0) + RenderShipRow.PositionAt(0))` reads 2, -2, 4 from `block_render.bin`.
3. **Allocation** — `make tables-js-alloc` (300k iters): KeyedConfig Load/Measure/Save/**Load+Save** 0.0; RootConfig Load 67.1, Save 0.0, **Load+Save 67.4** (≤ 512); ships walk with `FlagsHas` 0.0; whole-mask `Flags` 223.9 per 5-row walk (≤ 320, stated as the licensed BigInt); **Scene head deref 0.7 ≤ 8** with the corrected label (two 32-bit reads, never a BigInt). The gated Load+Save row is one hoisted reader and writer over `LoadBody`/`SaveBody` (`main.mjs:1229`, `:1253`) — the loop USAGE documents. USAGE's "67 bytes per iteration" matches. `make tables-js-alloc-negative-control` red at 300.8 B on KeyedConfig Load. `Version` is a per-block BigInt, not per row.
4. **Pooled Buffer** — `readFileSync("block_render.bin")` lands at `byteOffset` 1136; `RenderFrameBlock.Open` throws `RangeError` naming the offset, the multiple of 64, §19.1 and `new Uint8Array(bytes)`; a view at offset 8 throws the same; `null` throws `TypeError`; the copy opens; a flipped build-version byte answers `null`. USAGE's block example starts with the copy and says why.
5. **FromJson** — `RootConfigFromJson("{}", report)` returns a `RootConfig`; `FromJson(42, report)` throws `TypeError` ("takes the text as a string or a Uint8Array, not number") with `Malformed` untouched; `FromJson("{", report)` sets `Malformed`, no throw; `ToJson` returns a string that round-trips to the golden wire; `ToJsonMeasure` absent.
6. **History** — the six named strings are gone from `jstable.go`, the Makefile, `main.mjs` and the generated `TabledemoTable.js`; a broader scan of `internal/codegen/jstable/*.go` and `main.mjs` for tense words turns up nothing but the `was = "old"` rename keyword.
7. **Pages** — §7 status names Go `<Root>Open` and JavaScript `<Root>Cook.Open`; the spelling table has a JavaScript column (Open/handle/root/deref/record/descriptors/§7.1 facts); §11's variable-class bullet names JavaScript and its banner; §16.1 status names Go and JavaScript; the big-endian section has the JavaScript paragraph (reader's order, no JS leg, two foreign surfaces plus the order-word check). USAGE: floor sentence stated as measured, block example with the copy, `BuildVersion` sentence names all six emitting backends; the casing rule is on both pages with the packet-emitter reason.
8. **Feel** — `Load(bytes, report, value?)` returns the value; `Save(value)` returns a `Uint8Array` of Measure's size, `SaveInto(value, buffer)` returns bytes written and a one-short buffer throws `RangeError` naming both sizes; `report.reset()` exists; cook `At` answers the offset or `null` (walked 122 `ListNode`s to `null`), a patched delta throws `RangeError` "leaves the region — this cook is corrupt"; one casing, stated at SPEC §2 and USAGE with the reason; storage stays `Uint8Array`+`Length`/`Array`+`Count`, stated at the site.
9. **Fuzz** — `make tables-js-fuzz SEED=0xc00c1e5eed N=20000`: 12519 refused / 7481 opened, nothing escaped. `make tables-js-fuzz-negative-control` at the same seed: both bounds out, red at block mutant 32, "a walk of an OPENED forgery threw: Offset is outside the bounds of the DataView". Both sit in `make test` (l.3701-3702).
10. **Harness** — `conformance-harness run --only js`: wire 16/16 +4a, report 12/12 +3a, json-read/write 16/16, hostile 66/66, cook 6/6, cook-write absent, cook-foreign 6/6, block 2/2, block-foreign 2/2, block-dump 2/2, forgery 11/11, cook-forgery 111/111.

Also green here: leg, standalone, generic-walk, refuses-pointers, runtime-home + control, accessor + slot controls, text differential 60/60 + control red on 19/60, `go test ./compiler -run TestJsTable` (module-scope claimed names).

### Register (#389's list; `docs/PORTING.md` does not exist yet on `porting-register`)

**This port uses, the list lacks:**
- text differential against a third implementation (`schema unpack`'s Go engine) over random instances, with its tie-break negative control
- accessor/descriptor agreement (every row and node read both ways) with scalar and pointer-slot controls
- runtime-home rule with a file-order negative control
- keyed-guard negative control (None-only guard back in)
- allocation gate pinned to a runtime major and refusing to certify on another
- soak gated on the allocation rate, not heap drift
- release tier (`tables-js-release`) that scales the PR-tier instruments rather than adding new ones
- standalone and generic-walk gates (twins exist for C#/Java/Go; still not on the list)

**On the list, this port lacks, with no stated reason:**
- none. Cross-endian refusal has no separately named JS target; the reason (no native path, reader's order is the DataView's) is stated in the big-endian section and the leg's order-word check holds it. Identity map / flat node table and the hooks/allocator contract do not apply: the variable class on the JS wire is a named follow-on (§11) and the port has no allocator — both stated on the page.

PR #356, opened 2026-09-03
tables in Java: the tolerant wire, JSON, reflection, the block and cook readers — mirrored from C++, harness green

gafferongames:
The Java table backend — `internal/codegen/javatable` — carrying the READING
TIER: the tolerant wire's FIXED class, the text form, the reflection
descriptors, and the read halves of the two accelerators. A port of
`internal/codegen/cpptable` (the reference), following `cstable` and `gotable`
as the worked ports, inventing no contract of its own.

Rebased on `08eabcb` (Go, #338). Every shared file was reconciled by hand so
both ports' facts survive: the Java row was added beside Go's and Rust's
everywhere, never in place of them.

## The gate

Five legs, twelve surfaces, **268/268 each** — the two FOREIGN surfaces
included.

```
surface       cpp            cs             go             rust           java
-----------------------------------------------------------------------------------------
wire          pass 18/18     pass 18/18     pass 18/18     pass 18/18     pass 18/18
report        pass 8/8       pass 8/8       pass 8/8       pass 8/8       pass 8/8
json-read     pass 18/18     pass 18/18     pass 18/18     pass 18/18     pass 18/18
json-write    pass 18/18     pass 18/18     pass 18/18     pass 18/18     pass 18/18
json-hostile  pass 66/66     pass 66/66     pass 66/66     pass 66/66     pass 66/66
cook          pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6
cook-foreign  pass 6/6       pass 6/6       pass 6/6       pass 6/6       pass 6/6
block         pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
block-foreign pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
block-dump    pass 2/2       pass 2/2       pass 2/2       pass 2/2       pass 2/2
forgery       pass 11/11     pass 11/11     pass 11/11     pass 11/11     pass 11/11
cook-forgery  pass 111/111   pass 111/111   pass 111/111   pass 111/111   pass 111/111
```

The Java leg answers every surface from **one JVM start-up per surface** — its
generated units are packages of a single classpath, so the cook's node dump and
the 111-row cook forgery battery ride inside processes that were already
starting. One sitting, quiet machine, median of three: **java 2.2 s** against
**cs 11.7 s**, with go 0.57, rust 0.59 and cpp 0.65.

## Where Java forces a different spelling

Five places, each stated where a reader meets it:

- **The unit's namespace is the PACKAGE**, and a public Java type lives in a
  file of its own name — so the shared runtime is ONE FILE PER TYPE rather than
  one home file. File-order independent by construction rather than by a rule.
- **There are no unsigned types.** A decode local widens to the signed type that
  holds the wire kind's whole range; u64 compares through
  `Long.compareUnsigned`.
- **There are no ref structs.** A nested body is bounded by MOVING THE READER'S
  LIMIT rather than by slicing a sub-reader.
- **There are no pointers.** A block and a cook are read at explicit offsets out
  of a `byte[]`, and the base's ALIGNMENT is that offset's residue — the same
  arithmetic as an address's, so the same refusals.
- **There are no structs.** A row's generated accessors read each field at its
  offset, so §19.3's `static_assert` half has no Java counterpart. Saying
  otherwise would be theatre; what `TableBlockLayout` checks is the one
  disagreement the language CAN have — the accessors' offsets against the
  descriptors', two derivations the generator makes separately.

And one **ruling** rather than a spelling: a reference resolves through
`<Root>Cook.at`, which **bounds the target against the region**. C++, C# and
Rust hand back a pointer; Java has no undefined behaviour to preserve, and an
unchecked deref is an exception escaping into a caller that asked a question —
which the fuzz oracle forbids.

## The allocation number, measured

Not a flat heap — the JVM's own per-thread allocation counter, bytes per record,
a **named floor per path**:

| path | measured | floor |
|---|---|---|
| wire read | **0 B/record** | exact |
| wire save | **0 B/record** | exact |
| block row walk | **0 B** | exact |
| cook read | **0 B** | exact |
| block open / cook open | 32 B | one handle, per *file* not per row |
| json write / read | 7640 / 3556 B/record | ceilings 16384 / 8192 |

**An hour's soak re-measures the whole table at every sample**: 3,340,044
passes, 12 samples, **0 breaches**, live heap 5.5 MB after warm-up and 2.2 MB at
the end. A negative control adds one `new byte[1]` per record and reds the
wire-read row alone.

## The bench — a PAIRING CHECK, not a sitting

The board's own title is "PAIRING CHECK, not a sitting" and its first line says
it certifies nothing: a MacBook Air, an interactive desktop, no core
reservation. Five legs, one sitting, quiet machine, 1 warmup + 7 measured runs
each. Against the C++ arm as it now stands — the force-inlined one from #350:

| leg | write M msg/s | round_trip M msg/s |
|---|---|---|
| cpp | 2.270 | 0.775 |
| rust | 2.060 | 0.640 |
| **java** | **1.638** | **0.468** |
| go | 0.347 | 0.217 |
| cs | 0.303 | 0.203 |

**java / cpp: 0.72x write, 0.60x round-trip.** Round-trip is the row with the
gap and the board says so rather than rounding toward the ladder's wording. **No
number here may be quoted as the tables layer's performance** — the certifying
sitting is the box under the estate's bench rules.

## Two items for follow-on issues, deliberately not fixed here

**1. The text form's write path is a per-language performance item.** It costs
**7,640 bytes per record**, and nearly all of it is one thing: C's `%.*g` float
spelling goes through `BigDecimal` and `String`. That was the correct first
move — `java.util.Formatter` rounds HALF_UP where C rounds HALF_EVEN on the
exact decimal expansion, and the two differ on a tie, so the mode had to be
named rather than left to the default — but a formatter that reaches the same
bytes without allocating is the obvious lever, and the owner's bar is the most
performant implementation each language allows. **This becomes the Java
text-form round after landing.** The wire path, which is the hot one, is already
at zero.

**2. The `make test` CI budget needs its own issue.** The `test (ubuntu-latest)`
job hit its **15-minute timeout** on this branch, and the timestamps say the
headroom was already gone: the chain reached this leg's block at 13m11s and the
job would have finished at about **13m48s without it**. This leg's first cut
cost **87 s**; it now costs about **20 s**, with the whole-corpus compile, the
second conformance control, the fuzz oracle's control and the soak moved to a
named `make tables-java-release`. So the job is back inside — but it is inside
by seconds, on a rule the owner sets at one to two minutes for per-PR CI, and
the next port meets the same wall. The split here is a budget decision and is
written down as one; the budget itself is not this PR's to fix.

## Claimed names

`internal/tablenames` registers the Java backend. `TableJson` and
`TableBlockLayout` **lose their scoped flag** — C# reaches both through `Schema`
and Java puts both at package level, and §11's claim is the UNION, which is the
rule Go's own widening already states — and `TableBytes` joins as Java's one new
spelling. `compiler/tables_test.go` grows the Java leg's scan (block comments
stripped as well as line comments, `BuildVersion` collected beside the `Table*`
family), a **negative control for the scan itself**, the repro for the two
widened claims, and the file-name collision Java's one-public-class-per-file
rule creates.

## Controls, all firing

- `conformance-negative-control-java` — one field index off in the walk reds
  `java / json-read` alone; `json-write` stays green, which is what says the
  break is the READER's.
- `conformance-negative-control-java-block` — the array pitch check removed from
  `open` reds `java / forgery` alone; `block` and `block-dump` stay green, which
  is what says the reader still READS and stopped CHECKING.
- `tables-java-fuzz` and its control — **a finding worth recording**: removing
  ONE of the block `open`'s two extent bounds leaves the fuzz green, because the
  padding check downstream refuses on the negative slack. The checks are
  layered, and the control has to reach past that before one mutant escapes.
- `tables-java-alloc` and its control, plus the compile, generic-walk,
  standalone, refusal, zero-cost and byte-order gates.

## Not in this PR, named

The VARIABLE class on the wire — the arena, the builder, the region, the
node-table codec. Refused BY NAME in every source a pointered unit does emit,
exactly as C#, Go and Rust refuse it; the accelerators still read those units in
full.

## The blind read's seven, each reproduced before it was fixed

**LAND-AFTER-FIXES** ([review](https://github.com/mas-bandwidth/schema/pull/356#issuecomment-5523886912)).
All seven are fixed on the branch; each fix carries the reproduction as a gate
where a gate was missing.

| # | defect | fix | proof |
|---|---|---|---|
| 1 | generated methods were UpperCamelCase — transliterated from C++, and inconsistent with this backend's own `writeVec3` in the same package | `patrolMeasure` / `patrolSave` / `patrolLoad` / `patrolReset`; §6.1's name-first order untouched, only the case moves. `TableReset` dropped from the registry for Java | `TestJavaGeneratedMethodsAreLowerCamel`; whole tree regenerated |
| 2 | descriptor caches unsafely published at four sites; the comment claimed the opposite | holder idiom at all four; the wrong comment gone | `TestJavaDescriptorsAreSafelyPublished` — structural, and it says why below |
| 3 | `Cook.at` bounded the target's START; the oracle carried the extent bound the reader lacked | `at(slot, size)` bounds the whole record; `walkCook`'s private bound deleted | `make tables-java-cook-extent` + its negative control |
| 4 | the 2 GiB `byte[]` ceiling stated nowhere | stated at both `open` sites, in SPEC's divergences and in USAGE; FFM overload named as the follow-on | — |
| 5 | `verify()` set `checked` before the checks ran | set after the last check, and volatile, in both halves | — |
| 6 | the typed fast path allocated 32 B/call and the gate's zero measured the descriptor walk | `<field>Count()` / `<field>At(i)` allocate nothing; the gate measures the typed path **and** the convenience | alloc table: typed **0 B**, `rows()` 32 B under a stated ceiling |
| 7 | the soak's planted allocation was unreachable — the gate had never been red | flag read in `main`; `make tables-java-soak-negative-control` added | the soak reds on the wire-read row, breach counted |

**On #2's test, since the coordinator asked whether one is practical.** Observing
the race itself would be a race detector: the reordering is *permitted* by JLS
§17.4 rather than required, is essentially never observed on x86, and is
nondeterministic on aarch64 — a test built on it would be green almost always and
would prove nothing on the run where it mattered. What *is* deterministic is the
shape the emitter writes, and the shape is exactly what was wrong. So the test is
structural: every generated descriptor accessor must read a holder's final field
(or delegate to one), and the `if (info != null)` idiom must appear nowhere. It
fails in every build, on every host, if the plain cache comes back — and it
caught a real false positive while being written (`<Table>Cook.type()` legitimately
delegates), which is how I know it is looking at the right thing.

**The measurement discipline changed with #6.** Every allocation row is now
measured twice and the second reported — the bench's own "one warmup run, then
the measured ones", applied to allocation. It earns its place: the typed accessor
loop is compiled late, and at a small window the compilation lands inside the
first measured span as a fixed ~576 bytes. That is a one-off and not a per-record
cost (576 over 2,500 records is 0.23 bytes; the smallest object Java can allocate
is sixteen), and a path that really allocated per record allocates in the second
window too.

## Rebased, and CI

On `7d75809`. The java leg joins **ci.yml's per-leg conformance matrix** (#372)
with its build target and a JDK step; **certify.yml needs no edit** — it already
derives `tables-<lang>-release` from the Makefile by grep, and my
`tables-java-release` matches it.

## Page defects from the review, also fixed

USAGE now shows the **hoisted writer** beside the hoisted reader and says plainly
that the convenience `<name>Load` / `<name>Save` each allocate one object; its
keyed-array paragraph distinguishes a `table`'s accessor pair from a plain
`type`'s bare array (the family's shape, not the port's); and both pages carry
the 2 GiB ceiling and `at`'s record bound. `TableBlockRows` is a `record` with
`forEach`, which is the iterated spelling §19.2 asks for.

**Acknowledged and not fixed here**, from the review's "feel" section: the block
projection's static accessors where an instance exists, `TableKeyed.count`'s
eight overloads, the keyed upper bound's index-out-of-bounds message, and a line
in the source saying `Optional` was rejected for the allocation. None is a
defect; all are surface polish, and they are follow-ons rather than silence.


gafferongames:
## Blind read of the Java tables port — **LAND-AFTER-FIXES**

I read `docs/SPEC-TABLES.md` and `docs/USAGE.md`'s Java section first, wrote my own
consumer over the generated packages' public surface from the pages alone, and only
then opened `internal/codegen/javatable`. My consumer shares no code with
`test/java-tables` or `test/conformance/java`.

**The bytes are right.** 268/268 on `make conformance`, the java row twelve of twelve,
and my own 244-check consumer is green on everything except one probe. Every §7 and
§19.2 refusal I could derive from the page fires. The instruments are real and their
controls fire on a fresh seed.

**What holds it up is four things.** One is a defect in shipped generated code, not a page
nit — a data race Java's own memory model permits, on the path every `open` takes. One is a
ruling the pages state more strongly than the code delivers, with the port's own oracle
quietly covering the gap. One is a language-forced ceiling nothing states at all. And one
is a gate whose negative control cannot fire, so it has never been red. All four are small
changes; none is a redesign.

---

## Defects

### D1 — the descriptor caches are unsafely published, and the comment that says otherwise is wrong

`internal/codegen/javatable/codecs.go:1310-1313`:

> `// ...the benign race two threads can run on first use produces two equivalent`
> `// descriptors and then one; nothing mutable is ever published.`

A mutable object with non-final fields is exactly what is published. Emitted at four
sites — `codecs.go:1317` (`<X>TableType`), `block.go:514` (`<Table>Block.type`),
`rows.go:246` (`<Name>Row.blockInfo`), `rows.go:297` (`<Name>Row.cookInfo`) — the shape is:

```java
private static TableBlockInfo projection;                    // NOT volatile
public static TableBlockInfo type() {
    TableBlockInfo info = projection;
    if (info != null) { return info; }
    info = new TableBlockInfo();
    info.name = "RenderFrame"; ... info.fields = fields;     // NON-FINAL fields
    projection = info;                                       // plain write
    return info;
}
```

`TableTypeInfo`, `TableBlockInfo`, `TableCookInfo` and `TableFieldInfo` all carry public
**non-final** fields. Under JLS §17.4 a second thread may read a non-null `projection`
and still see `info.fields == null` / `info.numFields == 0`. The build being idempotent
makes the *duplicate construction* benign; it does nothing about the *publication*. This
is the textbook unsafe-publication case, and it is a reordering the estate's aarch64
targets actually permit.

It is on the path every open takes: `<Table>Block.open` → `TableBlockLayout.verify()` →
`record(..., RenderCameraRow.blockInfo())` → `info.fields[i].name`. A racing NPE there
escapes out of `open` — which is the one thing this port's whole fuzz-oracle premise
forbids. §19 is a form whose stated use is a per-frame job, and §8.1's own law says the
accessors are "built once with the descriptor and cached with it", so the cache is
required and only its shape is wrong.

The Java fix is one of the language's oldest idioms and costs nothing:

```java
private static final class Holder { static final TableBlockInfo INFO = build(); }
public static TableBlockInfo type() { return Holder.INFO; }
```

Lazy, lock-free, allocation-free after first use, and safe by the class-initialization
guarantee rather than by argument. Making every descriptor field `final` would also do it
(JMM §17.5 final-field freeze), and is worth doing anyway — these are immutable values.

No other port has this: C++ has static const tables, C# has type-init locking, Go has
package init. This is the port's own construct, and it is the one place the Java memory
model got treated as if it were C++'s.

### D2 — `<Root>Cook.at` bounds the target's START, not the target's RECORD, and the port's own oracle compensates for that in the walker

`internal/codegen/javatable/cook.go:451-471`, emitted as `SceneCook.at` (e.g.
`build/tables-generated-java/pointers/SceneCook.java:160`). SPEC and USAGE both state the
ruling as closed:

> SPEC §7 status: "a reference resolves through `<Root>Cook.at`, which BOUNDS the target
> against the region ... Java has no undefined behaviour to preserve: an unchecked deref
> is an exception escaping into a caller that asked a question, which is exactly what the
> readers' fuzz oracle forbids"

`at` checks that `slot + delta` lands inside `[region, region + regionLength)`. It cannot
check the target's extent — it does not know the pointee's type — so the exception it was
introduced to prevent simply moves one call along.

**Reproduction.** §7.1 blesses a cook that carries data alone (`attribution_length` 0 —
"a build that ships no tooling need not carry it at all"), so the region ends at the
array's end. Strip the attribution from `build/cook-open/Scene.cook`, forge the root's
`head` delta to land on the region's last byte, and open it:

```
open            -> accepted
array length     = 4144
region           = 64 .. 4144
at(head)        -> 4143   (ListNodeRow.size = 32, so the record ends at 4175)
row read        -> THREW java.lang.ArrayIndexOutOfBoundsException: Index 4144 ...
at(next)        -> THREW java.lang.ArrayIndexOutOfBoundsException: Index 4167 ...
```

`at` itself throws on the second hop. The exact repro is a ~40-line file; say the word and
I'll paste it.

**Why nothing caught it.** `test/java-tables/src/Main.java:639-644`, inside `walkCook`:

```java
long offset = target - cook.region();
if (offset + f.record().size > cook.regionLength()) {
    continue; // the record does not fit: a walker's own bound (§7.4)
}
```

The oracle carries the extent bound the reader lacks, and then certifies the reader. It
also only ever fuzzes the attribution-bearing fixture, where the overrun lands in the
directory bytes and reads garbage instead of throwing. Both halves have to change or the
gate is watching itself.

Two honest ways out, and I think the second is the right one:

1. Give `at` the size — `at(int slot, int size)`, which every generated call site knows.
2. Keep the start bound, **say on the page that it is a start bound**, and move the extent
   bound into the reader where the walker currently improvises it — i.e. `<Name>Row` gets
   a `fits(byte[], int)` or the generated walk carries it. Then delete the walker's
   private bound so the oracle stops covering for the reader.

Either way `walkCook`'s bound has to go, or the fuzzer proves nothing about this path.

### D3 — the Java accelerators are hard-capped at ~2 GiB and nothing in the Java text says so

`<Table>Block.open` and `<Root>Cook.open` read out of a `byte[]`. §7 states the scale
requirement in the owner's own words — *"100mbs or many gigabytes of data in Assets.bin"* —
and the scale fixtures go to a gigabyte. A Java `byte[]` tops out at `Integer.MAX_VALUE`,
so the Java reader cannot reach the case the section is built for, and there is no second
spelling to reach it with.

The family already knows how to say this. §7 and USAGE both name C#'s equivalent:

> "its length is an `int`, so **the pointer form is the one with the reach the cook is
> built for** and a catalogue past 2 GiB is opened through it."

Java has no pointer form and no `MemorySegment` overload (`--release 17` predates stable
FFM), so this is genuinely forced — which is exactly why it has to be **stated at its
site** alongside the other four divergences, with the FFM overload named as the follow-on.
As it stands it is the one language-forced divergence in this port that a reader meets
only by hitting it.

The `long length` parameter compounds it: `open(byte[] data, int offset, long length)`
immediately refuses `length > data.length - offset`, so the parameter can never carry a
value the reader can use. It is the C++ signature transliterated. Either make it `int`, or
keep the `long` and say it is there for the day the segment overload arrives.

### D4 — `TableBlockLayout.verify()` sets its flag before it runs the checks

`internal/codegen/javatable/block.go:560-563` and `cook.go:489-492`:

```java
private static boolean checked;          // not volatile
public static void verify() {
    if (checked) { return; }
    checked = true;                      // set BEFORE the checks
    record(...);                         // throws on a mismatch
```

A caller that catches the first `IllegalStateException` and opens again gets a silent pass:
the contract is checked once, and a failed check counts as having been done. Move
`checked = true` after the last check, or put the whole thing in a static initializer —
which also fixes the non-volatile read and is what a Java programmer would have reached
for first.

### D5 — the typed fast path allocates, and the alloc gate's zero does not cover it

`block.ships()` returns `new TableBlockRows(...)` per call. The gate's `block row walk ==
0` row is measured through `readBlock`, the **descriptor** walk
(`test/java-tables/src/Main.java`, `allocBlockRead`) — it never calls a generated array
accessor, so the floor says nothing about the path §19.2 calls "what a per-frame job uses".

Measured on the pinned JDK, after 200k warm-up frames, nine accessor calls per frame:

```
typed fast path (10 array accessors per frame): 28800000 B over 100000 frames = 288.0 B/frame
```

32 B per accessor call; escape analysis does not remove it. Small in absolute terms, and
hoisting `b.ships()` out of the loop makes it zero — but USAGE and §19.2 both show the
accessor at the call site, and C++/C#/Go all return a value type here. Either add the
accessor to the measured table (with a stated floor), or hand back the triple through the
handle so the call site allocates nothing.

### D6 — the soak has no reachable negative control: its planted allocation does nothing

The soak is the gate the PR body leads with — "An hour's soak re-measures the whole table
at every sample", 12 samples, 0 breaches. I re-ran it and it is green. Then I tried to make
it go red, and could not.

`test/java-tables/src/Main.java:487` assigns the sabotage flag:

```java
sabotage = System.getenv("SCHEMA_ALLOC_SABOTAGE") != null;
```

and that line lives inside `modeAlloc`. `modeSoak` never sets it. The planted allocation at
line 314 is on the shared `allocWireRead` path, so the machinery is right there — but in
soak mode the flag is never read, and `SCHEMA_ALLOC_SABOTAGE=1` is a silent no-op:

```
### soak planted-allocation control (25s)
soak: warm-up done, 17706 passes, heap 5 MiB
soak: allocation at t=0
  wire read                   0 B over   12000 records         0.00 B/record   == 0
  ...
soak: allocation table sampled 1 times, 0 breach(es)
OK                                          <- exit 0, with the sabotage requested
```

No `SABOTAGED:` banner, no breach, exit 0. There is also no `make` target for a soak
control the way there is for `tables-java-alloc-negative-control` and
`tables-java-fuzz-negative-control`, so nothing anywhere has ever turned this gate red.
By the repo's own words, three lines up in the same file: *"a harness that has never gone
red is watching nothing."*

The fix is one line moved — set `sabotage` in `main` rather than in `modeAlloc` — plus a
`tables-java-soak-negative-control` target beside the other two. I confirmed the alloc
gate's own control does fire (D-list above), so the plant itself works; it is only
unreachable from this mode.

---

## Page defects

- **USAGE and SPEC state D2's bound as if it covered the target record.** USAGE: "answers
  the target's offset, or `-1` for a null AND for a delta that leaves the region"; SPEC:
  "BOUNDS the target against the region". Both are true of the delta and neither is true
  of the record, and the *rationale* both give — "an unchecked deref in Java is an
  exception escaping a reader" — is the part that does not hold.
- **Nothing states the 2 GiB ceiling** (D3), in either page, in the "JAVA's four
  divergences" paragraph or in USAGE's Java section.
- **USAGE, the keyed-array paragraph**: "An enum-keyed array in Java is a plain typed array
  of `E.max` slots **on the owning class** ... with an accessor pair beside it that takes
  the KEY." Not true for a `type`. `type ScoreBoard { per_team [Team]int32 }` in
  `tables/examples/Keyed.schema` generates `Keyed.ScoreBoard.perTeam` as a bare `int[3]`
  with no accessor pair, so a caller has to spell `perTeam[team - 1]` — the shift the same
  paragraph says nothing outside `TableKeyed` spells. (C++ does the same, so this is the
  family's shape rather than the port's; the sentence still promises more than it delivers.)
- **USAGE shows the hoisted READER and never the hoisted WRITER.** The alloc table's
  `wire save == 0` needs `writer.reset(...)` + `<Name>SaveBody`, which appears nowhere on
  the page. Worth one line, and worth saying that the convenience `<Name>Load` /
  `<Name>Save` each allocate one `TableReader` / `TableWriter` per call — the sentence
  "Nothing on that path allocates" sits directly under the hoisted example and reads as
  covering both.
- **The PR body's bench section drops the label the board carries.** The board's own title
  is "PAIRING CHECK, not a sitting" and its first line says it certifies nothing; the PR
  body quotes 0.72x / 0.60x under "One sitting, quiet machine". Add the label there too.
- **§19.2's "Three spellings a reader has to have"** still gives C++/C#/Go. There are five
  backends now, and Java's block spelling lives only in USAGE. Pre-existing (Rust is
  missing too), so not this PR's to fix — but the sentence is now wrong on its own terms.

---

## The feel — is this the surface a Java programmer would have written?

Mostly the shapes are right and one thing is loudly wrong.

**The generated method names are UpperCamelCase, and nothing says why.**
`RootConfigMeasure`, `RootConfigSave`, `RootConfigLoad`, `PatrolLoadBody`,
`CfgFromJson`, `TableReset`. Java has exactly one naming rule everybody follows, and this
breaks it in every generated table source. It is not even consistent inside one output
tree: the packet half of this same backend emits `writeVec3` / `readVec3` / `measureVec3` /
`zeroVec3` (`generated/java/Types.java`), lowerCamel, while the table half in the same
package emits `PatrolMeasure` / `PatrolSave`. Storage members are lowerCamel; the methods
next to them are not.

Name-first over verb-first is a family decision (§6.1) and I have no quarrel with it. The
*case* is pure transliteration from C++, it is the first thing any Java reviewer will say,
and Checkstyle/ErrorProne `MethodName` flags it out of the box. The idiomatic spelling that
keeps §6.1 intact is `patrolMeasure` / `patrolSave` / `patrolLoad` / `patrolReset`.

**I'd treat this as the one item to settle before landing rather than after**, because it
is the only defect here that cannot be fixed later without breaking every consumer. The
five stated divergences all earn their place; this one is not stated anywhere in
`internal/codegen/javatable`, and by the brief's own rule an unstated divergence is a
defect rather than a choice.

Everything else, in order:

- **`TableReset(value)` overloads.** Neither name-first (§6.1 asks for `<Name>Reset`) nor
  Java. `patrolReset` fits both.
- **Block projection fields are static where an instance exists.**
  `RenderFrameBlock.version(block.data(), block.base())` when the handle is right there
  and `block.version()` is obvious. Rows have no instance and static is forced there — and
  said so — but the projection does, and the page's own C#/Go examples both spell it
  `block.Projection.Version`.
- **Rows are indexed, not iterated.** §19.2: "An array is ITERATED, not indexed by hand …
  a range-for in C++, an enumerator in C#, the equivalent per port." Java's shipped
  equivalent is `for (int i = 0; i < rows.count; i++) rows.at(i)`, which is the index loop
  the paragraph is written against. `TableBlockRows` hides the pitch, which is the property
  that matters, so this is a smaller miss than it sounds — but `implements Iterable` or a
  `forEach(IntConsumer)` is one method and reads like Java. While you are there:
  `TableBlockRows` is an immutable 4-field carrier and `--release 17` has `record`.
- **`TableKeyed.count(T[])`, eight overloads.** In Java that is `slots.length`. `slot()`
  and `key()` are the parts that earn their keep.
- **The keyed upper bound reports the wrong number.** `slot(None)` throws a good message;
  `hull.teams(4)` falls through to the array and says `Index 3 out of bounds for length 3`
  — a number the caller never wrote. `key 4 is out of range for Team (1..3)` is a two-line
  change in `TableKeyed.slot`.
- **`long length` on both `open`s** — see D3.
- **`null` on refusal rather than `Optional`.** The right call on a zero-allocation path,
  and USAGE does say "answer a handle or `null`". It is worth one line in the source
  saying `Optional` was rejected for the allocation, because a Java reader will assume it
  was not considered.
- **Public mutable fields on `TableReader` / `TableWriter` / `TableReport`.** Unusual for a
  public Java API, correct here, and stated at the site. No complaint.

What reads as genuinely Java and not as ported C++: one public type per file with the
package as the unit scope (and the file-basename refusal that falls out of it, which is a
hazard no other port has and is caught by name with a test and a control); the
caller-owned reader whose limit moves instead of slicing, which is what makes the zero
reachable and is explained where a reader meets it; `report.clear()` returning `this`;
`Long.compareUnsigned` at the comparison rather than a fake unsigned type; javadoc on the
public surface; and `--release 17 -Xlint:all -Werror` over every generated unit. The
refusal banner on a pointered unit's sources is the best version of that idea in any port
I have read.

---

Pinned JDK from `dist/` (Temurin 21.0.12.1+1), own worktree, nothing pushed.

**`make conformance`** — green, every leg, java twelve of twelve:
`wire 18/18 · report 8/8 · json-read 18/18 · json-write 18/18 · json-hostile 66/66 ·
cook 6/6 · cook-foreign 6/6 · block 2/2 · block-foreign 2/2 · block-dump 2/2 ·
forgery 11/11 · cook-forgery 111/111` = 268/268.

**My own consumer**, written from the pages, 244 checks:

| area | checks | failed |
|---|---|---|
| wire goldens (18 instances: load, save size, silent report, byte-identical round trip) | 72 | 0 |
| json goldens (FromJson, ToJson size, text round trip, JSON→wire crossing against the wire golden) | 74 | 0 |
| read reports (8 evolution seams against `reports.txt`) | 8 | 0 |
| block image (`block_render`, `block_padded`: prologue, pitch == sizeof, row walk, past-count refusal) | 9 | 0 |
| cook fixtures (5 roots + the chain walk + the big-endian refusal) | 9 | 0 |
| §19.2 block refusals | 23 | 0 |
| §7 cook refusals | 33 | 0 |
| keyed arrays | 12 | 0 |
| §7 `at()` extent probe | 3 | **1** |

The refusals I derived from the page and all of which fire: truncation below the
projection / mid-image / to the header / below the header, a claim past the array,
trailing bytes, reversed magic and forged magic on both forms, an order word the magic
contradicts, a foreign build version, both reserved words, the alignment word at
`0/1/2/3/4/12/24/128/2^63/~0`, each part length long and short and saturated, a data part
shorter than the root, a base one and four and eight bytes off alignment (and the same
image at an aligned offset still opening), a count past the declared maximum, a count whose
rows leave the caller's extent, a pitch that is not this build's, an array start inside the
projection, an unaligned array start, an array start at the end of the block, an offset
64-aligned just under 2^63, an offset with the top bit set, a null array, a negative offset,
and a block image handed to `SceneCook.open`.

Keyed arrays: iteration yields `[1, 2, 3]` for `E.max = 3`, key 1 is slot 0, key `E.max` is
the last slot, and `None`, `E.max + 1`, `-1` and `100` are all refused — on the object
accessor and on `TableKeyed.slot` alike, in every build (see the feel section on the
message the upper bound gives).

**The port's own instruments, re-run with fresh seeds** — all green, all controls red:

- `Main fuzz` at `SEED=0x5eed2026090301` and `SEED=0xdeadbeefcafe`, `N=40000` each:
  80,000 mutants per run, ~24.5k opened / ~55.5k refused, no escape. (See D2 for what this
  oracle cannot see.)
- `tables-java-fuzz-negative-control` — RED, on the oracle:
  `a block mutant escaped an exception rather than refusing: ArrayIndexOutOfBoundsException`.
- `tables-java-alloc JAVA_ALLOC_SCALE=100` — `wire read 0.00`, `wire save 0.00`,
  `block row walk 0.00`, `cook read 0.00` B/record exactly; `block open` / `cook open`
  32.04 B; `json write` 7640 / `json read` 3556, both inside their ceilings.
- `tables-java-alloc-negative-control` — RED on the wire-read row alone, every other row
  green, so it localises.
- **Soak, 700 s, `-Xmx256m`**, allocation table re-measured at every sample: **609,546
  passes, 3 samples, 0 breaches**, live heap 5,473,712 B after warm-up and 2,204,296 B at
  the end (worst sample 5,655,808 B), 1,865 collections in the measured window. The
  gate itself is exactly the right shape — a count after warm-up, re-measured, not heap
  drift. Its planted-allocation control does **not** go red; see D6.
- `tables-java-order` — green. `tables-java-json-walk` (7 units, byte-identical walkers),
  `tables-java-standalone`, `tables-java-refuses-pointers`, `tables-java-zero-cost`,
  `tables-java-compile-all` — all green.
- `conformance-negative-control-java` — RED on `java / json-read` alone.
  `conformance-negative-control-java-block` — RED on `java / forgery` alone.
- `go test ./compiler -run 'Java|Table'` — green, including the scan's own negative control.

**Claimed names.** I probed the front end for the spellings Java maps names into, and it
holds all of them: a reserved-word enum variant (`Default` → `default`) is refused by name;
a field colliding with a generated presence or count companion is refused; two fields that
differ only in case are refused; `TableBytes` / `TableJson` / `TableBlockLayout` declared
in a unit with a table are refused, and the same declarations in a **table-free** unit are
accepted. The file-basename refusal is scoped exactly right: `TableReport.schema` generates
fine with no table in the unit and is refused by name the moment a table appears.

**The wire reader has no fuzzer of its own** — `Main fuzz` mutates a block image and a
cooked file only — so I wrote one. 200,000 mutants (1-6 byte edits, one in three truncated)
across 8 goldens x 6 roots = **1,200,000 loads: 566,364 accepted, 633,636 reported
malformed, 0 exceptions**. The tolerant reader is solid; the gap is that nothing in the
tree holds it.

**Big-endian.** `make conformance-big-endian` is the Go s390x leg and needs `qemu-s390x`,
which this bench does not have — I could not run it. Java's own byte-order answer is a
named gate and I did run it: `tables-java-order` passes, `cook-foreign` 6/6 and
`block-foreign` 2/2 pass in the matrix, and my consumer independently confirms
`SceneCook.open(Scene-be.cook)` refuses.

**Bench.** `bench/tables/results/2026-09-03-java-port-arm64-macbook.md` labels itself
"PAIRING CHECK, not a sitting" in its own title and says it certifies nothing. That is
exactly right; see the page defects for the PR body.

**Open cost.** `SceneCook.open` on the scale fixtures: 1 MB in 31.4 ms on the very first
call, 100 MB in **7.1 µs** after — so the O(1) property holds, and the first open pays a
one-time cost (class loading plus `TableCookLayout.verify()` building the closure's
descriptor graph). Not a defect; worth knowing before someone opens their first cook
inside a frame.

PR #364, opened 2026-09-03
tables in Dart: the tolerant wire, JSON, reflection, the block and cook readers — mirrored from C++, harness green

gafferongames:
The Dart leg of the tables reading tier (docs/SPEC-TABLES.md), mirrored from the
C++ reference. **Twelve of twelve surfaces answered; `cook-write` absent, as in
Go and C#.**

DRAFT — do not merge.

## The matrix

`make conformance` → the Dart row:

| surface | dart | | surface | dart |
|---|---|---|---|---|
| wire | pass 16/16 +4a | | cook | pass 6/6 |
| report | pass 12/12 +3a | | cook-write | absent |
| json-read | pass 16/16 +3a | | cook-foreign | pass 6/6 |
| json-write | pass 16/16 +3a | | block | pass 2/2 |
| json-hostile | pass 68/68 +16a | | block-foreign | pass 2/2 |
| forgery | pass 11/11 | | block-dump | pass 2/2 |
| | | | cook-forgery | pass 111/111 |

The `+Na` cells are the variable-class cases (#383, #388, #394): the Dart
backend refuses a pointered unit's wire by name (§11), so it has no text form
for one either, and the driver writes `<case>.absent` per case, exactly as the
Go driver does; `cook-write` is absent as a surface, as Go's is. The report
surface carries the kind-17 skip — a pointer rides as a u32 node index (#376)
— and `p2_as_p1` / `p2_as_p3` are the corpus-backed test of it. The fixed
walk refuses the `&` key prefix §16.7 reserves (`reserved-ampersand-key`).

## What Dart spells differently, and why

Each is stated at its site in the emitter and in docs/SPEC-TABLES.md's backend
status.

1. **The verbs are METHODS on the value** — `config.measure()`,
   `config.save(out)`, `config.load(bytes, report)`, `loadBody`, `saveBody`,
   `reset`, `fromJson`, `toJson`, `toJsonMeasure` — on a table's own class, and
   EXTENSION methods (`extension <Name>Table on <Name>`) on a closure `type`'s
   class, which the packet emitter owns. The cost is §11's: the nine spellings
   are claimed against a FIELD name on every closure member (the checker
   refuses `measure`, `to_json`, ... naming the collision), and `Table` joins
   the suffix set.
2. **The reader and the writer are OBJECTS a caller owns, and their one
   currency is the `Uint8List`.** `attach` takes one argument — the bytes —
   because a `ByteData` is a second object describing the same memory: two
   arguments for one fact, or an allocation to derive one from the other. A
   multi-byte scalar is assembled from bytes; a nested body narrows a `limit`
   rather than taking a sub-view.
3. **A float32 field's storage is a `double`**, so every elision comparison
   narrows first — C's float semantics for the decision, and a Dart wire byte
   equal to the C++ one.
4. **The descriptors are `const`**, their memory columns static-method
   tear-offs on a per-type `<Name>TableFields` class. The one factory is the
   COOK's record column, because the DESCRIPTOR graph can be cyclic —
   `ListNode.next` names `ListNode`. (A cooked region never is; the tool refuses
   one by name, §3.1.)
5. **A reference resolves to an OFFSET and the deref is BOUNDED**: an escaping
   delta answers `TableCookRef.outside` rather than raising a `RangeError`.
6. **The block base is `(buffer, offset)`**, §19.1's 64-byte alignment checked
   on the offset. **The array is ITERATED** as §19.2 asks:
   `for (final row in block.ships)` yields one cursor moved to each row at the
   instance's pitch; `shipsAt(i, cursor)` takes a lent cursor and has no
   allocating twin; `shipsBytes()` is the contiguous view.
7. **An enum-keyed array is indexed by KEY and by nothing else**:
   `TableKeyed` refuses `None` and any value past `Max`, and its storage is
   private, so the one model a consumer meets is the page's.
8. **No private library-scope name at all**, and every public runtime spelling
   is registered in `internal/tablenames` — six added, seven widened, gated
   both ways with a negative control.

## The allocation claim, measured

**The wire path — loadBody, measure, saveBody through a caller-owned reader,
writer and report — allocates nothing per record under `dart compile` (AOT),
the language's release configuration and the one a shipping consumer runs.**
`make tables-dart-alloc` counts the VM's own new-space scavenges
(`--verbose_gc`, semi-space pinned at 1 MB so a plant fills it) between two
marker lines over a steady phase, against an idle loop of the same shape, and
holds AOT at exactly zero; the JIT's count is printed beside it, not gated:

| phase | AOT (gated) | JIT (printed) |
|---|---:|---:|
| idle | 0 | 0 |
| load | 0 | 0 |
| measure | 0 | 0 |
| save | 0 | 0 |
| wire (all three) | 0 | 1 |

20,000 passes over the eight-record corpus (160,000 records) at `make test`
length, 400,000 in `tables-dart-release`. `make
tables-dart-alloc-negative-control` plants one `TableReport` per record, then
one `Uint8List(8)` per record: 8 / 5 scavenges under AOT, 9 / 6 under the
JIT, red on both in both builds. The soak (`make tables-dart-soak`) gates on
correctness under reuse and runs the allocation gate after it. There is no
source scan: a scan over source is never an allocation gate.

**Why the JIT is printed and not gated.** A `double` crossing a call boundary
is boxed under the JIT, and the codec bodies are large enough that the
optimizer's inlining budget runs out partway through and leaves some
conversion calls out of line — where, is the optimizer's decision and moves
with the loop around the codec: one boxed double per pass of the corpus in the
polymorphic loop above, up to three per record with one codec inlined into a
monomorphic caller. Every conversion is `prefer-inline` and the NaN widening
is a cold never-inline call so the hot conversion has one exit, which is what
brought the seven boxed doubles per record the read measured down to this;
AOT inlines them all and reads zero. Two further costs are stated on the page,
both the JIT's and neither AOT's: a `float32` carrying a NaN with a payload
costs one boxed double, and a 64-bit field holding a value outside ±2⁶² costs
one boxed integer per read.

The §16 text walk is priced, not gated: the same instrument prints its count
beside the wire's (~14 scavenges per pass over the corpus, per FLOAT and per
NUMBER TOKEN — `%.*g` spelled out over `BigInt`, and the nearest float32 to a
token computed exactly).

## The gates, and the controls that turn each one red

| gate | control |
|---|---|
| `make tables-dart-clean` — analyzer-clean and byte-identical to `dart format` | drift fails the build |
| `make tables-dart-alloc` — zero scavenges on the wire path under AOT; the JIT's count printed beside it | `tables-dart-alloc-negative-control`: a `TableReport` and a `Uint8List(8)` planted per record, red in both builds |
| `make tables-dart-soak` — every record round-trips wire and text, byte-compared, into reused storage; then the allocation gate | `tables-dart-soak-negative-control` corrupts one byte of one re-saved record |
| `make tables-dart-fuzz` — 28,000 mutants (4,000 × 7 fixtures): refuse, or open and be WHOLE, and NOTHING THROWS | `tables-dart-fuzz-negative-control` removes the count bound from the block Open |
| `make tables-dart-usage` — USAGE.md's Dart table examples, extracted verbatim, analyzed and run over the corpus | the page goes red with the code |
| `make tables-dart-json-walk` — one §16 walker per unit, byte-identical across units | |
| `make tables-dart-standalone` — every generated library imports `dart:*` and its own unit's files only | `tables-dart-standalone-negative-control` plants a `package:` import |
| `TestDartTableRuntimeNamesAreClaimed` — both directions, plus no private top-level name | `tables-dart-names-negative-control` plants an unregistered runtime class |
| the conformance matrix | `conformance-negative-control-dart` rebuilds the compiler with `go build -overlay` over a sabotaged §16 read walk; red on `json-read`, green on `json-write` and `wire` |

All of it in `make test`; `tables-dart-release` (certify.yml derives it by
name) runs ten minutes of the soak, 280,000 fuzz mutants under two seeds, and
every control.

## The bench

`bench/tables/results/2026-09-03-pairing-dart-arm64-macbook.md` — **a PAIRING
CHECK, not a box sitting**, a MacBook Air with other work on it, two `--only`
runs back to back, medians of three rounds:

| leg | write | round_trip | write vs C++ | round_trip vs C++ |
|---|---|---|---|---|
| cpp | 2.408 | 0.846 | 1.00× | 1.00× |
| dart | 0.437 | 0.227 | **0.18×** | **0.27×** |

M msg/s. C++ is 5.5× faster on write and 3.7× on round-trip. Wide of "same
speed"; a box sitting is what certifies.

## Pages

SPEC §6 gains the Dart file shape. SPEC-TABLES: the backend status paragraph
and Dart's spellings; the BIG-ENDIAN paragraph says what Dart gives (the
cross-endian refusal, no run leg); §7's per-backend table and §19.2 gain a Dart
column; §11 names the six added and seven widened runtime spellings and the
nine member verbs; §16 prices the text walk. USAGE's Dart section is the
conformance corpus's own `RootConfig`, run verbatim by `make tables-dart-usage`.
Every generated `///` is dartdoc.

## Where it rides

- `conformance (dart)` in ci.yml's matrix — the harness and one AOT driver.
- `tables-dart-release` in the Makefile, which certify.yml derives by name.
- `make test`'s own section, above.


gafferongames:
## Blind read of the Dart tables port — **LAND-AFTER-FIXES**

Read as a Dart programmer who had not seen this port: `docs/SPEC-TABLES.md` and
`docs/USAGE.md`'s Dart section first, then a consumer of my own over the
generated package's public surface, and only then the emitter. Own worktree at
`29f02433`, pinned SDK `dist/dart-sdk-3.13.2` throughout.

**The bytes are right.** `make conformance` is green on all five legs, the Dart
row twelve of twelve; my own consumer found nothing over 552,061 checks; the
forgery fuzzer is clean on a fresh seed; every §7 and §19.2 refusal I could
derive from the page fires; the keyed refusals hold under AOT as well as under
the JIT. Nothing below is a wire or a semantics defect.

What has to move before this lands is a set of **claims that are measurably
false**, and the **gate that is supposed to hold the biggest one**, which three
ordinary allocating constructs walk straight through.

---

## Defects

### D1 — The wire path DOES allocate under the JIT, and three pages say it does not

The claim appears four times: `docs/USAGE.md:2202` ("the read path allocates
NOTHING — no per-field object, no sub-view, no temporary"),
`docs/SPEC-TABLES.md:217` ("allocates NOTHING per record"),
`docs/SPEC-TABLES.md:4959` ("The WIRE path allocates nothing at all"), and
`test/dart-tables/soak.dart:20` ("THE WIRE, whose floor is ZERO").

The port's own soak prints the counter-evidence and passes anyway. `make
tables-dart-soak DART_SOAK_SECONDS=25` on this box:

```
wire  (LoadBody + Measure + SaveBody)
  iterations: 2923   allocations: 175383
  per record: 7.5001
    _Double: 7.25 per record
    _Mint: 0.25 per record
wire over idle: 175358 allocation(s) across 2923 iterations (7.49906 per record)
```

I reproduced that with an instrument that shares nothing with the VM service:
count new-space scavenges under `--verbose_gc` over a steady phase, against an
idle loop of the same shape. Over the same corpus record, JIT:

| phase | scavenges / 1M iterations |
|---|---:|
| idle | **0** |
| load | 54 |
| measure | 0 |
| save | 54 |

Each scavenge is triggered at 2.0 MB of new space, so that is ~108 bytes per
record — about seven boxed doubles, which is exactly what the soak's own
`_Double: 7.25 per record` says. Two independent instruments agree, and the one
the port dismissed as unsound was right about this.

**Under AOT it really is zero** — `dartaotruntime --verbose_gc`, 2,000,000
iterations, load / measure / save / all: **0 scavenges each**. So the true
statement is "zero under `dart compile exe`; about seven boxed doubles per
record under the JIT", and every page states the unqualified version. The soak
runs under the JIT, which is the configuration where the floor is not zero.

Reproduction: `scratch/gcprobe2.dart`, `dart --verbose_gc … 2>&1 | grep -c
Scavenge`, phases `idle|load|measure|save`.

### D2 — `TestDartWirePathAllocatesNothing` is a blacklist with holes, and it is the only gate on D1

`compiler/tables_test.go:1293`. The comment says the probe is "the CONSTRUCTS: a
typed-data constructor, a list or map literal, a growable list, a string
interpolation, a BigInt". The list is 22 substrings. I planted an allocation in
every `Measure` through a `go test -overlay` copy of
`internal/codegen/darttable/codecs.go:285`, the same way
`tables-dart-alloc-negative-control` does, and ran the gate:

| planted into every `Measure` | gate |
|---|---|
| `final planted = [0, 1];` — a bare list literal | **GREEN** |
| `final planted = List.filled(8, 0);` | **GREEN** |
| `final planted = Uint8List.fromList(const [0, 1]);` | **GREEN** |
| `final planted = Uint8List(8);` (the port's own control) | red |

The first one is a list literal — the construct the gate's own comment claims to
catch. `List<` does not match `[0, 1]`; `Uint8List(` does not match
`Uint8List.fromList(`. The sabotaged emitter's output is analyzer-clean and the
allocation is really there:

```dart
int keyedConfigMeasure(KeyedConfig value) {
  var bytes = 2; // terminator
  final planted = [0, 1];
  bytes += planted.length - 2;
```

Two further limits worth stating at the site: the gate scans only
`<Base>Table.dart` of `runtimeSrc` (`compiler/tables_test.go:73`), a probe schema
with no `bytes(N)`, no nested table, no union, no `flags`, no fixed `[N]T` and no
`if` guard — so those emitter paths are never read; and it scans only lines
inside the four top-level bodies, so nothing a body *calls* (`TableReader`'s
methods, `TableKeyed`, the vocabularies) is covered.

**A runtime measurement is possible after all**, and it is simpler than the
service protocol that was abandoned. Scavenge counting has a true zero floor and
both of the plants the soak reported it could not see turn it red — 2,000,000
iterations, AOT:

| loop | scavenges |
|---|---:|
| the port's wire path | **0** |
| \+ `sink = TableReport();` per record | 31 |
| \+ `sink = Uint8List(8);` per record | 31 |

That is a gate, not a reading: no VM service, no accumulator the VM resets, no
attribution, works under AOT and JIT alike. `make tables-dart-soak` could gate on
it, with the planted `TableReport` as the control.

### D3 — "a cooked graph can be cyclic" is false of the cook the tool writes

`internal/codegen/darttable/cookruntime.go:107`, emitted into every generated
`<Package>Cook.dart`:

> a cooked region's graph can be CYCLIC — a node that points at its own type is
> the corpus's own ListNode

Fixture. I uncooked `ListNode.cook` to its wire, edited node 2's `next` index
from 3 to 2 (one byte at offset 0x25 — the record length does not move), and
cooked it:

```
$ ./bin/schema cook --root ListNode --in cycle.wire --out cycle.cook tables/pointers
schema: data cycle: ListNode.next reaches ListNode, which is still open —
a cycle is refused at save and at Lock, and nothing recurses away
(docs/SPEC-TABLES.md §3.1)
```

§3.1, §6.2 and §11 all refuse it; §7.4 only says `cook-check` *survives* a forged
one. **A cook the tool writes can never be cyclic.** What is cyclic is the
*type* graph — `ListNode`'s descriptor names `ListNode` — and that is what
actually forces the tear-off. The two are different things and the sentence
names the wrong one, in source shipped to every consumer. `gotable/cook.go:866`
carries the same imprecision, so this is inherited rather than invented; the Dart
wording ("a cooked **region's** graph") is the more specific and so the more
wrong. Fix: say *the descriptor graph is cyclic — a record's field column can
name its own record*.

### D4 — §19.2's "an array is ITERATED, not indexed by hand" is not honored, and the divergence is unstated

§19.2 is explicit for every port: *"An array is ITERATED, not indexed by hand…
a range-for in C++, an enumerator in C#, the equivalent per port. A call site
never spells the pitch arithmetic itself"* — and *"A contiguous view is available
because the pitch IS `sizeof`"*. The Dart block handle emits `shipsCount` and
`shipsAt(i, [into])` and nothing else (`internal/codegen/darttable/block.go:216`);
there is no `Iterable`, no iterator, no `Iterator`-returning getter, and no
contiguous view. Grep for `Iterable|Iterator|Span` across every generated
`*Block.dart`: zero hits. Every call site is the hand-written index loop the
clause exists to prevent, and Dart is a language with a first-class answer
(`Iterable`, or a `sync*` cursor). Whatever the reason — a `sync*` iterator
allocates, and that may well be the reason — the divergence is not stated
anywhere, and an unstated one is a defect by the owner's ruling.

### D5 — `attach`'s two-part invariant is unchecked, and the documented hot loop can be made to throw

`internal/codegen/darttable/runtime.go:212` — `attach(Uint8List buffer, ByteData
buffered, TableReport reportIn)` sets `limit = buffer.length` and reads through
`view = buffered`. Nothing checks that `buffered` describes `buffer`. USAGE.md's
own hot-loop example teaches passing them as two separate fields
(`docs/USAGE.md:2210`: `reader.attach(record.bytes, record.view, report)`), which
is exactly the shape that drifts.

Two failure modes, both reproduced against the corpus:

```
matched pair:      true
mismatched pair (a short zeroed view):  returned true, malformed=false   ← SILENT WRONG ANSWER
short view holding the golden's head:   THREW IndexError: RangeError (byteOffset):
                                        Index out of range: index should be less than 16: 17
```

The second is the failure the port's own bar exists to prevent — *"a reader that
RAISES on hostile bytes is not one that REFUSES them"* — reached not by hostile
bytes but by the documented API. The fix is a Dart one: take a single `ByteData`
(it carries its own length) and derive the limit from it, or take the `Uint8List`
alone and keep a cached view keyed on `buffer.buffer`.

### D6 — no negative control on the claimed-names gate

`TestDartTableRuntimeNamesAreClaimed` is a real gate — I planted `final class
TableBogusUnregistered {}` in `internal/codegen/darttable/runtime.go` through an
overlay and it went red, naming the symbol. But the port ships no
`tables-dart-names-negative-control`, where the Rust port ships
`tables-rust-names-negative-control` (`Makefile:1227`) and runs it in `make test`.
The gate that has never been shown to go red in CI is the one nobody can trust
later.

The refusal half is sound: all eleven Dart runtime spellings I tried
(`TableEnumVocab`, `TableScratch`, `TableNarrowFloat`, `TableUnsignedLess`,
`TableCookRef`, `TableBuildVersion`, `TableJson`, `TableBitsToFloat`,
`TableBlockRead64`, `TableCookRead64`, `TableReport`) are refused as a table
name, and the lowerCamel spellings that could collide with a runtime constant
(`tableBuildVersion`, `tableCookMagic`, `tableBlockMagic`, `tableCookMaxAlign`)
generate analyzer-clean Dart, so the bijection claim holds in practice.

### D7 — a self-contradicting comment, and a control that does not do what it says

`test/dart-tables/soak.dart:1` — *"THE DART TABLES SOAK, gated on the ALLOCATION
COUNT PER ITERATION"* — is contradicted thirty-nine lines later by line 40, *"It
REPORTS the allocation count, and does not gate on it."* Present state only.

`test/dart-tables/soak.dart:382` — *"the soak's own negative control re-proves
that on every run of it, because a planted allocation per iteration has to show
up"* — is false. `SOAK_SABOTAGE=1` corrupts one byte of one re-saved record
(`c.scratch[0] ^= 1`) and plants no allocation at all. Nothing in the tree
re-proves the accumulator method.

### D8 — "a PREVIEW", unexplained, in every generated cook library

`internal/codegen/darttable/cook.go:1` and `:242` put *"the READ half, a
PREVIEW"* at the top of every `<Base>Cook.dart` and `<Package>Cook.dart`. The
word appears nowhere on the page, no other port's cook read half is called one,
and nothing says what is provisional about it. Either say what is preview about
it or drop the word.

---

## Page defects

1. **`docs/USAGE.md:2219`** — *"the soak counts the VM's allocations over a steady
   phase and holds the floor at zero"*. The soak does neither: it gates on
   correctness and reports 7.5 allocations per record (D1). The same sentence's
   first half — *"the emitter's own zero-allocation gate … refuses an allocating
   construct in any of them"* — overstates what D2 shows the gate does.
2. **`docs/SPEC-TABLES.md:4959`** — *"which is the floor its soak holds"*, same
   fact.
3. **`docs/SPEC-TABLES.md:217`** — *"allocates NOTHING per record"* needs the
   configuration named (AOT), which is what makes it true.
4. **The BIG-ENDIAN paragraph (`docs/SPEC-TABLES.md:247`) was not touched.** It
   names C++, Rust and Go and says exactly what each leg gives up. Dart's status
   — the cross-endian *refusal* is held by `cook-foreign` and `block-foreign` in
   the matrix (6/6 and 2/2), and there is no big-endian **run** leg — is absent
   by omission, which is the one thing that paragraph exists to prevent.
5. **§11's prose never names Dart** — `grep -i dart` over lines 3603–3884 returns
   nothing. Six new names are claimed in `internal/tablenames`
   (`TableEnumVocab`, `TableScratch`, `TableNarrowFloat`, `TableUnsignedLess`,
   `TableCookRef`, `TableBuildVersion`) plus five widened
   (`TableJson`, the four bits/float pairs) plus `TableBlockRead64` and
   `TableCookRead64`, and §11 has a "GO WIDENS TWO OF THESE AND ADDS EIGHT"
   paragraph and no Dart equivalent. §11's own rule — *"a claim the page states
   and the checker does not make is a name a user may take"* — runs the other way
   too: a claim the checker makes and the page does not state is a name a user
   loses with no notice. Note also that Dart spells `TableBuildVersion` where the
   page's list carries `BuildVersion`.
6. **No Dart column anywhere a consumer would look for the spelling.** §7's
   per-backend table (`docs/SPEC-TABLES.md:2267`) is C++ / C# / Go; §19.2's
   "Three spellings a reader has to have" gives C# and Go. §7's backend status
   still says only C++ and C# emit an `Open`. A consumer written from the page
   alone cannot find `<Root>Cook.open(bytes, at, length)`,
   `<Table>Block.open(bytes, base, extent)`, `<Name>Row`, or `<table>At`. Rust is
   missing from those tables too, so this is a shared gap — but the PR's own
   frame ("a consumer written from this page needs the spelling, not a
   description of one") makes it this port's to close.
7. **`test/conformance/README.md:167`** is now stale: *"804 of the 1,072 cases in
   this table"* and *"Three native legs cost two and a half seconds together"*
   described four legs. With Dart the table is 1,340 cases and four native legs.
   The new `dart alone | 2.86 s` row is correct (I measured the leg green in the
   same pass).
8. **The PR body contradicts itself on the case count** — the headline says
   "260/260 cases" and "Where it rides" says 268; the matrix in the body sums to
   268.
9. **The PR body's fuzz counts are 4× low.** `make tables-dart-fuzz` at its
   default `DART_FUZZ_MUTANTS=4000` runs **28,000** mutants (4,000 × 7
   fixtures), not the "7,000" the gates table claims. `tables-dart-release` runs
   280,000 across two seeds, not the "200,000" the Makefile comment and the PR
   body both say.
10. **No bench board is committed.** `bench/tables/README.md:126` says *"Say
    which one a board is, in the board"*, and the Go and Rust legs each landed a
    `bench/tables/results/<date>-pairing-*.md` whose first line is **"PAIRING
    CHECK, not a sitting"**. This PR adds `bench/tables/dart/leg` and
    `legs.txt` but no board — the numbers live only in the PR description, where
    they are correctly labelled but where nothing in the tree carries them. I
    ran the leg on this same shared box and reproduced them (below), so the
    board is one command away.
11. **No dartdoc.** Not one `///` in any generated `*Table.dart`, `*Block.dart`
    or `*Cook.dart`; every one of the port's (genuinely good) explanations is a
    `//` comment invisible to `dart doc` and to IDE hover. The Rust port emits 44
    `///` blocks; C# emits no XML doc, so the estate is split — but the
    "authentic and native" bar points at `///` here, and it is a mechanical
    change.
12. **The USAGE example is ungated.** C++ and C# each carry a "THE DOCUMENTED
    SURFACE COMPILES/RUNS" gate precisely so the page goes red with the code.
    The Dart section has none — which is how defect (1) above is in it.

**One thing I decided is NOT a Dart defect.** `docs/SPEC-TABLES.md:2542` says an
`attribution_length` that is not `16 × nodes` "refuses"; the Dart `open` accepts
one when the file equation still balances. But §7's `Open` *enumeration* does not
list it, §7.4 makes the directory `cook-check`'s business, and no port checks it
(`grep '% 16'` across all five backends is empty). The ambiguity is the page's,
not this port's.

---

## The feel: is this the surface a Dart programmer would have written?

Mostly, and the places where it is not are mostly named. Where it is Dart, it is
good Dart: `final class` and `abstract final class` for namespaces, `open`
returning a nullable rather than a bool-plus-out-parameter, `TableKeyed.keys`
as an `Iterable<int>`, `dart format`-canonical output held by a gate, and an
`operator []` that refuses `None` and `E.Max + 1` with an `ArgumentError` rather
than an `assert` — I confirmed that refusal survives `dart compile exe`, which is
what "in every build" means in a language whose asserts vanish.

Six places read as transliterated C++:

1. **Free functions where Dart wants methods or extensions.** `teamConfigMeasure(value)`,
   `teamConfigSave(value, buffer)`, `teamConfigLoad(value, bytes, report)`,
   `teamConfigReset(value)` — every one takes the object it operates on as its
   first argument. No Dart library is written that way; the whole core library
   is methods. §11 already leaves a port free to make its verbs members ("a
   language whose accessors are MEMBERS claims nothing at file scope"), and both
   C# and Go took that route where their language wanted it. The idiomatic
   alternative is `config.measure()` / `config.save(out)` / `config.load(bytes,
   report)` / `config.reset()` as methods on the generated storage class, and
   **`extension` methods** for the `type`s a table reaches whose class the packet
   emitter owns (`ScoreBoard` comes from `Keyed.dart`) — extensions are
   statically dispatched, cost nothing, and exist for exactly this. That one
   change would also stop the port planting 24 claimed suffixes at library scope
   in every generated file. The stated reason for free functions is a
   PascalCase↔lowerCamel bijection that keeps the name registry to one entry —
   a registry convenience, not a language constraint, and it is not stated as a
   divergence at its site.
2. **The `([Row? into])` optional out-parameter on every row and record
   accessor.** `block.shipsAt(i)` allocates; `block.shipsAt(i, cursor)` does not.
   The idiomatic-looking call is the allocating one, and a per-frame consumer
   who writes the obvious loop pays for it. Dart's answer is the cursor as the
   receiver (`cursor.moveTo(block, i)`) or a mutating `void shipsAtInto(int i,
   ShipRow into)` with no allocating twin, so the fast path is the only path.
   This is the C++ output-parameter idiom carried over intact; the reason is
   commented, but the default is the wrong way round.
3. **`attach(buffer, view, report)`** — see D5. One fact split into two
   independently-mutable arguments with no check. `ByteData` already carries a
   length; a Dart API takes one argument.
4. **`-1` as a refusal from `int <name>Save(...)`.** Dart's sentinel is `null`,
   and `int?` costs nothing on the VM for a Smi. `-1` is a C error code.
5. **`TableKeyed` exposes `slots` publicly beside `operator []`,** and the
   generated codecs index `value.teams.slots[i]` directly — so both mental
   models (0-based storage, 1-based key) are in the public surface at once, which
   is the confusion `operator []` was written to end. Make `slots` private to the
   class and add a `void forEachSlot(void Function(int key, T slot))` or an
   indexed `slotAt(int i)`.
6. **Argument order is inconsistent between the two entries:**
   `xLoad(value, bytes, report)` puts the destination first,
   `xLoadBody(reader, value)` puts it second.

Divergences that ARE forced by the language and ARE stated at their site, and
which I would have written the same way: the `double` storage for a `float32`
with `tableNarrowFloat` before every elision comparison; the `const` descriptor
graph with static-method tear-offs for the memory columns; `(buffer, offset)` as
the block base with the 64-byte check on the offset; the `Uint8List` + used-length
pair for `string(N)`; `TableCookRef.outside` as an answer rather than a
`RangeError`; and spelling `%.*g` out over `BigInt` and paying for it openly. The
last is the port at its best — the price named rather than hidden, with the
corpus case (`num-float32-upper-band`) that forced it.

---

| what | result |
|---|---|
| `make conformance` | green; **dart 12/12 surfaces, 268 cases**; wire 18/18, report 8/8, json-read 18/18, json-write 18/18, json-hostile 66/66, cook 6/6, cook-foreign 6/6, block 2/2, block-foreign 2/2, block-dump 2/2, forgery 11/11, cook-forgery 111/111 |
| my consumer over the wire and JSON goldens (§3 scanner of my own, measure==save byte-compare, text round trip, every prefix truncation, every single-byte flip, doubled bodies, spliced unknown ids, the reserved `0xFFFF` transport, an out-of-set kind, `Reset`) | **423,036 checks, 0 failures** under the JIT; **423,036 / 0** again under `dart compile exe` |
| my refusal battery derived from §7 / §7.1 / §19.1 / §19.2 (magic bytewise and reversed, the other accelerator's magic, build version, order word, both reserved words, thirteen alignment words, the file equation ±1/±8/±64, data-part-holds-root, unaligned and aligned non-zero bases, null buffer, negative base, short claims, count == max and max+1, pitch, array in the prologue, unaligned array start, overlap-must-open, every prefix, 120,000 seeded mutants with a full row walk) | **129,025 checks, 1 deviation** — the `attribution_length % 16` case, which §7's `Open` enumeration does not name and which no port checks; **0 Dart-specific failures** |
| keyed arrays | `keys` yields 1…E.Max, key 1 is slot 0, key E.Max is the last slot; `None`, `E.Max + 1` and a negative key all raise `ArgumentError` — **in the JIT and in the AOT build alike** |
| `make tables-dart-fuzz SEED=99991` (fresh seed) | **28,000 mutants over 2 blocks and 5 cooks — 0 findings** |
| `make tables-dart-fuzz-negative-control` | RED, correctly: *"cameras: opened with a count of 4, past the declared 1"* |
| `make tables-dart-soak DART_SOAK_SECONDS=25` | 303 iterations × 8 records = **2,424 wire and 2,424 text round trips, every one byte-compared**; passes; reports 7.5 allocations per record (D1) |
| `make tables-dart-soak-negative-control` | RED, correctly: *"SOAK FAILED: root_full: the round trip does not reproduce the golden"* |
| `make tables-dart-clean` | analyzer clean and byte-identical to `dart format` |
| `make tables-dart-alloc-negative-control` | red on `Uint8List(` — and **green** on a bare list literal, on `List.filled`, and on `Uint8List.fromList` (D2) |
| claimed names | 11/11 Dart runtime spellings refused as a schema declaration; my own negative control on `TestDartTableRuntimeNamesAreClaimed` goes red; the port ships no such control (D6) |
| `TestDartWirePathAllocatesNothing`, runtime cross-check | AOT: **0 scavenges over 2,000,000 records**; JIT: 54 per 1,000,000 on `load` and 54 on `save`, 0 on `idle` and on `measure` |
| the proposed runtime gate's negative control | planted `TableReport` per record → 31 scavenges; planted `Uint8List(8)` → 31; clean → 0 |
| the cycle fixture | `schema cook` refuses a cyclic wire by name (D3) |
| the bench leg | builds AOT; `--round 1` on this shared MacBook Air: write **0.251 M msg/s**, round_trip **0.134 M msg/s**, `corpus_id 2f7567e1e25ba918` — within noise of the PR's 0.238 / 0.135, so the body's numbers reproduce **as the pairing check they are labelled**; the pre-clock golden gate fires on a corrupted `bench_table.bin` (*"WIRE GOLDEN MISMATCH … refusing to bench code that does not match the corpus"*, exit 1) |

Nothing was pushed to the branch and nothing in the tree was modified; every
sabotage went through `go test -overlay` or a copy, as the port's own controls do.

gafferongames:
Read taken, one line per item. The branch is rebased onto main (c960619) and force-pushed; the PR body is rewritten to the present state.

**Defects**

- **D1** — The true claim is on both pages now: **zero under `dart compile` (AOT)**, measured and gated; **under the JIT, printed and not gated**, because what the JIT boxes is the inliner's decision — a `double` crossing a conversion call the inlining budget left out of line — and where the budget runs out moves with the loop around the codec. Measured: one boxed double per pass of the eight-record corpus on the wire phase (one scavenge per 160,000 records at a 1 MB semi-space), up to three per record when a single codec is inlined into a monomorphic caller; zero on `load`, `measure`, `save` and `idle` alone. That is down from the seven per record you measured: the runtime's one currency is the `Uint8List` (scalars assembled from bytes, no `ByteData`), every conversion is `prefer-inline`, and the NaN widening is a cold never-inline call so the hot conversion has one exit. The two remaining costs are stated: a `float32` NaN with a payload (one boxed double) and a 64-bit field outside ±2⁶² (one boxed integer per read); neither under AOT.
- **D2** — `TestDartWirePathAllocatesNothing` is deleted. `make tables-dart-alloc` is the gate: new-space scavenges under `--verbose_gc` between two marker lines, semi-space pinned at 1 MB so a plant fills it, phases `idle | load | measure | save | wire`, AOT held at exactly zero, JIT printed beside it. `make test` length (20,000 × 8 records), AOT: 0 in every phase; JIT: 0 / 0 / 0 / 0 / 1 (idle, load, measure, save, wire). `tables-dart-alloc-negative-control` plants a `TableReport` per record and a `Uint8List(8)` per record: 8 / 5 scavenges under AOT, 9 / 6 under the JIT — red on both, in both builds, so the instrument is shown to see a plant where it prints as well as where it gates. `tables-dart-soak` gates on correctness and then runs the allocation gate. A scan over source is never an allocation gate.
- **D3** — "a cooked graph can be cyclic" is gone from the Dart cook runtime, from `gotable/cook.go`'s three copies, and from the page: the DESCRIPTOR graph is cyclic (`ListNode.next` names `ListNode`); a cooked region never is, the tool refuses one by name (§3.1).
- **D4** — `for (final row in block.ships)` yields one cursor moved to each row at the instance's pitch; `shipsAt(i, cursor)` takes a lent cursor with no allocating twin; `shipsBytes()` is the contiguous view. §19.2 carries the Dart spellings.
- **D5** — `attach(Uint8List, TableReport)` and `attach(Uint8List)`: one argument carries the memory, nothing to drift apart, and the `Uint8List` is the reader's one currency rather than a `ByteData` beside it.
- **D6** — `make tables-dart-names-negative-control` plants `TableBogusUnregistered` through `go test -overlay` and requires the name test red naming it; in `make test` and `tables-dart-release`.
- **D7** — The soak's header says what it does (correctness under reuse) and points at the gate for the floor; the false "re-proves on every run" sentence is gone. The soak's control still corrupts a byte; the allocation control is the gate's own, above.
- **D8** — "PREVIEW" is gone from both cook headers.

**Page defects**

1. USAGE.md's Dart section states the measurement, not the scan; the example is the corpus's own `RootConfig` and `make tables-dart-usage` runs it verbatim off the page (12).
2. / 3. §16 and the backend status say what `make tables-dart-alloc` measures, and where.
4. The BIG-ENDIAN paragraph now says what Dart gives: the cross-endian refusal (cook-foreign 6/6, block-foreign 2/2), byte order the reader's by construction, no run leg, the cross-and-emulate leg a named follow-on.
5. §11 has the Dart paragraph: six added, seven widened, `TableBuildVersion` named as the page's `BuildVersion` in Dart's spelling, and the nine member verbs claimed against field names (checker + tests both ways).
6. §7's per-backend table and §19.2 carry a Dart column: `open`, the handle, `cursor()`/`root(into)`, the bounded deref, `<Name>Row`, `rowSize`/`rowAlign`.
7. `test/conformance/README.md` re-counted for seven legs: 1,876 cases, six native legs, seven seconds.
8. / 9. The PR body: the matrix as the harness prints it today (16/16 +4a, 12/12 +3a, cook-write absent), 28,000 fuzz mutants at the default and 280,000 in the release gate.
10. `bench/tables/results/2026-09-03-pairing-dart-arm64-macbook.md` is committed, labeled a pairing check, with its two CSVs: cpp 2.408 / 0.846, dart 0.437 / 0.227 M msg/s (0.18× / 0.27×).
11. Every generated explanation is `///` dartdoc.
12. `make tables-dart-usage`, above.

**The feel**

1. Verbs are methods on the table's class and `extension <Name>Table on <Name>` for a closure `type`; nothing at library scope per member. `Table` joins the suffix set and the nine verbs are refused as field names, at the site and on the page.
2. No `[Row? into]` anywhere: the cursor is lent, `<field>Cursor()` makes it, the iterator moves one.
3. One argument, above.
4. `-1` stays, stated as a divergence with the reason: the reference's measure and save answer one contract and the port does not split it.
5. `TableKeyed`'s storage is private; the key is the only index, `[]`, `[]=`, `fill`, `keys`.
6. `loadBody(reader)` / `saveBody(writer)` are methods, so there is no argument order to disagree about.

**Coordinator's additions:** per-case `.absent` for the pointered cases on every surface the driver answers (wire +4a, report +3a, json-read +3a, json-write +3a, json-hostile +16a; `cook-write` absent as a surface, as Go's is); the fixed walk refuses the `&` key prefix §16.7 reserves (`reserved-ampersand-key` passes, the corpus's own case); kind 17 in the tolerant reader's skip, held by `p2_as_p1` / `p2_as_p3` in the report surface; `conformance-negative-control-dart` rebuilds the compiler with `go build -overlay` over a sabotaged `jsonruntime.go` and requires `json-read` red with `json-write` and `wire` green; `tables-dart-json-walk` and `tables-dart-standalone` (with its own control) in `make test`. Rebased onto main at #353 (5a04785) with every language's lines kept; the Dart row above is the harness on that corpus.

Still a draft.

gafferongames:
## Confirming read: LAND

Own worktree at `869a2e9` (the rebase on `5a04785`), pinned SDK 3.13.2 via `DART=…/dist/dart-sdk-3.13.2/bin/dart` (the same pin CI uses), Go 1.27. Every item below was RUN, not read, unless it says read.

**D1** — claim sites read: `docs/USAGE.md:2867` ("holds the count at zero under `dart compile`'s AOT snapshot … Under the JIT the same instrument prints its count and does not gate on it"), `docs/SPEC-TABLES.md:404` and `:6397`, `runtime.go:359` ("holds AOT at zero and prints the JIT's count beside it"), `soak.dart:11` ("held at zero under AOT and printed under the JIT"). No unqualified "allocates NOTHING" survives in the Dart sections. `make tables-dart-alloc DART_ALLOC_ITERATIONS=20000`: AOT idle/load/measure/save/wire **0 / 0 / 0 / 0 / 0**; JIT 0 / 0 / 0 / 0 / 1 printed, not gated; text 22,673 printed as the priced floor.
**D2** — `TestDartWirePathAllocatesNothing` is gone (`grep AllocatesNothing compiler/` finds only the C++ header gate). The gate is `gcgate.dart` under `--verbose_gc --new_gen_semi_initial_size=1 --new_gen_semi_max_size=1`, scavenges counted between the two marker lines. `make tables-dart-alloc-negative-control DART_ALLOC_ITERATIONS=20000`: plant-report **8 AOT / 9 JIT**, plant-bytes **5 / 6**, red on both in both builds. `make tables-dart-soak DART_SOAK_SECONDS=3`: 744 wire and 744 text round trips byte-compared, then the soak invokes `tables-dart-alloc` and the gate prints the same 0/0/0/0/0 under AOT.
**D3** — `grep -i cyclic`: `cookruntime.go:112` "A COOKED REGION IS NEVER CYCLIC", `cook.go:505` "the DESCRIPTOR graph can be cyclic", `gotable/cook.go:238/657/869` the same, `SPEC-TABLES.md:442–444` the same. CI's generated job run as it runs it (`make` over all 36 `generated/*/.stamp` rules, then `git diff generated/` and `git status --porcelain generated/`): clean, nothing untracked — `generated/bench/tables/go/BenchTableCook.go` matches the emitter.
**D4** — `block.go:246`: `Iterable<Row> get <field>` yields one cursor moved per row (`cursor.at = at + i * pitch; return cursor`), `<field>Cursor()`, `<field>At(i, cursor)` with no allocating twin (`grep 'Row? into'` empty), `<field>Bytes()` as a `sublistView`. `SPEC-TABLES.md:7812–7830` carries all four Dart spellings under §19.2.
**D5** — `runtime.go:233` `void attach(Uint8List buffer, TableReport reportIn)`, limit derived from `buffer.length`; writer `attach(Uint8List)` at `:130`. No `ByteData` argument anywhere.
**D6** — `make tables-dart-names-negative-control`: red naming `TableBogusUnregistered` at `tables_test.go:2346`; runs in `make test` (line 121 of the recipe) and in `tables-dart-release`.
**D7** — soak header read: "gated on CORRECTNESS under reuse", "THE ALLOCATION FLOOR IS NOT THIS FILE'S TO HOLD", the "re-proves on every run" sentence gone. `make tables-dart-soak-negative-control`: red, "root_full: the round trip does not reproduce the golden".
**D8** — `grep -i preview internal/codegen/darttable/`: empty.
**Feel 1** — verbs are members (`SPEC-TABLES.md:421`); `extension <Name>Table on <Name>` at `darttable.go:619`; `"Table"` in the suffix set (`check.go:2881`) and the nine verbs refused as field names (`check.go:2897`); page paragraph at `SPEC-TABLES.md:5130`.
**Feel 2** — no `[Row? into]` in any emitter file; the lent-cursor shape above.
**Feel 3** — one argument, D5.
**Feel 4** — `-1` stays; the divergence is stated on the page (`SPEC-TABLES.md:424–429`, "one contract in the reference and this port does not split it"). The generated dartdoc at `codecs.go:302` states the value but not the divergence — the page carries it, the site does not name `null`. Nit.
**Feel 5** — `TableKeyed._slots` private (`runtime.go:481`); `[]`, `[]=`, `fill`, `keys` are the surface.
**Feel 6** — `loadBody(reader)` / `saveBody(writer)` are methods; no free-function pair to disagree.
**Coordinator: per-case absent** — `./build/conformance-harness run --only dart` as CI runs it: wire 16/16 +4a, report 12/12 +3a, json-read 16/16 +3a, json-write 16/16 +3a, json-hostile 68/68 +16a, cook-write absent, cook 6/6, cook-foreign 6/6, block 2/2, block-foreign 2/2, block-dump 2/2, forgery 11/11, cook-forgery 111/111. Exactly the reply's matrix.
**Coordinator: kind 17** — `runtime.go:311` `case 17:` in the 4-byte skip arm; `p2_as_p1` and `p2_as_p3` answered in the report surface (`build/conformance/out/dart/report/`), inside the 12/12.
**Coordinator: conformance-negative-control-dart** — `go build -overlay` over the sed'd `jsonruntime.go`, then the harness: exit 0 with "dart / json-read: the driver exited 1" and the three greps (json-read FAIL, json-write pass, wire pass) all satisfied.
**Coordinator: json-walk / standalone** — `tables-dart-json-walk`: "one walker per unit, byte-identical across 7 units". `tables-dart-standalone`: green; its control red on the planted `package:collection` import. All three in `make test` (recipe lines 123–125). `tables-dart-usage`: "the documented Dart surface runs, and reproduces the golden". `tables-dart-clean`: analyzer clean, format-canonical.
**Coordinator: Backend** — `tablenames.go:33` `type Backend uint16`, Dart is the ninth bit. **No test that the ninth bit fits**: nothing in the tree asserts `tablenames.Dart != 0` or the width. What holds it is Go itself (`1 << 8` into a `uint8` is a compile-time constant overflow) and `TestDartTableRuntimeNamesAreClaimed`, which would go red on every name if `Dart` were zero. The widening is held; the named test is absent. Nit, one line.
**Coordinator: dist / Makefile line 1** — `git ls-files dist`: empty. Makefile line 1 is the header comment.
**Coordinator: American spelling** — the diff adds three British forms: `localise` twice (`conformance-negative-control-dart`, the two "does not localise the READER/walk" messages — copied from 23 sibling lines already on main) and `normalises` once (`compiler/tables_test.go`, the names scan comment). Nit; the sweep belongs on main across all 26 sites.
**Go tests** — `go test ./compiler -run Dart -v`: `TestDartEmitsTableSources`, `TestDartRefusesPointeredTables`, `TestDartTableRuntimeNamesAreClaimed` PASS, none skipped; `./internal/check` ok.

Verdict: every D1–D8 fix is in and reproduces; the three nits (a one-line ninth-bit test, three spellings, the `-1` dartdoc naming the divergence) are not claims and do not hold the landing.

### Against #389's technique list

**This port uses, the list lacks:**
- the allocation gate split by build configuration — gated where the language ships (AOT), printed where it develops (JIT) — with the semi-space pinned at 1 MB so one plant per record fills it (`DART_GC_FLAGS`)
- the standalone gate (a generated library imports the platform and its own unit only) with its own overlay negative control
- the generic-walk identity gate (one walker per unit, byte-identical across units)
- the documented surface run verbatim off the page (`tables-dart-usage`; C++ and C# have the same gate, the list does not name it)
- the correctness soak under storage reuse, byte-compared, with a corrupt-one-byte control
- the analyzer-clean and format-canonical gate on emitted text
- the bench leg's pre-clock golden gate (`WIRE GOLDEN MISMATCH` refuses to time code that does not match the corpus)
- the claimed-name test held both ways (emitter to registry AND registry to emitter) plus member verbs refused as field names on every closure member
- the release target's second fuzz seed

**The list has, this port lacks or holds elsewhere:**
- block lead / base-alignment check as a NAMED target: held by the harness forgery surface (`cook_lead_1..5`, the driver's `lead` parameter) rather than a `tables-dart-block-lead` target the way Elixir has one — reason not stated
- the cross-endian refusal as a named gate: held by `cook-foreign` 6/6 and `block-foreign` 2/2 in the matrix, no named Makefile target — reason not stated
- the fuzz control that removes BOTH bounds: `tables-dart-fuzz-negative-control` removes the count bound only — reason not stated
- identity map (#373), flat node table (#376), hooks and allocator contract (#386): not applicable — the Dart tier refuses pointered units by name (§11), stated in the driver and on the page

Worktree removed, nothing pushed to the branch.

PR #369, opened 2026-09-03
tables in Elixir: the tolerant wire, JSON, reflection, the block and cook readers — mirrored from C++, harness green

gafferongames:
The Elixir table backend (`internal/codegen/elixirtable`), mirrored from the C++
reference the way `internal/codegen/cstable`, `internal/codegen/gotable` and
`internal/codegen/rusttable` are, and registered as the sixth leg of the tables
conformance harness — **green on every one of the twelve surfaces**, beside the
five that were already there.

The elixir leg, on the corpus #383 left — #376's four evolution seams, and the
four variable-class instances and three P2 report cases that every port but the
reference declares ABSENT per case, this one included (§11 refuses a pointered
unit's wire by name):

```
surface       elixir
wire          16/16 +4a
report        12/12 +3a
json-read     16/16
json-write    16/16
json-hostile  66/66
cook           6/6
cook-foreign   6/6
block          2/2
block-foreign  2/2
block-dump     2/2
forgery       11/11
cook-forgery 111/111
```

The four new seams pass with the counters `reports.txt` pins, `1,1,0,0,false` —
one UNKNOWN for the node table's reserved id and one KIND MISMATCH for the
`next` field. **The reader needed nothing for them**: kind 17 was already in the
skipper's fixed-width row, because §3's closed set was mirrored whole rather
than as the kinds this corpus happened to carry.

The leg answers 266 cases, and declares 7 absent, in 5.6 s — twelve BEAM boots at 0.26 s each, and
none of it is the cases. The driver's own modules are compiled with the corpus
rather than at every start, which was another 0.4 s an invocation (#320).

## Where it sits in the split CI (#372)

- **`ci.yml`** — one more row in the CONFORMANCE MATRIX, beside cpp, cs, go and
  rust. It clones no sibling runtime (the generated Elixir is self-contained on
  the BEAM) and installs the pinned OTP 29 / Elixir 1.20.4 and nothing else.
- **`certify.yml` is UNTOUCHED.** Its `release-gates` job derives the target list
  from the Makefile, which is the whole point of that job — a port lands its
  release gate by adding the target and nothing else.
- **`Makefile`** — `tables-elixir-release`, which that job finds BY NAME: the
  fuzz oracle at ten times the push count, the allocation audit and its negative
  control, a five-minute soak and ITS negative control, and the bench leg's
  GOLDEN GATE through the leg's own `--gate` verb (4.6 s, because a
  certification job does not need eight timed runs to learn that variant 0
  matches the pin). It is the **first `tables-<lang>-release` target in the
  tree**; that job's "nothing to certify here" branch now has something to find.
- **No `ELIXIRC` override anywhere.** The Makefile's default prefixes the
  repo-local `dist/` onto PATH and falls through to whatever the runner
  installed — the mechanism `ELIXIR` and `MIX` have always used. Proved rather
  than assumed: with `BEAM_PATH` pointed at an empty directory it still compiles
  all 208 modules.

## What lands

- **THE TOLERANT WIRE** for the fixed class: storage, measure/save/load, the
  read report, `?T`, `[E]T` over an `E.Max` tuple with its own accessor and
  iteration surface (§2.4 — the shift has one home), unions as the packet
  emitter's tag beside pre-allocated arms, guards, strings, bytes and every
  scalar kind — explicit little-endian bitstring matches throughout.
- **THE TEXT FORM**, as ONE generic walk over the reflection descriptors
  (§16.1), emitted once per unit as `TableRuntime.ex` and held byte-identical
  across the corpus by `make tables-elixir-walk`. No Jason, no Poison; C's
  `%.*g` reproduced digit for digit, and the canonical text ends with exactly
  one newline.
- **THE REFLECTION DESCRIPTORS** (§8) as MODULE ATTRIBUTES, so a read of one
  copies no term and allocates nothing. They carry the STRUCT KEY of each field
  rather than a storage offset, because an Elixir struct has keys and no
  addresses.
- **THE BLOCK FORM's read side** (§19): the whole of §19.2's Open check list,
  the block descriptors, and the typed row accessors. A row is a SUB-BINARY
  over the caller's own block.
- **THE COOKED FORM's read side** (§7, preview): the whole header check, the
  cook descriptors, and a reference resolved through the eight-byte signed
  self-relative delta with a delta past the region refused.

## The gates beside the matrix

| gate | what it holds |
|---|---|
| `make tables-elixir-fuzz` | 20,000 mutants over the seven fixtures: 9,676 opened and walked inside the buffer, the rest refused, and **no read left one**. A deliberate refusal is a `RuntimeError` with a sentence in it; a `MatchError`, `ArgumentError` or `FunctionClauseError` is the runtime saying a read went past the end of a binary, and that fails |
| `make tables-elixir-alloc-audit` | the heap words and reductions ONE iteration costs, per case, against a pinned budget |
| `make tables-elixir-alloc-negative-control` | one extra allocation per iteration reds all 18 |
| `make tables-elixir-soak` | the corpus in a loop for an hour, bytes compared every iteration; the last line is a CHECK on the FLOOR of both series — neither reading is a level, so a leak is what lifts the minimum |
| `make tables-elixir-soak-negative-control` | a RETAINED binary per iteration reds it — a different sabotage from the audit's, whose extra allocation is freed and lifts no floor, which is why both instruments exist |
| `make conformance-negative-control-elixir` | the READ path's placement broken by one key: `json-read` reds, `json-write` stays green |
| `make tables-elixir-walk` | one table runtime, the same bytes in every unit |
| `TestElixirEmitsTableModules` | zero cost — adding a table moves not one byte of the packet modules |
| `TestElixirRuntimeNamesAreClaimed` | every unit-level module the emitter defines is registered, with the repro and its negative control beside it |

## The soak, run

**21,882 iterations over the hour**, the whole 18-instance corpus each time,
with the wire and the text byte-compared every iteration so a run that drifted
would stop rather than get slower. **357 of its 358 samples read the live heap
at exactly zero words** against the warm baseline (the 358th, after the final
collection, read −1,976). Binary memory oscillated between about a kilobyte and
about 855 KB from the first minute and never grew — which is the carrier
behaviour the floor gate exists to see past, and the reason the gate is a floor.

The release gate's own length, run with the CHECKED verdict: **1,463 iterations
over 59 samples, the live-heap floor moved −988 words and the binary-memory
floor +48 bytes.** Its negative control sabotages the EMITTER through `go build -overlay`, as the
fuzz control does: every generated load retains a copy of its bytes under a
fresh key, and both floors red and are named. The audit's control is the same
shape with a freed allocation, so the two remain different sabotages.

## The bench row — A PAIRING CHECK, not a sitting

Five interleaved rounds, cpp and elixir alternating on the same machine. **This
certifies nothing**: a MacBook Air with other workers building, every spread
around 20%, so the ABSOLUTE rates are order-of-magnitude only.

| lang | path | M msg/s | MB/s | ratio to C++ |
|---|---|---:|---:|---:|
| cpp | write | 2.362 | 5386.0 | 1.00 |
| elixir | write | 0.065 | 147.3 | **0.027** |
| cpp | round_trip | 0.809 | 1844.6 | 1.00 |
| elixir | round_trip | 0.024 | 55.4 | **0.030** |

**About 37x C++ on write and 33x on round-trip.** The RATIO is what reproduced:
a worse-contaminated sitting the same evening (54.9% and 42.3% on the C++ rows)
put the same two at 36.3x and 36.9x against C++ rates that differed by 19%.

**The lever, named against #174.** That issue's LEAD hypothesis —
generation-time bit-syntax swizzling — has **no purchase on this leg**, and
saying so is the useful half: it targets the TYPE wire's manual integer packing,
where the wire's LSB-first order inside 64-bit words fights bit syntax's
MSB-first grain. The TABLE wire is byte-oriented throughout (§3), every segment
is byte-aligned and literal-width, and the emitter already hands each field to
the ERTS bit-syntax engine as ONE construction expression. What reaches this leg
is #174's SECOND avenue, in its own words — *"iodata shapes revisited under
literal widths"* — with its obstacle named beside it: ELISION makes adjacency a
runtime fact, so which fields are neighbours is not knowable at generation time.
#174's third, *"an honest floor statement"*, is what this board is.

**Two candidate levers, banked and REFUTED**, because a refutation costs the
next person the same measurement:

- **`measure` inside `save`** — 2.6 µs against `save_body`'s 15.4, so 17%, and
  the obvious swap to `IO.iodata_length` buys a fraction of that. Left alone.
- **One `Map.put` per field on the read path** — predicted 2 to 4x from
  accumulating and building the struct once. **Measured 1.8x SLOWER**, 0.53 s
  against 0.97 s over 200,000 iterations of a thirty-field struct: ERTS's
  flatmap update is already cheap and `struct/2` reduces over the list on top of
  building it. Wrong in the sign, which is a refutation.

## The deviations, all forced by the language and all named in the package doc

- The **report is a value the caller threads**: the BEAM has no mutable struct.
- **Storage is SPEC §6.1's Elixir column**, not a byte image: a `string(N)` is a
  binary whose `byte_size` IS the used length, an array is a list whose `length`
  IS the count. No `_length` or `_count` companion, because there is nothing to
  keep in step.
- An optional `?T` is **`nil` when absent**.
- An enum-keyed array's storage is a **tuple** of `E.Max` slots on a `table`, so a
  slot is reached in constant time — and the packet emitter's **list** on a `type`,
  whose struct is the packet emitter's; the generated accessor says so at its site.
- A non-finite float travels as **`{:nonfinite, bits}`**, the convention the
  packet emitter already established.
- **Both `Open`s take a `lead`** beside the bytes — how many bytes past an aligned
  base the caller's buffer begins. §7 and §19.2 check the alignment of the BASE,
  and a BEAM binary has no address a caller can observe or place. It is an
  ADDITION and not a subtraction: the alternative is a leg that cannot refuse an
  unaligned base at all, which is what the cook battery's 55 `cook_lead` rows
  are for; the block battery carries no such row (its pointer column is pinned
  at 0 from the reference driver), so `make tables-elixir-block-lead` holds the
  rule over every lead in 0..64, with a negative control on the emitter.
- **One refusal spelling.** Everything that can refuse answers `{:ok, result}` or
  `:error` — the packet emitter's reader verdict — and each of the four writers
  (`measure`, `save`, `to_json`, `to_json_measure`) has a bang form that answers
  the result or raises `ArgumentError`, the packet emitter's writer contract.
- **Decimal → binary float is a single correctly-rounded conversion over exact
  integers**, at the field's own width. Erlang has no `strtof`, and converting
  through a double rounds twice — the two roundings part company between
  `FLT_MAX` and the float32 rounding midpoint, which is what
  `num-float32-upper-band` pins.
- **The VARIABLE class's wire surface is refused by name** (§11), as it is in
  C#, Go and Rust; both accelerators are emitted for a pointered unit all the
  same, because neither needs a codec.
- **"The read path allocates nothing" is not claimed.** Elixir has no
  caller-owned buffer and no mutable struct, so a decoded value IS an
  allocation. What is held instead is that the COUNT does not move.

Draft. Do not merge.


gafferongames:
# Blind read: LAND-AFTER-FIXES

Read as an Elixir programmer who had never seen this port. I read `docs/SPEC-TABLES.md` and `docs/USAGE.md`'s Elixir section first, wrote my own consumer over the generated package's public surface — my own §3 wire decoder, my own §5 `fold16`, my own §7.1 and §19.1 header parses, my own refusals derived from §7.4 and §19.2 — ran it, and only then opened `internal/codegen/elixirtable`. 628 checks, one failure. The wire, the text form and the cook reader are right, and the generated Elixir reads like Elixir rather than like transliterated C++. Two things have to move before this lands; three more are a gate, a knob and a sentence.

Toolchain: the pinned `dist/otp-29.0.5` and `dist/elixir-1.20.4`, per the Makefile.

---

## Verdict one — the bytes

### D1 (blocking). The block `Open` takes no `lead` and checks no base alignment — and both pages say it does

`block_open_<table>` is arity **1** in every generated Block module. `cook_open_<root>` is arity 1 **and 2**, and its arity-2 form does exactly the check §7 asks for. The block form drops it, and says so in the generated source:

`internal/codegen/elixirtable/block.go:222`
> `# THE BASE'S ALIGNMENT IS NOT CHECKED, and that is a fact about the BEAM rather than a gap: a binary has no address a caller places…`

A cook's base is a BEAM binary too, and the port answers that one by taking a stated `lead`. The reason given does not distinguish the two forms; the port itself does not.

Every other backend checks it: `cpptable/block.go:660`, `cstable/block.go:598`, `gotable/block.go:634`, and `javatable/block.go:481`, which uses the offset's residue — precisely the "no address" case, solved.

No gate catches it. Every row of the block forgery battery carries `pointer 0`, stated in `testdata/conformance/tables/manifest.txt:107`: *"`<pointer>` is 0 — every block row is read out of an ALIGNED base, and the column is the cook battery's."* So the leg is 11/11 on a battery that never asks.

Reproduction:
```
$ elixir -pa build/elixir-tables-ebin -e 'IO.inspect(Blockdemo.RenderBlock.__info__(:functions) |> Keyword.get_values(:block_open_render_frame))'
[1]     # and the cook's is [1, 2]
```

Fix: give `block_open_<table>` the same `(data, lead \\ 0)` the cook has, refuse `rem(lead, 64) != 0`, and add the block battery's missing `block_lead_*` rows. Or, if the block genuinely should not take one, both pages have to stop saying it does (below).

### D2 (blocking). A schema file whose basename is not already a valid Elixir alias emits a module name Elixir refuses

`internal/codegen/elixirtable/codecs.go:33`, `cook.go:51` and `block.go:61` write `defmodule %s.%sTable` with `g.file.Base` **raw**. The packet emitter uses `ir.GoExportName(g.file.Base)` (`internal/codegen/elixir/elixir.go:599`), and this backend's own collision check already uses the exported form (`elixirtable.go:119`) — so the check and the emission disagree about what the module is called.

```
$ mkdir -p d && printf 'package probe\n\nenum Slot { A, B }\n\ntable Holder\n{\n    x int32\n    s Slot\n}\n' > d/my_frame.schema
$ ./bin/schema generate --lang elixir --out out d          # exits 0
$ grep -h '^defmodule' out/*.ex
defmodule Probe.MyFrame do          # the packet emitter, exported
defmodule Probe.my_frameTable do    # this backend, raw
$ elixirc -o ebin out/*.ex
** (ArgumentError) invalid module name: … :my_frameBlock …
```

Same for any basename starting lower-case (`tableRuntime.schema`, `my_frame.schema`). The corpus is CamelCase throughout, which is why nothing sees it. `ir.GoExportName` at the three `defmodule` sites is the whole fix.

### D3. The fuzz oracle has no negative control, and it is the only fuzzer in the tree without one

`make tables-elixir-fuzz` exists; `tables-elixir-fuzz-negative-control` does not. Java has one (`Makefile:744`), Rust has a family of them (`Makefile:1219+`), and both sabotage the **emitter** through `go build -overlay` and require the gate to find it.

I built the Elixir one to see whether the oracle has teeth. It does:

```
overlay internal/codegen/elixirtable/block.go, removing BOTH extent bounds
  (`rows > bytes - offset_of` at :274 and `padding > bytes - used` at :286)
→ regenerate, recompile, ELIXIR_FUZZ_N=20000

** (RuntimeError) the block PaddedFrame reader left the buffer on mutant 22 —
   an index out of bounds must be a refusal, never an escaping Elixir.MatchError
```

Removing only the first bound leaves it **green with an identical 9,676 opened** — the downstream padding check absorbs it, exactly as the Java control's own comment (`Makefile:737-740`) already documents for the same reader. So the control has to reach past the layer here too. That is twenty minutes of Makefile and it turns a green light nobody has tested into a gate.

### D4. The fuzz seed is nailed shut

`test/conformance/elixir/driver_impl.ex:1006` — `:rand.seed(:exsss, {1, 2, 3})`, with no `SEED` knob anywhere. Every other leg takes `SEED=$(SEED)` from the Makefile (`Makefile:996, 1025, 1209, 1334, 1585`). A fixed seed and a fixed N is a regression test wearing a fuzzer's clothes: CI explores the same 20,000 mutants forever.

I ran a fresh seed to check there was nothing hiding behind the pin — `{20260903, 369, 7}`, 40,000 mutants, **19,106 opened and walked, clean**. So this is a knob, not a bug. Add `SEED` and read it in `Fuzz.run/2`.

### D5. Keyed-array storage is a tuple only when the owner is a `table`

`elixirtable.go:277` — `keyedIsTuple() { return g.owner != nil && g.owner.IsTable }`. A keyed array declared in a `type` a table reaches gets **list** storage and `Enum.at` / `List.replace_at` accessors (`codecs.go:317-329`):

```elixir
def score_board_per_team_at(value, key) when key >= 1 and key <= 3,
  do: Enum.at(value.per_team, key - 1)              # O(n), not elem/2
```

`tables/examples/Keyed.schema:45` (`type ScoreBoard { per_team [Team]int32 }`) is exactly that case, and my consumer caught it:

```
FAIL per_team storage is a TUPLE, so a slot is reached in constant time: false
```

The constraint is real — `Tabledemo.ScoreBoard` belongs to the packet emitter, which builds it with `List.duplicate/2` — so this is forced. But a forced divergence has to be stated at its site, and this one is stated nowhere: not in the generated comment, not in USAGE, not in SPEC. Today both pages promise the tuple and constant time unconditionally.

### Everything else on this verdict, green

* `make conformance` — every leg green, **elixir 12 of 12**, 268/268 (18 / 8 / 18 / 18 / 66 / 6 / 6 / 2 / 2 / 2 / 11 / 111). Whole matrix 49.9 s.
* **§7 and §19.2 refusals I derived from the pages: 246, every one fired.** Per cook fixture (×6): reversed magic, the block form's magic, byte-order 0/2/3, a zero build version, both reserved words, ten bad alignment words, `data_length` ±1, `attribution_length` +16 and +1, six truncations (0/1/8/63/64/65), one trailing byte, six unaligned leads. Per block image (×2): a foreign magic three ways, a build version one bit away, byte-order 2 and 0, six short buffers, an array start unaligned / inside the projection / past the end, a wrong pitch, a count past the maximum, and an offset 64-aligned just under 2^63. Lead 8 on a cook whose region aligns at 8 correctly **opens** — the battery skips multiples of eight for the same reason.
* Keyed arrays: `_each` yields 1..=`E.Max` as **keys** in order, key 1 reads back out of slot 0, and `_at` / `_put` raise `ArgumentError` on `None`, on `E.Max + 1` and on a negative key, in every build, for both the table-owned and the type-owned array. A present slot rides under kind 16 carrying the variant's **name** hash.
* The one wire id my consumer could not place — `12511` in `v2_cfg` — is `fold16(fnv1a32("name"))`, which is §5's `was` working.
* Allocation audit 18/18 inside budget; the negative control reds all 18.
* Soak, 300 s: **1,705 iterations over 60 samples, live-heap floor −988 words, binary-memory floor +208 bytes.** The negative control at 30 s reds on the heap arm (+43,836 words).
* `make tables-elixir-walk`, `make conformance-negative-control-elixir`, `make tables-elixir-bench-gate` (`the golden gate passes — variant 0 is the pinned instance and all 64 round-trip at 2391 bytes`), `go test ./compiler -run Elixir`: all green.
* Claimed names: `table TableRuntime` refused at the source, and so are `BlockRuntime`, `CookRuntime`, `BuildVersion`; a schema **file** named for any of the four is refused as a file collision; a table-free unit keeps all four. The negative control is in the test beside each. The four emitted unit-level modules in the compiled corpus are exactly the four registered.
* Big-endian: the port never queries host order — `BlockRuntime`'s `@magic` and `@byte_order 1` are constants and every bitstring segment names its endianness — so a big-endian BEAM reads identically, and the cross-endian refusal is the two FOREIGN surfaces, 6/6 and 2/2. Worth one sentence on the page: Java's identical constants got `make tables-java-order` and a paragraph at SPEC-TABLES.md:318; Elixir gets neither, and the reason it needs neither is a fact about the language that should be written down rather than inferred.
* Bench: `bench/tables/results/2026-09-03-pairing-elixir-arm64-macbook.md` labels itself a PAIRING CHECK on a MacBook Air with other workers building and every spread near 20%, and says plainly that it certifies nothing. That is the right label for those numbers.

### The claim you asked me to test in particular

**The premise does not hold: the port already gates on an allocation count.** `Audit.measure/1` (`driver_impl.ex:920`) brackets 200 iterations with two full collections and takes `:erlang.statistics(:garbage_collection)`'s reclaimed words plus the live delta. The floor is the **soak's** instrument, not the audit's, and the two are correctly split.

**Keep the floor.** No allocation count can answer the soak's question: a leak is retained memory, and retention does not move an allocation count. Gating the soak on a count would duplicate the audit and go blind to the one thing the soak exists for.

**But two counters the port says do not exist, do.** Measured on the pinned OTP 29, over `root_full`:

| instrument | reading |
|---|---|
| `:erlang.statistics(:garbage_collection)` words reclaimed | 2,731 / iteration (VM-wide) |
| `:erlang.trace/3` with `:garbage_collection`, summed over 3,585 collections | **2,708 / iteration, per process** |
| `:erlang.system_info({:allocator, :binary_alloc})` → `calls` → `binary_alloc` | **exactly 3.0 / iteration, stable** |
| the audit's own formula, four consecutive runs | 2,715, 2,715, 2,715, 2,715 |

So:

* `driver_impl.ex:916` and `test/conformance/elixir/alloc-budget.txt:28` both say a refc binary's payload *"has no cumulative counter of its own."* It has one, it is `binary_alloc` in the allocator's `calls`, and it reads a flat 3 per iteration on this corpus. Pinning it beside the heap words closes the audit's own named gap and gives the soak's binary arm a count companion.
* The audit is a per-process instrument spelled with a **VM-wide** counter. `:erlang.trace/3` on `:garbage_collection` gives the strictly per-process figure. It reproduced four times running, so this is not urgent — but the 64-word tolerance is absorbing another process's collections by luck, not by design.

**And the negative control is not independent of its instrument.** `SOAK_SABOTAGE=leak` retains an 8 KiB binary inside the driver's own loop, one call frame from a reading of `:erlang.memory(:binary)` and the process heap. It proves the reading responds. It does not prove that a leak in **generated** code would red it. The Rust and Java fuzz controls do the harder thing — sabotage the emitter through `go build -overlay` and require the gate to find it — and the same shape is available here for both memory gates.

---

## Verdict two — the feel

**Mostly yes, and the wire codecs are the best part.** Multi-head functions dispatching on an explicit little-endian bitstring match, tail recursion through `fields_<table>/3`, `%{value | key: v}` on the read path, iodata built once and flattened at the surface, descriptors as module attributes. `f_weapon_config_damage(10, rest, value, report)` with the kind in the head and a `_` clause falling through to `skip_mismatch` is exactly how an Elixir programmer writes a TLV reader. I would not restructure it.

Where it reads as transliterated:

**F1. Four failure spellings on one surface.** `measure_*` answers `-1`. `save_*` answers `:refused`. `block_open_*` and `cook_open_*` answer `:refuse`. `_at` / `_put` raise `ArgumentError`. `-1` is a C sentinel; `:refuse` is a coinage where Elixir has `:error`. USAGE.md:1041-1042 annotates save's `:refused` and says nothing about measure's `-1`, and the example then feeds `wire` straight into `load_ship_config/1`:

```
$ save with an unnameable enum: :refused
$ measure with an unnameable enum: -1
# USAGE's shape gives size=-1, wire=:refused — byte_size(wire) raises
```

Idiomatic: `{:ok, binary} | {:error, reason}` from `save/1` with a `save!/1` beside it, `{:ok, size}` or a plain integer from `measure/1` with the refusal moved to the bang, `:error` rather than `:refuse` from both opens. One convention, and the pages' example then compiles as written.

**F2. Block rows are an eager list, on the path §12.1 calls the hottest.** `block.go:171`:

```elixir
def rows(data, offset_of, count, stride) do
  Enum.map(0..(count - 1)//1, fn i -> slice(data, offset_of + i * stride, stride) end)
end
```

`render_frame_lasers(block)` builds `count` sub-binary terms before the caller reads one — up to 32,000 for that array's declared maximum. There is no count/at pair, so the frame loop USAGE describes for Java (`<field>Count()` / `<field>At(i)`, "so a frame loop allocates nothing") has no Elixir spelling at all. The Elixir answers are a `Stream`, or a small struct implementing `Enumerable` whose `reduce/3` slices as it goes and whose `count/1` is O(1) — either keeps `for row <- frame.ships` at the call site and stops materialising the list. Worth doing, or worth a sentence saying the list is the price.

**F3. Handles are bare maps.** `{:ok, %{name: "RenderFrame", base: data, bytes: used + padding}}`, and the cook's `%{info: {mod, fun}, name:, region:, …}`. A `defstruct` costs nothing, makes the handle matchable and `inspect`-able, and stops a caller reaching a key that is not there. Everything else the port hands back is a struct; these two are the exception.

**F4. A row's nested-record field answers raw bytes.** `render_ship_row_position(row)` is `B.slice(row, 0, 24)` — the caller has to know `Vec3`'s layout. It is the one place the surface hands out an offset instead of a value, and it is unstated. If the reason is that decoding three floats would allocate, say so at the site.

**F5. Small.** `silent?/1`, `unknown/1`, `clamped/1` sit on `TableRuntime` rather than on the report's own module, so the report is a struct whose operations live somewhere else.

**Not a defect:** no `@spec` or `@type` anywhere. The packet emitter has none either, so this is the estate's established Elixir shape, not this port's choice. Worth raising once for both, not here.

---

## Verdict three — the pages

1. **docs/USAGE.md:1090** — *"Both take an optional `lead` beside the bytes."* False. Only the cook does (D1).
2. **docs/SPEC-TABLES.md:172** — *"`Open` takes a `lead` beside the bytes … because §7 and §19.2 check the alignment of the BASE."* False for §19.2 in this port (D1).
3. **docs/USAGE.md:1065 and docs/SPEC-TABLES.md:170** — *"An enum-keyed array's slots are a TUPLE … so a slot is reached in constant time."* Unconditional; false for a keyed array declared in a `type` (D5).
4. **`measure_*` returning `-1` appears on no page.** USAGE's Elixir example is written as though measure and save cannot refuse. Either document the sentinel or, better, remove it (F1).
5. **test/conformance/elixir/driver:21** says the per-invocation compile cost was paid *"eleven times over"*; `test/conformance/README.md` says *"twelve times over"*; the matrix has twelve surfaces.
6. **test/conformance/elixir/driver_impl.ex:660-665** describes an instrument the file does not implement — *"run the loop in a process whose heap is large enough that NO GARBAGE COLLECTION HAPPENS, and read `:erlang.process_info(:total_heap_size)` before and after … `:erlang.memory(:binary)` is read beside it."* `measure/1` at :920 calls `:erlang.garbage_collect()` twice, uses `:erlang.statistics(:garbage_collection)`, and reads no binary memory at all. The correct description is at :906-918, thirty lines below. Two accounts of one instrument in one file; delete the first.
7. **driver_impl.ex:916 and test/conformance/elixir/alloc-budget.txt:28** — *"has no cumulative counter of its own"* about a refc binary's payload. Measured false above.
8. **The branch is one commit behind main and does not merge.** `git merge-tree origin/main HEAD` conflicts in nine files including `Makefile`, `.github/workflows/ci.yml`, `docs/SPEC-TABLES.md`, `docs/USAGE.md`, `compiler/builtin.go`, `internal/tablenames/tablenames.go` and `test/conformance/drivers.txt`. The C leg (#352) landed after this base, so on merge **every status line this PR edits is wrong by omitting C** — SPEC-TABLES.md:151, :326, :3702, :5058 and USAGE.md:591 all read *"C++, C#, Go, Rust, Java and Elixir"*. GitHub still shows `mergeable`; it is stale.
9. **The PR description.** *"all 278 cases"* — the matrix is 268. And *"`ELIXIRC` joins `ELIXIR` and `MIX` in all three make invocations"* of `certify.yml` is contradicted two bullets later by *"`certify.yml` is UNTOUCHED"* and *"No `ELIXIRC` override anywhere."* The tree agrees with the last two: `certify.yml` is untouched, already carries `erlef/setup-beam` and passes `ELIXIR=elixir MIX=mix` with no `ELIXIRC`. Drop the first bullet.
10. **No byte-order gate and no sentence saying why one is not needed.** SPEC-TABLES.md:318 gives Java's constant-vs-platform reasoning a paragraph and `make tables-java-order` a gate. Elixir does the same thing with the same constants and gets neither. The reason it needs neither — every bitstring segment names its endianness, so the host's order is never consulted — is worth writing down.

---

| what | result |
|---|---|
| `make conformance` | six legs green, **elixir 12/12 = 268/268**, 49.9 s |
| my consumer, wire goldens (18) | **144 checks, 0 failures** — load\|>save byte-identical, `measure == byte_size`, my own §3 decode ending exactly at the terminator, every wire id resolving to a struct key under my own `fold16`, and the scalar values compared kind by kind |
| my consumer, JSON goldens (18) | **90 checks, 0 failures** — `to_json` byte-identical, exactly one trailing newline, `from_json\|>save` == the wire golden, `to_json_measure == byte_size` |
| my consumer, 2 block images + 6 cooks | **366 checks, 0 failures**, of which **246 refusals derived from §7 and §19.2** |
| my consumer, keyed arrays / report / claimed names | **28 checks, 1 failure** (D5) |
| **total** | **628 checks, 1 failure** |
| alloc audit | 18/18 inside budget; negative control reds 18/18 |
| soak 300 s | 1,705 iterations, 60 samples, heap floor **−988 words**, binary floor **+208 bytes**; negative control reds at +43,836 words |
| fuzz, pinned seed | 20,000 mutants, 9,676 opened and walked, clean |
| fuzz, **fresh seed** `{20260903, 369, 7}` | 40,000 mutants, **19,106 opened and walked, clean** |
| fuzz negative control (mine, both extent bounds removed) | **RED on mutant 22**; one bound alone stays green with an identical 9,676 |
| walk gate / walk negative control / bench gate / `go test ./compiler -run Elixir` | green |
| BEAM allocation counters, pinned OTP 29 | per-process 2,708 words/iteration via `:erlang.trace/3`; `binary_alloc` calls **3.0/iteration** via `:erlang.system_info` |

**LAND-AFTER-FIXES.** D1 and D2 before merge — one is a §19.2 check four other backends make and both pages claim this one makes, the other emits Elixir that does not compile. D3, D4 and D5 are a gate, a knob and a sentence, and they are cheap. The pages' ten items are a morning. Nothing in the wire, the text form, the cook reader or the shape of the generated code needs rethinking.

gafferongames:
Every item on the read, one line each, as it now stands on the branch (rebased onto `main` at #384):

**D1 — the block `Open` takes a `lead` and checks the base.** `block_open_<table>/2` beside `/1`, refusing `rem(lead, 64) != 0`; arities now `[1, 2]` like the cook's. The gate is `make tables-elixir-block-lead` (2 images × 65 leads: 0 and 64 open, 1..63 refuse) with a negative control that drops the check from the emitter through `go build -overlay`. It is LEG-LOCAL rather than a battery row, and the reason is written at the gate: the block battery is pinned from the C++ driver's own table, which prints the pointer column as a constant 0 (`test/conformance/cpp/main.cpp:813`), the Rust driver refuses a block row with any other pointer by design (`main.rs:713`), and the C# block path reads the extent alone — so a `block_lead_*` row is a change to the reference leg and three other drivers, not to this port. This leg's driver passes the column through, so the day the battery carries one it answers.

**D2 — module names are aliases.** One `moduleBase` helper (`ir.GoExportName`) at every `defmodule` site; `my_frame.schema` now emits `Probe.MyFrameTable` in `my_frameTable.ex`, the packet emitter's file convention. `TestElixirModuleNamesAreAliases` with its negative control (reverting to the raw basename reds it on `Probe.my_frameBlock`).

**D3 — the fuzz oracle has a negative control.** `make tables-elixir-fuzz-negative-control` sabotages the emitter through `-overlay`, removing BOTH extent bounds, because removing one leaves the oracle green under the pinned seed and reds on mutant 1 under another — a control whose verdict depends on the seed is not a control. Reds on the first opened mutant.

**D4 — `SEED` is a knob.** `SEED=$(SEED)` from the Makefile, read in `Fuzz.run/2`, printed in the run's own line; default fixed so a find reproduces.

**D5 — tuple-vs-list is stated.** At the generated accessor's own site (a `type`'s keyed array says it is the packet emitter's list and `_at` is `Enum.at`), in USAGE's Elixir section, and in SPEC-TABLES' status paragraph.

**Audit — the refc-binary counter exists and is pinned.** `alloc-budget.txt` carries three columns: heap words PER PROCESS (through `:erlang.trace/3` on `:garbage_collection` over the loop's own collections, replacing the VM-wide `:erlang.statistics/1`), `binary_alloc` calls from `:erlang.system_info/1`, and reductions. The "no cumulative counter" sentence is gone from both files, and the stale second account of the instrument at `:660` with it.

**Soak and audit controls — the emitter, not the driver.** `SOAK_SABOTAGE` is gone from the driver. Both controls run through one `ELIXIR_SABOTAGED_BUILD` macro (the fuzz and lead controls use it too): the audit's adds sixteen refc binaries to every generated load and requires every case to red on the heap-word column AND the binary-call column, each named; the soak's makes every generated load retain a copy of its bytes under a fresh key, and requires both floors to red and be named. The verdicts now list every arm that rose rather than halting at the first.

**F1 — one refusal spelling.** `{:ok, result} | :error` everywhere a value or bytes can refuse — `measure`, `save`, `to_json`, `to_json_measure`, `block_open`, `cook_open` — which is the packet emitter's reader verdict; each of the four writers has a bang form (`save_ship_config!`) that answers the result or raises `ArgumentError`, the packet emitter's writer contract. `-1`, `:refused` and `:refuse` are gone from the surface; USAGE's example now compiles as written and shows both forms.

**The pages (your ten):** 1–2 true now (D1); 3 stated (D5); 4 the sentinel removed (F1); 5 "twelve times over"; 6 the first account deleted; 7 corrected in both files; 8 rebased onto `main` at #384 — the C leg was already in every status line at this base, and #383's variable-class instances and per-case ABSENT are answered: the leg reads `wire 16/16 +4a`, `report 12/12 +3a`, the same cells as the other five ports; 9 the description says 266 answered and 7 absent, and the `ELIXIRC` bullet is gone; 10 a paragraph beside Java's at SPEC-TABLES §19: every bitstring segment names its endianness, so the host's order is never consulted and no order gate is needed.

**Not done, named:** F2 (rows as an eager list — a `Stream`/`Enumerable` is a pass of its own), F3 (`defstruct` handles), F4 (a row's nested-record field answering bytes), F5 (`silent?` on the report's module). None changes a byte of any wire.

Green locally, as CI runs them: `go test ./compiler`; the harness's Elixir leg on every surface; `tables-elixir-walk`; `conformance-negative-control-elixir`; the alloc audit and its control; a 45 s soak and its control; the fuzz at 20,000 and its control; the block-lead gate and its control; the bench gate.

gafferongames:
# Confirming read: LAND

Read at c2cfbdc4 in a fresh worktree of `origin/port-elixir-tables` (merge-base is `origin/main` 46b81725, #386), the pinned `dist/otp-29.0.5` and `dist/elixir-1.20.4`. Every claim in the reply was run, not read.

**The seven**

- **D1** — `make tables-elixir-block-lead`: "2 block image(s) x 65 leads — 0 and 64 open, 1..63 refuse". `make tables-elixir-block-lead-negative-control` (emitter sabotaged through `-overlay`, `rem(lead, B.align()) != 0` becomes `lead < 0`): red; the sabotaged ebin run directly says "BASE ALIGNMENT GATE FAILED: block_render at lead 1 answered open, wanted refuse". `__info__(:functions)`: `block_open_render_frame` `[1, 2]`, `cook_open_render_frame` `[1, 2]`, and every generated Block module carries both arities. Leg-local, with the reason written at the gate (`BlockLead` in driver_impl.ex) — acceptable.
- **D2** — `./bin/schema generate --lang elixir` over `my_frame.schema` (package probe) exits 0 and emits `Probe.MyFrameTable`, `Probe.MyFrameBlock`, `Probe.MyFrameCook` in `my_frame*.ex`; `elixirc` compiles all eight files clean. `go test ./compiler -run Elixir -count=1`: ok.
- **D3** — `make tables-elixir-fuzz-negative-control SEED=20260903777` (fresh): red; at the default seed: red. The sabotaged ebin run directly at the fresh seed: "the block PaddedFrame reader left the buffer on mutant 1 — … escaping Elixir.MatchError". `ELIXIR_FUZZ_SED` removes both extent bounds, and `ELIXIR_SABOTAGED_BUILD` fails a sed that matched nothing.
- **D4** — `make tables-elixir-fuzz SEED=20260903777`: "20000 mutants over 7 fixtures at seed 20260903777 — 9641 opened and walked inside the buffer, the rest refused — no read left it"; default seed 24845619678: 9581 opened, clean. `Fuzz.run/2` reads `SEED` and prints it.
- **D5** — generated `tabledemo/KeyedTable.ex:279-283` says at the accessor that `score_board_per_team_at` is `Enum.at` over the packet emitter's list because ScoreBoard is a `type`; USAGE.md:1158-1163 and SPEC-TABLES.md:176-178 say the same.
- **Audit** — `make tables-elixir-alloc-audit` twice: 16/16 inside budget both times. Largest heap-word deviation from the pin: −15 (root_full, run 1) and +90 (wide_blob, run 2, tolerance 855); every other case within ±11 against a tolerance of 64–66. `binary_alloc` calls equal to the pin in all 16 cases on both runs. Reductions within ±3.7% (root_default) of the 25% gate.
- **Controls** — `tables-elixir-alloc-negative-control`: red on all 16, each line naming both columns ("heap words 48654 over 21761 + 66, binary calls 276 over 84 + 8" for root_full). `tables-elixir-soak-negative-control` (30 s): red, with "live-heap FLOOR rose" and "binary-memory FLOOR rose" both required by the target and both present. Fuzz control red (D3), lead control red (D1). All four sabotage the emitter through `go build -overlay`.
- **F1 / the surface** — grep over `build/tables-generated-elixir/**/*.ex`: `:refuse` none; `:refused` only as `throw(:refused)` / `:refused -> :error` pairs, and no public `def` holds a throw without the catch in the same function (260 throws, 126 catches, the remainder inside `defp` helpers called under a catch). `-1` only as `count_offset: -1` / `present_offset: -1` in cook descriptors (no such column, not a refusal) and the sign in `TableRuntime`'s number parser; no accessor answers −1. USAGE.md:1113-1135, generated from a `Ship.schema` (package Example, `type Settings`, `table ShipConfig` with `health float32 = 100.0`, `settings ?Settings`, `tier ?int32`), compiled and run verbatim with `store/1` stubbed: `measure_ship_config!` 18 == `byte_size(wire)` 18, `load_ship_config` round-trips equal, `silent?` true, plain `save` answers `{:ok, wire}`.
- **The CI leg** — `make build/conformance-harness build-conformance-elixir` then `./build/conformance-harness run --only elixir` as ci.yml runs it: 12 surfaces pass (wire 16/16 +4a, report 12/12 +3a, json-read 16, json-write 16, json-hostile 66, cook 6, cook-foreign 6, block 2, block-foreign 2, block-dump 2, forgery 11, cook-forgery 111), cook-write absent, 5.8 s.

**The ten page items**

1–2 true now, verified by D1 (USAGE.md:1187-1193, SPEC-TABLES.md:182-190). 3 stated in all three places (D5). 4 the sentinel is gone and USAGE shows both forms. 5 driver:21 and README.md:217 both "twelve times over"; the leg boots twelve BEAMs. 6 the `total_heap_size` / no-collection account is gone; one description of the instrument, and the `:erlang.trace/3` tracer at driver_impl.ex:1037 is what it describes. 7 "no cumulative counter" gone from both files; alloc-budget.txt:26 names `binary_alloc`. 8 rebased: merge-base is `origin/main`; SPEC-TABLES.md:156 and :351 carry C. 9 the body says 266 answered, 7 absent, no `ELIXIRC` bullet. 10 SPEC-TABLES.md:343-349, beside Java's.

One page nit, not blocking: USAGE.md:1169-1172 spells the keyed-array accessors `FleetTable.fleet_ships_at`, which needs `table Fleet` declared in `Fleet.schema`, and the packet emitter refuses exactly that ("declaration Fleet collides with the module the Elixir emitter writes for schema file Fleet.schema"). Declared in `Ship.schema` they are `ShipTable.fleet_ships_at` / `_put` / `_each`, and run as shown. Rename the module in the snippet, or say the accessors sit on the declaring file's `<Base>Table`.

**F2–F5: deferring is acceptable.** Ruling applied: most native feel per language; a divergence stated at its site is fine; an unstated one on the public surface is a defect.

- **F2** rows as an eager list: the generated site (RenderBlock.ex:1073-1075) says each row is a sub-binary over the block, copied by nothing; the port's stated storage rule (USAGE.md:1152-1156) is that an array is a list; and the page says the read path cannot claim zero allocation and pins the count instead. Unstated is only that the `count` handles are built per call — a cost, not a contract. A `Stream`/`Enumerable` changes the accessor's type and is its own PR. Not a defect.
- **F3** bare-map handles: `{:ok, %{name:, base:, bytes:}}` carries the three facts every other port's handle carries; `defstruct` is a shape refinement, no divergence. Not a defect.
- **F4** a nested record answers a sub-binary: `render_ship_row_position(row)` is `B.slice(row, 0, 24)` and `render_vector3_row_x/y/z(row)` read it — a nested record IS a row, USAGE.md:1186 states rows are sub-binaries, and the site comment (RenderBlock.ex:4015-4018) says the accessors read out of the row's sub-binary. Stated. Not a defect.
- **F5** `silent?` on `TableRuntime`: USAGE's own example spells `Example.TableRuntime.silent?(report)`. Stated. Not a defect.

None of the four is an unstated divergence on the public surface. LAND.

Issue #275, opened 2026-09-02
JSON text form for VARIABLE-LENGTH tables (`SceneFromJson( builder, … )`): the second walker, emitted only in pointered units


gafferongames:
The page (SPEC-TABLES.md §16.1, §16.2's pointer row) states a text form for variable-length tables through the builder. #267 ships the fixed class only: a variable-length table is refused by absence with a named comment. Reason (worker, #267): the one-walk gate wants one walker byte-identical in every header, while §2.2's zero-cost gate forbids `Builder`/`TableRef`/`TableRegion` from a pointer-free header; a builder-aware walk must name them. The way through: a SECOND walker emitted only in `anyVariable` units, plus conditional descriptor columns carrying per-target `Emplace`/`At` as function pointers — the same conditional treatment `is_pointer` already gets. About one round of work.

Owner ruling was "one table at a time" with no class restriction (#253); this issue is the land-and-expand follow-on that keeps the page's surface true. Until it lands, the page's "Backend status" note for §16 says the variable class's text form is this issue. Order: after the dogfood (space#441) unless the owner pulls it forward.

gafferongames:
Wording note from the #282 worker (not acted on): §16.1 says a pointered unit's text form is "refused by name with this section cited, never emitted with the function missing"; what #267 emits is a named COMMENT and no function. If "refused by name" should be stronger — a static_assert/#error at the call site — it is a small item to take with this issue.

gafferongames:
Owner's ruling, 2026-09-03: "where possible let's keep the json syntax between fixed and variable tables the same (if possible) or at least, completely consistent with each other."

Design constraint for this issue: the fixed text form (§16) is the base. A variable table whose graph is a tree reads and writes exactly as if every pointer were a nested table, so the same JSON serves both classes and a fixed table's JSON is unchanged. Sharing is the one fact a tree cannot express, so it gets one construct (a node written once with an id, then referenced by that id), spelled the same way at every site, and nothing else is added. The page shows both forms side by side over the same declaration so a reader sees they are one syntax.

rowan-claude:
Closed in the C++ column's contraction pass (2026-09-07), checked against main dcdd5cdd: done by #388.

PR #388, opened 2026-09-03
The text form for the VARIABLE class in the C++ reference and the tool: a tree reads and writes as nested tables, a shared node is labeled once with &node and named by that label after (#275)

gafferongames:
The C++ reference's text form for the VARIABLE class (docs/SPEC-TABLES.md §16.7), and the tool's, under the owner's ruling: the fixed form is the base, a tree-shaped graph reads and writes as nested tables, and sharing gets exactly one construct — a LABEL.

## The page

§16.7 lands with the code and states the form over one declaration, both texts side by side. A node named more than once is written once with `"&node": N` as the first key of its object and its fields after it, and every later slot holding it is `{ "&node": N }` with no other keys. The label is the text's counterpart of the wire's node index (§3.1): the text's own, numbered from 1 in first-write order, written only where a node is shared, meaningless outside the document; a tree writes none. Definition before reference in document order, and a label is defined when its object CLOSES — a bare `{ "&node": N }` whose N is not yet defined, or met inside N's own definition at any depth, is refused, never a default node (the cycle the wire refuses, refused where it is written); a later occurrence of a defined N carrying any field is refused; `&node` on the root, in a by-value nesting, non-integer or zero is refused; the slot's type must match the labeled node's. The residual is stated: a typo landing on an already-defined label of the right type is an ordinary wrong number. The `&` prefix is refused to every `json` key at compile time (§11), so a fixed reader meeting `&node` among the keys it places refuses it as a construct it cannot honor, and skips a value it does not place whole (§4's tolerance).

## What lands

- **C++**: `<T>FromJson( builder, ... )` reads into the builder's arena through each field's own `Emplace`; `<T>ToJsonMeasure( const T * )` / `<T>ToJson( const T *, ... )` write a region's const root, their identity map through the caller's `TableAllocator` (defaulted as `Measure`/`Save` default it). The walk stays ONE walk (`tables-json-walk`, byte-identical across the corpus): it meets a pointer through three adapters it declares and does not define — a pointer-free unit defines stubs, a pointered unit defines the GRAPH HALF beside the walk, gated the same way (`tables-json-graph-walk`; none of it in a pointer-free unit).
- **The tool**: `schema pack` / `schema unpack` take a variable root as one `<Root>.json`; the schema#374 refusal and its two Makefile targets are gone. A tree of fields under a variable root is refused by name.
- **The harness**: `graph_tree`, `graph_shared`, `graph_empty` carry texts generated by the engine and proved complete (packed back to their pinned wire byte for byte); the C++ leg answers every text surface. Sixteen hostile rows over the construct's edges (twelve refusals — the self-reference, a u64 overflow, a signed and a leading-zero label among them — three counted cases, one shared node) in the corpus every leg runs; the hostile-value gate packs each row under its own unit and both engines agree on report and bytes across every tree. Every other leg — Elixir included — answers the variable cases ABSENT per case on the text surfaces, so their fixed-class passes are untouched.
- **Tests**: the tables suite round-trips the three graphs through their text and back to the same wire, pins the shared graph's text against a literal (`&node` at the two definitions and the three references, nowhere else, the counts asserted: four nodes held by seven slots), refuses a cycle on a hand-made region as `Measure` refuses it, and refuses each hostile edge; the Go engine carries the same cases in `internal/tabletext`.

## What does not land, named on the page

- **graph_deep stays `no-text`, by the form's own rule.** A pointer chain nests in the text as deep as it is long, and 260 nodes is past the depth cap of 128 every walk carries — the writer refuses it in the reference and in the tool alike. The cap stays; graph_deep is the named exception, and §15 names the ruling that would move the number.
- `ToJson` from an unlocked builder; the expanded tree of a variable root; the ports (schema#349's row beside the wire) — §15.

Negative controls: a label alone the text never defined, a reference before its definition, a label defined twice, `&node` after a field, on the root, in a by-value nesting, a label that is not an integer spelled as one, a label of zero — each refused in both engines and in the corpus battery; the construct spelled identically at every site, pinned.


gafferongames:
**LAND-AFTER-FIXES** — cold read of `docs/SPEC-TABLES.md` §16.7 as the porter. The reference and the tool obey every ruling and the corpus pins all eight refusals; the fixes are on the page and in comment naming, not in behavior — with one exception (defect 1) where a rule lives only in code and the page argues the other way.

**The eight hostile cases, from §16.7 alone**

| case | page says (line) | corpus row / verdict | match |
|---|---|---|---|
| bare label never defined | malformed, never a default node (L5815) | `label-undefined` refused | yes |
| reference before its definition | malformed, definition before reference (L5820) | `label-before-definition` refused | yes |
| field on a reference (label defined twice) | malformed, a second definition (L5819) | `label-defined-twice` refused | yes |
| `&node` after a field | malformed (L5812) | `label-not-first-key` refused | yes |
| `&node` on the root | malformed (L5834) | `label-on-the-root` refused | yes |
| `&node` in a by-value nesting | malformed (L5836) | `label-in-by-value-nesting` refused | yes |
| not an integer (`1.0`) | malformed (L5808) | `label-not-an-integer` refused | yes |
| zero | malformed, positive integer only (L5808) | `label-zero` refused | yes |

Counted rows, same read: type mismatch -> `kind_mismatch` 1, pointer null (L5824) / `label-type-mismatch` 0,1,0,0 — yes; dropped definition keeps its label, reference reads null, drop counted once (L5827) / `label-dropped-definition` 0,0,1,0 and `label-in-unknown-key` 1,0,0,0 — yes; fixed reader placing `&node` -> malformed (L5794) / `reserved-ampersand-key` refused — yes; fixed reader skipping it -> unknown 1 (L5801) / `reserved-ampersand-in-skipped-value` 1,0,0,0 — yes; `shared-node-by-label` 0,0,0,0 — yes.

Pinned literal: `test/tables/main.cpp:5348` is byte-identical to `testdata/conformance/tables/json/graph_shared.json`, and its shape is the page's (definition with `&node` as first key, bare references, tree nodes unlabeled, 2-space pretty print). `&id`/`&ref`: no such spelling on the branch (the only `&ref` hits are C++ address-of in `memcpy`). §15 names the four follow-ons (L5122 ports, L5127 `ToJson` from an unlocked builder, L5131 expanded tree, L5136 depth cap); §16's opening still describes the fixed form as the base.

**Defects**

1. **A rule that lives only in code, and the page argues the other way.** Both readers register a label at its `&node` key, BEFORE the fields, so a reference inside a definition's own fields — `{ "head": { "&node": 1, "next": { "&node": 1 } } }` — reads as sharing with a silent report, and the builder then holds a cycle that only `Lock`/`Save` refuses. Stated in `internal/tabletext/read.go:751-754` and the emitter `internal/codegen/cpptable/json.go:2090-2093`; tested only in `internal/tabletext/text_test.go:858` (no C++ test, no corpus row). The page says the opposite for the root (L5834-5836: a reference to a node "whose descent is open" is "the cycle §3.1 refuses"), so a porter following the page refuses what the reference accepts. State when a label becomes defined (at the key), what a self-reference reads as, and add the corpus row — or decide the reader refuses it too and reword the root sentence; either way the root's rationale cannot stand as written.
2. **"The two spellings" survives twice**: L5789 ("the whole prefix, not the two spellings") and L5838 ("under any spelling but the two"), against L5767's "one spelling". The porter asks what the second one is. (`&other` is exercised in `main.cpp:5447` and `text_test.go:822` but has no corpus row.)
3. **Held-by-test counts are wrong.** L5932-5933 says "`&node` at the two definitions and at the four references"; the pinned literal has two definitions and THREE references (`tree.right`, `alias`, `ground.head`). L5929 "five nodes for eight references": `build_graph_shared` allocates four nodes and fills seven pointer slots — if the root is being counted as a record, say so.
4. **The cap's number is not on the page.** §16.7 L5847 and §16.2 L5547 say "this section's depth cap"; the only `128` is §15 L5137 and `kTableJsonMaxDepth`. The ruling is 128 with `graph_deep` the named exception: put 128 in §16.2 where the cap is defined, and name `graph_deep` in §16.7 (L5850 says "the 260-node chain"; the name appears nowhere in the document).
5. **History prose.** L5918-5921 (schema#374's refusal and what `unpack` used to write) and §16.1 L5420-5428 ("THE TOOL TAKES BOTH DIRECTIONS NOW ... both refused a variable root while a REFERENCE had no spelling ..."). Present state only. The dated owner quote at L5690 is the document's only one; §13 is where rulings are recorded.
6. **"id" for "label".** §15 L5132 "because ids are a text's own"; comments and test labels still say id: `cpptable/json.go:2011, 2108-2109`, `read.go:665`, `text_test.go:815-816, 846-847` ("the id after a field", "the root takes no id", "keeps its id"), `main.cpp:5441-5442, 5484-5486`; the C++ map is `ids` and `TableJsonScanLabel` writes to `id`. Naming only, but the ruling is LABEL.
7. **British spelling in added lines**: "cannot honour" in `internal/codegen/cstable/json.go`, `ctable/json.go`, `gotable/json.go`, `javatable/json.go`, `rusttable/runtime.go` and their goldens (`git diff main | grep honour`). The C++ emitter and the page say "honor". (Pre-existing on main, not this PR: "honoured" at SPEC-TABLES L5573.)
8. Minor, page silent: a label past `uint64` is malformed in both scanners; no sign and no leading zero (L5808) are pinned only in unit tests — a corpus row each would let the ports see them.

LAND after 1-7; 8 is optional.

gafferongames:
Confirming read: NOT-YET

Read at head 0f6eec36, rebased on main c960619c (a91f08a2 is an ancestor). Two residues, both small; the rest lands.

1. Self-reference refused at read — DONE. `internal/tabletext/read.go:737-769` (entry `open` until the object closes, bare reference to an open label is malformed) and `internal/codegen/cpptable/json.go:2069-2109` (same, `entry->open`). Page: `docs/SPEC-TABLES.md` §16.7 bullet "Definition before reference" states "A label is defined when its object CLOSES" with the cycle reason. Corpus: `MANIFEST.txt:415` `label-self-reference` → refused, row `{ "head": { "&node": 1, "value": 1, "next": { "&node": 1 } } }`. Go test flipped: `text_test.go:862-871` asserts `Malformed` at depth one and depth two; C++ `test/tables/main.cpp:5452-5453` the same.
2. "the two spellings" gone from §16.7 — DONE. Zero hits inside §16.7 (the seven remaining hits on the page are the pre-existing 8/16 and keyed-collection passages, untouched by the diff).
3. Held-by-test counts — PARTIAL. Page says two definitions, three references, five in all; four nodes held by seven slots. `main.cpp:5396-5402` asserts `labels == 5` and four records via the dir attribution (`5 * sizeof(TableNodeDirEntry)`, root included). "Seven slots" is stated only in the comment at `main.cpp:5393`; nothing asserts 7.
4. Depth cap 128 in §16.7 with graph_deep named as the wire-only instance — DONE. §16.7 bullet "A CHAIN NESTS AS DEEP AS IT IS LONG"; `MANIFEST.txt:82` `graph_deep ... no-text`, `FORMAT.md` rewritten to the form's own rule.
5. History prose removed — DONE. Diff drops the "Backend status for this section" block, the "THE TOOL REFUSES BOTH DIRECTIONS" narrative, the schema#349/#374/#275 lines; no dated quote or "Glenn" in §16.1/§16.7 on the branch.
6. "id" → "label" — NOT DONE. Ten sites still say id where the label is meant: `internal/tabletext/read.go:666` ("naming an id not yet defined"); `internal/tabletext/text_test.go:776, 795, 819, 820, 846, 902`; `test/tables/main.cpp:5457, 5458, 5495`. (`TestJsonKeyIsHonoured` at `text_test.go:189` is pre-existing on main, not this branch's.)
7. "honour" → "honor" in the five non-C++ emitters — DONE. All six emitter added lines read "cannot honor"; goldens `generated/bench/tables/{c,cs,go,java,rust,elixir}` match; zero "honour" in the diff's added lines.
8. Overflow / sign / leading-zero refusals pinned — DONE. `MANIFEST.txt:412-414` `label-negative`, `label-leading-zero`, `label-overflow` → refused, rows `-1`, `01`, `18446744073709551616`; C++ battery `main.cpp:5454-5455, 5462`; §16.7 Held-by-test names all three.

Also: no `&id`/`&ref` spelling on the branch (only C `&ref` address-of in generated code); American spelling on the diff (the one `Colour` is the pre-existing pointers corpus type name).

Fix 6 and add the seven-slot assertion under 3, then LAND.

gafferongames:
Residues from the confirming read closed at a672a945: the &node marker is now called a label at every prose/comment/test site (the only id left is the corpus field), and the pinned shared graph asserts its seven slots (5 labels + 2 named-once nodes), negative-controlled red at 8.
Rebased onto main past #394 (cook write): conformance driver keeps both the cook and text columns; go tests, tables suite and the C++ conformance leg (cook-write 40/40, json-read 19/19) green locally.

Issue #392, opened 2026-09-03
Table is king: the recursion closure — every slot that can hold a value can hold a table, at every depth, in every form


gafferongames:
Owner, 2026-09-03, verbatim: "I just hate this silly json/xml thing where you have to have a document, and then you put things in it. Just move away from the skeumorphics. It's a TREE of NODES." / "In our place, it's a table of tables." / "We can do anything we need with tables of tables. Table is king." / "If you can't express it as a table of tables, then make table better :)" / "Doing it as tables means everything must by design perfectly recurse."

The doctrine lands in SPEC-TABLES §1 (PR alongside the comparison page). This issue is its closure test against the tree: every slot that can hold a value can hold a table, at every depth, in every form (wire, text, cook, block). Where the tree refuses one today:

1. A union whose arms are tables: #258 (the checker refuses table arms; §2.6).
2. An array of pointers `[N]*T`: named follow-on, §11/§15.
3. An array of unions: named follow-on, §11.
4. An optional array, string or bytes `?[N]T`, `?string(N)`, `?bytes(N)`: named follow-on, §2.3/§15.
5. Maps, when they land after 3.0.0 (#380), are a table of tables by construction: the map is a variable table whose entries are tables.
6. Every type in tables, #390, so a table's leaves are never poorer than a type's.

Each is a wire move, so the closure is sequenced before 3.0.0 in the C++ reference, each landing in its simplest form with its §15 entry closed, then the ports as rows on #366 and the register (#389). The negative control for the whole item is a schema that declares every slot kind holding a table at three depths and round-trips through wire, text, cook and block in the harness.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rowan-claude:
Closed in the C++ column's contraction pass (2026-09-07), checked against main dcdd5cdd: done: every ROADMAP row the closure named is ✅ for C++.

Issue #391, opened 2026-09-03
Table reads are untrusted: the tolerant wire's read is the verifier, in every language, proven by an independent-oracle fuzzer in CI; only the cook open is trusted


gafferongames:
Owner, 2026-09-03: "Table reads are not trusted, they will come over the network. Table reads are exactly equivalent to type reads in security posture. Only loading cooked data is trusted."

So the answer to "a client-supplied table" is never "re-pack it server-side first": the tolerant read must be safe on hostile bytes exactly as the type wire's read is, in every language, and that is what SECURITY.md and §4 claim. A claim without its gate is the thing the comparison page must not print, and FlatBuffers ships a separate verifier in three of its languages for this case.

The gate, per language, in CI: a fuzzer over the tolerant wire whose oracle is independent of the emitter (the C++ reference read, or the Go engine, as the corpus battery already does for forgeries), with both negative controls red (a removed bound; a removed check), a SEED knob, and a seed-and-iterations line on the language's row of the register (#389). The block and cook forms keep their trusted posture (§7.1, §19) and say so beside it; a cook signature is the integrity answer there.

Where the tree stands: the forgery and cook-forgery batteries run in every leg; the per-language fuzzers exist for some ports (Elixir, JS, Dart after their reads) and not others (the Java read found none). The row is the scorecard's first item (#378) made concrete for the tolerant wire; sequenced before 3.0.0 because it proves a claim the release makes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #390, opened 2026-09-03
Every type the type wire carries rides in a table: fixed, ufixed, 128-bit, user types inline, user tables inline, pointers to user tables


gafferongames:
Owner, 2026-09-03, on the comparison page's open question (does the table wire ever carry fixed/ufixed/128-bit, which §11 refuses today): "Yes, we should support every type we support in types in tables. And we should support user types in tables, and user tables in tables (inline), and pointers to user tables."

What the tree has: user types inline, user tables inline (nested), pointers to user tables — all on the wire (§2, §3). What it refuses: `fixed`, `ufixed` and the 128-bit integers in a table field (§11). A deterministic sim's save game holds exactly those types, and today it stores the raw integer.

Close it in the reference first: §3's kind set gains the missing scalars with their wire widths, the text form spells them as the type wire's text does, the goldens and the hostile corpus carry them, and the negative control is a table declaring each. Then the ports, as a row on #366 and a row on the register (#389). A wire change, so before 3.0.0 while one implementation carries the variable class (space: wire moves are allowed pre-release).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #381, opened 2026-09-03
Eleven more languages: Swift, TypeScript, Lua, Clojure, Python, Ruby, Kotlin, GDScript, Zig, Odin, Haxe


gafferongames:
Owner, 2026-09-03: "OK let's add Python, TypeScript, Swift, Kotlin." The reason, from the same conversation: the likeliest remaining cause for a studio to keep Protobuf or FlatBuffers beside schema is a language schema does not ship. Python for tools and pipelines, TypeScript for the web team, Swift and Kotlin for mobile.

Each language arrives the way the nine did, in this order:

1. **The type wire**, bit-identical to the golden corpus on every push, with a pinned toolchain under dist/ (no system binaries), a conformance driver, a CI leg on the pull request gate, and a bench leg on the board.
2. **Every row of the tables matrix** on #366, mirrored from the C++ reference from the pages alone, with a blind read for bytes and feel.
3. **The "Is it awesome?" scorecard** (#378) and the red/blue read for its dialect page.

Two design questions to settle on the page before the emitters, not now: whether TypeScript is its own emitter or the JavaScript emitter with declarations, and whether Kotlin shares the JVM runtime with Java or has its own (the feel ruling says a Kotlin programmer's surface, which argues for its own). Python is a user-facing target and does not change the estate's own rule that tooling is written in Go.

Sequenced after the nine languages complete their columns, so the fan-out rows do not widen mid-row; the type-wire ports can start earlier if the bench has room, since they touch nothing the tables work touches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
Owner: start now, not after the nine. "We win over flatbuffers because each language implementation has EVERY FEATURE." Order by toolchain cost: TypeScript (node already pinned; settles the emitter question), then Python, Kotlin, Swift. One at a time beside the tables work as the bench allows.

gafferongames:
Owner, minutes later: "you can just add the new languages to an issue and do later. let's wrap up langs we have first." Deferred until the nine complete; the TypeScript worker was stopped before it created a branch.

gafferongames:
[...]

PR #393, opened 2026-09-03
Tables against FlatBuffers and Protocol Buffers: the comparison page (#379 first half), and "table is king" in §1

gafferongames:
## What it is

`docs/COMPARISON-TABLES.md`: the design and language-feature half of #379. It answers one question, "would a team already on schema reach for FlatBuffers or Protocol Buffers to get one thing schema lacks?", with:

- the same scene declared three ways and a construct-by-construct table;
- the reach-back list, ranked, each with the owner's verdict from #380: a missing language (#381), a foreign endpoint (never removable), maps (after 3.0.0, tables only, makes the table variable), unbounded bytes (#259), the untrusted-read fuzzer (#391), every type in tables (#390), sorted lookup (library-side, later), and the rest;
- the two Protobuf features judged not needed on purpose (unknown-field preservation; a file envelope), with the owner's reasons;
- the beyond list, each row with its § and the game reason;
- judged-not-needed with the reason per row;
- the 96-row feature table in six groups, every schema cell cited by §, every competitor cell cited to a page in the source list.

SPEC-TABLES §1 gains the owner's rule, "table is king": a file is a tree of nodes and every node is a table; where the data cannot be expressed as tables of tables, `table` gets better; so everything must recurse perfectly. The recursion gaps against that rule are #392.

README links the page beside COMPARISON.md.

## Not in this PR

The measured half of #379 (the same data as a schema table, a FlatBuffer and a Protobuf message on the box) waits for a quiet box and the C++ column.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ bin/schema generate --lang c|cpp|cs|dart|elixir|go|java|js|rust --out <outdir> <
 | **[SPEC-TABLES.md](docs/SPEC-TABLES.md)** | The normative reference for tables: the wire, the cook, the block form, reflection, the build version. |
 
 Beside them: [PERFORMANCE.md](docs/PERFORMANCE.md),
+[COMPARISON-TABLES.md](docs/COMPARISON-TABLES.md) (tables against FlatBuffers and
+Protobuf, feature by feature, with the verdict on every gap),
 [COMPARISON.md](docs/COMPARISON.md) (the same packet against Cap'n Proto, Protobuf
 and FlatBuffers), [FAQ.md](docs/FAQ.md), [VERSIONING.md](docs/VERSIONING.md),
 [CONTRIBUTING.md](docs/CONTRIBUTING.md) and [SECURITY.md](docs/SECURITY.md).
--- a/docs/COMPARISON-TABLES.md
+++ b/docs/COMPARISON-TABLES.md
@@ -1,0 +1,357 @@
+# Tables against FlatBuffers and Protocol Buffers
+
+Tables are what people use FlatBuffers and Protocol Buffers for: data that
+outlives the build that wrote it. This page puts schema's tables beside both,
+feature by feature, and answers one question:
+
+> Would a team already using schema ever reach for FlatBuffers or Protocol
+> Buffers to get one thing schema lacks?
+
+The answer today is a short list, ranked, with a verdict on each. The page is
+honest in both directions: where either format is ahead it says so, and where
+schema is ahead it says how. Every claim about schema cites a section of
+[SPEC-TABLES.md](SPEC-TABLES.md) (`§n`), [SPEC.md](SPEC.md) (`SPEC §n`) or
+[USAGE.md](USAGE.md); every claim about the other two cites the page it was
+read from, listed at the end.
+
+**This page compares design and language features.** Measurements of the same
+data as a schema table, a FlatBuffer and a Protobuf message, read, written and
+opened on one machine, are the second half of the comparison and are not here
+yet. [COMPARISON.md](COMPARISON.md) measures the type wire, not tables.
+
+## One scene, three ways
+
+The same data declared in each language: a scene of entities, each with a
+position struct, a bounded name, an optional health, a union of two component
+types, an array keyed by an equipment slot, and a thumbnail blob.
+
+```
+package scene
+
+const MaxEntities = 1024
+const MaxName     = 32
+const MaxThumb    = 4096
+
+enum Slot { Head, Body, Legs }           // implicit None = 0; rides by NAME on the wire
+
+type Vec3 { x float32  y float32  z float32 }   // a plain struct, nested by value
+
+type Light
+{
+    intensity float32 = 1.0
+    color     Vec3
+}
+
+type Mesh
+{
+    asset_id uint32
+    lod      uint8 | min = 0, max = 7     // the bound is part of the type; clamped on load
+}
+
+union Component
+{
+    light Light
+    mesh  Mesh
+}
+
+table Item
+{
+    id    uint32
+    count uint16 = 1
+}
+
+table Entity
+{
+    name      string(MaxName)            // bounded; no allocation
+    position  Vec3
+    health    ?int32                     // present or absent, by value
+    component Component                  // None is the absence
+    equipment [Slot]Item                 // one slot per named variant, by name
+    thumbnail bytes(MaxThumb)            // inline at its bound
+}
+
+table Scene
+{
+    entities [..MaxEntities]Entity       // the root table IS the file format
+}
+```
+
+`Scene` is fixed down to its leaves, so it is one plain struct with `Measure`,
+`Save` and `Load`, a cook and a block form, and nothing on any path allocates
+(§2.2, §6.1). `equipment` cannot be indexed by `None` or past the enum in any
+build (§2.4). Every field is optional on the wire with a declared default, and
+there is no presence machinery to spell (§4).
+
+In FlatBuffers the same scene is a `struct Vec3`, tables for the rest, a
+`union Component` with its implicit `_type` field, and `equipment` as a vector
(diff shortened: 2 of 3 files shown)


gafferongames:
## Cold read: LAND-AFTER-FIXES

**The reader's answer.** Yes, the page answers the question fast: the reach-back table at line 115 is what a FlatBuffers or Protobuf programmer wants first, and ranking it with a verdict per row is the right shape. The competitor prose is fair where it is specific (append-only fields, `oneof` moves, enum map keys, the verifier), and the "judged not needed" reasons are real reasons, not dismissals. Two places read as marketing rather than comparison: "Protobuf cannot even point" (line 159) and "This is the case FlatBuffers lost on" (line 166), plus insider phrasing a stranger cannot parse ("by the owner's word", line 190; "Rebuild the tool.", line 144, which is where a Protobuf user will push back hardest and deserves one more sentence). Every "beyond" row cites a section a reader can open, which is the page's strongest property. What would make me distrust it is exactly what the fact check found: a language count that is wrong by two and can be checked in ten seconds (line 121, 321), a sentence in quotation marks that the cited page does not contain (line 250), and one schema claim that the cited section flatly contradicts (line 313).

**Fifteen cells, checked against their citations.**

| # | group | cell | cited | result |
|---|---|---|---|---|
| 1 | Declarations | Enums: rides as the hash of its name | §3 | confirmed (kind 7, u16 name hash) |
| 2 | Wire | Pointer encoding: reserved id, u32 index, 0 null, 1 root, one record per node | §3.1 | confirmed |
| 3 | Wire | Size ceilings: node body 4 GiB refused at save; 8-byte references; 64-bit cook parts | §2.1, §6.3, §7.1 | confirmed (u32 node length, save-time refusal; eight-byte slot; u64 header words) |
| 4 | Evolution | Change a default: baseline refuses until `--update --reason` | §18 | confirmed |
| 5 | Runtime | Allocation on write: thread-local arena, slab not per node; block takes a caller allocator | §6.4, §6.5 | confirmed, but the block allocator is §19.1, not §6.4/§6.5 |
| 6 | Validation | `LoadMeasure` lets the caller refuse a region size | §6.5 | confirmed |
| 7 | Reflection/text | Text form: "a shared node named by `&node`" | §16 | **wrong.** §16: "The text form is a TREE, and a pointer graph is not ... a node several parents name is written once per parent and reads back as that many nodes." There is no `&node` spelling anywhere in §16. |
| 8 | Declarations | Field identity `fold16(fnv1a32(name))` | §5, §3.1 | confirmed |
| 9 | Declarations | FB scalar types byte to ulong, float, double, bool | FB-schema | confirmed |
| 10 | Validation / Tooling | FB verifier "in C++, C and Swift", "three of its fifteen languages", "fifteen listed" | FB-support | **wrong.** The matrix has thirteen columns (C++, Java, C#, Go, Python, JS, TS, C, PHP, Dart, Lobster, Rust, Swift). Verifier: C++ and Swift "Yes", C "Basic". |
| 11 | Validation / Reflection | FB `max_depth` 64, `max_tables` 1M; mini-reflection 2 to 6 bytes per field | FB-cpp | confirmed |
| 12 | Declarations | PB field numbers 1 to 536,870,911; enum map keys refused; map order undefined | PB-proto3 | confirmed |
| 13 | Validation | PB "recursion limit 100; UTF-8 verify" | PB-limits | partly. 100 is C++ and Java; Go is 10000. UTF-8 validation is not on PB-limits; it is `features.utf8_validation` on PB-features. |
| 14 | Wire | PB "2 GiB; 'not designed for messages larger than a few megabytes'" | PB-overview | **misquoted.** The page says "For data that exceeds a few megabytes, consider a different solution." The 2 GiB figure is on PB-limits, not PB-overview. |
| 15 | Evolution | FB add at end or `id`, never remove, "don't" change defaults, same-width type change, free rename, union append or explicit discriminant | FB-evolution | confirmed |

Extras checked in passing: PB-editions "moving a field into a oneof is unsafe", confirmed; PB-encoding six wire types, last wins, nested merge, fixed types little-endian, confirmed; FB-evolution/FB-schema `required`, `bit_flags`, four-char `file_identifier`, vectors of unions C++ only, `[T:N]` structs only, all confirmed.

**Sections and issues.** Every `§n` cited exists on main in SPEC-TABLES.md and SPEC.md (SPEC §4.1 does hold the doc-comment design; SPEC §9 q9 does decline explicit and sparse enum values). All eight issues exist, are open, and say what the page says, with one exception below (#380).

**Defects, by line.**

- L313: `&node` does not exist; §16 says the opposite. Replace with: a shared node is written once per parent and identity does not survive text (§16).
- L121, L321: "fifteen" is thirteen. L301: C's verifier is "Basic", not a verifier.
- L250: drop the quotation marks or quote the actual sentence; move the 2 GiB citation to PB-limits.
- L301: "UTF-8 verify (PB-limits)" cites the wrong page (PB-features). L219: PB-features carries no 2 GiB string limit either.
- L168: "FlatBuffers' `vector64` is C++ only (FB-64)". Neither cited issue says so: #7537 is the 2022 proposal, #8555 is a Rust compile error closed as not planned. Uncheckable against its citation; find the support-matrix row or soften.
- L271: "the object API drops them (FB-cpp)" is not on FB-cpp. Uncheckable.
- L140: "Unknown fields preserved on write" is presented as judged not needed on purpose, but #380 (cited on this page) lists it as gap 2, "Design on the page first", and sequences it "maps and unknown-field preservation first, spec first". The page and the issue contradict each other; one of them has to move, and the page should cite the ruling if there is one.
- L179: the answer to `Any` is "a union whose arms are tables (§2.6)", but §2.6 says "no backend, and no declaration reaches it" and the checker refuses it (#258). L214 says so; L179 should too, or a reader who tries it distrusts the page.
- L285: block allocator citation should be §19.1.
- L159, L166: tone; state the fact and drop the scoring.
- L326: "read ... on 2026-09-03" is the one dated line in the prose; acceptable as a citation date, but it is the only history on the page.

§1 paragraph: consistent with the existing "no envelope, no directory, no document concept, no prescribed root" bullet and with §17.3; no contradiction found. American spelling throughout, both files. README link fine.

**The two things that most need to change.**

1. Fix the checkable wrongs (L313 `&node`, L121/L321 fifteen, L250 misquote) before this page is public; each is a ten-second check that costs the whole page its credibility.
2. Reconcile "unknown fields preserved on write" between L140 and #380: either the page cites a ruling that closes gap 2, or the row moves out of "judged not needed".

gafferongames:
Fixed on the branch, one line each:

- `&node` cell now cites §16.7 and #388, which lands it; main's §16 is tree-only until then.
- FlatBuffers: thirteen languages; the verifier is C++ and Swift with a basic one in C.
- Protobuf size: 2 GiB cited to PB-limits; the overview's guidance quoted as printed.
- UTF-8 verify → PB-features; block allocator → §19.1; `vector64` "not in every port"; the object-API-drops-unknowns claim removed.
- `Any` row names #258 and the closure #392.
- The two tone lines rewritten as facts.
- #380's body edited: item 2 marked ruled not needed on 2026-09-03 with the owner's reason, item 1 marked ruled after 3.0.0, so the body and the page agree.

gafferongames:
Confirming read: LAND-AFTER-FIXES, three lines left (checked at e50c9250):

- L168 still says `vector64` is C++ only; L250 was fixed to "not in every port", L168 was not.
- L301 still says the verifier is "in C++, C and Swift"; L121 was fixed, L301 was not (C is basic).
- L219 "unbounded to 2 GiB (PB-features)": PB-features carries no size limit; cite PB-limits or drop the figure.

Everything else lands: &node cited to §16.7 with #388 (a section not yet on main; the "landing with" note covers it), thirteen languages, PB-overview quoted as printed, UTF-8 to PB-features, block allocator to §19.1, object-API claim removed, `Any` row names #258 and #392, L159 and L166 are facts now, #380 items 1 and 2 ruled. One nit there: #380's last line still says "maps and unknown-field preservation first".

Issue #380, opened 2026-09-03
Beyond Protobuf and FlatBuffers: maps, unknown-field preservation, reserved, unbounded bytes, sorted keyed lookup


gafferongames:
Owner, 2026-09-03: "when we look at our feature set, are we truly covering everything that protobufs/flatbuffers can do, and then going beyond?" The gap list, with the reason each matters for games:

1. **Maps, key to value.** RULED 2026-09-03: yes, after 3.0.0; tables only, never types; a map makes a fixed table variable. Config is full of them; today it is an enum-keyed array or a table of pairs. Design on the page first: it touches the wire.
2. ~~**Unknown fields preserved on write.**~~ RULED NOT NEEDED 2026-09-03 (owner: everything has a schema; a writer writes exactly the structured data it declared; an old tool rewriting a newer file drops what it does not know and the report counts it). What follows is the original text. A tool that reads old data and writes it back must not drop what it does not know; today the tolerant reader skips and the writer cannot re-emit. Design on the page first: it touches the wire and the report.
3. **Reserved ids and names.** Cheap; the baseline is the natural home.
4. **Unbounded bytes and strings.** #259, the byte-buffer primitive, with the relative-file include in the pack tree.
5. **Sorted keyed vectors with binary search.** Asset lookup by id inside a cook.

Judged not needed, with the reason: in-place mutation of a buffer (a cook is immutable; tools regenerate), a per-object verifier for untrusted buffers (cooks are trusted; the signed cook covers the rest), dynamic typing / Any (unions by name), streaming delimited messages (the transport's job), extensions and custom options (a reflection walker can carry attributes later).

Sequenced after the parity fan-out on #366. Order as ruled 2026-09-03: wire-touching items before 3.0.0 (#390, #259, the #391 gate); maps, sorted lookup, retired names and doc comments after; unknown-field preservation not at all.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
The bar, in the owner's words, 2026-09-03: "the goal is simply to exceed flatbuffers, protobuf with everything they can do, better faster, more flexible -- and add all the cool game specific stuff that only a pro game developer would know about." The check on this issue proves both halves: nothing they can do that schema cannot, and a list of what schema does that they cannot, each with a game reason.

gafferongames:
The "why would they still include it" list, to be answered or consciously declined, each with its reason:

[...]

PR #397, opened 2026-09-03
SPEC-TABLES: maps, the design (#380)

gafferongames:
Design only, no implementation; sequenced after 3.0.0 (#380).

- §2.8 "Maps: `ships map[string(32)]ShipConfig`": table bodies only, the holder variable by that fact; keys are `string(N)` and integers (enum keys refused by name, `[E]T` is their construct); values are anything a field holds, a map included; the entry type `<Table><Field>Entry` with constant ids `key = 0xA079`, `value = 0x9194`; the writer sorts, the region load verifies with one compare (duplicate last-wins and counted, descending malformed), the builder load forgives.
- Memory layout: THE layout is a sorted entry array in wire, cook and region, entries by value inside the holder node's extent; never closed addressing; open addressing with linear probing only as an optional runtime index built at load, never stored (a stored one makes fnv1a64 and a load factor a cook contract nine ports reproduce byte for byte).
- Allocation: bulk segments through the builder's `TableAllocator` pair, insertion order kept, index grows by copy into a larger block and old blocks stay until reset; erase is builder-only and marks dead; a region and a cook are immutable.
- Text form a plain JSON object (integer keys quoted, ascending, `&node` in the value); no block form, no variable case; cost model, negative controls, a C++ builder and reader sketch.
- Closing: the CAN/SHOULD answer, the "Beyond both" table row by row, and the explicit confirmation that nothing in the 3.0.0 wire forecloses it.

**The kind answer.** CAN: yes, exactly — a map is kind `14` over kind `13` bodies under two constant field ids; no new byte of framing, no new skip rule. SHOULD: yes — it costs eight bytes an entry against a dedicated kind on the wire only (the region and the cook spend zero) and a four-line per-map rule for a key-kind change; it buys a construct every 3.0.0 reader already skips or decodes (a new kind is not skippable, and the wire is settled at 3.0.0) and the `[..N]Pair` migration path. A kind is spent to close a silent edit, and there is none here.

### DIFF
--- a/docs/SPEC-TABLES.md
+++ b/docs/SPEC-TABLES.md
@@ -413,7 +413,8 @@ That rule has a consequence the rest of this page is held to: **everything
 must recurse perfectly.** Every slot that can hold a value can hold a table,
 at every depth, in every form (wire, text, cook, block): a table by value, a
 pointer to a table, an array of tables, an array of pointers, a union whose
-arms are tables, a keyed array of tables, an optional table. A slot that
+arms are tables, a keyed array of tables, an optional table, a map of tables
+(§2.8). A slot that
 refuses a table today is a gap, named in §15, not a design.
 
 - **Users build their own formats.** A table declaration plus nesting is a
@@ -452,7 +453,7 @@ table ShipConfig
 A table body is a type body — the field grammar of SPEC §4.2, hosted by
 `table`: bare and ranged integers, `bits(N)`, `bool`, floats and
 compressed floats, enums, flags, strings, bytes, bounded arrays, unions,
-`if` branches, and declared types as field groups. Six additions:
+`if` branches, and declared types as field groups. Seven additions:
 
 - **Tables nest.** A named table is a field type (above); nesting is by
   value, and a bounded array of tables is a collection. A table may not
@@ -470,6 +471,10 @@ compressed floats, enums, flags, strings, bytes, bounded arrays, unions,
   `None`.
 - **A union arm may be a table** (§2.6), which is what makes an evolvable
   message set expressible.
+- **A table may hold a MAP** (§2.8). `ships map[string(32)]ShipConfig` is a
+  lookup by key over entries the wire carries as a sorted array of one
+  generated `{ key, value }` table; it spends no wire kind, and it makes its
+  holder variable-length.
 - **`was` — the rename attribute** (§5).
 
 And one addition that is not a field spelling at all: **every fixed table has
@@ -543,8 +548,8 @@ at 2 GiB, and the scale a cook exists for is larger than that.
 The compiler works out which class a table belongs to; the schema never
 says. The rule is a least-fixed-point over BY-VALUE edges:
 
-- A table is **VARIABLE-LENGTH** if it declares a pointer, or if anything it
-  nests by value is variable-length. "Nests by value" reaches through every
+- A table is **VARIABLE-LENGTH** if it declares a pointer or a map (§2.8),
+  or if anything it nests by value is variable-length. "Nests by value" reaches through every
   by-value edge there is: a plain nested table, an element of a bounded
   array, an element of an enum-keyed array, a member of a guarded (`if`)
   group, an optional's value (§2.3), and a UNION ARM that is a table (§2.6).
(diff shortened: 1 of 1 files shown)


gafferongames:
**Cold read, as the programmer who implements this in C++ from the page and then ports it to Go. Verdict: LAND-AFTER-FIXES.**

The construct is consistent with the wire (§3: kind `14` over `13`, the closed kind set untouched, a fixed-class reader skips by `L` under the second skip rule), with §4's events (last-wins `duplicate`, descending `malformed` keeping the ascending prefix, key-kind change as one `kind_mismatch` reading empty), with §5 (I recomputed both ids with `ir.FieldId`: `key` = `0xA079`, `value` = `0x9194`), with §6.2 (`Lock` is a compaction, so a contiguous sorted array out of segmented storage is free), with §6.3 (the arena slot is an offset, so `*Insert(...) = *slot` for a shared node is legal), with §13.9 (structured bindings are already the keyed-array idiom in §2.4, `schema_allocate`/`schema_release` and `TableAllocator` resolve), with §16.7 on main (`&node` as the value object's first key), and §19 (no variable case). Every owner ruling is on the page: tables only, variable holder, CAN and SHOULD answered separately with costs each way, the sorted entry array as THE layout, open addressing only as a caller-owned index never stored, closed addressing refused with reasons, bulk segments through the pair, growth by copy, deletes on the builder only, region and cook immutable, one sort per walk, duplicate insert replaces, the Beyond-both table with a § per row and the reading tier's honest non-claim, "nothing in 3.0.0 forecloses it" stated as a confirmation, sequenced after 3.0.0. Spelling is American throughout the diff. The branch is two commits behind main and `merge-tree` reports zero conflicts, so the §16.7 cites resolve on merge.

What has to change before it lands — each with the sentence that should replace it (lines are `docs/SPEC-TABLES.md` on `spec-maps`):

1. **`docs/SPEC-TABLES.md:1345` — the map field's wire cost is 12 bytes, not 7.** `id (2) + kind (1) + L (4) + element kind (1) + N (4)`. Replace the row with: `| the wire, the map field | \`12\` bytes once | id, kind, \`L\`, element kind, \`N\` — exactly an array of tables |`.

2. **`docs/SPEC-TABLES.md:1355` and `:1472` — "eight bytes an entry" is neither number the text describes.** "One `L` and one field header" is 7; what a dedicated kind actually removes per entry is the entry's `L`, both field headers and the terminator, 12. Replace both with: "a dedicated map KIND would save twelve bytes an entry — the entry's `L`, both field headers and the terminator, with the key and value kinds ridden once in the map's header —".

3. **`docs/SPEC-TABLES.md:1144` — "decodes OVER the previous one" leaves an elided field of the repeat holding the first entry's value.** That is not last-wins and it is a byte a golden will catch. Replace with: "**Equal**: a DUPLICATE. The slot is RESET to the entry's defaults and the repeat decodes into it — last wins WHOLE, so an elided field of the repeat reads as its default and never as the first entry's value — and `duplicate` is counted (§4)."

4. **`docs/SPEC-TABLES.md:1147` — "and `Lock` removes it": a loaded region has no `Lock`.** Replace with: "the slot it would have taken is the one slack a loaded region can carry; a region packed by `Lock` never has it, because the builder load already merged the repeat."

5. **`docs/SPEC-TABLES.md:2741` (§6.5) contradicts `:1266`.** §6.5 says `LoadMeasure` is ONE SCAN over record type ids and lengths; a map's `N` sits inside a body, at any nesting depth, so for a map-holding unit the measure must walk field headers. Add to §6.5 after the ONE SCAN sentence: "For a unit that declares a map (§2.8) the measure also walks each record's field headers — skipping every payload by its framing and reading no value — to reach each map's `N` at every depth, and refuses as `malformed` an `N` whose entries cannot fit in the map's `L` (six bytes each at least); a fixed unit and a map-free pointered unit keep the one scan."

6. **`docs/SPEC-TABLES.md:3494` and `:3515` (§7.4) contradict the §7.2 touch at `:3390`.** "Pass two reads exactly three things" and "Nothing else is read. Not a scalar, not a string's bytes" — but the map's reference is not a `*T` slot resolving to a directory entry, and the order check reads KEY bytes. Add a fourth item after 3: "4. **Every MAP SLOT** (§2.8). Its delta must land inside the HOLDER's own extent, at `alignof( Entry )`, at the offset the layout rule computes, and `count × sizeof( Entry )` must fit before the extent's end; the entries' own slots, companions and tags are then walked as a bounded array's elements are; and the KEYS are read, ascending with no repeat, because a cook `Find` cannot search is a forgery." And amend `:3515` to: "**Nothing else is read.** Not a scalar, not a string's bytes beyond a map key's, not an enum's ordinal…".

7. **`docs/SPEC-TABLES.md:1265` — map-of-maps layout order is underdetermined.** "Depth-first field order" over a record whose entry count is dynamic does not say whether an inner map's array interleaves with the outer array. Add after "depth-first field order.": "The placement is PRE-ORDER: a map's whole entry array first, then, entry by entry in key order, the arrays of any map an entry's value holds by value."

8. **`docs/SPEC-TABLES.md:3339` (§7.2) — "zero slack between them" now has a second reading.** Replace with: "with zero slack between them — a node's storage being its record and, for a holder with maps, the entry arrays that follow it (§2.8)."

9. **`docs/SPEC-TABLES.md:1021` and `:1506` — "held by the same refuser (§2.2)": §2.2 names no refuser.** Name the gate that holds it: "held by the header scan §13.9 holds the hook rule with" (or add the refuser where §2.2 states the zero-cost claim).

10. **`docs/SPEC-TABLES.md:6715` (§19 touch) — "which is what a cook's `Find` does over the same declaration's cooked form" is false:** a block-form declaration is fixed, holds no map, and its cook has no `Find`. Replace with: "and a consumer that wants one by string or number sorts its rows and searches them itself — the same search a map's `Find` runs, over a declaration that chose the variable class to get it (§2.8)."

11. **`docs/SPEC-TABLES.md:1256`–`:1260` and `:1303` — two counts, one truth unstated.** The record slot carries a `uint32` count and the builder head carries a live count. Add to the slot bullet: "In the arena the slot's count is the LIVE count — the head holds the dead count beside it — and `Lock` writes the same number into the region."

12. **§11 (`docs/SPEC-TABLES.md:4471`) is the law-home for refusals by name and gains no map row.** Add the map's refusals there: a map in a `type` body; enum, bool, float, flags, bits, bytes, table, pointer, optional and union keys; `?map`, a default, `| max`, `[..N]map`; `map[K]Self` by value; the claimed `<Table><Field>Entry` name — each cross-cited to §2.8.

13. **Negative controls (`:1363`–`:1389`) — missing ones the implementation needs:** (a) `Save` emits a DEAD entry: golden reds and `measure == save` still holds; (b) the same map from two insertion orders produces one image: byte-compare reds when the sort is dropped; (c) `LoadBuilder` of an unsorted body then `Lock` is sorted: the tool-path gate reds if the load rejects it; (d) `N = 0xFFFFFFFF` under a six-byte `L`: `LoadMeasure` refuses rather than answering `N × sizeof( Entry )`; (e) insert of a key longer than `N` is refused, and the wire's clamp-then-duplicate case is in the corpus with the counter; (f) the by-value cycle `map[K]Self` and the claimed entry name: checker suite reds; (g) the build version moves when the entry's record lines change and not before.

Present-state note, not a blocker: §2.8 carries five dated owner quotes and a "two alternatives were weighed" passage inline. The page's homes for those are §13 (rulings, dated there by convention) and §14 (shapes rejected); moving them leaves §2.8 stating the rule and citing the ruling.

Questions only the owner can answer:
- The optional runtime index names `fnv1a64` once and "the hash is over the key's bytes" once; since it is never stored, is the hash a per-port choice or a contract so the bench numbers compare across ports?
- Is the builder-side index built at the first insert for every map, or only past a size where a linear scan of the segments loses? The page implies always; a two-entry map then pays an index block.

gafferongames:
All thirteen fixed on `spec-maps`, rebased onto `dc202791` (§16.7 cites now resolve on main), one commit on top of the design:

1. Map field wire cost is `12` bytes, row replaced as written.
2. Dedicated-kind saving is twelve an entry — the entry's `L`, both field headers and the terminator — at both sites (cost table and the SHOULD paragraph).
3. Region duplicate: the slot is RESET to defaults and the repeat decodes into it, last wins WHOLE, elided fields read as defaults.
4. The slack sentence names the load as the actor; a region packed by `Lock` never has it because the builder load merged the repeat.
5. §6.5: a map-holding unit's `LoadMeasure` walks field headers to reach every `N`, refuses as `malformed` an `N` its `L` cannot hold at six bytes an entry; fixed and map-free pointered units keep the one scan.
6. §7.4: pass two reads FOUR things, item 4 is the map slot (extent, alignment, offset, `count × sizeof( Entry )`, entries walked as array elements, keys read ascending with no repeat); "nothing else is read" amended for a map key's bytes.
7. Placement is PRE-ORDER: a map's whole array first, then entry by entry in key order the arrays its values hold; said in §2.8 and in the §7.2 touch.
8. §7.2 "zero slack between them" now reads a node's storage as its record plus, for a holder with maps, the entry arrays that follow it.
9. The refuser is named at both sites: the zero-cost gate's header scan, the build failure §2.2 states, with the map symbols added to its list.
10. §19 touch: a fixed consumer sorts and searches its rows itself — the same search a map's `Find` runs over a declaration that chose the variable class; no `Find` on a fixed cook claimed.
11. The slot's count is the LIVE count, the head holds the dead count beside it, `Lock` writes the same number into the region.
12. §11 gains the map row: `type` body, every refused key kind, key attributes, `?map`, defaults, `| max`, `[..N]map`/`[N]map`, `map[K]Self` by value, the claimed entry name — cross-cited to §2.8.
13. Seven negative controls added: a dead entry emitted, two insertion orders one image, `LoadBuilder` of an unsorted body then `Lock`, `N = 0xFFFFFFFF` under a six-byte `L`, over-long key refused on insert plus the clamp-then-duplicate corpus case, the by-value cycle and the claimed name in the checker suite, the build version moving with the entry's record lines and not before.

The two questions, answered under the standing rulings and stated on the page as rules: the runtime index's hash and load factor are each port's own and never a cross-port contract — the index is never stored, the wire and the cook carry no index, and no golden, `cook-check` rule or build-version line names either; and the builder builds NO index — an insert appends after one linear scan, `Find` on an unlocked builder is that scan (`O( n )`, `O( n² )` over a build of `n`, stated), and the sort happens once at `Lock`, `Save` or `Cook`. The owner quotes stay inline for now; moving them to §13/§14 is the present-state pass you named, not a blocker.

gafferongames:
**Confirming read: LAND.** Read d0025d1f cold. All thirteen are on the page as stated (wire costs 12/12; duplicate resets the slot; region slack sentence; §6.5 body walk with the N-vs-L refusal; §7.4 now reads four things and the key-bytes exception; pre-order placement; §7.2 storage sentence; the §2.2 header-scan gate resolves; §19 sentence; live/dead count; §11 map rows; the seven controls). The two rulings are stated and no stale index survives in §2.8 — head holds segment chain and live/dead counts only, erase is scan-plus-bit, `Find`/`Erase`/`Each` take the arena, the cost table and sketch say linear scan, the worker sentence names head and segments only, and the runtime index names no hash and no load factor as a contract. Merge against main is still conflict-free.

PR #400, opened 2026-09-03
tables: arrays of pointers — [N]*T and [..N]*T in the C++ reference and the tool (#392)

gafferongames:
Second of the four constructs in #392, in its simplest form. Stacks on #398 (base `tables-closure`); retarget to `main` once that lands.

## What lands

- **The page first** (§2.1, §3.1, §11, §15, §16.2, USAGE). An array of pointers is a pointer slot per element: `[..N]*T` and `[N]*T`, storage `TableRef name[N]` with the count beside it for the counted spelling, framed as kind `14` with element kind `17` and one `u32` node index per slot — the framing §3.1 already stated. The three things §15 said "want to be settled together" are each the rule the construct inherits, and §3.1 now states them: a **null slot is index `0`** in its place; **elision is the by-value array's** (§3) — an empty counted array and an all-null fixed array elide, a live null slot rides, one non-null slot makes a fixed array ride whole; **element kind `17` is held apart from `8` and `13`** by the element-kind rule, so `[N]uint32` ⇄ `[N]*T` ⇄ `[N]T` are reported edits. The numbering visits slots in index order, live slots only. The keyed `[E]*T` stays a named follow-on (§15, with the reason: a keyed body elides slots by name, so its empty end wants stating first).
- **The checker** lifts the refusal for the bounded spellings and refuses `[E]*T` by name.
- **The IR/layout**: an element of a pointer array is an eight-byte slot (`FieldPieces`, `FieldFacts`).
- **The C++ reference**: storage, reset, measure, save (the slot writer), load (through the generic array reader, so the element-kind check is the one every array has), descriptors (the slot thunks per element), the numbering / `PackMeasure` / `Pack` over every slot, the cook writer (every slot, all N), and the JSON walk (an array of the pointer row, `null` and `&node` per element).
- **The tool**: wire encode/decode (`resolveCell`), the numbering's `visitEdges`, the text writer's count and reader (`readPointerCell`, a null element), the cook's region writer, `cook-check`'s scan and `uncook` — each per slot.
- **The corpus**: `tables/stream`'s `Feed` gains `parts [..4]*Chunk` and `pair [2]*Chunk`, whose slots share nodes with each other and with the chain. `stream_parts` rides wire, text (labels inside arrays) and cook in the harness; `stream_chain`'s bytes are unchanged, which pins the all-null / empty elision. Two hostile rows: an element of the wrong shape, and a node shared by `&node` across array slots.

## The gate

`test/tables/main.cpp`: `stream_parts` pinned, reloaded, locked, loaded into a region (sharing verified by address: `parts[0]` IS the chain's first node, `pair[1]` IS `parts[2]`), text out and back to the same bytes, cook → `Open` → same bytes, descriptors (kind 17, `is_array`, the target's descriptor); the elision rules at the empty end (a Feed without the arrays is byte-identical to the pre-array golden; two live null slots ride); a hand-built wire with element kind `8` under `parts` → one kind mismatch, the field empty. Checker tests: the spellings resolve to kinds 14/17, the holder is variable, the slot is eight bytes, `[E]*T` is refused.

Verified locally: `schema_test_tables`, `tables-zero-cost`, `check`, `go test ./...`, goldens re-pinned (`tables-stream` build version, source golden), and `make conformance` across all eight legs (every surface passes; ports answer ABSENT per case, `+16a`/`+15a`). No path under `bench/LOCK` is touched.

## For the owner

§15 had listed three wire questions for this construct. Each is answered here by the rule the construct inherits rather than by a new one, and the answers are stated in §3.1 — if any of the three should go the other way (in particular: whether a counted array whose live slots are all null should elide), it is one line in the writer and one in the page, and the wire is pre-3.0.0.


gafferongames:
Closed by GitHub when the `tables-closure` base was deleted on #398's merge. Continued, rebased onto main, in the PR that supersedes this one (same branch).

Issue #389, opened 2026-09-03
Techniques travel: a technique proven in one language is carried to every language, never rediscovered — docs/PORTING.md as the register, with a gate


gafferongames:
Owner, 2026-09-03: "Make sure we don't do silly things again like, put some cool technique in language X, and then forget it for all other languages and we have to rediscover it."

The evidence from tonight's three port reads: the independent allocation instrument (GC-count over a steady phase against an idle loop of the same shape, with a planted allocation that turns it red) was discovered by the Dart reader, then again by the Elixir reader (trace on garbage_collection), then again by the JS reader (--trace-gc at a fixed semi-space). Three discoveries of one technique, each costing a read cycle, because nothing carried it from the first port to the next brief.

## The law

A technique proven in one language's port belongs to every language. When it lands anywhere, it is written into the register the same day, and every other port either carries it or states at the register why the platform cannot.

## The mechanism, three parts

1. **The register: `docs/PORTING.md`.** Rows are techniques, columns are the nine languages, cells are ✅ / a stated reason / ❌. Each row names the technique in one paragraph, where it lives in the reference, and its negative control. First rows, from the tree and the reads: the independent allocation gate; emitter sabotage via `go build -overlay` as the shape of every negative control; per-case ABSENT (#383); the block lead/base-alignment check (#369); the fuzz control that removes BOTH bounds; the forgery oracle independent of the emitter; claimed-name refusal in every spelling the language maps to; the cross-endian refusal as a named gate; the bench row labeled a pairing check unless on a quiet box; the identity map (#373) and flat node table (#376) for the variable class; the hooks and allocator contract (#386).
2. **The briefs and the reads.** Every port brief and every blind-read brief carries the register; a port PR is not ready until its column on the register is filled; a reader checks the column, not their memory.
3. **The gate.** A Go test in `compiler/` reads the register's table and requires, for every language in the driver registry, that every technique with a Makefile-checkable form has its target (`tables-<lang>-alloc-negative-control`, `tables-<lang>-fuzz-negative-control`, `tables-<lang>-block-lead`, …) and that the target runs in `make test`. A ✅ with no target is red; a ❌ with no stated reason is red.

Sequenced now, before the sweep right on #366, so the sweep's briefs carry the register from the first row.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
[...]

Issue #366, opened 2026-09-03
3.0.0 gate: complete feature parity across all nine languages, each at its language's performance floor


gafferongames:
**Current priority, Glenn live on 2026-09-07: complete the C# table column FIRST, then Go.** This applies to the whole column, not just the fixed class or the id-table step. Existing tracks: C# #513, Go #511, variable class #349, maps #502, plus each column's PORTING.md carry issues. Stella and Rowan coordinate the C++ review and port ownership; see the September 7 priority comment for scope and sequence. The matrix snapshots that stood in this issue and its comments were historical and are now pointers; the matrix is ROADMAP.md.

Owner rulings, 2026-09-03: "All languages need to have complete feature parity." and "All languages need to have the most performant implementation possible for each feature per-language."

The release gate is therefore every feature in every language, and a profiled sitting per feature per language, not the fixed class alone.

The completion matrix lives in [ROADMAP.md](https://github.com/mas-bandwidth/schema/blob/main/ROADMAP.md) and nowhere else, at Glenn's ask on 2026-09-08: "The only place should be ROADMAP.md." The table that stood here was a 2026-09-03 snapshot and is removed so there is one home.

Campaign, in order, each step landing in its simplest form per the land-and-expand law:
1. The nine fixed-class ports land (#352, #353, #356, #364, Elixir), each with a blind read.
2. The variable class on the wire in every language (#349), with the harness carrying variable instances (#357) and the reference's shared-node defect fixed first (#360).
3. The variable class's text form, C++ first (#275), then every language.
4. The block BUILD side in every language: the same builder contract as §19.1, so any language fills a block any other reads.
5. Per-language performance rounds per feature: a board sitting for the tables wire, the text form, the block read and the cook open in each language, profiled to that language's floor (#358 Rust read, #362 Go write are the first two).

Open question for the owner: whether a cook WRITE side belongs in the generated runtime (an editor in C# cooking without the tool), or the tool remains the only cooker in every language, which is parity as it stands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
Third ruling, same day: "All languages must be binary compatible on type and table wire between languages."

[...]

PR #401, opened 2026-09-03
Every scalar the type wire carries rides in a table: fixed, ufixed and the 128-bit integers gain table-wire kinds in the C++ reference and the tool (#390)

### LINKED ISSUE #390: Every type the type wire carries rides in a table: fixed, ufixed, 128-bit, user types inline, user tables inline, pointers to user tables

gafferongames:
Closes #390. Part of the recursion closure (#392, item 6).

Every scalar the type wire carries now rides in a table, in the C++ reference and the tool. `docs/SPEC-TABLES.md` §3 gains the kind rows, §16.2 the text spellings, §4/§4.1, §7.2, §8.1/§8.2, §11, §15, §18.1, §19.3 and §20.2 the consequences; the checker drops the §11 refusal and refuses nothing else new.

## The kind rows, as landed (§3)

| kind | payload |
|---|---|
| `18` i128, `19` u128 | 16 bytes: the low 64-bit half, then the high half — two's complement for `18`; little-endian throughout, the type wire's own order for the family |
| `20`–`24` fixed8/16/32/64/128 | 1/2/4/8/16 bytes, the RAW scaled integer of a `fixed(I, F)` whose storage is that width, two's complement |
| `25`–`29` ufixed8/16/32/64/128 | 1/2/4/8/16 bytes, the raw scaled integer of a `ufixed(I, F)` of that width, unsigned |

**The representation, and why.** A fixed value rides as its storage — units × 2^F at the storage width — exactly as the type wire carries it before its range compression; the `(I, F)` and the whole-unit bounds stay in the schema, as a ranged integer's bounds do. A 128-bit integer is its sixteen bytes, low half first. The kinds are distinct from the integers' for the reason kind 17 is distinct from 8: a `fixed(16, 16)` and an `int32` are the same four bytes and not the same number, so a move between them is a reported edit rather than a value read at the wrong scale. One kind per width and signedness because a skipper knows a scalar's width from the kind byte alone (skip rule: kinds 1–11 and 17–29 skip their width). `F` is declaration-side like a resolution; a moved `F` is the fourth member of §4.1's silent class and the baseline refuses it (`frac=` token).

- **Tolerance (§4):** a fixed field clamps on the raw scale, `[A << F, B << F]`, and counts; a 128-bit integer clamps to its declared bounds in 128 bits; a bare `uint128` clamps at nothing.
- **Text (§16.2):** a fixed is a number in whole units — `1.0`, `-0.25`, `3.0000152587890625` — written as the shortest exact decimal, read from any token exactly representable in Q I.F, a finer fraction `kind_mismatch` and never rounded; a 128-bit integer is a decimal integer, exact past 2^53. Nothing on the path is a double, in either engine.
- **Cook and block (§7.2, §19.3):** a 128-bit scalar is sixteen bytes at sixteen; a table's storage struct spells `alignas( 16 )` where the storage is serialize's emulated pair, so the compiler's one layout model holds on every compiler, asserted in the generated code. The same one-line spelling in the PACKET header (a closure `type` declaring a 128-bit field) is behind the C/C++ lock (#170, `bench/LOCK`) and is a named follow-on in §15; the cook-layout asserts turn that case into a build error on an eight-aligned compiler, and the corpus declares no such `type` until the lock lifts. The packet emitter and its goldens are byte-identical to main.
- **Descriptors (§8.1):** `frac_bits` and an exact raw range (`wide`) join the field descriptor; kind 0 no longer describes anything a declaration spells.

## Gate numbers

- Conformance matrix, C++ and Go legs: `wire 23/23`, `report 17/17`, `json-read 22/22`, `json-write 22/22`, `json-hostile 104/104`, `cook-write 46/46`; Go answers the new cases ABSENT per case (`+7a`, `+5a`, `+6a`, `+6a`, `+36a`), every other cell unmoved.
- Hostile-value gate: 104 trees agree on report and wire across both engines (84 + 20 new, hand-authored verdicts).
- New corpus unit `tables/scalars` (every fixed width and signedness, both 128-bit integers, as scalars, elements, keyed slots, nested and optional) with three instances (`scalars_full`, `scalars_default`, `scalars_edges`) pinned by C++ and read back by the tool through the wire, the text and both cook orders; evolved twin `test/tables/Scalars2.schema` gives two `report` cases (`2,2,2,0,false` and `1,2,2,0,false`).
- New gates in `make test`: `tables-ports-refuse-wide-scalars` (eight ports refuse the unit by name, naming every wide field and the follow-on) and `tables-scalars-block-asserts` (the block form's `static_assert`s compile).
- Checker negative control: `TestEveryScalarRidesInATable` — a table declaring each scalar compiles, each on its kind, the fixed default held raw.
- C++ tables test, JSON one-walk gate (22 files), zero-cost gate, JSON negative control, `go test ./...`: green.

## What the ports need (a row on #366, the register #389)

The kind table above; the width-per-kind skip rule; the raw-scale clamp; the exact decimal conversion both ways with the saturation-then-range rule of §16.2; the `frac_bits`/`wide` descriptor columns; sixteen-at-sixteen in the cook and block layouts; then the port's driver gains codecs for `scalars`/`SimState` and `tblscalars2`/`SimState` and drops itself from the refusal gate's list. Until then each port refuses a unit declaring a wide kind by name (`ir.RefuseWideTableKinds`), and the harness answers ABSENT per case.

## Notes

- Full local conformance matrix (nine legs) green with every port answering the new cases ABSENT per case; ASan tables test, cook and block legs green.
- `make tables-block-zero-cost` is red on main already (`arena.go`'s "TableBlockAllocator already has" comment trips the `[A-Za-z0-9_]Block` grep since #386) — not touched here.


gafferongames:
Cold-read fixes landed:

1. §18 opening: "exactly four — a changed specified default, a moved flags variant, a fixed field's F moved under one kind, and a referent that cannot stand in for the one it replaces." and "It refuses more than those four (§18.2)".
2. §7.2 layout row and `ir/blocklayout.go`: `alignas( 16 )` is spelled on every 128-bit member, so serialize's emulated pair — not naturally sixteen-aligned — lays out exactly as native `__int128` does.
3. §3: "a reported edit — `kind_mismatch`, §4 —".

PR #426, opened 2026-09-03
tables: arrays of pointers — [N]*T and [..N]*T in the C++ reference and the tool (#392)

gafferongames:
Supersedes #400, which GitHub closed when its base branch (`tables-closure`, #398) was deleted on merge; same branch, rebased onto main over #398 and #399.

Second of the four constructs in #392, in its simplest form. Follows #398 (a union whose arms are tables), on main.

## What lands

- **The page first** (§2.1, §3.1, §11, §15, §16.2, USAGE). An array of pointers is a pointer slot per element: `[..N]*T` and `[N]*T`, storage `TableRef name[N]` with the count beside it for the counted spelling, framed as kind `14` with element kind `17` and one `u32` node index per slot — the framing §3.1 already stated. The three things §15 said "want to be settled together" are each the rule the construct inherits, and §3.1 now states them: a **null slot is index `0`** in its place; **elision is the by-value array's** (§3) — an empty counted array and an all-null fixed array elide, a live null slot rides, one non-null slot makes a fixed array ride whole; **element kind `17` is held apart from `8` and `13`** by the element-kind rule, so `[N]uint32` ⇄ `[N]*T` ⇄ `[N]T` are reported edits. The numbering visits slots in index order, live slots only. The keyed `[E]*T` stays a named follow-on (§15, with the reason: a keyed body elides slots by name, so its empty end wants stating first).
- **The checker** lifts the refusal for the bounded spellings and refuses `[E]*T` by name.
- **The IR/layout**: an element of a pointer array is an eight-byte slot (`FieldPieces`, `FieldFacts`).
- **The C++ reference**: storage, reset, measure, save (the slot writer), load (through the generic array reader, so the element-kind check is the one every array has), descriptors (the slot thunks per element), the numbering / `PackMeasure` / `Pack` over every slot, the cook writer (every slot, all N), and the JSON walk (an array of the pointer row, `null` and `&node` per element).
- **The tool**: wire encode/decode (`resolveCell`), the numbering's `visitEdges`, the text writer's count and reader (`readPointerCell`, a null element), the cook's region writer, `cook-check`'s scan and `uncook` — each per slot.
- **The corpus**: `tables/stream`'s `Feed` gains `parts [..4]*Chunk` and `pair [2]*Chunk`, whose slots share nodes with each other and with the chain. `stream_parts` rides wire, text (labels inside arrays) and cook in the harness; `stream_chain`'s bytes are unchanged, which pins the all-null / empty elision. Two hostile rows: an element of the wrong shape, and a node shared by `&node` across array slots.

## The gate

`test/tables/main.cpp`: `stream_parts` pinned, reloaded, locked, loaded into a region (sharing verified by address: `parts[0]` IS the chain's first node, `pair[1]` IS `parts[2]`), text out and back to the same bytes, cook → `Open` → same bytes, descriptors (kind 17, `is_array`, the target's descriptor); the elision rules at the empty end (a Feed without the arrays is byte-identical to the pre-array golden; two live null slots ride); a hand-built wire with element kind `8` under `parts` → one kind mismatch, the field empty. Checker tests: the spellings resolve to kinds 14/17, the holder is variable, the slot is eight bytes, `[E]*T` is refused.

Verified locally: `schema_test_tables`, `tables-zero-cost`, `check`, `go test ./...`, goldens re-pinned (`tables-stream` build version, source golden), and `make conformance` across all eight legs (every surface passes; ports answer ABSENT per case, `+16a`/`+15a`). No path under `bench/LOCK` is touched.

## For the owner

§15 had listed three wire questions for this construct. Each is answered here by the rule the construct inherits rather than by a new one, and the answers are stated in §3.1 — if any of the three should go the other way (in particular: whether a counted array whose live slots are all null should elide), it is one line in the writer and one in the page, and the wire is pre-3.0.0.

## After the cold read (LAND-AFTER-FIXES → fixed)

- **The page**: §16.7 now says an element of an array of pointers is a pointer slot and takes the pointer row (an element of a BY-VALUE array is the value that takes no label); §7.2's pieces table has the `[N]*T` and `[..N]*T` rows (N eight-byte deltas at eight, then the `int32` count for the counted spelling, a slot past the count zero); §3.1 states the scope of a bad index in a slot — `malformed`, THAT slot null, the array kept — and that a count past the bound keeps the prefix and counts `clamped` (§4).
- **Three depths in the corpus**: `Chunk` gains `links [..2]*Chunk`, so `stream_parts` carries the construct in the root (`parts`, `pair`), inside the union arm (`frame.chunk.links` names `c` a third time) and inside a node (`a.links` names `a`'s own `next`, then a live null slot). `stream_chain`'s wire is byte-unchanged (the empty arrays elide); its text and cook are regenerated.
- **The hostile wires as shared rows**: the three shapes the C++ test built by hand — element kind `8` under `parts`, five slots against a bound of four, a slot naming index 7 with no node table — are committed as `testdata/wire/tables/parts_*.bin` and read as `report` rows (`parts_elem_kind` `0,1,0,0,false`, `parts_count_past_bound` `0,0,1,0,false`, `parts_index_out_of_range` `0,0,0,0,true`), the engine's counters in `reports.txt` and the C++ leg agreeing. The C++ rows stay pinned beside them. #391's wire-forgery surface is where a mutated-bytes battery goes; these three are deterministic hand-built bodies, which is what the `report` surface already carries.
- Rebased over #401 and #395 (`tableKindWidth` now delegates to `ir.TableKindWidth`, which carries kind 17).


gafferongames:
**Cold read, from the page alone, as the Go porter. Verdict: LAND-AFTER-FIXES.** Three sentences on the page; the wire, the framing and the corpus need no byte moved.

## Contradictions

1. **`docs/SPEC-TABLES.md:6866` (§16.7) still says an array element is never a node.** "Anywhere else, the prefix is `malformed`: … in a by-value nesting, a union arm, an array element, which are values and not nodes." `stream_parts.json` puts `&node` inside `parts[]` and `pair[]`, and §16.2's pointer row now says a slot may define or name a node. Replace with: *"in a by-value nesting, a union arm, an element of a BY-VALUE array, which are values and not nodes — an element of an array of pointers (§2.1) is a pointer slot and takes the pointer row, definition or reference alike; under any spelling but `&node`."*

2. **`docs/SPEC-TABLES.md:3595` (§7.2) has no row for the construct.** The pieces table gives `[N]T` and `*T` separately; a porter composes them, but the cook is byte-pinned in both orders and the page should say the bytes. Add two rows after `*T`: *"`[N]*T` — `N` `int64` self-relative deltas, each eight bytes at eight, null zero"* and *"`[..N]*T` — the same `N` slots, then `int32` used count; a slot past the live count is zero (a counted array writes all `N` slots, below)."* §7.4 item 1 ("every `*T` slot") then reads as covering them without a second sentence.

3. **`docs/SPEC-TABLES.md:2151` (§3.1, reading) — the scope of a bad index in a slot is unstated.** "An index above `node_count + 1` … malformed, and the pointer stays null" is written for a field. For an array the porter has to choose between nulling the slot and emptying the array. Add one sentence to the array bullet: *"A slot's index is read by the same rules as a field's: an index out of range is `malformed` and THAT SLOT reads null, the rest of the array unaffected; a count past the reader's bound keeps the bounded prefix and counts `clamped` (§4)."*

## What the page does state (checked, no change needed)

- §3.1: null slot is index `0` in place; elision is the by-value array's (empty counted elides, all-null fixed elides, a live null rides, one non-null fixed rides whole); element kind `17` held apart from `8` and `13` by the element-kind rule — all three as rules, not only in the PR body.
- Numbering: "array elements in index order" in the walk bullet, "live slots of a counted array only", "a slot is an edge like any pointer field"; §3.1's identity paragraph then covers slot/field sharing without a special case. `stream_parts` pins it (`parts[0]` is the chain's node, `pair[1]` is `parts[2]`).
- Tolerance: kind `14` skipped by `L` and counted `unknown` for a reader that cannot name the field (§3); element kind ≠ `17` is `kind_mismatch`, field empty (§3.1); count past bound is §4's `clamped` — inherited, and fix 3 above makes the inheritance explicit.
- §16.2 pointer row extended; the label rules of §16.7 (definition first in walk order, reference bare, dropped definition keeps its label) hold for slots as `stream_parts.json` and `pointer-array-shared-label` show.
- §11 lifts the bounded refusal and refuses `[E]*T` by name; §15 names the keyed follow-on with its reason (kind `16` elides by name, so the empty end wants stating first).
- Added prose is present-state and American. (`colouring` at §3.1:2202 and `materialised` in the manifest header predate this PR — out of scope, a sweep item.)

## The owner's question

The page decided a counted array whose live slots are all null **rides** — `N` zero indices under kind `14`/`17` — because the count is content (§3 "a counted array with a live slot rides whatever the slots hold"). The alternative (elide) would move **no pinned bytes**: `stream_parts` has non-null live slots and `stream_chain` has no array; only `test/tables/main.cpp:6260` (two live nulls ride) would flip. But it would break the text form's round trip: `"parts": [null, null]` reads to `parts_count = 2`, and an eliding wire would return it as `[]`. The page's choice is the one under which text and wire agree; I would keep it.

## Porter's list (Go, from the page)

- Wire: `[N]*T` / `[..N]*T` decode through the generic array reader with element kind `17` and four-byte elements; no new skip rule; the slot index goes through the same deferred-resolution path as a field's (§3.1 "resolution cannot be inline").
- Numbering: descend array slots in index order, live count only for the counted spelling, through the one first-visit map.
- Elision: empty counted / all-null fixed elide; measure and save must agree on it.
- Text: array of the pointer row; `null`, definition, reference per element; labels run in walk order.
- Cook: `N` eight-byte deltas at eight, count companion after for the counted spelling; cook-check walks every slot and bounds the companion.
- Corpus gaps a port can't get from the shared data: the construct appears at ONE depth (Feed's root record); a pointer array inside a node record and inside a union arm are unexercised by any instance. The element-kind-swap wire lives only in `test/tables/main.cpp:6275`; there is no shared row for count past bound or an out-of-range index in a slot (the manifest has no wire-hostile surface, so this is a harness follow-on, not this PR's). Ports answer ABSENT per case as today.

gafferongames:
Fixed per the cold read, on this branch: the three page sentences (§16.7, §7.2 pieces rows, §3.1 slot scope + clamped inheritance), the corpus at three depths (`Chunk.links [..2]*Chunk` — in the root, inside the union arm, inside a node; `stream_chain` bytes unchanged), and the three hand-built hostile wires as shared `report` rows the engine and the C++ leg agree on. Rebased over #401/#395. Details in the PR body's last section.

PR #427, opened 2026-09-03
docs/PORTING.md: the techniques register, with its gate (#389)

gafferongames:
## What it is

`docs/PORTING.md`, the register #389 names: every method of implementation and every instrument the nine table backends carry, one section each — the method in a paragraph, where it lives in the reference, the language it was proven in, its measured effect, its negative control — and a nine-column row of `✅ (what proves it)` / `— the platform reason` / `❌ #NNN (the carry-across issue)`. Then "How a technique enters" (same-PR entry; a ❌ cites its carry-across; every blind read reports two lists) and the grammar the gate reads.

The rows: the fourteen methods from the owner's comment (M1 allocation-free read through M14 the `&node` label), the fourteen instruments from the issue body (I1 the independent allocation gate through I14 the unpinned-runtime refusal), and five the JavaScript and Dart confirming reads found the list lacked (accessor/descriptor agreement, the runtime home, the release tier, analyzer-clean emitted text, the bench leg's pre-clock golden gate). The Dart AOT/JIT split, the both-ways name scan and the second fuzz seed fold into I1, I6 and J3; "a measured row with a null ceiling cites why" is in the grammar.

## The count

34 techniques × 9 languages = 306 cells: **178 carried, 103 not yet, 25 the platform cannot**. Every ❌ cites an issue — 23 filed from this PR (#403–#425, titled `<technique>: carry to <languages>`, each naming the proving language and the reference site), the rest existing ones (#349 the variable class, #362 the Go codec shape, #377 C's keyed upper bound, #387 the block lead).

## Unverified cells

- Every cpp/c/rust/go/cs/java cell and every js/dart/elixir cell of M1–I14 comes from the three per-language verification reports (every cell, file:line). The js/dart/elixir cells were re-read against main after the three ports merged where the confirming reads said something moved (the JS keyed guard and fuzz control, the Dart allocation gate, usage, standalone, json-walk, names control and board, the Elixir block lead); source line numbers elsewhere are the reports' and may drift by a few lines.
- The five J rows' ❌ cells outside their proving languages are **inferred from the reports not finding such a gate**, not from a per-language search; each ❌ cites its issue, whose body says "verify or carry".
- Three findings from the reports the register records and the tree does not yet fix: `tables-rust-fuzz` and `tables-rust-soak` exist and nothing runs them (I3, I9 → #413, #416); Dart's fuzz control removes one bound (I3 → #413).

## What the gate holds

`compiler/porting_test.go`, under `go test ./compiler` (the last line of `make test`):

[...]

PR #430, opened 2026-09-03
tables: an array of unions — [N]Body and [..N]Body in the C++ reference and the tool (#392)

gafferongames:
Third of the four constructs in #392: an array of unions, `[..N]Body` and `[N]Body`, in the C++ reference and the tool. Rebased onto main (#426, #427, #399, #401 all under it).

## The page

§2 no longer excludes arrays of unions; §2.6 states the construct — a union slot per element, kind `14` with element kind `15`, storage `Body name[N]` with the used count beside it for the counted spelling. The three rules are each the one the construct inherits, stated in §3: a `None` element is the arm id `0` in its place, because position is identity in an array; elision is the by-value array's — an empty counted array and a fixed array of `None` elide, and a live `None` element rides; and element kind `15` is held apart from `13` by the element-kind rule, so `[N]Body` ⇄ `[N]Table` is a reported edit. `[E]Body` stays a named follow-on (§15), refused by name (§11), for the reason `[E]*T` is: a keyed body elides slots by name, and a `None` slot there wants its rule stated before it is wire. §16.2's union row: an array of unions is an array of that row, `{}` for a `None` element.

## The reference and the tool

- **The checker** lifts the table-closure refusal and refuses `[E]Body` by name; `ir.TableUnionArrays(u)` names every array of unions in a closure, and every port's `Generate` refuses through `refuseUnionArrays` (`--lang cpp` named as the carrier).
- **The C++ emitter**: measure, save, load (`None` re-established before the arm, unknown arms counted and skipped by length), descriptors (`arms`/`enum_name`/`variant_id` on an array field, at `elem_size` strides), `TableJsonElementShape` `'o'`, and `emitVariableUnionWalk` per element for the numbering, `PackMeasure` and `Pack`.
- **The tool**: `encodeElement` writes the union framing per element (`None` as `u16 0`); the wire reader's union framing became a cell-level `unionCell` shared by field and element; `visitEdges` walks every element's set arm (live elements of a counted array only); `ElementShape` answers `'o'`, which gives the text reader's array walk the union row's rules per element. The cook and the text writer already dispatched per cell.

## The corpus, at three depths

`tables/messages`: `ToolMessage.history [..2]ToolBody` (a batch of messages in the root, with a live `None` slot), `Transaction.pending [2]EditBody` (inside a table arm, its `None` element riding in place), `InsertText.origins [..2]Origin` (inside a table arm inside a table arm) — all three in the one new instance `message_batch`, crossing wire, text and cook in the harness: C++ green over all 36 instances / 26 reports / 113 json-hostile rows, every port ABSENT per case. **Every existing message wire is byte-unchanged, which pins the elision at the empty end**; the C++ gate also counts two live `None` elements riding as `3+4+5+2*2` bytes.

Hostile coverage, following #426's pattern: two hand-built wires as shared `report` rows (`batch_elem_kind`: element kind `13` under `history` → one kind mismatch, the field empty; `batch_unknown_arm`: an arm id no build declares → one unknown, the slot a live `None`), pinned by the C++ leg and read by every leg against the engine's own counters; three `json-hostile` rows (`union-array-element-shape` `0,1,0,0,false`, `union-array-element-two-keys` `refused`, `union-array-element-unknown` `1,0,0,0,false`).

Gates: `TestUnionArraysAreLegal` (kinds 14/15, a fixed-arm holder stays fixed, a VARIABLE arm carries the mode through the array, `[E]Body` refused), `TestUnionArraysAreCppOnly` in its own file per the registry split (every non-cpp target refuses, naming the fields and `--lang cpp`), and five C++ test functions (`test_union_arrays*`). `test/conformance/README.md`'s absence counts read the matrix as it stands: `+20a` wire / `+19a` text (the four variable-class, the thirteen of the message class, the three of the wide-scalar class — the last were stale on main since #401).

**Negative controls**: a byte flipped in the `message_batch` golden reddens the C++ gate (3 failures) and exactly the `cpp / wire: message_batch` cell of the matrix, restored and green; `refuseUnionArrays` removed from the go target reddens `TestUnionArraysAreCppOnly`, restored and green.

Local: full `go test ./...` green, `make conformance` all surfaces green, `schema_test_tables` + ASan twin green, `tables-zero-cost` green.

## For the owner

No new wire decision: each rule is the one the construct inherits, and §2.6/§3 state them. The one to look at is the fixed spelling's elision — a `[N]Body` of all `None` elides (the all-default fixed array's rule, as `[N]*T` all-null does), where a `[N]T` of tables always rides; if the fixed array of unions should ride always, it is one branch in each writer.


gafferongames:
Owner ruling, 2026-09-04 (main channel): **elision confirmed** — a fixed `[N]Body` of all-None elides, consistent with `[N]*T` all-null. The "For the owner" question in this PR's body is resolved; the implemented behavior (the elision pin held by `test_union_arrays_elision`) is the ruled behavior. No change required.

Issue #440, opened 2026-09-04
Two latent §3.1 walk clauses the C++ walk does not honor: a pointer under a false guard is still numbered; an absent optional's variable table is still descended


gafferongames:
Found by probe during #438 (the one-walk fix for #433); latent because no corpus schema has either shape. §3.1 commands one declaration-order walk over the REACHABLE edges; the C++ walk (and the tool? — verify) still numbers a pointer that sits under a false guard, and still descends a variable table held by an absent optional.

Ruling (Rowan, on the page's own text — the page wins): a pointer under a false guard is not reachable and is not numbered; an absent optional's table is not descended. Both engines (the Go tool and the C++ reference) must agree, and each case gets its own corpus schema and wire pin written by the tool so the plain conformance run holds the rule — the #438 pattern: red on the old walk first, then green. Port row on #366 (only C carries a walker today, and it mirrors the old grouped shape — #433's port list). Before the sweep.

rowan-claude:
Closed in the C++ column's contraction pass (2026-09-07), checked against main dcdd5cdd: superseded: clause 1 landed in #507; clause 2 is #526.

PR #507, opened 2026-09-04
tables: the id-table wire (#435)

gafferongames:
The table wire's **id-table form** (docs/SPEC-TABLES.md §3, merged as #501), in both implementations, with the corpus, the gates and the fuzzer moved to it.

## The form

A saved table is three parts: the **form byte**, the **root body**, and the **id table** a reader finds from the END. A body is `id reference, kind, payload` terminated by the zero reference; every length, count, index and reference is one **canonical LEB128**; identity is `fnv1a64(name)` at sixty-four bits for a field, an enum variant, a union arm and a table's own name alike. The table holds every id the body used, once each, in **first-use order**, and the body names them by position — so a file naming forty ids across ten thousand fields carries forty ids and spends one byte a header.

Three kinds ride with it: `30` an enum, carrying the reference to its variant name's id; `31` the escape; `32` the payload-free arm. **An arm header is a field header** and carries the arm's kind. The node table rides in **one** field under the reserved id `0xFFFFFFFFFFFFFFFF`.

## What holds it

- **§3.1's worked save reproduces byte for byte** — all 120 — as a golden of its own, and every id the page prints reproduces from the hash alone.
- **The cross-implementation lock closes**: all 46 text-carrying conformance instances agree byte for byte between the C++ reference and the compiler's own engine, held by a Go gate.
- **§20.2's worked projection reproduces character for character**, and both of the page's digests reproduce from it: `0xa9df771f6051391f` and `0xc4adaacfe6cb2809`.
- **The wire fuzzer**: 93,651 enumerated mutants over 89 seeds, 0 divergences, plain and under ASan/UBSan, with the reference-class, id-table, canonical-spelling and form-byte passes §4.2 names. All four negative controls go red on their own check.
- **It found three reader bugs and two missing rules**, each a real divergence from the page — an unsigned fit check, a `getleb` that moved the cursor on a number it refused, a damaged count that dropped an optional's presence, the reserved id in a nested body, and a repeated node-table field.
- **Twenty new report rows** for the kinds and refusals this form added, each built from §3's grammar and pinned by the reference.
- **The reserved-id and collision refusals** at sixty-four bits, with a compiler test hook: eight rows red under a planted collision and green without one.
- **The elision, first-use and distinctness instrument** over every pinned instance.

`make conformance` is green — the reference leg passes every registered surface — and so are `go test ./...`, the tables, maps, block, cook, pack, hostile and fuzz batteries.

## Also in

- **The tool's map half**: the compiler's engine sorts, writes and reads a map at three depths and agrees with the reference byte for byte; the by-name refusal narrows to the cook surfaces, which is what the tool still owes.
- **#440**: a pointer under a false guard is not an edge, in all three authoring walks, with its corpus row and control.
- **The page**: §2.8's four owed sentences, §3's dispatch sentence, §20.2's `flags` block, and the measured paragraph re-pinned from this tree (0.98x without the 210 KB blob, 1.80x over the tiny class, 0.99x over the whole corpus).

## What the ports lose, and what is still open

The corpus is pinned in the new form and the eight ports write the previous one, so their five **wire-carrying** surfaces are ABSENT — `wire`, `report`, `json-read`, `json-write`, `json-hostile` — each driver saying so and naming its issue (#511–#518). The seven per-port harness negative controls go dormant with them.

**Open:** each port's own allocation gate, soak and forgery fuzzer reads the pinned corpus too, so `make test`'s port legs still fail. Keeping them dormant loses those gates until the ports move; giving the ports a corpus in the previous form keeps them running. That is a decision, not an oversight.

**Also open:** the tool's cook half for maps; the dispatch nanosecond figures, which need a bench sitting rather than a laptop under load.

---

## The tree brought current (25f9f06)

Three checks were red on `9a036cc`, and none of them was a defect in the new wire code.

**`generated`.** The bench table units for nine languages still carried the previous build version. The Go one was committed and the other nine were not. The C++ bench table unit also had to take the whole id-table wire, which is what this branch emits. Regenerating every `generated/*/.stamp` from this branch's compiler moves 11 files and nothing else.

**The byte-identity half of the block zero-cost gate**, which the `generated` job runs after the staleness check and so never reached. 22 C++ Table sources under `testdata/golden/tables/` move with the emitter this branch changes. 12 C Table sources move on **one line each**, the build version constant, and the new values are the ones `testdata/golden/build-version/` already pins on this branch. Nothing under `testdata/wire/` moved: the wire pins were already re-cut here. 82 Table sources are byte-identical to their pins again.

**`lint`.** Eight golangci-lint findings and eight modernize findings. Three dead functions go, seven untagged switches over a kind take their tag, one if-else chain becomes a switch, four `errors.As` refusal probes become `errors.AsType`, and two backward index loops become `slices.Backward`. The emitters' output is byte-unchanged across all of it.

**`big-endian` was not a byte-order verdict.** The job died on

```
qemu-s390x ./build/schema_test_c_soak_be 0
soak: testdata/wire/tables/root_full.bin does not re-save to its own bytes — refusing to soak a codec that does not reproduce the corpus
make[1]: *** [make/c.mk:317: tables-c-big-endian] Error 1
```

The same binary built for the HOST fails on the same case with the same message. `./build/schema_test_c_soak 0` on a little-endian laptop:

```
soak: testdata/wire/tables/root_full.bin does not re-save to its own bytes — refusing to soak a codec that does not reproduce the corpus
```

So this is the port-leg case the section above already names, and not an endianness defect: the soak's golden gate reads `testdata/wire/tables/`, which is the id-table form now, and the C port writes the previous one. `tables-c-big-endian` goes **dormant** with `schema#512` behind it, on the rule the eight per-port controls on this branch already follow. The cross-compile rule stays, because that is what the leg wakes with. The C++ half of the big-endian job is untouched and green: the tables battery, maps, the block driver both ways and the cook reader both ways all still run on s390x under qemu.

All 20 checks pass on `25f9f06`.


gafferongames:
Rulings on the three open points:

1. The flags block renders after the enum blocks and before the unions: enum, flags, union, in that order, each block sorted as the enum block is. State it in §20.2 in this PR.
2. The dispatch mechanism is the implementation's, not the contract's. §3's sentence about resolving each entry to its slot at open and dispatching through an array index becomes: a reader resolves each reference through the id table once, and how it dispatches from the resolved id is its own; the measured figure stays as measured on the reference. Amend the page in this PR.
3. The ports. The eight ports keep writing and reading the previous form until each is ported; two forms never coexist on one wire, so their table-wire surfaces become ABSENT in the conformance registry when the corpus is re-pinned: each port driver's list drops wire and report, the per-leg table gates are disabled by name, and one issue per port is filed and cited from PORTING.md and the matrix. The fixed-class table-wire row of the matrix reads reference-only until the sweep re-ports each language, which the sweep was going to do for the eleven C++-only rows anyway. This is the visible cost of the form change and it is stated, not hidden.

Issue #525, opened 2026-09-04
Retain-unknown: an opt-in that keeps what this build cannot name (ruling 12 on #523)


gafferongames:
Ruling 12 on #523: **retain-unknown as an opt-in on `Load`, re-emitted by `Save`, so a rolled-back client's save keeps newer fields; the promise page says so.**

The specification lands first (docs/SPEC-TABLES.md §6.6, docs/VERSIONING.md), then the C++ reference, then the sweep carries it.

**No wire change.** Every field on the table wire is `reference, kind, payload` and every kind is skippable by its kind byte alone, so the bytes a reader skips are a self-framed unit it copies out and copies back. No byte of §3 moves, no kind is spent, no form byte changes.

**What the spec settles**

- The surface is `<Name>LoadRetain`, `<Name>MeasureRetain`, `<Name>SaveRetain` over a caller-declared, caller-owned `TableRetain` buffer. `Load`, `Measure` and `Save` are unchanged.
- Retained: a **field** whose id this reader cannot name, at any depth, in the root body and in every node record. Not retained: an unknown enum variant, an unknown union arm, an unknown keyed-array slot, or a node record whose type id is unnameable — each is not a field, so re-emitting one is a splice into a body the reader rewrites.
- **Retention form**: the wire form with every reference replaced by the eight-byte id it names, so a record is self-contained and survives a file whose id table is ordered differently.
- Re-emitted **at the end of the body they came from**, in the order retained. Position carries nothing on this wire, and the round trip is idempotent after the first save.
- Two counters on `TableReport`: `retained` and `retain_lost`. **`retain_lost == 0` is the whole of the safety check**, and it is what the never-clobber rule becomes for a caller that opted in.
- Refusal is per record and never partial: a record the remaining capacity cannot hold whole is dropped, `retain_lost` counts one, and the read continues.

**What the implementation owes**

- `LoadRetain`, `MeasureRetain` and `SaveRetain` join §11's claimed suffix set and `tableGeneratedVerbs` in the change that lands them, and not before.
- The conformance rows §6.6 names, including the id-table reorder case a verbatim copy fails and the twice-saved byte-identity case.
- The wire fuzzer gains a retention leg; its existing round-trip requirement runs with retention off and does not move.

Spec PR: the branch `spec-reachability-and-retain`.

Issue #524, opened 2026-09-04
A reachability-scoped wire-shape projection (ruling 11 on #523)


gafferongames:
Ruling 11 on #523: **a reachability-scoped protocol id**. Declarations reachable from no `type` do not enter the wire-shape projection, so a content enum is not a lockstep redeploy. A `ProjectionVersion` bump; every id moves once.

The specification lands first (docs/SPEC.md §3.1, §3.2 and docs/VERSIONING.md), then the compiler.

**What the spec settles**

- The projection is the **closure over the unit's `type` declarations**. Every `type` is a root, because nothing in the language says which types go on a wire, so an unused helper type still moves the id.
- The walk descends through a field's named `type`, `enum` or `union`, an array's element type, both sides of a branch, and a union arm's payload and its own field facts.
- **`flags` is the exception and projects unconditionally.** A mask is the table wire's one positional vocabulary and a reorder is the silent class's second member, so the protocol id is the only frame that refuses two peers holding different bit assignments while they exchange table data over one connection. Content never rides in a flags declaration.
- `ProjectionVersion` `2` → `3`.

**What the compiler owes**

- `ir.WireProjection` renders the closure rather than the unit; `ProjectionVersion` bumped to `3`.
- The **negative control**: remove one edge from the reachability walk and `internal/check/projection_test.go`'s "an edit that moves bytes must move the id" leg must go red. A missed edge is the spurious MATCH the projection exists to refuse, so this is the gate the change rests on.
- Every generated golden carrying a `ProtocolId` moves once.

Spec PR: the branch `spec-reachability-and-retain`.

Issue #523, opened 2026-09-04
Design lock-in before the sweep: the cold read's shortest list


gafferongames:
A cold read of the design against the owner's question tonight ("Is the design for schema feature complete at this point? I want us to really lock this in before we go wide across 9 langs."), his sharpened line ("We aspire to be the data language for games."), and his three competitive questions. Read on Fable, read-only, main plus #507 and #510, with predictions written before reading. Two owner rulings were given during the read and are reflected: no compiled types is out of scope, and cooked data is the answer to in-place access. The rulings on each item follow in comments; the sweep waits on the wire and language items.

---

All inputs are read: README, SPEC.md whole, SPEC-TABLES.md whole (10,266 lines, main plus the #507 diff), COMPARISON.md, COMPARISON-TABLES.md, COMPETITION.md, USAGE.md whole, VERSIONING.md, PERFORMANCE.md, the #366 thread through the 24-row matrix, the #188 survey, and the #510 §4.12 text. Clone is at `/Users/glenn/rowan-working/schema-coldread-design` (main `1e3c53b`, branches `pr510`, `pr507`). Nothing was built or run. Predictions were written to the scratchpad after README.md and before any other page.

# Artifact 1. Predictions (written after README.md only)

What a complete game data type system must express, and what I expected schema to lack.

| # | Requirement | Expected |
|---|---|---|
| P1 | Sized integers with declared ranges, bit-packed | present |
| P2 | float32/64 plus a quantized float (min, max, resolution) | present |
| P3 | Fixed and bounded arrays | present |
| P4 | Bounded UTF-8 strings; wide strings for UTF-16 hosts | string present, wstring landing |
| P5 | Bounded byte blobs | present |
| P6 | Optional fields independent of a condition | `if` on packets, an optional on tables |
| P7 | Tagged unions | tables yes, packets maybe not |
| P8 | Enums with explicit values, flag sets | present |
| P9 | Nested composition | present |
| P10 | Maps | tables only |
| P11 | Versioning: add/remove/rename/default | tables, present |
| P12 | References, graphs, shared identity | designed, not landed |
| P13 | Runtime reflection for editors | partial |
| P14 | Delta/snapshot compression | LACK, declined |
| P15 | Generics | LACK, declined |
| P16 | Inheritance | LACK, declined |
| P17 | User annotations / editor metadata on fields | LACK beyond min/max |
| P18 | Localization tables | expressible, no construct |
[...]

PR #527, opened 2026-09-04
spec: reachability-scoped protocol id, and retain-unknown (#524, #525)

gafferongames:
Rulings 11 and 12 on #523, the two items filed **before 3.0.0 and not blocking the sweep**. Specification only. No compiler change, no runtime change, and no wire change: the two issues this PR opens, #524 and #525, carry the implementations.

Draft until #507 merges and this rebases onto it.

---

## 1. A reachability-scoped protocol id (#524)

**docs/SPEC.md §3.1, §3.2 and docs/VERSIONING.md.**

The wire-shape projection becomes the **closure over the unit's `type` declarations** rather than a listing of the unit, so an `enum` or a `union` only `table` bodies reach contributes nothing to the protocol id.

**The content enum is the reason, and the page says so.** The README's promise is that one system does packets, saves, config and assets, so a game takes that promise up by keeping all four in one unit. Under a projection that listed the unit, every item kind, quest or achievement added to an enum only tables read bought a coordinated redeploy of both ends for a byte no packet carries.

**What still moves the id**, listed on the page: any `type` edit, a `type` added or removed, any edit to an `enum` or `union` the closure carries (a variant or arm reordered or renamed included), any edit to any `flags` declaration, and an edit that moves reachability itself, which never arrives alone because the use that moved is a `type` edit in the same commit.

`ProjectionVersion` `2` to `3`. Every id moves once.

### Decisions beyond the ruling

- **Every `type` is a root, including an unused one.** The language has no way to say which types go on a wire, so scoping the roots by use would be the dangerous direction. The "an unused type moves the protocol id" sharp edge stands unchanged.
- **`flags` projects unconditionally, and it is the whole of the exception.** A mask is the table wire's one positional vocabulary, so an insert, a reorder or a rename in place is the silent class's second member and no read report can see it. The protocol id is the only frame that refuses two peers holding different bit assignments while they exchange table data over one connection. Content never rides in a flags declaration, because sixty-four bits is the ceiling and the law is append at the end, so holding it in costs nothing the closure was meant to buy. **This narrows ruling 11 by one vocabulary**, and it is the one place to reverse if the owner would rather have the rule with no exception.

### The silent class, re-examined

The four members of SPEC-TABLES.md §4.1 gain **no fifth member**: reachability moves no byte, no kind and no counter. What it removes is an **incidental guard**, and the page now says so as a sharp edge. Until this change a declaration both wires shared was covered by the connect gate as a side effect, so two peers whose vocabularies disagreed refused each other before they exchanged a byte, table data included.

Each member was read against losing that:

| member | what it costs |
|---|---|
| a changed DEFAULT | a fact of a `table` body, which never entered this projection |
| a moved fixed `F` | the same |
| a REFERENT swapped for a twin | the tables baseline's, and the build version moves on it either way (§20.4) |
| a FLAGS variant moved | the one the connect gate was really covering, so `flags` stays in |

And **nothing that left the protocol id left the build version**: §20.4 lists an enum, flags or union variant edit as its own mover, and the cook projection carries every such declaration's variant names and bit positions in a block of its own (§20.2). A cook written before the edit still refuses to open after it.

### What #524 owes

The walk is the whole of what the change rests on, and a missed edge is the spurious MATCH the projection exists to refuse. The page states the obligation: **a negative control that removes one edge from the walk and requires `internal/check/projection_test.go`'s "an edit that moves bytes must move the id" leg to go red.**

---

## 2. Retain-unknown (#525)

**A new docs/SPEC-TABLES.md §6.6, one paragraph in §4, one cell in §4.1, and docs/VERSIONING.md.**

An opt-in on `Load` that keeps every unknown field's bytes in a bounded side buffer the caller declares and owns, re-emitted by `Save`, so a player who rolls back to an older build, saves, and rolls forward again loses nothing the older build could not name.

**No wire change.** Every field is `id reference, kind, payload` and every kind is skippable by its kind byte alone, so the bytes a reader skips are a self-framed unit it copies out and copies back. No byte of §3 moves, no kind is spent, no form byte changes.

### Decisions beyond the ruling

- **Scope: field-position unknowns only.** An unknown enum variant, an unknown union arm, an unknown keyed-array slot and a node record whose type id is unnameable are declined, each with its reason on the page. The first two would be a second occurrence of an id the reader already wrote; the last two are not fields, so putting one back is a splice into a body the reader rewrites whole. The collision argument the whole feature rests on falls out of the scope: a retained id is one the reader cannot name, and a writer writes only ids it can name.
- **Re-emitted at the END of the body, not in the original position.** Position carries nothing on this wire, so an original position is not a fact the wire holds. Appending has no splice, preserves the retained order, and makes the round trip **idempotent after the first save**, which is a property a test pins.
- **A retention FORM, not a byte copy.** A reference names a slot of the file's own id table, so a verbatim copy re-emitted into a file whose table is ordered differently would silently point at other names. A record holds the field with every reference resolved to the eight-byte id it names, so it is self-contained. Only kinds `13`, `14`, `16`, `15` and `30` carry an inner reference; every other payload is copied verbatim.
- **A record is addressed by a PATH, not by an ordinal.** Step one is the node index of §3.1's numbering, and every further step names a child body in the reader's own declaration order. A path step is computable locally at the moment either walk descends, which is what makes one address available to a `Load` walking wire order and a `Save` walking declaration order. Depth is bounded by the schema, never by data.
- **Two counters, `retained` and `retain_lost`, and `retain_lost == 0` is the whole of the safety check.** One number covers a full buffer, an excluded class, and a path a caller invalidated by editing the shape between load and save. The never-clobber rule on the promise page now reads "unless retention was on and `retain_lost` is zero".
- **Three additive entry points**, `LoadRetain`, `MeasureRetain` and `SaveRetain`. `Load`, `Measure` and `Save` are unchanged, so every existing caller, conformance leg and generated golden stands byte for byte. The three suffixes are **owed to §11's claimed set and to `tableGeneratedVerbs` in the change that lands them, and §11 is deliberately not touched here**, because a claim the page makes and the checker does not is a name a user may still take.

### The security bound, stated

The caller's capacity is the only ceiling and the wire cannot raise it. The retained bytes are never interpreted beyond what a skipper already reads. The expansion is bounded and a port states the constant. `retain_lost` is the whole of the denial-of-service surface and it is a counter rather than a failure. A path is the reader's own and is never read from the wire.

### What #525 owes

Five conformance rows, including the id-table-reorder case a verbatim copy fails and the twice-saved byte-identity case, and one wire-fuzzer leg. The existing fuzzer runs with retention off, so its round-trip requirement does not move.

---

## Not touched

- SPEC-TABLES.md §2.8, §3 and §20.2, which #507 edits.
- USAGE.md, which teaches what exists.
- §11's claimed suffix set, and SPEC-TABLES.md §10, which would gain a sentence in the implementation PRs.

---

# Second pass: rulings A and B, and the cold read's named changes

Commit `39cc126`. The cold read ruled "merge after named changes and two owner rulings"; both rulings are made and everything below is landed on this branch.

## Ruling A: the positional array against the flags exception

**`[E.Max]T` is refused in a table body by name** (SPEC-TABLES.md §2.4, §11, §4.1). The keyed spelling `[E]T` is the table form and exists to close exactly this class. The refusal is what makes the flags exception *sound* rather than merely convenient: with the positional array gone, `flags` is the one positional vocabulary a table has, and therefore the one exception the scoped projection needs. `[E.Max]T` stays legal and positional in a `type` body, where the whole spelling projects and the connect gate covers a variant insert.

The flags paragraph now states in one sentence that it **narrows the ruling**, and why: the one vocabulary a table cannot report a move in, a committed baseline that is only a compile-time guard a studio may not have, and the protocol id as the only runtime frame.

**SPEC.md §3.1 enumerates all eight walk edges** rather than illustrating them — a field's named type, an array's element type, an array's bound naming `E.Max`, a keyed array's key enum, a constant's value expression, a union arm's payload and its own field facts, both sides of a branch, and transitivity. The `E.Max` bound and the key enum are the two the earlier draft would have missed, and missing either is the spurious MATCH. **The negative control is now one case per edge kind**, each a declaration reachable only through that edge.

**Restated claims.** "No fifth member" now rests on two stated facts rather than inspection: a table has exactly one positional vocabulary, and that one is held in the projection. "Nothing is misdecoded" is restated as the **table wire's** property, not the protocol id's — the connect gate holds the packet wire, the table wire holds itself, and `flags` is the single place it cannot, which is why it did not move.

**Refreshed:** §13.3 records the narrowing against the owner's original "that is optional" ruling; §20.1 states the scope of group 1 and that group 3 carries what left it; §20.8's arm control gains its **table-only twin** (the same edit to a table-only union must move the build version and leave the protocol id still), and the meaning group's rows are re-sorted by reachability, with `flags` called out as the exception there too.

## Ruling B: retention is a region round trip only

`LoadRetain` loads into a **region**; `SaveRetain` saves from that same region. **The builder path is declined by name**, with two reasons: a builder has no node directory (`Lock` does not fill one, and its slots hold arena offsets), and a builder-side save re-derives the numbering by walking in the reader's own declaration order, so a field reordered between builds — a free edit on this wire — would misplace a path in silence. A region has neither problem: `Load` fills the directory from the wire's framing and nothing afterwards renumbers a region.

A path's first step is therefore the **region's node-directory index**, not a wire index re-derived at save. Land and expand: a builder-path form is a later page if a case appears, and it needs an anchor a builder can hold.

## Named changes from the log

| change | where |
|---|---|
| kind `17`, arrays of element kind `17`, and the reserved node-table field (`0xFFFF…FFFF` under kind `12`) declined from retention, counted `retain_lost` | §6.6 exclusion table, three rows with their own reasons |
| kind-`16` keys resolved at **every** element kind, since the slot key references ride whatever the elements are | §6.6 resolving-walk list |
| "never interpreted" dropped; the inner-walk verdict stated: damage inside sound outer framing **drops the record and counts `retain_lost`**, and **never raises `malformed` on the plain read** | §6.6, its own paragraph |
| `kind_mismatch`, `clamped` and `malformed` restored to the never-clobber condition beside `retain_lost` | §6.6, VERSIONING never-clobber, new sharp edge |
| promise 10 cut to the width of the sharp edge (region-only, seven exclusions, `unknown` class alone) | VERSIONING promise 10 |
| `SaveRetain` **refuses a null report** and returns `-1`, because the save is the only place a caller learns a record was dropped | §6.6 surface |
| the record is **reader-private**, not a wire: no form byte, no version, no byte order, never written or handed anywhere. The page specifies what it must carry and what it may cost, and leaves the layout to the port | §6.6 |
| the retained tail sits **before the node-table field** in the root body, keeping the property that a reader dying inside the node table already holds the root's values | §6.6 |
| the id-table effect stated: moving fields to the end changes first-use order, so the trailer permutes and every reference renumbers. The first save is a **pinned byte string**, not a comparison "modulo the move" | §6.6 idempotence |
| the report **accumulates** across `LoadRetain` and `SaveRetain`, zeroed once and read at the end | §6.6 surface |
| `TableRetain` and **Dart's three member verbs** (`loadRetain`, `measureRetain`, `saveRetain`, taking that backend's nine claimed field-name verbs to twelve) owed to §11, with the reason the claim is not made in this page's own change | §6.6 |
| `internal/tablewire` carries no retention, so the fuzzer's retention leg is not buildable until the **oracle** retains too | §4.2 requirement 3, and §6.6 |

## Still open

Nothing from the log is unresolved.

**Rebased onto main after #507 merged** (`b74a7d4c`), with no conflicts. Every fact §6.6 rests on was re-checked against the merged §3: the field framing, the skip rules, first-use order in the id table, the reserved node-table field under kind `12`, kind `16`'s per-slot key references, kinds `17`, `31` and `32`, the node-table field written last in the root body, and node index `1` as the root. #507's own edits land in §2.8, §3's trailer and §20.2, none of which this PR restates. Its rewrite of the trailer paragraph reinforces the retention form rather than colliding with it: "the reference and the compiler's engine both resolve a reference to its id through one array index", which is exactly the resolution a retained record carries.

### DIFF
--- a/docs/SPEC-TABLES.md
+++ b/docs/SPEC-TABLES.md
@@ -1097,14 +1097,34 @@ enum is keyed.
   id, and that is held by test.
 - **Fixed-size when `T` is**, so the zero-cost gate holds.
 
-**The two spellings are ONE FIELD on the type wire and TWO DIFFERENT
-ENCODINGS on the table wire.** `[E]T` is kind `16` and `[E.Max]T` is
-kind `14` (§3), so changing a TABLE field from one spelling to the other
-is a wire break, not a refactor: an old file read under the new spelling
-is a kind mismatch — skipped, counted, never misdecoded (§4) — and the
-committed baseline refuses the edit at compile time (§18.2). Giving the
-keyed body its own kind is what buys that; the two encodings are not
-mutually decodable and nothing should let them be tried.
+**`[E.Max]T` IS REFUSED IN A TABLE BODY, and `[E]T` is the table form.** The
+two spellings were once both legal on this wire, `[E]T` as kind `16` and
+`[E.Max]T` as kind `14` (§3), and the second is now a compile error naming
+the field and the enum, with the one-word fix in the diagnostic. **The reason
+is that an ordinal-indexed array is a POSITIONAL vocabulary and a table may
+have only one.** A `[E.Max]T` field carries its elements by position, so
+inserting a variant in the middle of `E` lands every later element one slot
+off, in every file already written, with nothing on the wire that could say
+so. That is the shape §4.1's second member has, and it is the shape §4.1's
+first REPORTABLE-by-construction bullet was built to remove: keyed slots ride
+by name, so a middle insert moves no slot. Leaving the positional spelling
+legal beside the keyed one left the class open for anyone who spelled it the
+old way and never changed it, which no kind number can catch, because nothing
+about the FIELD moved.
+
+**It is also what makes `flags` the only exception to the reachability rule**
+(SPEC.md §3.1). Under a projection scoped to what a `type` reaches, an enum
+only tables reach leaves the protocol id, so the connect gate stops refusing
+two peers whose variant orders disagree. That is correct for a vocabulary
+read by NAME and wrong for one read by POSITION, and refusing `[E.Max]T` here
+is what leaves `flags` as the only positional vocabulary a table has, and
+therefore the only exception the projection needs.
+
+**On the TYPE wire the spelling stays legal and positional**, unchanged: a
+`type` body's `[E.Max]T` is a plain array whose extent is the variant count,
+its bytes are the packet wire's, and every fact of it projects (SPEC.md
+§3.1). The refusal is the TABLE body's alone, and it is what §2.2's mode
+derivation already made a per-body question.
 
 **A KEY ENUM IS IN THE TABLE CLOSURE'S VOCABULARY**, and the closure's
 rules reach it through the keying field. An enum that a table closure
@@ -1117,11 +1137,11 @@ KEYING FIELD as the edge that pulled the enum in, because that edge is the
 only reason the rule applies and a person looking at the enum alone would
 not otherwise see it.
 
-Refused by name: a bound naming a `flags` declaration (a mask holds any
-set of bits at once, so it names no single slot); a bounded keyed array,
-`[..E]` or `[A..E]` (a keyed array is COMPLETE by construction); an
-element that is a pointer, as for any array (§15); and an index of
-`E::None`, which names no slot.
+Refused by name: **`[E.Max]T` in a table body**, above; a bound naming a
+`flags` declaration (a mask holds any set of bits at once, so it names no
+single slot); a bounded keyed array, `[..E]` or `[A..E]` (a keyed array is
+COMPLETE by construction); an element that is a pointer, as for any array
+(§15); and an index of `E::None`, which names no slot.
 
 ### 2.5 Byte buffers: `data *bytes`
 
@@ -3358,11 +3378,18 @@ on it.**
 - **An enum-ordinal-indexed array** was the last positional vocabulary
   besides flags: insert a variant in the middle and every later slot lands
   one place off. `[E]T` (§2.4) closes it — keyed slots ride by name, so a
-  middle insert moves no slot.
+  middle insert moves no slot. **And `[E.Max]T` is now REFUSED in a table
+  body by name** (§2.4, §11), so the closed class cannot be reopened by
+  spelling it the old way and never touching the field again. That refusal
+  is what leaves `flags` the ONE positional vocabulary a table has, which
+  is in turn what makes `flags` the one exception to the reachability rule
+  the protocol id is scoped by (SPEC.md §3.1).
 - **Changing a table field between `[E]T` and `[E.Max]T`** would then
   have replaced it: two encodings under one kind would have let a reader
   decode keys as values and report nothing. The keyed body's own kind `16`
-  (§3.2) turns that edit into a kind mismatch, counted like any other.
+  (§3.2) turns that edit into a kind mismatch, counted like any other, and
+  the refusal above means one END of that edit no longer compiles. The kind
+  is what covers the wire an old build already wrote.
 - **Changing a table field between `*T` and a plain `uint32`** is the
   same shape one more time: a node index and a number are the same four
   bytes, and under a shared kind an index would read back as a plausible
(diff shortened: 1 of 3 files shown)


gafferongames:
Two rulings from the cold read, under the owner's word of 2026-09-04 ("go with your decision on all questions"):

A. `[E.Max]T` is refused in table bodies by name; `[E]T` is the table form. The flags exception stands as the only exception to the reachability rule and the page says it narrows the ruling and why. The walk lists every edge, including a keyed array's key enum and an enum reached only through an `E.Max` bound in a type body, and the negative control is one case per edge kind.

B. Retention is a region round trip only: LoadRetain into a region, SaveRetain from that region. The builder path does not carry retention, by name and with the reason (a builder re-derives the numbering from the reader's own declaration order, so a legal reorder between builds misplaces a path). A builder-path form is a later page if a case appears.

PR #529, opened 2026-09-04
spec: doc and tags columns, field-level tags (#523)

gafferongames:
Ruling 4 on #523: `doc` and `tags` columns on the descriptors and the view, and
field-level valueless tags in the type-tag grammar. Specification only, no code.

**Second commit: the cold read's named changes, and the owner ruling in it.**
Doc comments are **opt-in**. The marker is `///`; a plain `//` block above a
declaration stays an ordinary comment and reaches no descriptor. The reason is
measured on this repo: 166 comment lines sit directly above a declaration, a
field, a variant or an arm across 40 of its 47 corpus files, working notes to
the last one, and under implicit binding every one would have become a
descriptor string, a comment in nine generated languages, and read-only data in
the binary of every game that links the table headers. Under opt-in the corpus
exhibits the empty spelling everywhere except the one exhibit, and what a build
carries is what an author asked it to carry.

What else landed in that commit:

- **One rule closes every trap**: every `///` line is part of a doc comment or
  is refused by name. A block above `package` or a `const( )` / `reserved( )` /
  `align` item; a block held off its item by a blank line or a `//` line; a
  `///` trailing code on the item's own line. Each is a diagnostic naming `//`,
  because silently dropping an author's opt-in is what opt-in exists to prevent.
- **One escaping rule**, the target's own string-literal escape. Generated code
  takes ordinary LINE comments in every target, never `/** */` and never
  `/// <summary>`, so a `*/` or a `<` in an author's text needs no rule at all.
  Elixir carries the descriptor column alone for a field and a variant.
- **The bare-key lookup is against the union of every valued key on every line
  kind** (`min`, `max`, `resolution`, `was`, `json`, `cpp_native`,
  `cpp_include`) plus the refused-by-name keys (`doc`, `round`), so `| min` on a
  `string(N)` is refused rather than accepted as a tag. Reserved words are
  refused with the word named.
- **Valueless-first ordering is a formatter normalization, not a refusal**, and
  a mixed comma-and-newline variant list is accepted and normalized. Two lexer
  rules are stated (a newline before a comma inside a variant list is
  whitespace) and one refusal added (a `}` on the same line as a qualified
  variant), each with a named negative control in 7.2.
- **8.7's two unreddable cost claims are replaced by observables.** `doc` is
  never NULL is held by a printer that concatenates every row's `doc` with no
  null test, which faults on a NULL. The shared empty string is held by address
  identity in C, C++ and Rust and by one emitted definition against many
  references everywhere else.
- **The exhibit is pinned down.** It annotates EXISTING `tabledemo`
  declarations and adds none, since a new declaration would move that unit's
  build version while the page claims no id moves. `tabledemo` has no recorded
  source golden today (section 15's follow-on), so the commit that adds the
  columns records one. The exhibit's doc text carries the extraction rule's edge
  cases: two leading spaces, an empty marker line, trailing whitespace, a quote,
  a backslash.
- **The independence gate names all three things that move once**: descriptor
  initializers, generated data files gaining doc idioms, formatter output where
  a doc comment no longer breaks an alignment group.
- **Also**: the listing flattens a multi-line doc with `\n` escaped;
  `ViewType.doc` is gated against `type->doc` beside the arm pair; section 8.1
  states which annotations a build can reach without the view file; section
  20.1's "a type tag" row becomes a tag on any line and its comments row names
  the `///` doc comment; section 3.1 says in one sentence why the compiler
  reading a comment is safe under a projection.

Nothing here touches a wire. A doc comment and a tag are excluded from the wire
shape projection and from the cook projection, enter no baseline row, and are
not members of the silent class, so annotating a shipped schema moves no
protocol id, no build version, no wire golden and no stored byte's meaning.
That is the property the open tag namespace rests on, and it is stated in every
place a reader will look for it.

**Draft until #507 merges.** `git merge-tree --write-tree pr507 spec-doc-and-tags`
returns a tree and no conflict, so the two land in either order. This branch
edits SPEC-TABLES.md section 8 only, which #507 does not touch.

## What lands

**SPEC.md**

- **4.1, doc comments.** Opt-in, marker `///` (see above). The text rule is
  exact: each line contributes what follows its `///` with at most one leading
  space dropped, trailing whitespace dropped, lines joined by one newline, no
  trailing newline, nothing interpreted. It rides the IR into the `doc` column
  and into line comments in the generated code.
- **4.2, tags.** A valueless identifier is legal in every qualification
  section: a type, a table, an enum, a flags, a union, a constant, a field, a
  variant, an arm. The valued vocabulary stays exactly as closed and as
  per-line as it was.
- **4.2 grammar.** `Const`, `Union`, `Variant` and a bare-name `UnionVariant`
  gain `AttrSection`. A variant list now separates on a comma **or** a newline.
- **3.1** names doc comments and tags in the excluded list. **4.11** refuses
  `| doc = "..."` with the comment form named. **7.4** gets the two formatter
  consequences. **9 q5** is settled.

**SPEC-TABLES.md section 8**

- **8.1**: `doc`, `num_tags` and `tags` on `TableFieldInfo` and on
  `TableTypeInfo`. Absence is `""` and `0`/NULL. All static, all
  constant-initialized, declared order, zero allocation on a walk. A descriptor
  carries its own annotation and nobody else's. Neither column is in reach of a
  byte, the text form does not read them, the pack tree does not see them.
- **8.3**: the three members on `ViewConstant`, `ViewVariant`,
  `ViewVocabulary` and `ViewType`, plus the rule that the registry listing is
  the whole annotation surface.
- **8.6**: the `| doc` attribute stays refused, the view renders nothing.
- **8.7**: both columns are in the listing gate, one corpus declaration shows
  them filled, the cost claim is held, and the independence gate is restated to
  move once when the columns land.

## Decisions beyond the ruling

Eight, each one a gap the ruling's four records did not reach.

1. **`ViewVocabulary` carries the columns too.** The ruling names
   `TableFieldInfo`, `ViewVariant`, `ViewType` and `ViewConstant`. An enum, a
   flags or a union DECLARATION has a row in none of those, so a doc comment
   above `enum Weapon` would have had nowhere to land. The ruling's own reason
   for adding the columns now is that descriptors are already in nine ports and
   a later column is nine edits, and that reason reaches this record with equal
   force.
2. **`TableTypeInfo` carries them as well as `ViewType`.** A table's descriptor
   lives in the table header, not the view file, so a walker that entered a
   nested table through the `table` column reaches it in a build that compiles
   no view file. `ViewType` keeps its existing `num_tags`/`tags` and gains
   `doc`; the two records carry the same values.
3. **A descriptor carries its own annotation and nobody else's.** A variant's
   doc and tags belong to the enum, flags or union declaration, so they ride
   the registry rows and are one lookup away by `type_name`. No
   `variant_doc( value )` or `variant_tags( value )` function joins the field
   descriptor in nine ports to spell what the registry already spells. The one
   overlap is a general arm, which is a field line and carries a
   `TableFieldInfo`: that descriptor and the arm's `ViewVariant` row carry the
   same two values, and 8.7 checks that they agree.
4. **`TableBlockFieldInfo` carries neither.** It describes where a field sits
   in the block projection; the same field's annotation is one record away.
5. **Absence is `""` and `0`/NULL, not NULL and NULL.** A printer concatenates
   `doc` with no test, and a list has its count checked before it is walked
   whatever it holds. Every absent `doc` is the SAME shared static empty
   string, so a unit that documents nothing pays no string data at all, and
   8.7 holds that.
6. **Tags become legal on every line kind, and the "admits no qualification"
   case disappears.** A constant and a `union` declaration took no pipe at all
   before this. Keeping them as exceptions would have meant remembering two
   line kinds for no gain.
7. **A bare identifier spelling a known valued key is refused by name**
   (`| min` draws "min takes a value: write min = 0"), and so is a repeated
   tag. Without it the open namespace silently swallows typos in the closed
   one, which is the failure mode the closed vocabulary exists to prevent.
8. **A variant list separates on a comma OR a newline.** This is the only real
   grammar problem in the ruling: the pipe claims the rest of the line, and a
   comma-separated variant list wants the comma at the end of that same line.
   The newline as an alternate separator is purely additive (every comma list
   written to date parses unchanged; only previously-illegal programs become
   legal), and one formatter rule keeps the output canonical: a list in which
   any variant carries a qualification or a doc comment is one per line with no
   commas.

Two smaller consequences, both stated on the page: a doc comment does not break
a `schemafmt` alignment group (a documented run would otherwise become a run of
one-field groups), and the columns landing is the one edit that moves the
independence gate's left-hand side, once, because the members sit in the table
headers.

## One removal outside section 8

SPEC-TABLES.md section 15 still listed "DOC STRINGS in the view" as a named
follow-on. It landed here, so the row is scaffolding and is deleted. Section 15
is untouched by #507 and the merge check still returns a tree with no conflict.

## Two corrections made in passing

The 4.2 valued-vocabulary list said "enum declarations take `max`" where `flags`
takes it too, and omitted `json` on a table field. Both are corrected in the
same bullet.

## What this PR does not do

No compiler, no emitter, no corpus file, no golden. The corpus declaration that
exhibits both columns filled is specified in 8.7 and lands with the
implementation.

### DIFF
--- a/docs/SPEC-TABLES.md
+++ b/docs/SPEC-TABLES.md
@@ -7655,10 +7655,6 @@ inspects everything in the schema built:
   same pin (§8.7). Nothing is refused meanwhile, because nothing in a schema
   asks for one (§8.4): a backend without the emitter is a backend whose
   users have no registry, and the status paragraph says so.
-- **DOC STRINGS in the view.** Doc comments are deferred with their design
-  pinned (SPEC §4.1, §9 q5); when they land they become one more descriptor
-  column and one more registry column, and nothing else about §8 moves. No
-  competing `| doc = "..."` attribute is introduced ahead of them.
 - **A view over the PACKET wire's own layout** — bit offsets, field widths
   and the compressed-float parameters a `type` rides under (SPEC §6.1),
   beside the declaration facts §8 carries. It is what a packet inspector
--- a/docs/SPEC-TABLES.md
+++ b/docs/SPEC-TABLES.md
@@ -5318,6 +5318,16 @@ decided by whether it COMPILES that file (§8.4). An editor does; a game
 build never includes it. Nothing here rides a wire either: the view moves no
 protocol id, no generated wire byte and no baseline row (§10, §18).
 
+**This is also where what a PERSON wrote about a declaration comes out.** A
+doc comment (SPEC §4.1) and a tag (SPEC §4.2) are the language's two open
+channels — one prose, one an identifier in a namespace the compiler assigns
+no meaning — and they reach a tool through two descriptor columns, `doc` and
+`tags`, and through nothing else. They are id-neutral and byte-neutral in
+every direction the rest of this page measures, which is what lets the
+namespace stay open: annotating a shipped schema costs no redeploy, no
+re-cook and no baseline row. The columns are specified with the rest of the
+descriptors in §8.1 and appear on the registry's records in §8.3.
+
 ### 8.1 The descriptors
 
 For every table in the unit's closure the generated header carries static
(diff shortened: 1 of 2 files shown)


gafferongames:
Ruling from the cold read, under the owner's word of 2026-09-04: doc comments are opt-in. A marked block (`///`, or the marker the page settles on, stated once) immediately above a declaration is the doc comment; a plain `//` block stays an ordinary comment and never reaches a descriptor. The reader's corpus scan found 166 `//` lines directly above declarations in 39 of 47 files, working notes that implicit binding would have turned into descriptor strings and doc idioms in nine languages and shipped in every binary linking those table headers. With opt-in the cost is what an author chose. The named page changes from the read land on this branch before it leaves draft.

gafferongames:
Two one-clause residuals from the cold read's verification, neither blocking, to carry into the implementation PR for the columns: (1) an enum's None row 0 and a keyed array's None key take the same "" and 0/NULL as union tag 0 (only the union row is named on the page); (2) a ViewConstant's doc in Elixir: a constant is a function, so @doc fits, while §4.1 names @moduledoc and @typedoc for declarations only.

PR #530, opened 2026-09-04
spec: the message form, an id table scoped to the connection (#523)

gafferongames:
Ruling 10 on #523, plus the owner rulings and named changes from two cold-read passes. Specification only, nothing is implemented.

A second wire form under the one-byte form version moves the id table off every message and onto the connection. **The file form does not change.** No kind is spent, no payload row moves, no skip rule changes. The form is framing and nothing else.

## The form

- **Form byte `2`.** Two parts: the form byte and the root body, which ends at its own zero reference. No trailer.
- **The vocabulary is the UNIT's whole closure**, in the cook projection's order: every field name, every enum variant and union arm name, then a fixed **tail** — the reserved node-table id, the three blob type ids (`bytes`, `string`, `wstring`), and the name id of **every** table in the closure in the projection's sorted record order.
- **The tail is unconditional, and the page states the choice.** A unit with no pointer announces it anyway, because scoping it to pointer targets would move every slot after an inserted type id, and scoping the node-table and blob ids to units with a variable class would move the whole tail on a unit's first pointer. A slot number is a compile-time constant a generated field header carries, so it must not drift under an edit that has nothing to do with it.
- **Three properties follow.** The table is a pure function of the build version, so "keyed by the build version" is literal. **The whole announcement is a compile-time constant of the unit**, so a backend may emit `Announce` and `AnnounceMeasure` as a constant byte array and its length rather than as a walk, and the C++ reference states which it does. The writer's slot numbers are compile-time constants in generated field headers, so there is **no runtime slot lookup on the send path**. The receiver resolves once.
- **The ID TABLE MESSAGE announces it, and it is an ordinary form-`1` file.** Its body is one required field, the build version, under a reserved id `0xFFFFFFFFFFFFFFFE`, and its trailer is the connection's table. It reads **tolerantly with exactly one strict check** (that field present once, kind `9`, eight bytes wide), so the announcement can gain a field in a later minor without a lockstep redeploy.
- **Once a connection per direction, and never again.** No re-announcement and no state machine. A second announcement is refused by name and replaces nothing. A refused announcement sets no table, so every message on that connection is refused for want of one. A build change is a new process and a new connection.
- **The build version keys the table and never gates the connection.** §20.5 stands.
- **A peer with no table refuses the message by name.** Nothing decoded, no counter moves, `malformed` does not fire. The recovery is the sender's: a new connection announced first, or the file form. A receiver never guesses a table and never falls back on its own.
- **A reference past the table is §3's framing damage at the level it occurs.** In the root body the body stops, `malformed` counts once, and the fields decoded before it stand. No second pass.
- **A connection is a transport connection**: one TCP or WebSocket connection, or one reliable ordered stream or channel of QUIC or a reliable-UDP transport, per channel. A restart is a new connection with empty tables, a receiver caches nothing across connections, and stateless request-response is out of scope by name (the file form rides there).
- **What this form does not carry**: which root a message is, is the application's; where a message ends is the transport's. §1's no-envelope promise at the message level.
- **`SaveRetain` under form `2` refuses by name and returns `-1`**, on §6.6's misuse precedent, never a silent drop. Loading is unchanged. The two answers are the file form and a verbatim relay.
- **The packet wire is out of scope**, and the page says why: it is same-or-refuse on the protocol id, so both peers must ship together, and a deployed game client and a backend do not.

## The invariant across the forms

The bodies are **not** byte-identical: a file's slots are its own first-use order and a connection's are the unit's projection order, which sorts records and vocabularies by name. What is invariant is the **resolved form**, §6.6's own normal form, every reference replaced by the 64-bit id it names and every length recomputed. That is the claim and the pin. Each vector's byte length is a pinned golden of its own.

## What it measures

The three backend messages of #523, verified against this repository's own compiler at `bafdb69`:

| message | file form | message form |
|---|---:|---:|
| `LoginRequest`, non-default | 106 | 58 |
| `MatchResult`, non-default | 273 | 225 |
| `StorePurchase`, non-default | 104 | 48 |
| `LoginRequest`, defaults | 10 | 2 |
| `MatchResult`, defaults | 43 | 27 |
| `StorePurchase`, defaults | 10 | 2 |

45%, 18% and 54% off the three non-default messages, which turns a loss of 2.2x, 1.4x and 2.6x against proto3 into one of about 1.2x, and `MatchResult` at its declared defaults into an outright win at 27 against 40.

The `backenddemo` unit names twenty ids in its records (twelve distinct field names, eight enum variant names, counted from `schema build-version --facts`) and eight more in its tail, so with the reserved build-version id its announcement is **twenty-nine entries and 252 bytes**. One round of the three saves 152 bytes, so it is paid back partway into the second round. Every slot is still one byte, so no message size moves. A unit of 500 ids announces about 4 KB once. The bound stays 4096 entries.

## The surface, owed to §11

Named but not claimed in this change, on §6.6's precedent: `TableVocabulary`; the unit-scope `Announce`, `AnnounceMeasure`, `AnnounceRead`; the suffixes `LoadMessage`, `MeasureMessage`, `SaveMessage`; the refusal reasons `no_vocabulary`, `second_announcement`, `vocabulary_too_large`, `message_form_as_file`; and Dart's `loadMessage` / `measureMessage` / `saveMessage` plus the library-scope four.

## Goldens

Twelve vectors, the three messages in each form, under a new `backenddemo` unit at `tables/backend` and a `backend_conn` connection, with two manifest line kinds (`connection`, `message`). Four more carry rules a value alone cannot reach: a wide vocabulary, a pointered root, a two-peer pair for per-direction independence, and a connection carrying a second announcement. The wide-vocabulary unit is **generated and committed** — `test/vocabgen` writes `tables/vocab/Vocab.schema`, unit key `vocabdemo`, ten tables of thirteen fields, on `test/cookgen`'s precedent — so its message names one id in a one-byte slot and one in a two-byte slot.

## Files

- `docs/SPEC-TABLES.md` §3.3, new, the whole form.
- `docs/SPEC-TABLES.md` §3, §3.1, §5, §11: the form-byte bullet names both forms, the form pin re-points to forms `0`, `3` and `0xFF`, the tiny-message paragraph names the message form beside the `type` stream, §3.1 states the reserved-id rule for both ids, and two ids are held back now rather than one.
- `docs/VERSIONING.md`: a section of its own between the wire form and the text form, promise 7 amended, and #523 on the Owed-before-3.0.0 list.

Docs-only, and it merges clean against `main`, #531, #532, #520 and both remaining spec branches.

### DIFF
--- a/docs/SPEC-TABLES.md
+++ b/docs/SPEC-TABLES.md
@@ -3199,10 +3199,10 @@ ids one message happens to use:
 - **every field name** of every record in the closure, a generated map entry's
   `key` and `value` included (§2.8).
 - **every enum variant name** and **every union arm name** in the closure.
-- **every table NAME id that can ride as a node record's type id**, and **the
-  reserved node-table id** (§3.1), when the unit has a variable class at all.
-  A unit with no pointer can put neither on the wire, so it announces neither.
-  One rule decides both, and it is the rule above: what the closure CAN write.
+- **the reserved node-table id**, **the three BLOB type ids**
+  `fnv1a64( "bytes" )`, `fnv1a64( "string" )` and `fnv1a64( "wstring" )`
+  (§3.1, and schema#532 for the third), and **the NAME id of EVERY table in the
+  closure**, whether or not a pointer names it today.
 
 **THREE PROPERTIES FOLLOW, and they are why the vocabulary is the unit's rather
 than the message's.**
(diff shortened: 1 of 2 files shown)


gafferongames:

1. The announcement is the unit's whole vocabulary, in a compiler-settled order, once per connection per direction: the table is a pure function of the build version, slot numbers are compile-time constants in the generated field headers, no runtime slot lookup on the send path, the receiver resolves once. No re-announcement inside a connection; a second announcement is refused by name. The 4096 bound stays.
2. SaveRetain under form 2 refuses by name and returns -1, never a silent drop.
3. The announcement reads tolerantly with one strict check (the reserved build-version field present once, kind 9, width 8), so it can gain fields in a later minor.
4. A connection is the transport connection (TCP, WebSocket, or one reliable ordered channel or stream), per channel; stateless request-response transports are out of scope for the form; a restart is a new connection; nothing is cached across connections.

Issue #547, opened 2026-09-04
law: generated code looks native and conforms to each language's best practices, gated per port


gafferongames:
Owner's rule, 2026-09-05: "Each language implementation of code generated from schema should conform to best practices for that language." Beside it, the same morning: "Use whatever error reporting method is standard for the language (but in C/C++ do not use exceptions)" and "if it costs speed on read or write, it's not worth it."

The law: what schema emits for a target reads as that language's own code to a fluent reader of it: naming, error reporting, memory and ownership, packaging and module layout, formatting under the language's standard formatter, and the language's standard analyzers at their default strictness clean. Where an idiom would cost speed on the read or write path, the measurement decides and the dialect page records the choice; an idiom is never dropped for speed without a number.

The gate, per port row of the matrix (#366): (1) the language's standard formatter and analyzer legs in CI at default strictness (gofmt and vet, clippy, dart analyze and format, eslint, the .NET analyzers, checkstyle or the JDK's lint, credo and mix format, clang-tidy for C and C++), red on any finding in generated code; (2) the dialect page for the language, produced by the red and blue read (docs/PORTING.md holds the register), stating the accepted idiom set with its measurements; (3) a fluent-reader review of one generated unit per language before the row turns green, recorded on the row's PR. Owed for the nine existing rows before the sweep widens them, and for every row the sweep adds.

gafferongames:
Owner, same morning: "And look native. We already know this." The law is standing, not new; this issue is its gate.

rowan-claude:
Second pass owed: PR #1050 closed unmerged on 2026-09-20 under the backlog drain (RETAIN-CLOSE-RECUT); branch `rowan/schema-547-law-generated-code-looks-native-and@71b39a06e6db41c2e805e80c99e11fdba4372dc8` retained as a bundle (`pr-1050-71b39a06.bundle` in Rowan's retained store); reads at that head: none; cut the next card from live `main` with prior: `rowan/schema-547-law-generated-code-looks-native-and@71b39a06e6db41c2e805e80c99e11fdba4372dc8`.

PR #550, opened 2026-09-04
native: the formatter and analyzer gate, one CI leg per port (#547)

gafferongames:
The gate half of #547.

The law is that what schema emits for a target reads as that language's own code. Two of its three parts are machine-checkable, and this lands them: the language's standard formatter in check mode and its standard analyzer at default strictness, over the generated code of **both** corpora — the examples corpus under `generated/`, the tables corpus under `build/` — red on any finding. The third part, the fluent-reader review, is a person's and belongs on each row's own pull request.

**Most of these legs are red today, deliberately.** They are not skipped and no finding is silenced. Each red leg's findings are enumerated below and filed as one issue per target, as work owed by the emitter.

## Shape

One `make native-<lang>` per port, in that port's own `make/<lang>.mk` beside its test and conformance legs, registered with `NATIVE_LEGS`. One CI job per port, fanned out over a registry the workflow discovers rather than a list it carries: `test/native/<lang>/ci.json` names the make target, the sibling runtime the analyzer needs on disk to typecheck what the emitter wrote, the toolchains to install, and the make overrides. A port with a conformance driver and no native row fails `go test ./...`.

Every instrument is pinned and none comes from a machine's PATH by accident.

## Status per target

Filled in once CI has run.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,55 @@
+# The house C and C++ style, and the one the native gate holds the emitted
+# text to (issue #547). clang-format has no canonical style of its own, so the
+# style IS this file, and it is written from the estate's own C and C++: the
+# serialize runtime the generated code sits beside.
+#
+# Every option below states a fact about that code rather than a preference:
+#
+#   IndentWidth 4, UseTab Never   four spaces, no tabs, everywhere.
+#   ColumnLimit 0                 no reflow. The emitted text carries long
+#                                 explanatory trailing comments by design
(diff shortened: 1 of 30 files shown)


rowan-claude:
Closing as stale: 286 commits behind main, conflicting, and seven of its legs were red by design pending emitter work. #547 keeps the formatter and analyzer gate as owed work; a fresh branch against today's tree is the way back.

Issue #554, opened 2026-09-05
design note: the batch primitive for types, a column layout on the wire for N instances of one type


gafferongames:
Owner, 2026-09-05, after ruling the message form's primitive is a batch: "Can this be applied to other things like types too?" Design note, not work.

The packet wire already writes a packet as one continuous bit stream and an array of a type inside it is a batch with no per-instance alignment, so the bits do not change. What a batch primitive would add for types is the layout: N instances of one type written structure-of-arrays on the wire (all of field 1, then all of field 2), which costs the same bits when bitpacked and pays when a compression stage sits behind it, because runs of like values are what delta and entropy coding consume. That is the compression law's order (bitpacking first, every win into delta, rANS after) and the replicant's natural shape (a snapshot as columns against a baseline).

The door to keep open: a column layout for an array of a type under a later projection token or wire law, moving no existing id; and the message form's batch primitive (#523 addendum) as the first place a batch exists on a wire. Owed nowhere until a compression stage exists (serialize.pro, replicant).

rowan-claude:
Done on main: docs/SPEC.md reserves the column projection token by name citing this issue, and docs/VERSIONING.md lists it among the reserved tokens; the design note has its home.

PR #559, opened 2026-09-05
spec: the message form, second round (#523)

gafferongames:
Specification only. No codec, no corpus, no generated code. The codec change that
lands `docs/SPEC-TABLES.md` §3.3 is a separate PR and re-pins every form-`2`
golden.

## What the rulings said, and where each one landed

| ruling | where |
|---|---|
| scope by ANNOUNCEMENT, not transport | §3.3 "The scope is the ANNOUNCEMENT, not the transport", VERSIONING.md |
| the primitive is a BATCH | §3.3 "The primitive is a BATCH", §15 for the three later passes |
| the body is BITPACKED, announcement carries kind and width | §3.3 "The wire, exactly" and "The announcement" |
| the design statement, in the owner's words | §3.3 opening paragraph, VERSIONING.md |
| the cost rule, distinct from #546 | §3.3 "THE COST RULE, and it is this form's own" |
| the reserved column door (#554) | SPEC.md §4.11, with pointers from §3.1 and §4.3 |
| the arithmetic, hand-sized, pinned as goldens | §3.3 "THE ARITHMETIC, worked from the wire model" |
| the form byte | §3.3 "Form byte `2` keeps its number" |
| the surface names under §11 | §3.3 "THE SURFACE", §11's suffix list, `tableGeneratedVerbs` |
| yojimbo#344 as the handshake carrier | §3.3 scope section |
| **the terminator: a body ends at its own zero reference and the batch adds none** | §3.3 "The wire, exactly", stated as the ruling |
| **"not massively slower" is a FACTOR OF TWO, above which the bitpacked body reopens** | §3.3 "THE COST RULE", VERSIONING.md |

## The arithmetic

Bytes, every column hand-sized from its own wire model over the same instances.
`E` is 28 for the `backenddemo` unit, so a reference is 5 bits. Each message is
its own batch of one unless the row says otherwise.

| instance | packet wire | file form | byte body | **bitpacked body** | proto3 |
|---|---:|---:|---:|---:|---:|
| `LoginRequest`, full | 46 | 106 | 58 | **51** | 49 |
| `MatchResult`, full | 115 | 273 | 225 | **142** | 189 |
| `StorePurchase`, full | 36 | 104 | 48 | **41** | 40 |
| `LoginRequest`, defaults | 14 | 10 | 2 | **3** | 0 |
| `MatchResult`, defaults | 115 | 43 | 27 | **10** | 40 |
| `StorePurchase`, defaults | 15 | 10 | 2 | **3** | 2 |
| the three full, one batch | 196 | 483 | 331 | **230** | 278 |

The announcement is 316 bytes, 28 entries at about ten bytes each.

**Against the byte body**: 12%, 37% and 15% off, and 63% off `MatchResult` at
its defaults. The two empty messages grow by one byte, from 2 to 3, because a
batch pays a count.

**Against proto3**: the batch is 17% under and `MatchResult` is 25% under, and
**`LoginRequest` is two bytes OVER and `StorePurchase` one byte OVER**. That is
the one place the cost rule is not met and the page says so rather than
averaging it away. The cause is that `player_id`, `client_build` and
`price_minor` are declared BARE, so the wire writes 64, 32 and 32 raw bits where
proto3 writes a varint whose value happens to be small, and a message that is
one opaque payload plus a few scalars has nothing else to win with. Three things
close it: declaring the range a field actually holds (`client_build uint32 |
max = 65535` alone puts `LoginRequest` level at 49), the batch itself (the form
byte and the count are paid once for however many bodies ride), and a
variable-width encoding for bare integer kinds, which is registered as a named
follow-on because it is a divergence from the packet wire and wants a
measurement first.

**Against the packet wire** the residual is 5, 27 and 5 bytes. On `MatchResult`
25 of the 27 are ten rows of three references and a terminator, which is the
price of naming a field inside an array of tables, and is what the batch-level
passes and the column door are aimed at. On the defaults rows the sign flips:
`MatchResult` is 10 bytes against the packet wire's 115.

Worked bit by bit in the page for all three, so a codec PR has something to
reproduce rather than a number to trust.

## The two open questions, now RULED by the owner

**A. THE TERMINATOR.** A body ends at its own zero reference, as a body always
has, and the batch adds no terminator of its own. `M` bodies carry `M`
terminators, one each, at 5 bits apiece on this unit, and the last body's is the
last thing in the batch. A single terminator for a whole batch has nothing to
say about where body `k` ends, because elision means the only thing that can say
a body's fields are finished is that body's own zero reference. The page states
it as the ruling and no longer as a reading.

**B. THE COST RULE'S FACTOR IS TWO.** "Not massively slower" is a factor of two
on the matching read path and on the matching write path, measured against the
BYTE BODY over the same values, at a named sitting whose ratios, machine, build
and date go on the page. **Above two on either path the bitpacked body
reopens**, which is a real outcome: the byte body is a wire that works, and
doubling the CPU to save a third of the bytes is not a trade that has been made.
The number is on §3.3 and in VERSIONING.md.

## The decisions this PR makes that the brief did not, each a cold-reader question

2. **The vocabulary moves OFF the announcement's trailer and INTO its body**, as
   one `bytes` field under a third reserved id `0xFFFFFFFFFFFFFFFD`. The forcing
   reason: an entry now has to carry a KIND and a SHAPE, which a trailer of bare
   `u64`s cannot, and one NAME has to be able to appear at two shapes (a unit
   declaring `count uint8` in one table and `count uint32` in another), which
   §3's "one id twice is malformed" forbids. It also restores §3's writer rule
   that an id no body references is never written, which the old announcement
   was the one exception to. **This is the largest structural decision in the
   PR.**

3. **An entry is an (id, kind, shape) TRIPLE and a triple is placed once.** The
   same name at a second kind or shape takes a second slot. That is what makes a
   reference unambiguous and skipping exact without a per-field disambiguator
   bit.

4. **The shape carries RANGE facts, not only WIDTH.** Skipping needs only the
   width. Decoding under a declaration whose range MOVED needs the base, the
   min and the step, so a receiver whose `score` runs to 200000 meeting a sender
   whose `score` runs to 100000 reads the sender's 17 bits, reconstructs from the
   sender's base, and clamps to its own, counting `clamped`. Without it, a range
   edit would become a dropped field, which would move a row of §4's evolution
   table. The cost is a few bytes in an announcement paid once. **Worth it?**

5. **Enums and unions ride by NAME, not at the packet wire's width.** "Values at
   their declared widths as the packet wire writes them" would make an enum its
   declaration ORDINAL, and a variant inserted mid-enum would then read as a
   different variant on the two builds, in silence, which is the class this wire
   exists to prevent. So an enum value and a union arm are references at
   `bits_required(0, E)`. Cost on this unit: 5 bits where the packet wire spends
   **3**, since a four-variant enum is ranged over `[0, 4]` with `None` a value,
   so **the residual is two bits**. **The alternative that keeps both** is to
   announce each enum's variant name ids in declaration order inside the enum's
   shape and ride the ordinal, which costs 3 bits here and a second resolution
   path. Not taken, on land-and-expand, and now registered in §15 with the two
   bits it would remove.

6. **Damage is TERMINAL for the batch.** A bitpacked body carries no length, so
   §3's "the parent reads on past `L`" has nothing to stand on. The fields
   decoded before the damage stand, one `malformed` counts, and nothing after is
   read, which is the packet reader's own answer. **This moves §4's framing
   damage row for this form**, and §4 now says so. Every EVOLUTION row still
   stands, and an unknown field is still stepped over exactly, because the
   announcement gave the reader its width. **This is the one real tolerance
   regression in the design.**

7. **The batch count is a ranged integer over `[1, 256]`, 8 bits, and 256 is a
   WIRE CONSTANT.** The width depends on the bound, so the bound cannot be a
   receiver's policy the way the announcement's entry bound is. 256 keeps a batch
   of one at one byte and is generous for one call's worth of messages. A batch
   of zero is not spellable. **Is 256 the right constant, and should the count be
   there at all for a batch of one?** Dropping the form byte for a batch would
   save another byte and would cost the refuse-a-form-you-do-not-know property.

8. **`string(N)` and any array of element kind `6` ALIGN before their payload,
   and `wstring(N)` does not.** The align costs at most 7 bits and buys a
   `memcpy` on the largest payload on the wire, which is the "not massively
   slower" half of the cost rule. It makes `[..N]uint8` align where the packet
   wire does not, a one-rule divergence taken over a per-entry flag.

9. **The node table is the FIRST field of a form-`2` root body.** A pointer index
   is `bits_required(0, node count)` bits, so the count has to be read first.
   This is the only ordering constraint on a body and it inverts §3.1's
   file-form note that the node table is written last. §3.1 now says so.

10. **Form byte `2` is kept and the byte body is replaced in place.** No tagged
    release carries the byte body (`git merge-base --is-ancestor ddd9c2d0 v2.4.0`
    is false), no port ever grew a message verb, and the only readers are the
    reference and the tool in this tree. Taking `3` would leave `2` as a form no
    writer writes, carried or refused by name in nine languages forever.

11. **The singular verbs are removed, not kept beside the plural ones.**
    `MeasureMessage`, `SaveMessage` and `LoadMessage` leave §11's claimed set and
    `MeasureMessages`, `SaveMessages` and `LoadMessages` take their places, so the
    41 and 57 counts do not move.

12. **`E` for `backenddemo` is 28, not 29.** The announcement's two reserved ids
    are its own transport, live in its trailer, and take no vocabulary slot, so
    slot `1` is the first entry of the closure.

13. **A batch is of ONE root.** Mixed roots go through a union root, which keeps
    `<X>SaveMessages` well typed.

14. **The escape kind `31` on a bitpacked body is align, a 32-bit `L`, then `L`
    bytes.** It is the one kind with no declared bound, so it is the one place a
    width does not come from a declaration. It is also **the only path a later
    major has on this form**, because an announcement carrying a kind outside the
    closed set is refused whole.

## The sixteen named changes from the cold read, and where each landed

1. **The `flags` row.** A mask rides at its DECLARED W bits on this form, not as
   kind `9`'s raw `uint64`, which is the packet wire's rule taken for the packet
   wire's reason. **§4's mask ROUND-TRIP row is named as MOVED for the message
   form**: appended bits survive a file round trip and not a message one. Sixty
   four bits for a three-bit mask is the largest fixed overspend a small message
   has. That makes **two** rows of §4 this form moves, framing damage and this,
   and §3.3, §4 and VERSIONING.md all say so.
2. **The node table's bitpacked framing, stated exactly.** A thirty-two bit
   count, a record as a TYPE REFERENCE and a body ending at its own zero
   reference with NO length, a blob record's length at thirty-two bits then an
   align then the bytes, and a root reaching no node eliding the field like every
   other empty thing. §3.1's kind-`12`-with-an-`L` framing does not carry here.
3. **Unbounded arrays, maps and blob nodes are NOT refused on form 2.** Each
   carries THIRTY-TWO RAW BITS, announced as `min` `0` and `max` `2^32 - 1`, so
   the width falls out of the existing formula with no rule of its own. Refusing
   them would have made the message form a subset of the language. The cost is
   four bytes where a bounded array spends a handful of bits, which is what a
   declared bound buys.
4. **The `wstring` width is named as a departure.** Sixteen bits a code unit
   against SPEC.md §4.12's 32-bit group, because the group is the packet codec's
   word orientation showing through and a bit stream has no word to fill.
5. **The enum cost is corrected.** A four-variant enum is ranged `[0, 4]`, three
   bits on the packet wire, so the residual is **two** bits and not three.
6. **USAGE says LEVEL WITH proto3** for `client_build uint32 | max = 65535` at 49
   bytes, and states the other bound: `price_minor uint32 | max = 9999` puts
   `StorePurchase` at 39, under proto3's 40.
7. **`bits(N)` has a shape**: RANGED, `base` `0`, `bits` N, in the shape table,
   beside `flags` at RANGED, `base` `0`, `bits` W.
8. **The batch surface answers five questions.** `M` above 256 on the write side
   **REFUSES** by name at `MeasureMessages` and `SaveMessages`, returning `-1`,
   rather than concatenating batches into one buffer, because that would invent a
   second framing level and break one-batch-per-datagram silently. `M` above the
   caller's capacity on the read side refuses by name before a body is decoded.
   Both are `batch_too_large`, one name for two bounds on `vocabulary_too_large`'s
   precedent. Damage inside body `k` delivers bodies 1 to `k - 1` with the
   returned count stating `k - 1` and one `malformed`. Bytes after the pad are
   malformed as in the file form. **A pointered batch takes ONE region for the
   batch**, sized by `LoadMeasure`'s message overload so no new verb is claimed,
   with each body keeping its own numbering inside it, and USAGE shows the
   parameters.
9. **Alignment is relative to the batch buffer's byte `0`**, one sentence, since
   a body does not begin on a byte boundary.
10. **"Closes the connection" is gone.** The library returns a refusal and
    closing the connection is the application's act, in §3.3, USAGE.md and
    VERSIONING.md. A refused announcement sets no vocabulary, so the refusal is
    safe to hold rather than urgent to act on.
11. **§15 gains the ENUM-ORDINAL alternative** with the two bits it would remove
    and what it costs, and **the varint follow-on gains the owner's negative
    case**: a random 64-bit `player_id` is TEN varint bytes against EIGHT raw, so
    a varint loses on exactly the id-shaped values a backend message is full of.
    A **2-bit width class** is named as the shape to measure if it is ever taken.
12. **The plural spellings reach test/conformance/README.md and docs/PORTING.md.**
13. **Three rules stated that a codec would otherwise guess.** A variant or arm
    reference naming an entry of the wrong sort (a variant on a non-kind-`0`
    entry, an arm on a kind-`0` one) is MALFORMED, because the entry resolved and
    contradicts its position. A ranged offset above the SENDER's own `max`
    reconstructs and clamps to the reader's bound rather than being damage,
    because the position after it is known. An over-long array CLAMPS BY WALKING
    the surplus, arithmetic for a fixed-width element and element by element for
    a non-fixed-width one, counting one `clamped`.
14. **§3.1's nested-body reserved-id sentence says its recovery half is the file
    form's**, where §4 says it of the framing-damage row.
15. **The escape kind is the only path a later major has on this form**, one
    sentence, since an announcement carrying an unknown kind is refused whole.
16. **`schema unpack` on an announcement prints the vocabulary decoded**, slot,
    id, name, kind and shape, stated as owed by the codec PR.

## One code change, and why a spec-only PR has it

`internal/check/check.go`'s `tableGeneratedVerbs` moves the three message
suffixes to their plural spellings, because `TestSpecSection11EqualsTheChecker`
parses §11's fenced list and holds it to that variable in both directions. A
page edit alone would red the gate. The gate is green:

```
$ go test ./internal/check/ -run TestSpecSection11EqualsTheChecker
ok
$ go test ./compiler/ ./internal/goldens/ ./internal/format/
ok  ok  ok
```

`internal/codegen/cpptable/pointers.go` still emits `LoadMessage`, which is now
an unclaimed spelling. It is the codec PR's to move, along with the generated
headers under `generated/`.

## What is owed after this

- the codec: the reference's bit reader and writer on the tolerant path, the
  corpus re-pin, the controls red (schema#523)
- the measurement the cost rule asks for: read and write against the byte body,
  held to the FACTOR OF TWO now on the page, with the ratios, the machine, the
  build and the date recorded beside it. Above two on either path, the bitpacked
  body reopens
- yojimbo#344, after the codec lands
- USAGE.md's message-form section moved with the page, since a page that
  contradicts its own tutorial is worse than one that is silent

### DIFF
--- a/docs/PORTING.md
+++ b/docs/PORTING.md
@@ -786,17 +786,24 @@ instances agree byte for byte between the two, and the wire fuzzer runs 103,286
 enumerated mutants over 112 seeds with no divergence, plain and sanitized.
 
 **THE MESSAGE FORM IS THIS MILESTONE's FOLLOW-ON** (docs/SPEC-TABLES.md §3.3).
-Form byte `2` is the same body with its id table moved one level up, to the
-CONNECTION: a peer announces its unit's whole vocabulary once a direction as an
-ordinary form-`1` file, and every message after it is the form byte and the
-root body alone. What a port owes is `LoadMessage`, `MeasureMessage` and
-`SaveMessage` beside the file form's three, `TableVocabulary` and the three
-unit-scope entry points `Announce`, `AnnounceMeasure` and `AnnounceRead`, each
-in that language's own naming convention. C++ and the tool carry it today and
+Form byte `2` moves the id table one level up, to the CONNECTION, and BITPACKS
+what is left: a peer announces its unit's whole vocabulary once a direction as
+an ordinary form-`1` file, and every BATCH after it is a form byte, a body count
+and the bodies as one continuous bit stream. What a port owes is
+`LoadMessages`, `MeasureMessages` and `SaveMessages` beside the file form's
+three, the message overload of `LoadMeasure` for a pointered batch's one region,
+`TableVocabulary` and the three unit-scope entry points `Announce`,
+`AnnounceMeasure` and `AnnounceRead`, each in that language's own naming
+convention. **The verbs are PLURAL** because the form's primitive is a batch of
+bodies of one root and a single message is the batch of one. C++ and the tool
+carry a form-`2` path today, byte framed until the codec change lands §3.3, and
 the harness's `message` surface prints ABSENT for every port, so the cell is
-where the work is counted. The form adds no rule to the body: the references
-resolve against the announced table instead of a trailer, and every framing,
-tolerance and malformed rule above is unchanged.
+where the work is counted. The BODY's rules are the ones a port already has,
+read off a bit stream instead of a byte one: references resolve against the
+announced vocabulary instead of a trailer, elision and every tolerance rule
+above are unchanged, and the two rules that DO move are named on the page,
+damage being terminal for a batch and a `flags` mask riding at its declared
+width.
 
 **Measured effect.** 0.98x the corpus's bytes without its 210 KB blob and 0.99x
 over the whole of it, against 1.80x over the tiny class: an empty table is ten
--- a/docs/SPEC-TABLES.md
+++ b/docs/SPEC-TABLES.md
@@ -4474,18 +4474,23 @@ batch.** The fields decoded before it stand, ONE `malformed` counts, and nothing
 after it is read, which is the packet wire's own answer (SPEC.md §4.3) reached
 for the same reason.
 
-The three ways a batch is damaged, at bit granularity:
+The four ways a batch is damaged, at bit granularity:
 
 - **A REFERENCE ABOVE THE ENTRY COUNT `E`.** The width can spell values above
   `E` whenever `E` is not one less than a power of two, and every one of them is
   damage. So is a reference of `0` where an entry is REQUIRED, which is an
   enum-keyed array's slot key and a node record's type id (§3.1, §3.2).
+- **A REFERENCE NAMING AN ENTRY OF THE WRONG SORT**, which is a variant
+  reference on an entry that is not kind `0` and an arm reference on one that
+  is (above). The entry resolved and it contradicts the position it was used
+  in, so the next bit's meaning is what is in doubt.
 - **A LENGTH, COUNT OR INDEX WHOSE PAYLOAD RUNS PAST THE BATCH.** A string
   length, an array count, a node index or an escape's `L` that would read beyond
   the batch's bits is damage at the field that carried it.
-- **EXHAUSTION.** The bits run out inside a field, or the count promised `M`
-  bodies and fewer terminators arrive, or the trailing pad to the byte boundary
-  is not zero.
+- **EXHAUSTION, AND ITS OPPOSITE.** The bits run out inside a field, or the
+  count promised `M` bodies and fewer terminators arrive, or the trailing pad to
+  the byte boundary is not zero, or BYTES REMAIN AFTER THE PAD, which is §3's
+  rule for a file with bytes left over met on a batch.
 
 **ILL-FORMED TEXT IS DAMAGE HERE TOO, and it is terminal like the rest.** §3's
 one content rule holds unchanged in what it rejects, a kind `12` payload that is
@@ -4903,10 +4908,10 @@ entries announces about 5 KB once.
   boundary are not zero, and a buffer carrying a whole batch and then a byte
   more. Red if a leg reads either clean.
 - **The batch's five answers.** A `SaveMessages` and a `MeasureMessages` of 257
-  bodies, each refusing by name and writing nothing; a `LoadMessages` of a
+  bodies, each refusing by name and writing nothing. A `LoadMessages` of a
   256-body batch into storage for eight, refusing by name with no counter moved
-  and nothing decoded; a three-body batch damaged inside the second, whose
-  returned count must be one; and a pointered batch measured once and loaded
+  and nothing decoded. A three-body batch damaged inside the second, whose
+  returned count must be one. And a pointered batch measured once and loaded
   into ONE region. Red if a leg writes consecutive batches, decodes a body before
   refusing on capacity, returns two or three for the damaged batch, or asks for a
   region a body.
@@ -6868,8 +6873,8 @@ the wire, and keeps the flexibility that comes with it.
 
   **It is not the MESSAGE FORM's vocabulary, though, and the two are worth
   telling apart.** §3.3's `no_vocabulary`, `second_announcement`,
-  `vocabulary_too_large` and `message_form_as_file`, and the form byte's own
-  `newer_form`, ride on the message path and are stated there. A caller
+  `vocabulary_too_large`, `batch_too_large` and `message_form_as_file`, and the
+  form byte's own `newer_form`, ride on the message path and are stated there. A caller
   meeting a `TableRefuseReason` has been refused a FILE, by a header match or by
   a measure, and falls back or gives up; a caller meeting one of the message
   form's has been refused a MESSAGE on a connection, which is a different
--- a/docs/SPEC-TABLES.md
+++ b/docs/SPEC-TABLES.md
@@ -4027,8 +4027,12 @@ with one message passes one and pays a count of eight bits for it.
 
 - **`M` ABOVE 256 ON THE WRITE SIDE IS A REFUSAL BY NAME.** `MeasureMessages`
   and `SaveMessages` both refuse and return `-1`, and neither writes consecutive
-  batches into one buffer. The refusal is learned at MEASURE time, before a
-  buffer is allocated, which is the point of measuring first. Concatenating
+  batches into one buffer. **The `-1` carries its reason on the carrier
+  `LoadMeasure` already uses**: the `TableRefuseReason` enum out-parameter
+  (§7, §11), never an exception, so the write side and the read side answer a
+  refusal one way and neither is a special case. The refusal is learned at
+  MEASURE time, before a buffer is allocated, which is the point of measuring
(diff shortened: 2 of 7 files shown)


gafferongames:
Rulings from the cold read, under the owner's word: (A) the terminator reading stands as the page has it, each body ends at its own zero reference and the batch adds none. (B) not massively slower means a factor of two on the matching read and write path against the byte body, measured later at a named sitting, above which the bitpacked body reopens. (C) flags ride at their declared W bits on this form and §4's round-trip row is named as moved. The sixteen named changes from the read land on this branch before it leaves draft; every arithmetic row reproduced bit by bit.

Issue #604, opened 2026-09-06
Retain-unknown: the page contradicts itself in three places, settled before the reference builds it (#525)


gafferongames:
The #525 builder stopped before code on 2026-09-06 because SPEC-TABLES.md §6.6 cannot be implemented exactly as written. Spec first: these are settled on the page, cold-read, before the reference PR. Rowan's rulings under the reduce-cases law, for the owner to move if he sees otherwise:

1. **How many excluded classes.** §6.6's prose says five, its own table lists six, the flight brief said seven, and the HELD-BY-TEST list splits the first row three ways for eight cases. Ruling: the table is the law; the prose says six; the test list names each of the six once, one retain_lost pinned per class, and the pointer exclusion is recursive with the whole record rejected and one retain_lost counted (#575's ruling stands).

2. **TableIds cannot hold a retained id.** Its capacity is a compile-time constant (the distinct names the closure can spell, so a save allocates nothing), and a retained id is by definition not one of them; today ref() sets overflow and the save answers -1 for the whole file, so one retained field would refuse the save rather than count one retain_lost. That contradicts "retention allocates nothing, in any port." Ruling: the retention tail carries its own id list in caller-owned storage, declared with a capacity beside the retained records themselves (the same shape as the message form's resolved vocabulary: the caller places it, the codec never allocates), the fixed TableIds untouched; a retained id past the caller's capacity counts one retain_lost and the record is dropped, never a refused save. The page states the storage, the capacity, and the constant a caller can size exactly by.

3. **The path step.** The page wants one integer per step, an array element's step to be its index and a union's step to be the arm ordinal; with data-driven counts (maps, unbounded arrays) one flat ordinal per body and an arm-indexed one cannot both hold. Ruling: a step is a pair, the field ordinal in its body and the element index (zero for a scalar, the arm ordinal for a union, the key's slot for a map), computable locally on both sides; the page defines it once and the retention form encodes the pair.

4. **Fixed-class roots.** "A region round trip and only that" implies a fixed-class root (no region, no node directory) gets no retention. Ruling: stated plainly on the page; LoadRetain on a fixed-class root is refused by name; the conformance rows live on pointered units.

[...]

PR #615, opened 2026-09-06
tables: retain-unknown in the C++ reference (#525)

gafferongames:
Retain-unknown in the C++ reference, from docs/SPEC-TABLES.md §6.6 as it was
settled by #604 and landed in #608. Closes the last of the original three C++
table-wire rows.

## What lands

**A second, parallel family of body functions.** `Load`, `Measure` and `Save`
are not touched: the reference gains `<T>LoadBodyRetain`,
`<T>MeasureBodyRetain` and `<T>SaveBodyFieldsRetain` beside them, with the PATH
threaded at every child-body descent, and `<T>LoadRetain`, `<T>MeasureRetain`
and `<T>SaveRetain` on every variable-class root. A caller that does not ask
instantiates none of it.

**The caller's two stores.** `TableRetain` is a byte buffer, a capacity, an id
list, an id capacity and what has been used of each. Retention allocates
nothing, in any path, and the test holds that with an `operator new` audit.

**The retained tail, and the merged trailer.** A retained field goes back at
the END of its own body, in the order retained, and in the root body BEFORE the
node-table field. The generated id table is untouched: its capacity, its
overflow rule and its `-1` stand exactly as they are, and a retained id takes
its entry from the caller's list, the two stores numbered into ONE trailer in
merged first-use order.

**The resolving walk, both ways.** Every reference in a retained record is the
sixty-four-bit id it names and every length that frames one is recomputed. The
walk reads kind bytes, lengths and references and nothing else, and anything it
cannot frame drops the record, counts one `retain_lost` and never raises
`malformed` on the plain read.

**The six excluded classes, one row each**, and the two refusals: a fixed-class
root and a form 2 write, each a compile error that names itself.

### Load, Measure and Save are byte-identical

The control compares every committed C++ table golden against what this
compiler emits, and asks whether the old text is a strict SUBSEQUENCE of the
new, so every existing line is present, in order, with nothing moved and
nothing changed:

```
C++ table goldens compared against origin/main: 73
  byte-identical: 42
  pure insertion: 31
  a line moved or changed: 0
```

The 42 that did not move at all are the units with no variable table, which
carry not one byte of retention: the zero-cost gate (§2.2) is why, and the
fixed-class refusal is emitted where retention is rather than everywhere.

**And every pinned WIRE byte string is identical.** `make update-goldens`
rewrites `testdata/wire/` from the reference's own writers, and after this
change `git diff origin/main -- testdata/wire` is empty. That is the half of
the claim that matters: `Load`, `Measure` and `Save` write what they always
wrote, byte for byte, on every instance the corpus pins.

### Every gate the page names, with its negative control

| gate | red when |
|---|---|
| the round trip at eight depths | a field is lost, duplicated, or placed in another body |
| idempotence | the second save is not the first, byte for byte |
| six excluded classes, `retain_lost` at one each | a class is retained, or the counter lands anywhere but one |
| a node record whose type id is unnameable | it is counted anywhere but once |
| a buffer one byte short of the last record | a read counter moves, or the truncated record is written |
| an id list one entry short | the save is refused, or a record rides without its id |
| record lifetime | `7` is resurrected beside `known = 2`, or the discard is counted |
| damage inside sound outer framing | the read reports damage it did not see |
| the permuted trailer | the two id tables are the same sequence |
| the fixed-class root | `LoadRetain` compiles, or the refusal is a missing symbol |
| a form 2 write | `SaveRetainMessages` compiles, or the refusal is a missing symbol |

The two refusals are `make tables-retain-fixed-class-negative-control` and
`make tables-retain-message-form-negative-control`. Each compiles a call and
requires the compile to FAIL with its own sentence in the diagnostic. A
refusal that were a missing symbol would be a linker error with no reason in
it, which is exactly what §11's rule for a surface a class does not carry
refuses.

## The cost, measured

**The tables read and write benchmark, before and after, one sitting on an M-series
Air, `bench/tables/cpp/leg` at `-O3 -DNDEBUG`.** Three binaries, rotated so no
one of them sits at the same place in the sitting's own drift. `before` is
main at 14727765, `after` is this branch, and `ctl` is this branch with the two
report counters deleted from the header by hand, which is the only line of this
change the bench unit sees at all.

Round trip, M msg/s, five rounds:

| | 1 | 2 | 3 | 4 | 5 | median |
|---|---|---|---|---|---|---|
| before | 0.248 | 0.260 | 0.261 | 0.260 | 0.262 | **0.260** |
| after | 0.259 | 0.244 | 0.257 | 0.258 | 0.255 | **0.257** |
| ctl (no counters) | 0.262 | 0.252 | 0.258 | 0.260 | 0.259 | **0.259** |

Every difference between the three is inside each binary's OWN spread in that
sitting: `before` alone ranges 0.248 to 0.262, 5.6%. The control is the reading
that settles it, because it differs from `after` by exactly the two counters
and nothing else, and it lands 0.8% away in the same band. **No measured cost.**

The write row is untouched by construction: nothing on the write path changed
in a unit with no variable table, and the bench unit is one. `before write
0.892 / 0.887 / 0.888` against `after write 0.887 / 0.888 / 0.867` in the
paired sitting, which drifts downward through the sitting for both.

**And the retain family's own cost, stated as a factor, because a caller
choosing it should know.** One RT2 wire of 574 bytes carrying 24 retained
records, through RT1, `-O3 -DNDEBUG`:

```
Load         1393.2 ns/op
LoadRetain   2652.9 ns/op   x1.90
Save          954.6 ns/op
SaveRetain   5526.0 ns/op   x5.79
```

The load pays the resolving walk twice per record, once to size the resolved
form and once to write it. The save pays more: it walks the record list at
every body it visits to find the records that body owns, and each record it
places is walked twice again for the same reason the plain save walks a nested
body twice, because a canonical LEB128 length cannot be patched in place. Both
factors scale with the RECORD COUNT the caller admits, which is what the two
capacities bound.

## Silences the page left, and what this PR decided

Each of these is a place §6.6 does not reach the case. None was decided
quietly.

1. **The reserved node-table field's row cannot be built.** §6.6 lists it as an
   excluded class whose conformance row must sit on a POINTERED unit and pin
   `retain_lost` at one, and its own "why not" says a build with no kind `17`
   counts it one `unknown`, which is a FIXED-class build, and §6.6 denies
   those retention. On a pointered unit the id is the transport the reader
   consumes and is never unknown. The rule is implemented, since the retain
   family's unknown arm counts one `retain_lost` for any of the three reserved
   ids and captures none of them, and no row is in the corpus, because no wire this
   port can read produces one.
2. **`LoadRetain`'s signature.** §6.6's sample spells
   `SceneLoadRetain( scene, region, region_size, wire, wire_size, &retain, &report )`
   with a first argument the sample never declares, and `Load` takes no value
   and returns the root. The reference follows `Load`'s own shape with the two
   extra arguments, and `MeasureRetain` and `SaveRetain` are the sample's own.
3. **The resolving walk's nesting is CAPPED at 64.** A retained record's inner
   nesting is the WRITER's and is the one depth on this path a file can drive,
   and §6.6 bounds the PATH and says nothing about the walk. A record that nests
   past the cap is dropped with one `retain_lost`, on the same rule as anything
   else the walk cannot frame.
4. **Two more shapes are damage.** A reference at an id-table entry of ZERO,
   where zero is the wire's own no-id, the enum's `None` and the union's empty
   arm, and the resolved form spells that as the id zero. And a reference at one of
   the three RESERVED ids, which re-emitted into a nested body would be
   malformed. Each drops the record and counts one.
5. **An array of UNIONS takes two steps.** §6.6 gives one convention for an
   array's element index and another for a union's arm ordinal, and this
   construct needs both. The element is one step and the arm is the next, and
   the two carry the same field ordinal because no body with a field list of
   its own sits between them.
6. **Eleven suffixes, not three.** §6.6 owes §11 `LoadRetain`, `MeasureRetain`
   and `SaveRetain`. The reference emits eight more, the body family, the wire
   pair, the node dispatch and `SaveRetainMessages`, and §11's own rule is
   that its lists are the checker's claim entire, so all eleven are stated and
   claimed.
7. **`TableRetain` in the DART registry waits for the Dart port.** §6.6 asks
   for it now. That registry says what a backend DEFINES and a gate holds it to
   that, so registering a name Dart does not define would take the gate away.
   The three MEMBER spellings, which are field-name refusals with no such gate,
   are claimed now and take Dart's claimed verbs from nine to twelve.
8. **The fixed-class refusal is emitted where retention is.** A unit whose
   tables are all fixed-class carries none of the machinery, which is §2.2's
   zero-cost gate, so there the name is free rather than refused. The CHECKER
   claims the suffix in every unit either way, which is what protects a
   declaration from taking it.
9. **`retain_lost` at save is counted ONCE, after the save.** `MeasureRetain`
   takes no report and counts nothing. The save marks each record it places and
   counts what it did not. A body measured twice would otherwise count twice.
10. **A record's node is matched by ADDRESS at save.** The record carries the
    directory INDEX the page defines, and the save resolves that index through the
    region's directory in one add and compares addresses, so neither side ever
    searches a numbering.

## What is still owed

Named in §6.6's own status line, which this PR rewrites to the present state
rather than deleting:

- **the eight ports.** Retention is the C++ reference's alone today.
- **`internal/tablewire`'s retention, and the fuzzer's retention leg** (§4.2).
  The leg is not buildable until the oracle retains, and the oracle is a third
  reading of §3 written from the page: that is what the leg costs rather than a
  detail of it. The fuzzer still runs with retention OFF and its requirement is
  the one it has always had.
- **the MESSAGE form's `LoadRetain`** (§3.3). The form 2 WRITE refuses by name
  and is held by a control. The form 2 READ is not built.
- **the conformance manifest's `retain` surface.** The rows live in the
  reference's own gate for now, because the manifest names no language and a
  surface only the reference answers is one the eight ports would each have to
  say absent to. The surface lands with the second port.

Refs #525. The page is docs/SPEC-TABLES.md §6.6, settled by #604 and landed as
2fcc2ee7 in #608.


gafferongames:
`make test` whole, green, on the branch at 4ecc12fe (macOS, this Air):

```
MAKE TEST EXIT 0
```

The retain gate inside it, plain and under ASan/UBSan, and both refusals red by name:

```
./build/schema_test_retain
retain: ok
./build/schema_test_retain_asan
retain: ok
the fixed-class root refuses retention BY NAME (docs/SPEC-TABLES.md §6.6)
retention writing form 2 refuses BY NAME (docs/SPEC-TABLES.md §3.3)
```

Every CI check on the pull request passes, big-endian and msvc included.

gafferongames:
Not merged at the 2026-09-06 wrap-up: every item of the cold read is landed on the branch (the linear walk with its bound on the shape gate's ledger, the resolution row, the reserved branch deleted, the discard scoped, the page sentences, the measured factor, the regenerate over #616) at twenty green, and the branch went CONFLICTING against main the moment #617 (wstring on the table wire) merged, since both re-pinned the C++ table goldens. The next session merges main into impl-retain, regenerates goldens from the merged compiler rather than hand-merging, confirms twenty green, and lands it. Its honest state on landing is partial, so the cell on #366 stays ❌ with the note until the fuzzer leg and the oracle's retention exist.

Issue #589, opened 2026-09-06
Packet wire: string, bytes and flags defaults in the eight ports (#396, C++ landed in #576)


gafferongames:
Owner, 2026-09-05: the packet wire is fixed for all nine currently supported languages after the C++ column is complete and before the table-wire sweep. This row is one of four red packet-wire rows in the nine (#366's packet section): the value defaults for string(N), bytes(N) and flags fields ride in C++ only; compiler/builtin.go's refuseUnported turns away a unit carrying one in the other eight.

The work, mirrored from the C++ reference and SPEC.md §4.2: each port constructs a field at its declared default (the string bytes, the byte block, the flags mask), writes and reads it as an ordinary packet value with no default-based elision, and the refusal in refuseUnported is deleted for that target. The default rows in the conformance corpus (testdata/conformance/tables/MANIFEST.txt, w1_fleet/default) are TABLE rows and prove nothing about this packet row (Stella's correction, 2026-09-05). The first port therefore adds the packet oracle: an isolated packet fixture carrying the three defaults, its C++ packet goldens pinned including consumed size, a twin fixture without defaults initialized to the same values writing the same bytes (so accidental elision is caught), and the read overlay proving received values replace defaults under the used-prefix rule. examples/Wire.schema is not touched until every port carries the row, since adding defaults there breaks the other ports' generation. Each later port pins to the same oracle. The acceptance gate splits by wire: a target may carry packet defaults while its table backend does not, so refuseValueDefaults admits a packet-only unit for that target and keeps the named refusal for any default in a table closure (ir.TableClosure is the walk; a packet type reached through a table, array or union counts), and the page states the split in one sentence in the same PR. Eight targets, one PR per target or one PR for all eight, the builder's call stated in the PR body; each with its negative control red (revert the default's birth in one target and watch its leg go red by name). No em dashes in prose. American spelling.

Cells that move: the row "string, bytes and flags defaults (#396)" in the packet section, eight columns.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
Owner, 2026-09-06: "Can you ask her to please fix the packet wire X's for the original batch of languages, so all 9 first languages have all features." The packet-wire sweep across the nine is Stella's, starting now rather than after the C++ column; the four rows in the ruled order (#503, #589, #519, #590), one PR per language after the first row, each cold-read and merged on green by Rowan. Briefs at rowan-new/cairns/59d93707-briefs/.

[...]

PR #633, opened 2026-09-07
Support packet string, bytes and flags defaults in Rust

gafferongames:
Rust packet types now construct string, bytes and flags fields at their declared defaults. `new()` copies literal bytes into zeroed backing buffers; `Default` remains the zero form. The existing unsigned flags representation already carries the full 64-bit mask. Packet support is registered independently, and defaults reached through a table closure remain refused.

This is the Rust-only follow-on to #630 for #589, stacked on its C branch until Rowan merges it. One PR per language. It adds no wire elision and changes no existing generated or wire golden.

Validation: all Go tests (including the seeded fuzz corpus), `go vet ./...`, the existing Rust packet and 128-bit runtime legs, and Clippy on the new consumer pass. The eight C++ byte/consumed-bit pins pass in both debug and release; coverage includes a defaultless twin, fixed/counted backing arrays, explicit-empty and short defaults, independent 33/64-bit masks, reused prefixes and repeated union selection. Removing only constructor byte copies through `tools/sabotage` builds successfully and fails the exact constructor-bytes marker. The positive and negative checks are attached to `test-rust` in full certification.

The full nine-language local suite is pending the remaining toolchain setup; full branch certification and Rowan's independent read are still required before merge. An initial invocation of the existing Rust test from the repository root could not locate its relative goldens; it was rerun successfully from its required `test/rust` directory.


rowan-claude:
Cold read of the defaults row (#633 to #638) at these heads, by a Fable reader with none of the author's reasoning, in its own clone; the Rust leg run, plus the C++, C, Go and Rust consumers and their controls re-proved at the final head 87e5b8ec under #636's shared-fixture change (ChoiceEnvelope); C#, Java, JS, Dart and Elixir read, their blades verified mechanically with tools/sabotage, their runtime assertions unverified there. Each PR touches one emitter, its leg, its docs and its ROADMAP cell; `git diff df14c50e..87e5b8ec -- testdata/wire` is empty, so every port reads the bytes #630 pinned; the compile-time refusals live once in the front end (internal/check/check.go:1626, :1630, :4001, :4005) and no port adds its own; ROADMAP climbs 318 to 323. Every PR is mergeable as it stands. Optional: compiler/packetvaluedefaults_test.go's else branch expecting a refusal for cpp and go is now unreachable for every built-in target. Merge order bottom-up once #630 lands, each PR retargeted to main before the branch under it is deleted; #636 changes the fixture the earlier consumers compile against, so the order is not reorderable.

Issue #519, opened 2026-09-04
string(N): read-side UTF-8 refusal in every target, per the 2026-09-04 ruling


gafferongames:
Owner ruling, 2026-09-04: the asymmetry between `string(N)` and the reference resolves toward the reference. `string(N)` gains a read-side refusal of malformed UTF-8 in every target, mirroring classic `serialize_string` exactly as serialize's STANDARD.md states it, and `wstring(N)` keeps its read-side refusal of unpaired surrogates. The spec half lands in #510 (SPEC §4.7 rewritten, §4.12 states the symmetry, SECURITY.md carries the refusal class, USAGE.md's writer-contract paragraph replaced). This issue is the implementation half.

Before this ruling, §4.7 made UTF-8 well-formedness a writer contract, checked by a debug-only assert in C, C++ and Rust and by nothing at all in the other six. That is the exact class of cross-language acceptance divergence SECURITY.md calls a bug: a payload C++ let through in release and Elixir would have had no opinion on. After the ruling there is one rule, on the read path, in every target and every build mode.

## The rule, verbatim from the page

A `string(N)` payload that is not well-formed UTF-8 fails the read. Not well formed means not well formed under Unicode Table 3-7, so overlong forms, surrogates encoded in UTF-8, code points above `10FFFF`, truncated sequences and the bytes `0xFE` and `0xFF` all fail. The refusal is terminal (SPEC §5). The interior-null rule stays exactly as it is, the stricter check beneath the encoding rule, because NUL is valid UTF-8.

## The work

**The C++ reference validator.** C and C++ already carry a generated UTF-8 validator on the write path behind an assert. It moves to the read path, unconditional, in every build mode, and it becomes the reference implementation the other eight are read against. It must be a straight Table 3-7 DFA over the used bytes with no allocation and no locale dependency, and it runs after the length bound and before the interior-null scan (the two can share one pass).

[...]

PR #639, opened 2026-09-07
C: refuse malformed UTF-8 on packet reads

gafferongames:
C packet readers now reject malformed UTF-8 in every build mode, before the existing interior-NUL scan. The former writer-only debug assertion is removed, matching the C++ reference and SPEC §4.7. This addresses C's part of #519 and follows defaults PR638 in the landing order.

A shared packet-text harness reads all 33 applicable serialize corpus rows directly, checks their declared verdicts, decoded bytes, consumed bits and canonical writer bytes against C++, then compares 496 single-bit mutations. C agrees in debug and optimized NDEBUG builds, including 192 mutated refusals. Removing only its read validation compiles successfully and makes the mutation-only sweep fail the exact verdict marker.

`go test ./...`, `go vet ./...`, ordinary/128-bit C packet tests and benchmark corpus wire checks pass. Deliberate generated-source changes and their three source goldens are included; no wire pins changed. Independent review is complete. The required C reproduction is recorded in bench/results/2026-09-07-c-utf8-studio.md with its CSV, automatic Studio samples and exact source hashes. All four rows passed seven measured runs, maximum spread 1.51%. The shared corpus close and diagnostic lint repairs are included here at the first reader PR; pinned golangci-lint v2.12.2 and modernize pass, and the C corpus/control checks pass after the repair. Current CI also exposes the separate TextTable.cpp golden mismatch inherited from main, reported to its table owner.


rowan-claude:
Cold read of the UTF-8 row (#639 to #646) at these heads, by a Fable reader with none of the author's reasoning, in its own clone; C, Rust and Go legs run, the other five read. The harness is C++-anchored: the C++ driver is checked against the 33 pinned corpus rows first, then each port's whole output line per row and per flip against C++, 496 flips, 192 refusals, reproduced identically in C, Rust and Go; every sabotage anchor matches exactly once and every run control goes red on the same case (string-golden/flip-15). Wire pins empty across all eight; source goldens move only under each port's own directory; ROADMAP climbs one cell per PR, 323 to 331.

MUST on this PR: the C packet emitter moves (internal/codegen/c/fields.go; a validator call on the read path of generated/bench/c/BenchWire.h) and bench/LOCK's standing gate asks for the reproduction rows (bench_mixed write and round_trip, bitpacker bits) re-pinned in the same PR with the sitting stated; `git diff 87e5b8ec 383d7aff --stat -- bench/` is empty. The predecessor row did it in bench/results/2026-09-07-c-defaults-studio.md; a Studio sitting with its provenance stated is admitted.

MUST on #646 (the last of the row): docs/SPEC.md:1441-1450, the §4.7 status paragraph ("ONE TARGET ENFORCES IT ON READ TODAY ... this line is deleted by the last of them"), is untouched by the whole stack; #646 deletes it. Each earlier PR should narrow it, or #646 may take the whole change.

Per PR: #639 after the sitting; #640 Rust, #641 Go, #642 C#, #643 Java, #644 JS, #645 Dart mergeable as they stand (#642's test/packet-text/cs/.gitignore rides in #643, harmless in merge order); #646 after the page deletion. Optional, informational: the ports run the UTF-8 check before the interior-null scan where the C++ reference runs the null scan first; both refusals are terminal so verdicts agree. Merge order bottom-up once #638 lands, retargeting each PR to main before the branch under it is deleted.

PR #646, opened 2026-09-07
Elixir: refuse malformed UTF-8 on packet reads

gafferongames:
Elixir packet reads now reject malformed UTF-8 with `:error`. `String.valid?` validates the received binary before the interior-NUL scan. This completes the eight UTF-8 ports for #519 on the implementation stack, following Dart PR645.

The generated reader agrees with C++ on 33 pinned corpus rows and 496 bit flips, including decoded values, exact-bit/one-bit-short bounds and canonical writes. Removing only UTF-8 validation compiles with warnings as errors, then makes mutation-only agreement fail its exact marker. Formatter checks, the nested defaults fixture, ordinary and 128-bit Elixir packet legs, `go test ./...` and `go vet ./...` pass. The earlier ordinary test that expected malformed UTF-8 to pass now requires `:error`. Generated files and source goldens are updated; wire pins stay unchanged.

Independent review is complete. The obsolete SPEC section 4.7 paragraph claiming only C++ validates UTF-8 reads is deleted in this final reader PR. Current CI also exposes the separate TextTable.cpp golden mismatch inherited from main; the narrow harness lint repair is on predecessor PR639 and must propagate through the landing sequence.


rowan-claude:
From the row read (posted whole on #639): this PR is the last of the UTF-8 row, and docs/SPEC.md:1441-1450, the §4.7 status paragraph, says it is deleted by the last target that moves the check; the whole stack leaves it untouched (`git diff 87e5b8ec 5ebc1405 -- docs/SPEC.md` is empty). Deleting it here is the one thing before merge.

Issue #590, opened 2026-09-06
Packet wire: wstring(N) in the eight ports (#188, C++ landed in #556)


gafferongames:
Owner, 2026-09-05: the packet wire is fixed for all nine currently supported languages after the C++ column is complete and before the table-wire sweep. wstring(N) (SPEC.md §4.12) rides the packet wire in C++ only since #556; the other eight refuse it by name in their target files, and the wide corpus under examples-wide/ is C++'s proving ground.

The work, mirrored from the C++ reference: each port carries wstring(N) on the packet wire in its own idiom (char16_t in C, the language's UTF-16 string in the others, the packet-wire form exactly as §4.12 states it), its read-side refusal of malformed UTF-16 (unpaired surrogates) as the page states, the wide corpus generated and its goldens bit-identical to C++, and the refusal in the target file deleted. The table-wire kind is #522 and is not this issue. One PR per target, each with its negative control red. No em dashes in prose. American spelling.

Cells that move: the row "wstring(N) (#188)" in the packet section, eight columns.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

gafferongames:
Owner, 2026-09-06: "Can you ask her to please fix the packet wire X's for the original batch of languages, so all 9 first languages have all features." The packet-wire sweep across the nine is Stella's, starting now rather than after the C++ column; the four rows in the ruled order (#503, #589, #519, #590), one PR per language after the first row, each cold-read and merged on green by Rowan. Briefs at rowan-new/cairns/59d93707-briefs/.

rowan-claude:
Closed: the eight ports landed today under Stella's name (#647 to #654); every column of the packet section carries the row. Main 7ccd06fd and after.

PR #647, opened 2026-09-07
C: carry wstring on the packet wire

gafferongames:
C packet fields now carry `wstring(N)` as a fixed `uint16_t[N + 1]` buffer and used length. The codec writes unaligned 32-bit groups, rejects invalid lengths/null units on write, and refuses oversized groups, nulls, unpaired surrogates and exhaustion on read. Successful reads supply a terminator and retain later unused storage. This addresses C's part of #590, following Elixir UTF-8 PR646 in the landing order.

The shared harness now supports wide text: all 26 applicable serialize corpus rows, the six independently packed interop cases, and 1,016 bit flips agree with C++ in debug and NDEBUG. ASan/UBSan checks writer bounds/nulls, read pairing, terminators, reused tails, unaligned conditionals, branch clearing, fixed/counted arrays, born counts and repeated union tags. Removing pairing compiles and makes mutation-only agreement fail its marker. Narrow UTF-8/defaults checks, ordinary/128-bit C legs, full Go tests and vet pass.

Packet source and protocol-id pins are added. The packet declaration is staged unchanged away from `examples-wide`'s table declaration and baseline. Table-closure guards include scalar union arms and recommend only a table-capable target; unrelated tables remain allowed. Existing source and wire pins are unchanged. Independent review is complete. The required C reproduction is recorded in bench/results/2026-09-07-c-wide-studio.md with its CSV, automatic Studio samples and exact source hashes. All four rows passed seven measured runs, maximum spread 1.59%. The wide bit-packer integer-range spelling is included here at the first wide reader PR; pinned modernize and the C wide corpus/control checks pass after it. The narrow harness lint repair is on predecessor PR639 and must propagate through landing. The separate TextTable.cpp golden mismatch inherited from main is reported to its table owner.


rowan-claude:
Cold read of the wide row (#647 to #654) at these heads, by a Fable reader with none of the author's reasoning, in its own clone; C, Rust and Go legs run, the other five read. The harness is C++-anchored: 26 corpus rows at buffer sizes 8 and 5 (9 accepted, 456 flip bits) plus the six SPEC interop values packed bit by bit independently (560 bits), 1,016 flips, the C++ oracle checked against the pins first and each port compared line for line including the canonical re-encoding. Wire pins empty for all eight pairs; the C++ emitter never moves; the packet proving-ground unit's id 0x53835dbf9f05e4d9 and examples' id unchanged base to tip; kind 33 stays C++ only with the five diagnostics cases passing; the terminator and tail behaviours match §4.12 (C and C++ write the zero unit, the seven write nothing past the used length); every sabotage replaces exactly one line and the run controls go red on wstring-accept-group-ffff/flip-16.

MUST on this PR: the C packet emitter moves (internal/codegen/c/c.go, fields.go, the new wstring.go) and bench/LOCK's standing gate asks for the reproduction rows (bench_mixed write and round_trip, bitpacker bits) re-pinned in the same PR with the sitting stated; `git diff --name-only 5ebc1405..d8e80c09 -- bench/` is empty. generated/ is byte-identical (additive TWString cases only), which is the mitigation for the owner's judgment.

Per PR: #647 after the sitting; #648 Rust, #649 Go mergeable as they stand (run); #650 C#, #651 Java, #652 JS, #653 Dart, #654 Elixir mergeable on reading. Optional on #654: SPEC §4.12 "All nine packet targets now carry" narrates; compiler/widetext.go:63-66's "no code generator carries wide text yet" branch is dead; TestWideTextIsRefusedByEveryOtherTarget's comment says "has not landed the codec" where it is now the closure refusal. Merge order bottom-up once #646 lands, retargeting each PR to main before the branch under it is deleted.

PR #670, opened 2026-09-07
Refresh six map table header goldens against integrated main

gafferongames:
The generated-code check on current main fails for six map table headers added with the expanded map-value corpus. This also blocks unrelated PRs, including #669: CI run 34109282889 reports CellsTable.h, ChunksTable.h, DocsTable.h, RunsTable.h, SlotsTable.h and SpansTable.h as moved.

Refresh only those six snapshots with the compiler from main dcdd5cdd. The generated differences carry the existing retained-payload handling for union/enum references, plus the existing message-form UTF-8 validation and code-point clamp for string keys in Cells and Docs. No emitter, schema, workflow or runtime code changes.

Validation: reproduced `make tables-block-zero-cost` failing on exactly these six files before the refresh; after it, the gate passes and reports all 117 table sources byte-identical to their pins. Reviewed the generated diff and passed `git diff --check`.

This is separate from #669's certification queue policy. Land this correction and carry current main into that PR before expecting its generated check to pass.

### DIFF
--- a/testdata/golden/tables/maps/CellsTable.h
+++ b/testdata/golden/tables/maps/CellsTable.h
@@ -3752,6 +3752,16 @@ inline int64_t TableRetainInContent( TableRetainIn & s, uint8_t kind, int64_t le
             }
             break;
         }
+        case 15: case 30:
+            // A UNION ARM AND AN ENUM'S VARIANT REFERENCE RESOLVE AS A FRAMED
+            // CONTENT TOO (§6.6): a kind 15 arm whose own payload is a union,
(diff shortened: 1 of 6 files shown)


rowan-claude:
Closed as superseded, with thanks for isolating it: #668 landed first and regenerated the same six map table headers against the unit's new vocabulary (the constants moved again with its three units), so main's block zero-cost gate is green at 123 sources byte-identical. This branch's six files were pinned against the compiler before #668 and differ from main's now; merging them would make main stale again. Nothing else in the diff.

PR #704, opened 2026-09-07
Complete Go schema table implementation

gafferongames:
Completes the Go schema-table backend: fixed and variable table files, announced message batches, builders and packing, blobs, lists, maps, union payloads, unknown-field retention, cook writing, block storage/fill, and the unit reflection registry. The Go conformance driver now covers all 18 registered surfaces with zero absent cases, and the mutation driver binds every ordinary, message, retaining and builder root in its applicable roster.

The public API includes typed error-returning measure variants and cook/block Open methods while preserving existing calls. Caller allocators own variable authoring and node-proportional scratch; Go 1.26.0 allocation tests account for the remaining typed activation frames. Packet-only reflection descriptors follow actual packet storage, with private descriptors where a table Row has a different layout. The fifteen-unit listing comparison verifies descriptor content; `TestUnitViewPacketStorage` separately verifies native packet offsets, which the listings do not contain. USAGE contains a compiled example.

The independent oracle and C++ comparisons exposed shared framing/widening and repeated-field retention defects and three final Go defects: malformed message sizing stopped early, counted-array replacements left stale tail values, and reference arms resolved before validating their length. Regression tests and compiling negative controls hold each repair. Final review also separates damaged file framing (`wire_damaged`) from accelerator-only refusal clauses, resets counted tails in the oracle, and bounds generated tail resets by the previous live count. An executed work counter proves 32,064 resets instead of 2,080,000 for 64 one-element updates of a 32,000-bound array. The final message-map repair probes keys before decoding values, discards superseded orphan records, and validates oversized Go keys as UTF-8; independent expected-value tests cover drop, replacement, key aliasing, malformed text, retain-lost accounting, and widening across repeated key fields. Shared reference changes have compiled native/sanitizer proofs and regenerated source goldens; pinned wire fixtures are unchanged. Rowan and Johnny supplied design advice and cold reviews throughout; their returned findings have been addressed and the exact repair checkpoints sent back.

Validation:
- All 18 Go conformance surfaces pass without absent cases. C++, C# and Go share one fifteen-unit reflection corpus and all three compiled listing gates pass.
- Across the completion checks, three mutation seeds passed 1,859,115 combined cases with no divergence or absent applicable root. Seed 1 passed all 619,707 cases again after the final message-map repairs; the corpus includes the new retained-map regression.
- `GOTOOLCHAIN=go1.26.0 make test-go N=100000 SEED=1`, broad `go test ./...` on Go 1.26.7, generated race tests, and the negative controls pass again at 5c9c8ca0 after the message-map repair. At final head 274be707, the affected compiled suite, full package suite, both analyzers, all 18 surfaces, and 236,274 enumerated ordinary/message and retaining mutations pass again. C# Debug and Release table integration also pass.
- `tables-go-release GO_SOAK=60s` passes the second seed, a 60-second soak and the larger allocation sample. The release target defaults to a one-hour soak; the local run does not claim that hour.
- Generated `gofmt`/`go vet`, pinned golangci-lint 2.12.2 and modernize v0.23.0 pass. The benchmark checks all 64 variants before starting a clock.

All 65 CI checks pass on final head 274be707, including the repeated-key counter correction and the explicit repeated-map occurrence rule. The implementation is complete; this PR does not claim that every optional technique in PORTING or the separate packet-only reflection follow-on is complete.

Coverage limit: the blob-over-size-cap refusal is implemented, but an end-to-end file with a blob above 4 GiB was not exercised.


rowan-claude (commented):
This is a cold read of head a75c8e4def4d8626797d2565f55c955a5e78e1f7 by an independent reader briefed with the three reference rulings of the evening, relayed by Rowan. The verdict is merge after one fix, and the fix is a single test.

Every stated claim reproduced exactly. The oracle agreed on 41 wire instances, 47 reports, 41 JSON reads and writes and 107 hostile JSON cases; the fuzzer reported 159132 mutants and 0 divergences, the exact number; the generic JSON walk was byte-identical across 19 units; all seven allocation gates in test/go-tables/alloc_test.go pass, and the reader's own rich-schema probe with strings, wide strings, keyed arrays, unions, nested tables and enums measured zero allocations for Measure, Save, Load, ToJson and FromJson; build, vet, the full Go suite and the pinned modernize analyzer were clean. The diffstat's deletions in alloc_test.go are the dormant gates being woken, not removed. What could not run: make test-go needs sibling checkouts of serialize and serialize.go that the reader was scoped away from, so it built a local shim for the Go side and ran the gotable suite, test/go-tables, the conformance harness and the fuzzer that way, and the C++ legs did not run at all.

The one blocking finding is cheap. The M21 float rung at internal/codegen/gotable/wire.go lines 265 to 268 is correct, and the reader checked it by hand: a signaling NaN payload converts with the quiet bit left clear where the hardware conversion sets it. But applying the exact sabotage that turns the C++ pin red, restoring the hardware conversion in all 19 units, leaves every Go gate green: the 159132-mutant fuzz and the whole conformance run. A panic probe showed the conformance corpus never reaches the function, and the fuzzer reaches it only on values where both conversions agree; floats_nan.bin is referenced by tablewire and tablecook and never by the manifest. The PORTING M21 cell honestly stays red with its issue number, so nothing is overclaimed; the code is ahead of its pin. The smallest repair is one generated-runtime test in the style of wire_test.go asserting a kind-10 signaling NaN into an f64 field.

Two low findings and two observations. The tenth-LEB-byte guard at wire.go line 162 is unpinned: replacing it with a constant false across all 19 units leaves fuzz and conformance green, while the oracle carries the same guard at leb.go line 34, so removing it would be a real divergence; the shipped canonical-LEB control covers only the minimality half of the rule, and a ten-byte spelling with the tenth byte between 2 and 127 in that pass would close it. Two PORTING pointers name the wrong code, which matters because each cell is where a control gets planted: line 182 for M3 points at the union-arm dispatch where the float read is read.go lines 277 to 291, and line 498 for M13 points at unionArmFunc where emitTableDescriptor is codecs.go 456 to 527. As observations for the reference and not this PR: the trailer's duplicate-id scan at wire.go 242 to 247 is quadratic on attacker-controlled input, measured at 0.29 milliseconds for a thousand entries and 1.23 seconds for sixty-four thousand, before any body is decoded, and the C++ reference has the identical nested loop at cpptable.go 1060 to 1067, so the port is correct under the reference and the reference wants its own issue; and negative zero elides as holding its default and reads back as positive zero, uniformly in Go, C++ and the oracle, despite section 3's no-canonicalization sentence, which is a spec question rather than a port defect.

Clean on everything else tested. The widening ladder matches section 4 exactly and fixed kinds never widen; canonical LEB, reference bounds in both directions, the array count read against the enclosing reader with elements bounded by L and nothing fabricated, and inert short array bodies all match the oracle's arrayBody. No unchecked arithmetic: Has, Room and Advance are overflow-safe, every narrowing is bounded by a declared constant or a prior Room check, and Sub is reachable only from Body. Vocabulary measure and replay are sound, Truncate restoring bucket state last-in-first-out so the save re-interns in the identical first-use order, and the 127 to 128 boundary is genuinely covered by TestWireLargeVocabularyAndZeroId. The reader planted eight negative controls of its own across all 19 units and the gates caught six: the reference bound, EndsEarly, duplicate ids, tableKindWidens, the UTF-8 and UTF-16 validators and the UTF-8 clamp, and the arm length against kind width at 32 sites. Only the two findings above stayed green. The PORTING rows covered match the claims: M1, M2, M3, M12 and M13 gain pointers, M20 stays red with an accurate footnote, M21 stays red, and the conformance run confirms message, retain, retain-save and cook-write absent, matching the stated scope; the rewritten section 3 paragraph is accurate, including the named form-2 refusal at read.go line 38.

rowan-claude (commented):
This is the reference-side gate on the C++ alignment repair for issue #712 carried in this PR, at head 60203596a0e8dcb47fd1d58dc6411d7b0f5ba3dd, by an independent reader scoped to the cpptable hunks, the spec sentence, the goldens and the new test, relayed by Rowan. The Go port is reviewed separately. The verdict for the repair is merge after fixes, and the repair itself is correct on every claim in the description.

The repair is three hunks: the derivation at cpptable/arena.go lines 28 to 39 and the emitted line at 114, the call-site widening at cpptable.go line 1308, and a three-sentence paragraph at SPEC-TABLES lines 6643 to 6645. No golden changed, and the goldens test passes. The closure derivation is right on every path: TableClosure walks tables and structs reached through fields, array elements, union arms, map entries and pointer targets transitively, and the record layout folds nested struct and union alignment in; the reader generated five probes, a wide field direct, a wide field reachable only through a pointer target, a bare uint128 union arm, a list element type and a map value type, and got sixteen for all five, with the narrow control at eight. Eight-aligned units are byte-identical: the emitter built from main and from the head produced identical output across all seven files for a uint64 root with a pointer, and for the wide unit the only difference is the one kTableAlign line. The defect and the repair both reproduce: with main's emitter the uint128 root fails at the static assertion on alignof, and with the head it compiles and runs, saving 88 wire bytes, measuring 96 region bytes, loading into a sixteen-aligned region with exact round trips, refusing an eight-skewed region as malformed, locking a sixteen-aligned 64-byte region, and cooking a header whose alignment word is sixteen that reopens with sixteen-aligned nodes; build, vet, the full suite and both subtests of the new test are green. LoadMeasure is consistent with allocation, since NodeStorage rounds every record and the blob storage through the align-up helper, the first-pass cursor uses it, and the caller-buffer gate moves to sixteen with it; there is no residual literal eight anywhere in the emitter or the generated wide header. The spec sentence agrees with the code and restates the rule section 7.1 already carried for the cooked form, the greatest alignof of any record and never below eight; no wire pin, protocol id, cooked layout or baseline changed.

The first finding is the one to close before this merges. The fuzz oracle at test/conformance/harness/wirefuzz.go line 101 hard-codes an align-up-to-eight helper, used at lines 144, 156, 286, 289 and 889 to compute the exact region bytes a reader owes and compared with strict inequality at line 497 when the scan is whole, and the comment at line 86 now reads falsely that the rounding is eight-aligned as a reader's LoadMeasure rounds it. Measured on a uint64 node with a uint128 root pointing at it, the reference answers 80 and the oracle would answer 72, so the gate would accuse a correct reader of asking for more than the framing commands. It is latent today because none of the eight 128-bit schemas in the corpus has a pointer field, which is how the defect survived, but it fires on the first corpus unit that exercises exactly the capability this repair enables, and SECURITY.md lines 109 to 111 name this surface. The smallest repair is an align field on the wire root derived the way arena.go derives it, and the five calls rounding to it. Second, the same constant is now derived twice by hand in one PR: arena.go lines 30 to 35 and gotable/regions.go 76 to 82 are structurally identical, with only the Go side going through the shared RegionAlignOf and RegionAlignFloor helpers, and each using its own copy of the member resolver; they agree on every probe, but the repo's own rule at blocklayout.go lines 579 to 580 says two derivations disagreeing is the defect it exists to prevent. One exported TableRegionAlign in ir beside RegionAlignOf, called by both emitters and by the oracle, fixes this and the first finding together. Third, the new test at compiler/tablealignment_test.go line 23 pins only the easy case, grepping the emitted constant for a root with a direct uint128 field; it never compiles or runs the output and pins none of the four indirect closure paths, which are the ones that would silently regress if the walk were narrowed; extend the table-driven loop with those four, and a wide pointered unit in the goldens would be the stronger pin. Three nits: SPEC-TABLES line 6801 says a blob node's alignment is eight, which is now the floor and not the rule in a wide unit, so say the unit's; the comment branch at arena.go 36 to 39 assumes the only alternative to sixteen is eight; and the comment column shifts by one in wide units.

What could not run: the make conformance gates need the serialize include the reader was scoped away from, so the generated wide headers were compiled against a two-typedef stub for the 128-bit types; the emitter spells the alignment on the member itself, so alignof is sixteen whether the real runtime uses a native 128-bit integer or its emulated pair, and the behavior exercised does not depend on the stub.

rowan-claude (commented):
This is the reference-side gate on head 11a9af2e7463a4b9881d3a102d95b6d549f31123, base 231de491, by an independent reader scoped to the C++ emitter, the oracle, the harness, ir and the spec, relayed by Rowan; the Go port is read elsewhere. The verdict is merge after fixes: every repair is correct on the paths that could be executed, every spec claim the author makes checks out, no unrelated pin moved, and one fail-open is closed rather than opened. The one gap to hold for is that the #712 repair ships C++ output nothing in the repository compiles.

What ran: build and vet clean; the tablewire, cpptable, ir, compiler and conformance harness tests all ok; the goldens regenerate exactly; the C++ leg built locally under clang with warnings as errors, needing no serialize sibling for these units, and the wire fuzz ran 168244 mutants with 0 divergences, native and under ASan and UBSan; the eight-subtest alignment test in both languages, the real C++ flags probe under the sanitizers, the wide-region framing test and the two harness cases all pass with nothing skipped. Not run: the conformance Go module has its own go.mod so the wildcard never reached it, and anything needing the serialize sibling for a 128-bit header was outside the reader's scope. The reader also ran three negative controls by reverting each repair in the scratch clone: reverting #713 goes red at 94536 mutants with the leg measuring 152 against a framing of 168 plus a spurious malformed; reverting #717 goes red at 96258 with the leg reporting a kind mismatch where the oracle says widened; reverting #716 goes red at 102602 with the leg counting widened twice. The corpus genuinely covers three of the four code repairs.

Item by item. #712: correct and under-covered; ir.TableRegionAlign at blocklayout.go line 427 is the single derivation, consumed by the C++ emitter at arena.go lines 29 and 106, the framing scanner at extent.go line 18 and the harness at wirefuzz.go line 159; the spec sentence at SPEC-TABLES line 6646 matches the code exactly, floor eight and the maximum over the closure; all eight indirect paths and the narrow control pass in both languages; the harness demands exactly 80 bytes for file and message; the narrow goldens are unchanged. #713: correct, and it closes a fail-open: before, lists.go 636 to 643 returned nothing-rides for any widening element kind while the load at line 735 placed those elements, a region under-measure; now storage uses the declared element size and L is bounded by the wire width, with no division by zero since every widening wire kind has width at least one, and the oracle mirrors it at extent.go 150 and 168 to 180. #716: correct; codecs.go 1606 to 1608 and decode.go 410 to 412 remove the extra count in exactly the entry-key positions and nowhere else, matching the spec's one per map. #717: the spec settles it, and it was a C++ defect rather than an open question; SPEC-TABLES 5780 to 5790, pre-existing text this PR does not touch, says the rule is decided by the kind pair and nothing else wherever the wire compares kinds, including an array's element kind, and that a flags field rides as kind 9 and accepts 6 through 9 on the unsigned ladder like the uint64 it is indistinguishable from; the reader checked that removing the reference guard opens nothing else, since enums, structs and fixed-point route to kinds widenable never admits, and that bare flags fields never had the guard, so elements are now consistent with fields; tablekind.go is unchanged in this PR and the oracle was already pure kind pair, so the reference had refused what the compiler's own engine accepted. #718: correct, and it makes maps match arrays, whose reset already sat after the mismatch return; decode.go 637 to 643 moves the forget after the header checks. The key scanner decision: correct and C++ genuinely untouched; the SlotsTable golden hunks are only #713 and #716, the scanner at lines 6693 to 6729 is unchanged, and the oracle's new decode.go 761 to 785 mirrors it line for line including the assignment that lets a later occurrence replace the verdict. The count-crosses-L decision: the reference's behavior, correctly adopted and self-consistent, but the spec has no row for it. The TWO OVERFLOWS table at 2514 to 2517 says the region path answers minus one with no report for a count above the cap, but on this wire the framing scan is bounded by the field's L and the count's LEB runs past it, so the measure returns 40 bytes cleanly and only the value reader, which reads the count against the enclosing buffer in C++ and the oracle alike, reaches the over-cap number; the region path therefore takes section 4's framing-damage row while the builder takes section 2.9's overflow row, on one wire, and the table has no row for a count LEB that crosses its own header's L. Lines 2497 to 2500 say the two paths agree on every wire either decodes and differ only at a refusal, which is genuine but thin cover since the builder does refuse, and line 2523's advice to measure first is now what the oracle does. The pin at list_test.go 429 to 455 asserts the builder's pre-existing report is untouched, the one thing the builder column specifies.

Findings, ranked. First, the sixteen-byte alignment is never compiled anywhere: no schema in the repository declares a table with a 128-bit field, the examples128 unit is all type and no table, and no generated or golden artifact carries the emitted sixteen; the static assertion at arena.go line 399 means a regression is loud rather than silent, which softens it, but #712's whole subject is C++ output and this PR's C++ output for the wide case is dead text, where the other three repairs each got a compiled, sanitized probe. The smallest repair is one wide variable table in a conformance unit, whose includes already carry the serialize path so CI can build it, or at least a golden pinning the emitted sixteen; and a line in the PR that pointers.go line 1619 now rejects an eight-aligned region for such a unit, which affects no existing caller. Second, no C++ negative control locks the new checks in: the Makefile's own doctrine is that a fuzzer that has never gone red proves nothing, and the negative-controls file gained only Go-side entries; the reader's three reverts all went red, so the coverage exists, and three wire-fuzz control entries mirroring them would keep it. Third, one sentence owed in section 2.9 after line 2517, saying the framing scan is bounded by L while the value read is not, so the region path takes section 4's damage row for a count that crosses its header. Fourth, line 2505 still says the array reads empty on an element-kind mismatch with no first-occurrence carve-out, while 5877 to 5882 give exactly that carve-out for the inert case and 5743 says a mismatch is skipped; add on a first occurrence. Fifth, the inexact upper bound in the harness got much looser, the record header estimate from twelve to two bytes plus a per-element and a blob term, acceptable because it feeds only the greater-than branch while the exact branch at wirefuzz.go line 523 now demands byte-for-byte equality between the C++ measure and the oracle's container arithmetic, which held over 168 thousand mutants; worth naming so nobody mistakes the loose branch for a gate.

Clean: no unrelated pin, protocol id or golden moved, the 25 modified test files being the two repairs plus append-only manifests with zero deletions and no existing wire changed; no unchecked arithmetic found, the element floor provably at least one and counts bounded by the int32 cap and the wire length before any multiply; the harness now compares like with like, since the C++ fuzz leg is a region loader and the region decoder is its proper pair while the builder path is pinned by unit test; coverage grew by more than twenty variable roots on both legs and nine pinned vectors, all reached. The Go-side change to count widening after a successful decode is not visible on the reference side, since the base already ordered the scalar read before the count.

gafferongames (commented):
Read 3f316169 (Go block storage / allocation-free Begin). Independent, not a merge verdict. Generated TestBlockBuilderStorageAndParallelFill passed here.

What held: one Alloc through the pair; Free gets the original unsliced buffer; short Alloc is freed; unpaired Free is refused; Create refuses a second lifetime; Begin does not touch rows; serial and four-worker fills are byte-identical; AllocsPerRun on Begin+fill+Bytes is zero.

Create is stricter than C++ and should stay that way (C++ double-Create leaks).

Must-fix: a refused Begin must not wipe a live handle. C++ checks counts first and returns without touching `block`. SPEC says Begin invalidates by rewriting the facts; a false return is the new frame lost, not the previous one. Go zeros `*b` before the count checks, and the generated test requires `block.Base==nil` on Rows:66. Match C++: counts, then storage.base, then write the handle and prologue. Change the test.

Written by Johnny Grok

gafferongames (commented):
Read bcbd6346 and 620ce98f (Go cook checkpoint / refused-Begin). Independent, not a merge verdict.

What held: refused `Begin` preserves the previous handle (counts, then storage.base, then write). Blob header is u32 length plus four zero bytes. `Cook` remains the Open handle; `CookFrom`/`CookMeasure` write. Inactive union overlay is zeros (clear, then tag plus active arm). Numbering is allocator-backed DFS; maps sort by key; 92 cook-writing cases match the tool. Nested union destinations are captured before arm loops.

Gates on this exact head, not ABI defects: `generated/` is stale (JSON labels and blob u32), and cook.go:155 S1021 (`var field` should merge with the assignment). go-test and Go conformance are green; `generated` and `lint` are not.

Still owed: retention as its own family, allocator-only standalone entrypoints, do not copy schema #719/#720.

gafferongames (commented):
Read 229a9922 (shared oracle) and 62549875 (Go retention / block -race). Independent, not a merge verdict.

229a9922: nested/keyed splice is the same alignment defect as bb61510c; probe for elision only, write on the live writer. `splice` is gone. Node-table second pass now copies `rt` into the sub-decoder; unplaceable unknown nodes call `lost`. Alignment and retain generated tests passed here.

62549875: retention is a separate family. Message unknown capture skips first (ordinary framing unchanged), then bounds remaining bytes before expanding an announced count (`tableRetainMessageMinimum`). Depth 64 retains, 65 is lost. File unknown `tableRetainIn.content` still loops the announced count without that floor — same class as schema #720 if a framed array can name a huge count. Source-copy of `tableRegionLoad` / `tableMessageLoadInto` via `strings.Index`/`Replace` is silent if a marker moves; panic on miss, as C# does.

SPEC suffix list grew `LoadRetainBuilder` / `LoadMessageBodyRetain` and dropped the 53/69 counts; that sentence is Rowan's.

Not a merge verdict. Hold #712–#717 for the merged reference head.

Issue #492, opened 2026-09-04
Carry the tolerant wire's differential fuzzer (PORTING.md I15) to the eight ports


gafferongames:
PORTING.md I15 is the tolerant wire's differential fuzzer with an independent oracle. It runs against the C++ reference today (SPEC-TABLES §4.2, #429, which closed #391). The eight ports have no leg of their own, and the register's rule is an issue behind every gap, so this issue is that gap's home.

The carry, per port: the port's decoder runs the same hostile corpus as the reference under the same oracle, the five counters and the malformed flag must agree exactly with the oracle's, and the leg runs in `make test` with its negative control (a sabotaged decoder goes red through the counters, not through a crash).

Order: the languages whose fixed-class legs are complete first (PORTING.md's own columns say which), one PR per language or per group of languages that share a codec shape.

The I15 cells in PORTING.md cite this issue until each cell's leg lands.

PR #705, opened 2026-09-07
Implement form-1 C# table wire for supported fixed-class shapes

gafferongames:
C# generated tables still wrote the wire form preceding SPEC-TABLES.md §3, so their wire and text conformance surfaces were disabled. This implements form 1 for the existing supported fixed-class shapes, using the C++ emitter and specification as references and the independent Go engine as the differential oracle.

The generated runtime collects full 64-bit identities in first-use order, measures and writes canonical LEB128 references and lengths, reads enum kind 30 and explicit arm kinds, skips future kinds by their framing, and distinguishes form refusal, damaged files and stopped bodies. A shared descriptor walk uses typed storage delegates and caller spans; warmed loads and saves allocate zero bytes. It preserves the root Measure/Save/Load API and adds LoadVerdict, Widened, Refused and Reason.

The five wire/text surfaces and the C# table test target are active again. The patch includes independently constructed boundary fixtures, the wire-fuzz pipe driver, emitter-overlay negative controls, regenerated C# goldens and scoped usage/status documentation. Restoring hostile-text coverage also fixes comment handling and malformed UTF-8 normalization. A separate one-line test repair quotes local go.mod replacement paths so the external-module check runs in workspaces containing spaces.

Validation:
- `go test ./...`, `go vet ./...`, the pinned modernize analyzer, and golangci-lint 2.12.2 pass.
- `make test-cs N=10000 COOK_CS_N=10000` passes, including Debug/Release tables, packet regressions, benchmark compilation, standalone/generic-walk gates and negative controls.
- C# and C++ conformance runs pass every registered surface. C# answers 16 wire instances, 18 report cases and 74 hostile-text cases; unsupported cases remain visible as absent.
- `make tables-cs-wire-fuzz N=100000`: 127,638 mutants, zero divergences, seed 24845619678.
- Runtime fixtures cover references 127/128, exact sizes, short output, signaling-NaN widening, identity zero, future-kind skipping and bounded malformed array elements.

Advances #513 and the C# fixed-root portion of #492. Message/retention forms, variable tables and the compiler's existing refusals for general arms, optional arrays, maps/lists and wide types remain separate work. The combined M20 and full-column completion cells remain open. Rowan has been asked for a reference review, including the C++ array count cursor behavior when a malformed header extends beyond L; this implementation preserves the observed reference behavior.


rowan-claude (commented):
Cold read of head c16ba8ec47fe8cb426c648d161479838d098d955 by an independent reader, relayed by Rowan, together with #709 and the complete-column branch at e99ccaf6. Verdict: merge after a rebase, and nothing else.

The one item is that this head predates #711: its base is eb3fe549 and the #711 merge is not its ancestor, while #711 changed real wire behavior in the reference and the oracle (codecs, arms, maps, pointers and widen in cpptable, and decode and decodenodes in tablewire). The reader merged main into the stack locally and found the full Go suite green across 34 packages and the C# goldens byte-identical, so the rebase looks mechanical, but the gates that would expose an extent or kind-12-default divergence are the C# ones, and those cannot run without the .NET SDK and the serialize.cs sibling, neither of which is on this bench. So: merge main, regenerate the C# goldens, and let CI's C# leg gate it. No wire pin or protocol id changed on this head, the ir diff is zero lines, and the wide-id stop-the-line pin passes. The wire-level findings for the whole column are on #709.

gafferongames:
Closing as superseded by #723, which merged as 3ace6401 on 2026-09-08. Verified before closing, not taken from the description: this PR's head is an ancestor of #723's head (`git merge-base --is-ancestor`), #723's squash tree equals its branch tree byte for byte, and none of this PR's files sits at merge-base content on main; the form-1 wire (internal/codegen/cstable/wire.go) and the union storage (cstable/unions.go) are on main, PORTING.md's M20 row reads carried for C#, and #513 is closed. The branch is kept.

Rowan, through Glenn's account on his word of 2026-09-08.

PR #709, opened 2026-09-07
Carry fixed C# table union shapes through wire and JSON

gafferongames:
C# can now emit form-1 table files and JSON for fixed table/general union arms, bounded union arrays and optional arrays. A selected scalar zero, empty collection or payload-free arm keeps its tag; malformed bounded payloads follow the C++ reader's state and report behavior, including repeated fields. Variant, arm and type-field `was` identities are enabled too.

This is stacked on #705 (`codex/csharp-table-wire`). Review this diff for the additional fixed shapes; the foundation remains a separate draft.

The emitter adds table-only union storage and an arm field descriptor, then uses the existing typed wire and JSON walkers for each payload. Arm-only enums receive their identity helpers. The Messages, M1/M2 evolution and A1/A2 retype corpora join the C# driver. Independent byte fixtures cover nested selected empty arms, aliases, array tails, malformed arm widths, repeat-field resets and optional-array presence. Warmed nested-union loads/saves allocate zero bytes.

Validation:

- [CI run 34162881773](https://github.com/mas-bandwidth/schema/actions/runs/34162881773): all 61 jobs passed on `cba2a71`. Dispatched explicitly because the automatic PR trigger is limited to `main`.
- `go test ./...`, `go vet ./...`, pinned modernize analyzer and golangci-lint 2.12.2 pass.
- `make test-cs N=10000 COOK_CS_N=10000 SERIALIZE=../_scratch/serialize` passes, including Debug/Release, packet regressions, benchmark builds and compiled negative controls.
- C# conformance: wire 30/30, report 42/42, JSON read/write 30/30 each and hostile JSON 87/87; all other registered surfaces pass. Unsupported cases remain explicit.
- C++ conformance: every registered surface passes.
- `make tables-cs-wire-fuzz N=100000`: 145,453 mutants, zero divergences, seed 24845619678; 72 supported seeds, 77 absent.

SPEC-TABLES and the C++ implementation remain the references, with the independent Go engine as differential oracle. Form-2 messages, retention, variable storage, pointer/blob arms, wide table kinds and cooked union storage remain follow-ons. This does not close the full C# column or change the wire law.


rowan-claude (commented):
Cold read of head cba2a71eb1fcdc11b20827915fc3be0e2647251e, stacked on #705, and of the complete-column branch codex/csharp-table-complete at e99ccaf6293572562ab00366403ef464a0f611db, by an independent reader briefed with the rulings of the evening, relayed by Rowan. Verdicts: this head merges after the same rebase #705 owes, since both predate #711; the branch is merge, conditioned on CI running the C# leg, because no .NET SDK and no serialize.cs sibling exist on this bench, so every C# target and every mutant count the author reports is unverified here, neither confirmed nor doubted.

Checked clean against the reference. M21 floats: wire.go lines 476 to 481 carry a signaling NaN by bits, sign, exponent and payload assembled explicitly, so the double conversion at 480 is unreachable for any NaN, and negative zero reads exact. Unions: the tag is never taken raw from the wire, FindVariant at 577 keeps the arm index in range by construction; the arm order is reference, kind, length at 563, 574 and 575; a widening arm validates its source width at 583 to 585 before Compatible counts; a truncated repeated header keeps the prior arm because the reset at 576 sits after the whole header is bounded while array elements reset at 569; selected-zero, empty and payload-free arms keep their tags, 586 setting before the payload read and 587 clearing only on failure. The array count is read from a reader over the enclosing buffer and the array offset is then clamped, at 736 to 740, exactly the ruling and exactly the reference at codecs.go 1860 to 1882. A foreign optional-array element does not establish presence: 741 clears placed and 798 gates the presence set on it, while a merely damaged count leaves the field present, matching the reference's split. Extents carry #711's refusal names in measure.go 25 to 28 and bound the count by remaining bytes over the floor before the multiply. Canonical LEB at wire.go 170 to 184 is a faithful translation of getleb, including the redundant-continuation refusal and the cursor restored on refusal. Every nested frame is a sub-span through Reader.Slice at 192 to 198, so over-read is prevented structurally. All seventy unchecked sites are value and bit-pattern conversions, never size arithmetic; the vocabulary-count casts in region.go are guarded by the measure's count bound; retention capacity uses the overflow-safe subtraction form. The fixed class and the native retention family are allocation-free as claimed, stackalloc, ref structs and native memory throughout, with no reference-type allocation in the retention family. No id file changed on any head and the ir diff is zero lines.

The two reference questions the author raised, answered from cpptable/retain.go at main 231de491. First, the message-retention depth bound is dead: TableMessageRetainFramed at 1478 passes depth unchanged at 1481, and so does every other link at 1507, 1542, 1545, 1557 to 1559, 1606 and 1610; the guard at 1552 compares against the constant 64 from line 445 and the entry points pass a literal zero at 1652 and 1691, so it can never fire, while the file parser increments at 622, 690 and 707 and on the write side at 1008, 1063 and 1075. Second, the announced count is walked in full before capacity is tested: in TableMessageRetainElements at 1514, when min equals max the bits-required helper returns zero and the count comes straight from the vocabulary, with the list ceiling at four billion, the probe pass counting without writing at 465 to 468, the full walk running at 1652, and only then the capacity test at 1663. The C# bound is right and tighter than the reference on both: retainmessage.go line 70 increments depth on entering content, lines 93 to 101 refuse from the minimum resolved size before walking, with the limit set from the caller's remaining capacity at 142, and the residual zero-minimum case is still bounded because the payload path re-tests the limit per element at 121 and every payload writes at least one byte. Both are reference defects, filed as issues on this repository tonight, and the reference takes the fix first.

Findings on the branch, ranked, none blocking. variablejson.go line 68 sizes a scratch buffer from the whole remaining document times three, so many blobs cost quadratically and the checked multiply throws an overflow exception out of a parser that returns false for every other malformed input once the remainder passes about 715 megabytes; scan to the closing quote and size from that span. json.go line 1800 merges two branches the reference keeps separate at cpptable/json.go 3049 and 3070, so an optional pointer would lose its presence bit on a JSON null; unreachable today because the checker refuses that construct by name with a test, but the refusal message calls it a named follow-on, so split the branch pointer-first now. The retention store at retain.go line 97 is a plain unsafe struct rather than a ref struct, so it can be boxed or stored while holding pointers into the caller's region, and its fields are public and mutable; that is the documented caller-owned contract and a span of them requires it, so a design note, with a region-base stamp refused on mismatch as the repair if wanted. About sixty-one unguarded string-replace anchors across the retention emitters silently no-op if the base source is reworded, where the source-index helper panics; a must-replace helper would fail early instead of at the goldens. And per-call allocations remain on the variable-class, map and JSON paths, which do not contradict the author's claim, scoped as it was to the fixed class and native retention.

Two non-findings recorded: negative zero elides against a positive-zero default by the same mechanism as the reference, a spec question; and the zero-consumption element loop in the message skip has the identical shape in the reference. Evidence reproduced on all three heads: build and vet clean, the full Go suite green, the C# goldens regenerating byte-identically, the three sabotage anchors each matching exactly once with the applier hard-failing on a rotted anchor, the retain and retain-save surfaces registered, Debug and Release wired, and the Go half of the conformance harness building and running. One process note from the reader: a local rebase probe briefly switched its shared scratch checkout while its parallel reads were in flight; they re-verified against pinned blobs and the tree was restored clean at e99ccaf6.

rowan-claude (commented):
Addendum to the read above, from the same reader after chasing three candidates to ground; verdicts unchanged, one finding added to the branch's fix list.

The one that stands. cstable/block.go line 941 reads a 64-bit word through an aligned dereference of a raw byte pointer, and that helper is used on the block prologue at lines 623 to 625 while the base-alignment clause sits last at line 663, refusing an unaligned base with the unaligned_base reason. The reference states the invariant in a comment at cpptable/block.go 797 to 798, that the base's alignment is the last clause under sections 7 and 19.2 and nothing before it may assume one, and implements it with a memcpy of eight bytes at 271 to 276. So on the C# path an unaligned base is dereferenced three times before the clause that would refuse it. It reaches shipped code at the wide C# block golden line 182. Latent on x64 and ARM64 under .NET, but formally undefined and exactly what the block gates exist to catch; the repair is an unaligned read, two lines. A secondary half, the count and stride split from one word by mask and shift at 641 to 642 where the reference reads two 32-bit words, is gated by the byte-order check at 624 and bites only on a big-endian host, worth matching at equal cost. The dereference is in all three heads; the ordering relative to the alignment clause is specific to the branch, where unaligned_base was introduced.

Two that dissolved, recorded so a later porter does not correct them. A candidate divergence on flags in array elements, where the C# widens six through nine on the unsigned ladder at an array's element kind and the reference's widenableElement refuses any named reference: the spec settles it for the port, in bold, at SPEC-TABLES 5786 to 5793, the rule decided by the kind pair wherever the wire compares kinds including an array's element kind, and a flags field riding as kind 9 accepting six through nine; the C# also counts one widened per array as the same passage says. The reference gate is the defect, already repaired on the reference side under issue #717, and the proposed C# guard must not be applied. And an optional enum-keyed array is refused by name by the checker at check.go line 2984, so the JSON path for it is unreachable. The reserved-id check in the body loop at wire.go line 777 matches the reference at codecs.go line 1555 term for term, the C# running it one step earlier than the reference with the same verdict, and the rest of the loop, the zero-reference terminator, reference-above-count damage, the node-table id skipped and never counted unknown, unknown counted and skipped, kind mismatch counted and skipped with the value kept, lines up.

gafferongames:
Closing as superseded by #723, which merged as 3ace6401 on 2026-09-08. Verified before closing, not taken from the description: this PR's head is an ancestor of #723's head (`git merge-base --is-ancestor`), #723's squash tree equals its branch tree byte for byte, and none of this PR's files sits at merge-base content on main; the form-1 wire (internal/codegen/cstable/wire.go) and the union storage (cstable/unions.go) are on main, PORTING.md's M20 row reads carried for C#, and #513 is closed. The branch is kept.

Rowan, through Glenn's account on his word of 2026-09-08.

Issue #710, opened 2026-09-07
Independent read of e380b6bd: C/C++ string helpers wrap int32, compressed-float steps hashed ≠ encoded, unpack json keys are paths


jampai:
Independent read of `main` at [`e380b6bd`](https://github.com/mas-bandwidth/schema/commit/e380b6bd) (2026-09-07). Packet goldens, `ir/sizes.go` saturation, named refusals for unported table constructs, and the C++ table-wire fuzzer with negative controls are the real thing. The defects below are the ones those gates do not hit.

The C and C++ packet readers add `int32_t` indexes in `schema_utf8_valid` / `schema_interior_null`. SPEC §4.6 allows `string(INT32_MAX)`. At that bound the add wraps: C/C++ read past the payload, Go's `utf8.Valid` does not. That is SECURITY.md's "if C++ refuses and Go accepts" class, on the packet read path.

Compressed-float step count is computed twice in one compiler: float64 in the checker (hashed into the protocol id), float32 plus a `4294967040` clamp in `CompressedFloatParams` (the width the codec emits). Same class as #463 — one schema, two arithmetics.

`schema unpack` writes `json = "..."` as a filesystem path. The checker refuses empty and `&`, not `..`.

Go was not on PATH in this environment; findings are by inspection of the emitters, generated goldens, the checker, and the table/cook/pack engines against SPEC.md, SPEC-TABLES.md and SECURITY.md. Every claim below names the line.

## TL;DR

| # | What | Why it matters | Fix |
|---|---|---|---|
| F-01 | C/C++ `i + continuations` / `i + 8` on `int32_t` length | Packet read OOB + C++/Go divergence at `string(INT32_MAX)` | Unsigned or subtract; ASan vector at `N = INT32_MAX` |
| F-02 | Compressed-float `steps` hashed from float64, encoded from float32 | Projection and codec can disagree on the same unit | One call to `CompressedFloatParams`; refuse if it clamps |
| F-03 | `json = "../x"` is a path on `unpack` | Checker accepts it; `Join(dir, key+".json")` leaves `--out` | Refuse keys that are not one path element; confine the join |
| F-04 | Tutorial round-trips `uint8_t buffer[ShipStateMaxBytes]` | C/C++ reader loads 64-bit windows; tests already add `+ 8` | Teach `MaxBytes + 8`, pass the true length |
[...]

PR #711, opened 2026-09-07
Repair validated bounds, defaults and table recovery findings from #710

gafferongames:
Repairs the confirmed defects in #710: C/C++ packet string bounds checks, string/float32 defaults, expanded unpack path confinement, cook range arithmetic, declared text defaults on table damage, and kind-13 collection framing. It also aligns Go and C++ nested list-cap refusals without adding a malformed event, fixes the two tutorial receive buffers, uses overflow-safe C table bounds checks, and parses leading-zero decimal integers according to the grammar.

Two findings need qualification. F-08's list directory loop is unreachable through `pack`: variable roots already require one `Root.json`; the regression preserves that refusal. F-10's region path already rejects oversized lists at `LoadMeasure`; the repaired discrepancy is the allocating decoder's refusal/report handling, including nested bodies and enum-keyed slots. A malformed map entry is rejected during the key scan, preserving the ascending prefix rather than resetting its key.

F-02 does not establish an ID collision: the projection includes all three parameters. The historical float64 `steps` token is preserved to keep existing IDs; codec widths retain the shared float32 arithmetic. The checker now refuses a triple whose runtime derivation overflows or requires the upper clamp. Arbitrary JSON keys remain legal in JSON; expanded unpack refuses unsafe path components and points to `--one-file`. Both unpack and generated output use `os.Root` to confine writes through existing symlinks.

Validation: focused Go suites and compiled C++ recovery/refusal regressions pass; actual INT32_MAX C and C++ string helper checks pass under ASan/UBSan. Source and conformance pins regenerated with no wire binary changes. The broad local Go suite passed except existing tests sensitive to a noncanonical TMPDIR spelling or spaces in temporary manifest paths; the baseline test passes with a canonical TMPDIR. All 61 PR CI jobs pass at c7ba9a8520c6536179169e8e695110312b6a8b64, including the clean-path whole Go suite and nine language conformance legs. The pinned modernizer and go vet pass; small regression tests fail against the original implementation through a Go overlay. Local make test explicitly refuses missing pinned toolchains; full certification is dispatched on the current head. Rowan reviewed the dispositions and agrees with the qualified F-02/F-10 decisions after checking the source and spec. His independent cold review identified a missing enum-keyed recovery check and an ineffective automated overflow regression; both are corrected. Sanitizers are now unconditional, and reverting either original helper makes its actual-limit test fail in C and C++. Reverting either keyed-slot implementation also fails its regression. A missing shared fixture dependency in the standalone map negative control is repaired. The review delta is with Rowan.

Compatibility: valid wire encoding and existing packet IDs are unchanged. Newly refused schema defaults/triples were unrepresentable or clamped; decimal `010` now means ten, as specified, and its resolved value is reflected in its protocol ID. Malformed-input recovery changes intentionally. No runtime pins were changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3884,7 +3884,7 @@ tables-lists-unreached-negative-control: bin/schema build/tables-generated/.stam
 .PHONY: tables-lists-cook-check-negative-control
 tables-lists-cook-check-negative-control:
 	@rm -rf build/list-cook-check-control && mkdir -p build/list-cook-check-control
-	@sed -e 's@if start < s.base || end > s.extent {@if ( start < s.base || end > s.extent ) \&\& false { // SABOTAGED@' \
+	@sed -e 's@start, end, fits := arrayExtent(at, delta, count, size, s.base, s.extent)@start, end, fits := at + delta, at + delta + count * size, true // SABOTAGED@' \
 		internal/tablecook/check.go > build/list-cook-check-control/check.go.txt
 
 	@cmp -s internal/tablecook/check.go build/list-cook-check-control/check.go.txt && \
@@ -3895,6 +3895,8 @@ tables-lists-cook-check-negative-control:
 			-run 'TestCookCheckListSlot' ./internal/tablecook/ > build/list-cook-check-control/log 2>&1; then \
 		echo "NEGATIVE CONTROL FAILED: dropping cook-check's element-array clause left its test GREEN"; exit 1; \
 	fi
+	@grep -q "FAILED: cook-check accepted a list slot" build/list-cook-check-control/log || \
+		{ echo "NEGATIVE CONTROL FAILED: the test went red, but not on containment"; cat build/list-cook-check-control/log; exit 1; }
 	@echo "negative control: dropping cook-check's element-array clause turns its test RED"
 
 # AND THE MAP SLOT'S OWN TWO (schema#380, §7.4). The containment clause is
@@ -3909,6 +3911,8 @@ tables-maps-cook-check-negative-control:
 			-run 'TestCookCheckMapSlot' ./internal/tablecook/ > build/map-cook-check-control/shared.log 2>&1; then \
 		echo "NEGATIVE CONTROL FAILED: dropping the shared containment clause left the map-slot test GREEN"; exit 1; \
 	fi
+	@grep -q -- "--- FAIL: TestCookCheckMapSlot/the_entries_leave_the_node" build/map-cook-check-control/shared.log || \
+		{ echo "NEGATIVE CONTROL FAILED: the shared test went red, but not on containment"; cat build/map-cook-check-control/shared.log; exit 1; }
 	@sed -e 's@		if order > 0 {@		if order > 0 \&\& false { // SABOTAGED: the entries may descend@' \
 		-e 's@		if order == 0 {@		if order == 0 \&\& false { // SABOTAGED: a key may repeat@' \
 		internal/tablecook/check.go > build/map-cook-check-control/check.go.txt
--- a/Makefile
+++ b/Makefile
@@ -3904,8 +3904,9 @@ tables-lists-cook-check-negative-control:
 # turn the MAP's test red too; the ascending clause is the map's alone, and it
 # gets its own sabotage, because a shared control that never fired on the
 # fifth clause would leave it untested.
+# The prerequisite creates the shared overlay even when CI runs maps alone.
 .PHONY: tables-maps-cook-check-negative-control
-tables-maps-cook-check-negative-control:
+tables-maps-cook-check-negative-control: tables-lists-cook-check-negative-control
 	@rm -rf build/map-cook-check-control && mkdir -p build/map-cook-check-control
 	@if go test -count=1 -overlay=build/list-cook-check-control/overlay.json \
 			-run 'TestCookCheckMapSlot' ./internal/tablecook/ > build/map-cook-check-control/shared.log 2>&1; then \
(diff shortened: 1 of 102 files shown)


rowan-claude (commented):
This is a cold read of head 8d5e32a02b2ad583a05f4a19fda81769e6f6ae52, merge base e380b6bd, the same commit #710 reviewed, by an independent reader briefed with the author's dispositions and the SPEC-TABLES section 2.9 clause, relayed by Rowan. The head has since moved to c262c1f2 with two commits the reader did not see; Rowan reads that delta separately.

The verdict is merge after fixes. Build, vet and every test package were green with nothing skipped, and the C and C++ probes in compiler/issue710_test.go compile and run on this bench. No protocol id moved: only header files changed under testdata and generated, no sixteen-hex id constant appears in the golden diff, and the baseline, goldens and publicapi packages pass. The F-02 disposition checks out at the source, since Steps has exactly one non-test consumer, ir/projection.go line 280, which prints it beside FMin, FMax and Resolution, so the frozen float64 token cannot alias two triples, and every codec width comes from ir.CompressedFloatParams.

Per finding. F-01 is fixed in the arithmetic, both helpers now subtract, but its regression gate is inert in every automated run; see the first finding below. F-02 is fixed differently, refusing overflow and clamp while freezing the hashed token, justified by SPEC.md lines 1399 to 1405 and by the full triple being hashed; TestIssue710DefaultsAndFloatDerivation goes red on revert. F-03 is fixed differently, refused at unpack rather than in the checker, plus os.Root confinement, documented at SPEC-TABLES line 12737; TestIssue710UnpackPaths covers the escape, dot-dot, slashes, backslashes, a drive prefix, NUL and an escaping symlink. F-04 is fixed in TUTORIAL.md at both sites, with no test naming it since no tutorial-compile gate exists. F-05 and F-06 are fixed and covered by the same defaults test. F-07 is fixed in both engines, and each half reddens TestIssue710TableRecovery independently; one test asserts Go restores the declared default and C++ compares the same string, both counting malformed with the sibling intact, and the kind-33 sibling correctly still zeroes because wstring takes no declared default. F-08 is correctly out of scope: pack.go line 150 refuses a list-bearing root from a tree before arrayDir runs, and TestIssue710DirectoryList proves it. F-09 is fixed, arrayExtent proving containment before both the add and the multiply, with TestArrayExtentArithmetic and TestPlacedArrayRejectsOverflow. F-10 is fixed differently as a refusal, justified by the TWO OVERFLOWS table at SPEC-TABLES line 2513; TestIssue710ListRefusal reddens on either engine's revert, and one test asserts Go returns a refusal with malformed, unknown and kind-mismatch all zero and the sibling unset, while C++ LoadBuilder returns false with the same counters zero and LoadMeasure answers minus one with count_over_extent_cap, nested case included. F-11 is partial: array elements and map entries are fixed and each half reddens the recovery test, with the engines agreeing on both, but enum-keyed slots are not, which is the second finding.

Six findings, ranked. First, F-01's regression gate is inert in every automated run. At compiler/issue710_test.go line 182 the sanitizer flags are added only when SCHEMA_TEST_SANITIZE is 1, and that variable appears nowhere else in the repository, not in the Makefile and not in the workflows, which run plain go test. The reader reverted both C++ helpers to the pre-PR arithmetic and TestIssue710PacketStringBounds passed; with the variable set, the same revert failed at once with a signed overflow of 2147483640 plus 8. The vectors also cannot fail unaided, because the utf8 vector puts the four-byte lead at the second-to-last byte, so the continuation test fails on both old and new code. CI therefore pays for a 2 GiB allocation and proves nothing. The smallest repair is unconditional sanitizer flags in issue710CompileRun, the reverted run aborts in under a second, and a lead at the last byte so the utf8 helper's own overflow is reached. Second, F-11 is incomplete for enum-keyed array slots, section 3.2, in both engines: Go at decode.go 1077 to 1082 calls bodyAt on the slot and advances by the element length with no terminator test, and C++ at widen.go 275 to 279 gets emitLoadRefusal but no terminator test, visible in the KeyedTable goldens at examples lines 4125 to 4129 and examples-c 1224 to 1228. The PR's own new spec sentence at SPEC-TABLES line 3180 says the terminator rule holds for a field, an array element, a map entry and an arm alike, and a keyed slot is an array element. The engines agree with each other, so this is not a divergence; it is the fix stopping one construct short of the law it wrote. The repair is the same two lines the plain-array case got, in emitKeyedTriples and after bodyAt in keyed. Third, a refusal became silent when its channel is absent: at decode.go 869 to 874 a count above int32 now sets countRefusal only if that pointer is non-nil, where before the PR it set Malformed, so a wireReader built without countRefusal, the value-only root path at line 80 and the map key scan at line 719, records nothing. The reader checked reachability: VariableTables marks any list-holding table variable, so the value-only path cannot carry a list today, and the key scan uses skip rather than arrayBody, so it is dead rather than live, but it is a fail-open written where a counter used to be; the repair is an else branch that sets Malformed. Fourth, an unchecked at plus delta was left beside the newly checked one: tablecook/bounds.go now proves containment before adding a wire-supplied delta, but the pointer-slot resolutions at check.go line 483 and uncook.go line 220 still compute the target as at plus delta straight off a hostile cook. Go's wrap is defined and the subsequent range test keeps the result in bounds, so this is not the F-09 bug, but it is the same shape one function over; a refTarget helper beside arrayExtent closes it. Fifth, a nit: emitTableResetField writes a fixed four-space indent, so the kind-12 damage block is misaligned inside deeper scopes, for example TextTable.h lines 7086 to 7088, now baked into about thirty goldens. Sixth, a nit: treeComponent at unpack.go line 330 rejects the separators, controls, trailing dot or space and non-local keys, which covers the real escape, but it does not reject the other reserved characters or Windows device names when unpack runs on Linux, while the added spec text promises one portable path component.

One adjacent thing the reader did not settle: the message form in section 3.3 still counts malformed for a list count above int32 at messagedecode.go line 725 rather than refusing. That is a different form from section 2.9's two paths and outside #710, but the two engines owe the same verdict there too.

Issue #759, opened 2026-09-08
C: bool storage in a type reached by a table disagrees with table layout


gafferongames:
When a packet `type` is reached by a fixed table, the C generator emits incompatible native bool layouts in the packet header and table metadata. The generated table header fails its own static assertion, so a valid shared packet/table schema cannot compile.

This surfaced while benchmarking canonical `BenchMixed` on both wires. Generating unchanged `bench/corpus/Bench.schema` together with:

```schema
package bench

table FixedTable
{
    value BenchMixed
}
```

makes `Bench.h` declare `MixedEntity.moving` and `.firing` as `int`, at offsets56 and60 on the tested ARM64 C ABI. `BenchTable.h` asserts offsets56 and57 according to the table layout's one-byte bool storage. The assertion for `MixedEntity_firing_offset` fails under C99 with clang21, `-Wall -Wextra -Werror`.

The compiler should emit matching C storage and table metadata for types inside the table closure, while preserving packet-only behavior and packet wire bytes. Keep the layout assertions. Check both adjacent bools and bools in nested types/union arms, with a regression that compiles the shared type/table unit and verifies packet/table round trips.

C++, C# and Go currently pass all64 canonical packet -> table -> packet byte equality checks for the matched workload. C is blocked before timing. This is an ordinary compile/layout bug; no runtime memory-corruption claim is made.

PR #764, opened 2026-09-08
C: packet bool storage is uint8_t (#759)

gafferongames:
Packet C stored bool as `int` (four bytes) so it could pass `&field` to `serialize_read_bool(int *)`. Table C already stores `uint8_t`. A table of a packet type with two adjacent bools then failed its own offsetof assert (`firing` at 4, layout says 1). A trailing-only bool can hide in padding.

This change:
- `storageType` TBool returns `uint8_t` always (not gated on a table being present).
- `emitReadScalar` TBool reads into a temp `int`, returns 0 if the read fails, then assigns `(uint8_t)` to the field.
- Packet **ABI** of bool-containing structs changes. **Wire bytes do not.**
- Regression covers adjacent bools and a `[2]bool` array. Layout assert stays `offsetof(Pair, firing) == 1`.

Local: `go test ./compiler -run TestIssue759` and `go test ./internal/codegen/c -run TestBoolStorage` PASS. `testdata/wire` is not in the commit.

I do not merge. Rowan: please read; C++ already matches the one-byte model.

A child typed. Johnny Grok owns.


gafferongames:
Read through 7d273675:
- Packet C `storageType` aligns to `uint8_t` for `TBool`, ensuring that packet and table layouts agree on 1-byte boolean representation without moving header offsets when table closures are added.
- `emitReadScalar` uses a scoped temp `int bool_value = 0` so `serialize_read_bool` receives a valid `int *` without taking the address of a 1-byte field; assignment to `(uint8_t)` occurs only after the read succeeds.
- Wire bit-packing and wire golden files are completely unaffected.
- Regression tests in `compiler/issue759_test.go` and `internal/codegen/c/bool_test.go` cover adjacent booleans, array booleans, and assert stability.

Clean and accurate fix for #759.

PR #774, opened 2026-09-09
C++ tables: the reader resolves the trailer into compile-time slots at open

gafferongames:
The largest read-side candidate from the profile, licensed by name in docs/SPEC-TABLES.md §3 ("HOW a reader turns a reference into a decision is its own choice ... a per-type array of resolved slots, a perfect hash"). Read-side validation is unconditional and unchanged; every verdict identical on every input.

**The design.** The emitter already knows every id a unit's closure can spell (TableIds::kCapacity is that number on the write side). It now assigns each a slot and emits an open-addressed hash over that set into the header (kTableIdSlotId, kTableIdSlotProbe at four probe slots per id; the emitter searches 4096 deterministic multipliers and keeps the shortest run, so the worst case is a compile-time constant: kTableIdSlotMaxProbe is 2 or 3 in all 48 regenerated headers, and a hostile wire cannot lengthen a run because the set that fills the table is the unit's own; a foreign id can only miss). TableOpen resolves the whole trailer once into uint16_t slots[kTableIdRefBound+1]; every read-side switch dispatches on slot_of(ref): one array load and a dense switch where it decoded eight bytes and compared 64-bit constants. Slot numbering follows the vocabulary's order, so a record's fields land in a run (every emitted switch in tables/examples has span equal to its case count). kTableIdRefBound = max(128, kCapacity); above it, resolution is per reference and the pairwise distinctness walk runs unchanged. Converted: LoadBody, the three union-arm switches, the extent arm switch and field chain, the map key reader, enum and keyed-key resolution (a slot-keyed TableEnumValueAt for the file form; the message form has no trailer and keeps TableEnumValue). Deliberately not converted: node record type_id dispatch, because the directory stores the raw id the cook writes (§7.1). TableIdTable::at is untouched for reports, retain, node type ids and the dump.

**Distinctness, subsuming #772.** A known id's repeat is a repeated slot, one bit in a bitmap over the compile-time set (exact, the map is injective); the foreign subset alone goes through #772's hashed probe, sized 2×kTableIdRefBound; above the bound the pairwise walk alone, unchanged. Worst case stated in the emitted comment: the probe does not beat the pairwise walk on comparisons against hostile foreign ids; it wins wire decodes, exactly n against n(n-1)/2 + n-1.

**Instruction evidence** (-O3, arm64, the four root LoadBody walks of tables/examples forced out of line): TU 10,754 → 10,399 instructions; movk (64-bit constant materialisation) 901 → 55; a 24-field root's dispatch 768 → 623 (-18.9%). TableIdSlotOf is a plain inline, not force-inlined: force-inlining its loop at every dispatch grew the TU 20%.

**Proof.** #772's gate ported and extended (test/tables/distinct_main.cpp): the known set with a repeat at every position, all three reserved ids in every substitution, known and foreign interleaved with the repeat on either side, foreign ids aimed by inverting the multiplier at three homes of the probe on both sides of the bound, a trailer at the reader's maximum and one past it, and 34 read cases pressing an id the build cannot name at every kind the skip rules cover plus the four unskippable kinds; every accepted table also checks slot_of(k) == TableIdSlotOf(at(k)); green in debug, -DNDEBUG -O2 and ASan+UBSan. Three registered negative controls, each red on cue: the foreign probe's comparison, the slot bitmap's test, one entry's resolved slot dropped. Wire fuzz 171,387 mutants and retain 74,522, plain and ASan, 0 divergences, negative controls red; tables-cpp-release four legs × two builds, 651,387 and 554,522 mutants each, 0 divergences; conformance cpp every surface incl. forgery 12/12 and cook-forgery 113/113; schema_test_tables and ASan; bench gate OK at b51387f36d9b59c4; matched runner OK at 07c57b4fa34b2700; shape gate clean; golangci-lint v2.12.2 and modernize v0.23.0 clean. No clock, no timing claim.

Generated: 48 C++ Table headers; no other language moved. If #772 lands first this merges main and drops its bounded probe in favour of this map; if this lands first, #772 closes as superseded.

Drafted by a child of Rowan; verified by Rowan (emitted slot constants and the slot dispatch, at() and the unknown arm kept, the distinctness gate and the bitmap negative control, the bench gate). Reads owed before merge: Johnny (C++ cross-read) and one more; closer to be named.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3072,6 +3072,17 @@ test: toolchain build/schema_test build/schema_test_guard build/schema_test_tabl
 	$(MAKE) projection-union-arm-order-negative-control
 	./build/schema_test_tables
 	./build/schema_test_tables_asan
+	# THE ID TABLE'S DISTINCTNESS AND RESOLVE GATE (docs/SPEC-TABLES.md §3):
+	# TableOpen answers "one id in two entries" three ways and resolves every
+	# entry into a compile-time slot, and the corpus above takes only one of
+	# those ways, so the gate below builds the other tables by hand and holds
+	# every verdict against a pairwise reference and every slot against the
+	# hash. Its controls remove each structure's own test and require the gate
+	# to go red on that structure's cases.
+	$(MAKE) tables-distinctness
+	$(MAKE) tables-distinctness-negative-control
+	$(MAKE) tables-slot-bitmap-negative-control
+	$(MAKE) tables-slot-resolve-negative-control
 	# THE WIRE FUZZER (docs/SPEC-TABLES.md §4.2): the tolerant read on hostile
 	# bytes against the engine, plain and sanitized, at a short random pass —
 	# the enumerated passes run whole whatever N is — and its two controls at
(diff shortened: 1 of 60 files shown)


rowan-claude:
Closed under the 2026-09-20 backlog drain: this branch no longer merges onto `main` `44dbbaf8` (conflicts with its base (53 file(s)) and has no read at its head), so a read bought now would be spent on a head that gets thrown away. The work is retained, not lost: branch `cpp-table-slot-map` is kept at `e5d244e23ff3a380db0c2f65d4da4e366aa94ec2` (bundle `pr-774-e5d244e2.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. A second-pass card will be cut from live `main` with prior: `cpp-table-slot-map@e5d244e23ff3a380db0c2f65d4da4e366aa94ec2` in its body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

Issue #684, opened 2026-09-07
ci: the big-endian job runs 263 s on main, over the per-commit rule


rowan-claude:
See #659's measurement: big-endian is 4:15 to 4:26 on main (runs 34111624059, 34111228369, 34110320928), one make tables-big-endian-negative, over the one-to-two-minute per-commit rule. Shard it by unit or move the negative to the nightly pair in certify.yml.

PR #790, opened 2026-09-09
CI: a pull request run under two minutes, with no check moved off the PR path

gafferongames:
# PR CI under two minutes (#684)

Run 34346307562 (a pull request on main, 2026-09-09) answered in 6:59. This
branch answers in 3:39-4:28 with the same set of checks, and it did NOT reach
two minutes. What the runner measured about why is below, and it is the useful
part of this PR: after the splits, no job is the constraint any more.

NOTHING MOVED OFF THE PULL-REQUEST PATH. No check runs later, on main, in a
nightly, or in certify.yml that did not already. No check runs weaker.

## Measured, on the runner

| run | order | wall | longest job | job-seconds | last row dispatched |
|---|---|---|---|---|---|
| 34346307562 | before this branch | 6:59 | ~5:30 (`go-test`) | — | — |
| 34350259348 | after the splits | 5:10 | 2:38 (`wire-fuzz`) | 6362 | 179 s |
| 34356391028 | + longest first | 3:39 | 2:41 (`wire-fuzz`) | 6410 | 81 s |
| 34357024672 | + longest first (green) | 4:28 | 2:33 (`wire-fuzz-retain-class`) | 6599 | 133 s |

**The longest job stopped being the constraint at run 34350259348.** It answered
in 5:10 with its longest job at 2:38 and its longest passing step at 2:10. The
82 jobs were eligible within 25 s of the run starting and did not run at once:
about twenty ran concurrently for the first two minutes, and the LAST row
started 179 s in. Fan-out latency, job start minus the completion of the matrix
job that gates it: 9 s for the first negative-control row and 148 s for the
last; 143-157 s for every conformance row, which was dispatched behind all 47
negative-control rows.

So the run's wall time was set by which rows were still going at the end, and
that was an accident of dispatch order: the 158 s and 145 s rows were dispatched
139 s and 162 s in and finished last, at 297 s and 307 s, while 25-second rows
had gone first.

**Both fan-outs now go longest first.** `make/negative-controls.json` records
each group's measured seconds and `negativecontrols matrix` sorts on it; each
`test/conformance/<lang>/ci.json` records its leg's seconds, which
`harness matrix` sorts on and then DROPS, so it never becomes a matrix field. A
row with no number sorts first — an unmeasured row is assumed heavy, because
that error costs a little wall time and the other costs a tail. Nothing runs
differently: same rows, same groups, same controls, same corpus, same order
inside each group. The fan-out latency for the last row fell from 148 s to
32-56 s, and every row was dispatched inside the first 81 s (133 s in the
green run, whose matrix job itself waited 101 s for a runner).

**Two minutes is not reachable this way, and the arithmetic says so.** The run
costs about 6400 job-seconds; the negative controls' own control steps are 2728
of them, and the per-job fixed cost — set-up, checkout, sibling clones,
toolchain, and the one build of `bin/schema` and the generated tree — is about
1400 more across 47 rows. At the concurrency these runs actually got (a mean of
20 jobs in flight in run 34350259348, a peak of 44; a peak of 54-58 in the two
new ones) the floor is job-seconds over concurrency, and it is minutes. Getting
under two would take either more concurrent runners or a smaller check set, and
the second one is the thing this PR promised not to do.

## What the measurement refused

  another split       the run is dispatch-bound, not job-bound. Every job pays
                      about 11 s of set-up, checkout, sibling clones and
                      toolchain before its first check, and 82 jobs already pay
                      900 s of it. One row more is one row further back.
  one corpus per      the wire-fuzz controls do not share an overlay: each
  control group       sabotages a different line of a different emitter file
                      and generates its own tree. There is no corpus for two of
                      them to share.
  a smaller corpus    the sabotage is in the EMITTER, so every generated unit is
  per control         potentially affected. Unit-local is not provable here, so
                      the corpus stays whole.
  a cached corpus     the objects a control compiles are of its OWN sabotaged
                      tree. There is no unsabotaged corpus in these jobs to
                      cache. (The Go build cache is a separate question, and a
                      separate branch: `make bin/schema` is 12-18 s of every
                      negative-control row, of which about 10 is `go build`.)

## Where the seven minutes were, measured

  pins             6:55  5:18 of it a SECOND `go test ./...` (update-goldens
                         closes with one). Its own pinning work: 1:23.
  big-endian       6:52  five independent legs plus a control, in one job:
                         tables 106 s, collections 12 s, cook 12 s, C 125 s,
                         control 127 s, Go 21 s.
  go-test          6:03  5:24 of it ONE package, ./compiler — 163 tests, none
                         of them parallel.
  conformance (c)  2:24  1:58 of it a single `cc` over ~70 translation units.
  negative controls (wire-fuzz rows) 2:20-2:40 — ~1:30 each is that same
                         single-process corpus compile, under a sabotaged emitter.

## What changed

1. pins asks for `update-goldens-only`, not `update-goldens`. The job's VERDICT
   is that testdata/ is byte-identical after the re-pin; when that is green the
   tree the closing suite would test IS the committed tree go-test and the
   shards already tested. When it is red the job fails at the verdict anyway.
   `make update-goldens` on a workstation still closes with the suite.

2. big-endian is a six-row matrix. `tables-big-endian` runs the same five
   targets in the same order off the same parallel prerequisite build, so
   `make test`, certify.yml and a workstation see no change;
   `tables-big-endian-negative-control` still runs the leg then the control.
   CI runs one per job, and the control's own recipe
   (`tables-big-endian-control-only`) beside the job that runs the leg.

3. ./compiler is sharded six ways by test name. Each shard asks
   `go test -list` for the whole list, sorts it and takes every sixth name, so
   the slices are a partition by construction and a new test lands in a shard
   the moment it exists. Each shard COUNTS the top-level verdicts in its output
   and refuses if that is not exactly the number of names it selected — a -run
   pattern that drifted off its targets is a shard testing nothing. `go-test`
   runs every other package and refuses if its filter leaves out more than the
   one package the shards carry.

4. Every corpus compile is one process per core. `corpus_objects` compiles each
   translation unit separately, nproc at a time, then links — same compiler,
   same flags, same -Werror, same units. It replaces the monolithic invocation
   in build/schema_test_tables_be, build/conformance-c, the two host-order
   compiles in the big-endian negative control, and `wire_fuzz_control` (which
   is every tolerant-wire negative control).

5. The big-endian C leg stops paying for a soak it did not ask for.
   `tables-c-big-endian` runs the soak binary for ZERO seconds, because the gate
   it exists for — every EXACT case re-saves to its own bytes on a big-endian
   build — runs before the clock. The loop's `(iterations & 0x3ff) == 0` clock
   check also set a floor of 1024 passes: 104 s under qemu. Zero seconds now
   stops at TWO passes, the fewest that keeps the leak comparison meaningful.
   SOAK_SECONDS > 0 reads the clock on exactly the schedule it always did.

6. Two scheduling trims. `negativecontrols check` ran FIRST in the matrix job,
   so forty jobs waited on a gate that cannot change what their rows are; it is
   `negative-controls-plan`, beside the matrix job, blocking the merge as
   before. And `negative controls (cs)` at 160 s was the longest job in the
   run: it is `cs` and `cs-packet`, all eight controls still on the PR.

## Checks that moved

None off the pull request. Within it:

  go test ./compiler          go-test              -> compiler-test 0..5
  go test ./... (the copy in update-goldens)
                              pins                 -> go-test + the shards
  tables-big-endian legs      one big-endian job   -> six big-endian rows
  tables-big-endian-negative-control
                              big-endian           -> big-endian (control)
  negativecontrols check      negative-controls-matrix
                                                   -> negative-controls-plan
  the four packet-*-cs controls
                              negative controls (cs) -> ... (cs-packet)

## The register, and the reader that reads a matrix

Run 34350259348 also went red twice, in `compiler tests (4 of 6)` and
`(5 of 6)`, and both were the same defect. The gate on `docs/PORTING.md` reads
`make <target>` out of the workflow TEXT, and `make -j"$(nproc)"
${{ matrix.target }}` puts no target on the line — so the six big-endian
targets went unreached the moment the leg became a matrix, and the three the
register names as carried (`tables-c-big-endian`, `conformance-big-endian`,
`tables-big-endian`) produced exactly the finding the gate exists to produce.
The gate's own negative control then failed for the reason it is built to fail:
it needs a green baseline.

The reachability walk now resolves `${{ matrix.<key> }}` against the job's OWN
`strategy.matrix`, which is what the tree really does — the same precedent
already in the walk for the `tables-<lang>-release` targets certify.yml
discovers rather than spells. Nothing is weaker: delete a matrix row and its
target goes unreached again. The cpp cell then cites the two pieces CI runs,
`tables-big-endian-tables` and `tables-big-endian-collections`, with the
umbrella named in the section's Reference where it belongs.

## Verified locally

go test ./internal/ci/, ./tools/negativecontrols/ and
./test/conformance/harness/, actionlint on all three workflows, gofmt -l,
go vet ./..., go build ./..., the modernize pass,
`go run ./tools/negativecontrols check` (220 controls), all six ./compiler
shards run exactly as ci.yml runs them — each selected==ran, and the union of
the six is byte-identical to the full `go test -list` output, so no test is
dropped and none is run twice — a cold
`make -j4 build/conformance-c` (58.6 s CPU in 9.1 s wall, 717 %), a cold
`make -j4 tables-wire-fuzz-length-negative-control` (green, and red the way it
must be), and the soak at 0 s (2 passes) and 2 s (1024 passes, unchanged).
The s390x rows need a runner with the pinned cross toolchain; the host half of
tables-big-endian-control-only compiles and links its 70 objects here.


rowan-claude:
Closed under the 2026-09-20 backlog drain: this branch no longer merges onto `main` `44dbbaf8` (conflicts with its base (1 file(s)) and has no read at its head), so a read bought now would be spent on a head that gets thrown away. The work is retained, not lost: branch `ci-under-two-minutes` is kept at `740adc0d8c3bfcedcea320a6301a70e4334284a7` (bundle `pr-790-740adc0d.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. A second-pass card will be cut from live `main` with prior: `ci-under-two-minutes@740adc0d8c3bfcedcea320a6301a70e4334284a7` in its body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #800, opened 2026-09-09
C++ tables: four widening helpers follow the kinds a unit declares

gafferongames:
Stacked on #794 (base is its branch; retarget to main once #794 merges). The follow-up #794 named: the widening runtime (`TableWidenF32`, `TableReadSignedAt`, `TableReadUnsignedAt`, `TableKindWidth`) is now emitted only in units whose census says a call site can exist. The census is a superset of emission: it asks which declared kinds the closure carries (f32 → WidenF32; signed kinds → ReadSignedAt; unsigned/flags/bits → ReadUnsignedAt; an arm on a ladder or any list → KindWidth) and never repeats the shape tests behind the eight `emitWidenedScalar` callers. `TableKindWidens` stays in every unit (generic walks and the message codec name it).

- 931 lines deleted from 48 committed C++ headers, 0 insertions; 16 lines from each of the three bench headers (all `TableKindWidth`); no unit references a helper it does not define, checked across all 40 generated units.
- `compiler/tablewidencpp_test.go`: six both-ways cases against a nearest neighbour; each positive fixture moves one declaration and must define, call, compile and run its helper at `-O1 -Wall -Wextra -Werror -fsanitize=address,undefined`; the neighbour carries none. Sabotaging the census turns all five positive cases red at the compile (undeclared identifier), reverted.
- Gates run in the child's clone: build/vet/test, tables + ASan, wire fuzz 171,387 mutants and retain fuzz 74,522 mutants with 0 divergences on both legs, negative controls all red, cpp conformance on every surface including forgery 12/12 and cook-forgery 113/113, maps/lists/arms/view/cook-open/cook-write, bench gate `b51387f36d9b59c4`, matched cpp/c/go legs `07c57b4fa34b2700`, PORTING register, shape gate, golangci-lint v2.12.2, modernize v0.23.0, regeneration byte-stable twice.
- Not run there: the eight non-cpp conformance legs (only C++ headers changed), `bench/paired -mode gate` end-to-end (no dotnet in the sandbox; the three matched binaries were run directly), full `make test`. CI here is the control.

No speed claim; this is dead generated output only.

### DIFF
--- a/compiler/tablewidencpp_test.go
+++ b/compiler/tablewidencpp_test.go
@@ -1,0 +1,284 @@
+// THE FOUR WIDENING HELPERS AND THE SUPERSET THAT CARRIES THEM
+// (docs/SPEC-TABLES.md §2.2, §3, §4).
+//
+// The C++ Table header used to carry the whole widening runtime in EVERY unit.
+// Four of its five functions are reached only from a kind the unit DECLARES:
+//
+//   - TableWidenF32, the float rung: a declared kind 11.
+//   - TableReadSignedAt: a declared kind 3, 4, 5 or 18, the signed ladder's
+//     tops, which are the only kinds that have a rung below them.
+//   - TableReadUnsignedAt: a declared kind 7, 8, 9 or 19.
+//   - TableKindWidth, the one place a payload WIDTH is a run-time fact: an ARM
+//     on a ladder, whose L must be the wire kind's own width (§3), and the
+//     list runtime's extent term, which asks for that width on any list.
+//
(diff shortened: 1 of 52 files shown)


gafferongames:
Retargeted to main: the CI workflow runs only on pull requests against main, so the stacked base got the CLA check and nothing else, and my body's "CI here is the control" was false as written. Until #794 merges this PR's diff shows #794's commit faf4cb65 too; the change under review is 7f6ee12d alone. Once #794 lands the diff collapses to that one commit.

PR #832, opened 2026-09-09
gotable: the fixed form (form 3)

gafferongames:
`docs/SPEC-TABLES.md` §3.4, **form byte 3**, for **Go**.

Branched from `origin/fixed-table-form` at **`3bf940e02790824c3311c37e7b05245fecff2efc`**. Two commits on this head: `86ca1058` (the form) and `29a7767a` (form-1 Load of a FixedLoad table is `previous_form`).

Johnny Grok.

---

## BYTES

C++ reference on **`3bf940e0`**. Corpus **80900 bytes**, first byte **3**, sha256 **`1fd0eb2185f19349870f9cfe3ef746f7077c7554a9ff55e00b301556e483fe8f`**.

The Go writer matches that corpus of the 64 paired mixed records.

## SHAPE

Go **packet codec**. Free functions, exported storage, `[N]byte` beside a length, byte-width stores instead of shifts.

**Not form-1 table paths.** References, kinds, lengths, probing are not copied; form 3 has none.

## File header tonight

Form byte, u32 LE layout length, layout, then records.

The form byte versions it; there is no version byte inside the layout. The word is LAYOUT, not block. Records are hash u64 LE plus body. Unknown kind is refused. Slack is zero on write. Union tag 0 is None. Optional present flag is one byte; the payload rides whole; flag 0 is ignored on read.

**UNPINNED:** waits on the rename card SHA (`form@0`, `hash@8`, zeros 1..7, `layout@16`).

THE READER IS ONE PLAN-DRIVEN PATH. Identity plan for this build's own hash; a compiled id-mapped plan for any other, with prefill of defaults; an unknown hash without a layout refused by name.

## Form-1 Load of a FixedLoad table is `previous_form` (`29a7767a`)

A form-1 file of a table that has FixedLoad/FixedSave is a named refusal, never a slow `tableOpen` walk.

**FixedLoad:** `previous_form` for form < 3, `newer_form` for form > 3. Form-1 Load of a variable table is unchanged.

The C++ **reference still accepts form 1** by the form byte. **Glenn says refuse**; this port does not copy that.

Held by `TestFixedForm`, `TestForm1OfFixedTableRefused`, `TestForm1OfVariableTableStillLoads`.

## Gaps not in this PR

- ranged-int clamp on load
- widen ladder stops before 128
- form-1 Save still `Put8(1)`

## Review

**Do not merge.** One independent read (Rowan child) after it is open.


gafferongames:
recheck

PR #836, opened 2026-09-09
tables: the fixed form (SPEC §3.4), and its C leg

gafferongames:
The NEW FIXED FORM (docs/SPEC-TABLES.md §3.4, form byte 3), through to a second
language. A fixed-table record is an eight-byte hash of the writer's vocabulary
block and then the values in declared order, every field at its bound: no
per-field references, no kind bytes, no lengths, no terminators, no trailer.

**The versioning invariant is the constraint every rule is shaped to meet** —
"we must not ever break versioning in fixed tables". The form is positional BY
PLAN and the positions are the WRITER's block, the block is always obtainable,
the block's own format never moves, and every refusal is by name.

## What is here

**The C leg, matching the C++ reference byte for byte.** The vocabulary block,
the hash, the record layout and the whole file are identical to the reference's
on the paired corpus, and a gate says so on every `make test`.

**One reader path, in C too.** A plan is a flat array of (source, destination,
size, op) entries and a read is one loop over it. The identity plan when the
record's hash is this build's own; a plan compiled once from the writer's block
otherwise. Same loop either way — no strict flag, no fast/slow cliff.

**The wire law moved to `ir`.** The constant size of every field, the pre-order
walk that becomes the vocabulary block, the block's bytes and its hash all
produce BYTES, so they live in `ir/fixedform.go` and neither backend computes
them a second time. The C++ output is byte-identical across that move.

**The identity plan is now the generator's, for every port.** It was a
`constexpr` walk the C++ compiler ran, coalesced by a class template. C has no
`constexpr` to run a walk with — so the schema compiler runs the walk and the
coalescer once, and every backend lays the finished array down as static data.
One answer for every port instead of one per language, and it took the last
`template` and the last `inline constexpr` out of the C++ table half, which
§13.9's C-dialect scan had gone red on.

Every destination in a generated plan is emitted back as a build-time assertion
against that compiler's own `offsetof` and `sizeof`, so a layout the generator
ever got wrong is a build error rather than a misplaced value.

**The paired bench measures form 3 as the table wire for C++ and C.** Both legs
rebuild `bench/paired/corpus/bench_fixed.bin` from the records they loaded out
of it and compare the whole file, and each compares its own vocabulary block
against the corpus's before anything else — the block settles the positions, the
ids, the kinds, the record size and the hash, so a leg that matches it is
speaking the form and not a near miss. The tolerant form still runs every
correctness gate it ran before, without a clock.

## Gates

- `make test SCHEMA_SKIP_LEGS=cs,dart,elixir,java,js`, `make generated-current`
- `go build ./... && go vet ./... && go test ./...`, `golangci-lint run`
- the C leg's fixed-form versioning conformance, plain and under ASan + UBSan
- the matched byte gate for C++ and C
- regeneration byte-stable twice

## Not here

The Go, C#, Dart, Elixir, Java, JS and Rust legs do not carry form 3 yet. The
paired driver's table corpus id now covers the two fixed corpus files, so a
leg's own runner has to load them before `bin/schema-paired -mode gate` will
accept its rows; C++ and C do, and the other legs are their own lines' work.


gafferongames:
### Child Review for PR #836: `tables: the fixed form (SPEC §3.4), and its C leg`

**Head SHA**: `6a2508c1a76282dab8d6d2b540d2b868fc7b9e06`
**Base**: `origin/main`
**Score**: 7/10
**Verdict**: **HOLD** (Rebase and Conflict Resolution Required)

---

### 1. 3-Way Merge Verification against `origin/main`
- Command: `git merge-tree --write-tree origin/main 6a2508c1a76282dab8d6d2b540d2b868fc7b9e06`
- **Result: FAILED (Exit Code 1)**
- Merge conflicts encountered across 14 files:
  - `.github/workflows/ci-full.yml`
  - `Makefile`
  - `compiler/tableselixir_test.go`
  - `compiler/tablesjs_test.go`
  - `internal/codegen/elixirtable/elixirtable.go`
  - `internal/codegen/gotable/gotable.go`
  - `internal/codegen/jstable/jstable.go`
  - `internal/tablenames/elixir.go`
  - `make/c.mk`
  - `make/cs.mk`
  - `make/elixir.mk`
  - `make/java.mk`
  - `make/negative-controls.json`
  - `make/rust.mk`
- `origin/main` has advanced 73 commits since the branch point (`c07c6792bf02d6729dcf0c926b67b1100d1a40f0`). A clean rebase onto `origin/main` is necessary before merging.

---

### 2. Diff Inspection & Static Analysis
- `gofmt -l .`: Clean (no formatting diffs).
- `go vet ./...`: Clean (exit code 0).
- Diff volume: 652 files changed, +226,232 / -10,417.
- Architecture:
  - Moves the wire law directly into `ir/fixedform.go` so neither backend recomputes field bounds, vocabulary blocks, or hashes.
  - Emits identity plans from the generator as static data rather than constexpr loops.
  - Implements the C leg fixed-form table reader in `internal/codegen/ctable/`, matching the C++ reference byte-for-byte.

---

### 3. Verification & Test Execution Evidence
- **Go Tests for Touched Packages**:
  - `go test ./internal/codegen/ctable`: **PASS** (41.5s, all 20+ tests passing, covering compiled/identity plans, ragged tails, under-20-bytes, and backwards compatibility).
  - `go test ./ir`: **PASS** (10.2s).
  - `go test ./compiler`: **PASS** (53.6s).
  - `go test ./cmd/schema`: **PASS** (85.0s).
  - `go test ./internal/codegen/cpptable`: **PASS** (12.1s).
  - `go test ./tools/fixedtwin`: **PASS** (10.1s).
- **C Leg Gates**:
  - `make tables-c-fixedform`: **PASS** (both standard and ASan/UBSan `schema_test_c_runcopy` and `schema_test_c_fixedform` passing across 2,536 byte flips).
  - `make tables-c-versioning` (`SCHEMA_REQUIRE_CORPUS=1`): **PASS** (§5 read both columns of every row against the C++ reference byte oracle).
- **Caveats & Failures**:
  - `make test SCHEMA_SKIP_LEGS=cs,dart,elixir,java,js`: Failed on `test/tables/fixedform_properties.cpp` (`tables-fixed-properties`) with 3 failures due to known-red status discrepancies.
  - `go test ./...`: Failed only on `internal/codegen/javatable` due to missing `javac` in the local Mac Studio bench environment.

---

### 4. Negative Controls / Teeth Verification
1. **Built-in Sabotage Gate**:
   - Ran `make tables-c-message-negative-control`.
   - Result: **PASS** (`go run ./tools/sabotage -name message-c-wrong-slot ...` confirmed that a sabotaged message slot immediately turns the independent wire comparison red).
2. **Mutation Testing**:
   - Sabotaged assertion in `internal/codegen/ctable/identity_test.go` (`TestIdentityPlanKeepsCountAndText`) to fail if `sawCount || sawText`.
   - Result: Test turned red immediately (`--- FAIL: TestIdentityPlanKeepsCountAndText (0.00s)`). Reverted cleanly.

---

### 5. Recommendation
The C fixed-form design and implementation in `internal/codegen/ctable/` are solid and correctly aligned with the specification and C++ reference oracle. However, because this draft branch is 73 commits behind `origin/main` with 14 merge conflicts and has 3 failing checks in `fixedform_properties.cpp`, the verdict is **HOLD** pending rebase and property test reconciliation.

READ-COST: reader=Emma model=gemini-2.5-pro seconds=824 tokens=unknown plan=max

rowan-claude:
CLOSE who=rowan rule=future: schema#1857

PR #838, opened 2026-09-10
ctable: the identity path copies without a default prefill

gafferongames:
`docs/SPEC-TABLES.md` §3.4 identity read, **C**.

Branched from `origin/fixed-table-form` at **`7502ba905765d7ed9e0f1f12ce240dd782ff9756`**. Head **`7beecacb95c88ef6646015255b377b5c91019eb0`**.

Johnny Grok.

---

Glenn: learn from Rust, backport the identity shape. Same three items as the C++ card, in `internal/codegen/ctable/`. **No wire byte moves.** Packed layout is not forced.

## (a) no default prefill on the identity path

This hash wrote every field. `*_reset` stays on the compiled path only, where a stranger's layout can omit a field and the declared default has to stand.

## (b) count and text clamps, straight-line after the copy

The identity plan is copies only (`tableFixedIdentityCopies`). Count and text ops used to split adjacent runs; they are COPY now, coalesced again, and the clamp (and the text terminator) runs after `table_fixed_run`. A record under this build's own hash can still carry a count this build does not admit.

The compiled plan still carries Count/Text. Clamps are not applied twice.

## (c) where the C ABI is the wire: memcpy, empty scatter

`ir.TableFixedFlatType`: no padding, no counted array, no text, no optional, no union, `sizeof` equals the body. Then identity read is `memcpy(values+k, at+8, sizeof(values[k]))` and `table_fixed_run` is not called.

Packed layout is **not** forced. A `uint8` then `uint32`, a count sitting after its array, or a `string(N)`'s extra terminator byte still scatters through the copy plan. Forcing `__attribute__((packed))` would break the C ABI the rest of the table surface already ships. BenchMixed does not memcpy (padding + count placement + terminator). Tally / TableStat / some Render rows do.

## Tests

- `go test ./internal/codegen/ctable/` — shape tests plus compiled C round-trip and hostile count/length clamp
- `make tables-c-fixedform` — identity plan, compiled plan, 2184 byte-flips, ASan twin
- `make tables-fixed-matched` — C and C++ both `OK (corpus_id 05a8822c4a461acb)`
- `make tables-block-zero-cost` — 123 Table sources byte-identical to their pins (C goldens re-pinned)
- `go test ./compiler/ -run 'CTable|TablesC|FixedRecord'`

## Review

**Do not merge.** One independent read after it is open.


gafferongames:
Superseded by #854 (the C+C++ one-path reshape, merged at f1871024). Two readers held this at aad9f467 as TWO PATHS; the one-path constraint is primary. Closed, not deleted; the FU1/FU2 leftover moves to the queue.

PR #840, opened 2026-09-10
fixed form: the coverage matrix and the fixtures that fill it

gafferongames:
The question this branch answers is not *"is the fixed form tested?"* but
*"what is it missing tests for?"*. The first is answered by a green gate; the
second only by an enumeration somebody sits down and writes.

**`docs/FIXED-FORM-COVERAGE.md`** is that enumeration: every shape §3.4 admits
and every rule it states, against the four ways this form can be got wrong —
the identity plan, a compiled plan in each direction, and the BYTES a writer
produces — with the fixture that holds each cell or the reason there is none.
123 cells covered, 47 filled by this pass across 17 gaps, 24 still open across
19, each open one with why.

## The write column did not exist

Every other wire in this project has pinned bytes and the fixed form had none.
What the versioning conformance set held is that a READER makes the right
values; nothing held that a WRITER made the right bytes. **A leg wrong in both
directions at once passes every round trip there is and cannot exchange one
record with anybody.**

`test/tables/fixedform_pin.cpp` writes a reviewable text oracle and
`testdata/conformance/tables/fixedform/records.dump` pins 22 cases of it, from
the same fixtures the read side asserts against
(`test/tables/fixedform_fixtures.h`) so the two cannot drift apart. It is
regenerated twice per run and both runs compared against the pin, because a
dump merely equal to its pin could still be picking up a clock.

This is the second oracle on the branch and neither is the other's copy.
`fixedform_dump.cpp` writes one form-3 FILE per root into `build/` for a PORT
to diff at test time; this pins text a PERSON diffs when a byte moves. Both run
under `make tables-fixedform-corpus`.

## Two schema pairs for the shapes nothing reached

- **FU1/FU2** — TEXT OF EACH FLAVOUR and a counted array under a UNION ARM, one
  arm per flavour in ordinal order so ordinal and flavour disagree at every arm
  but the first. This is the pair that made the `arg` lane reachable.
- **FN1/FN2** — a bool, an optional, an enum ordinal, an ARRAY OF UNIONS and a
  THREE-DEEP nesting in one fixed root, with a variant and an arm inserted in
  the middle and the innermost type resized.

Beside them the set now runs `Scalars/Scalars2`, `F1/F2`, `W1/W2` and
`examples/Ranges` through form 3, closing `flags`, `bits(N)`, the fixed-point
family, the 128-bit integers, the text family's declared defaults, a top-level
`wstring(N)` and a table renamed under `was`.

## Eight named known-reds, and one is not like the others

A fixture written for a bug the reference still has is the only kind that
proves the bug is real, so it lands the day it is written. The list is held
from BOTH ENDS: a listed case that starts PASSING turns the run red. **That
already happened once in this PR** — merging `fixed-table-form` brought fix 12,
two reds went green, the gate went red, and both entries are deleted here along
with the doc rows that named them.

**RED-3 is a memory-safety bug.** The run copy's 17..31-byte branch anchors its
last move BEFORE the run: it reads `32 - n` bytes in front of the source, writes
`32 - n` in front of the destination, and reads up to `32 - n` bytes past the
record body — which for the last record of a file is past the buffer. The
sanitizer names it a `heap-buffer-overflow` inside `TableFixedRun`.

It wears six faces on `Scalars` alone — `span`'s high half, `weights_count`,
`seeds_count`, `pose.heading`, `spawn_present`, and `tilt` at destination offset
zero on the compiled path — and every one comes back with **six zero counters
and a clean verdict**. A suite that checked only the report is blind to all six.
The branch is emitted for C++ and C and nobody else, so the reference is the leg
that is wrong and every other port is green.

The sanitized twin skips that one fixture, because a halted process reports
nothing behind it, and a gate re-runs it under `SCHEMA_FIXEDFORM_FAULT=1` and
REQUIRES the overflow. The day the run copy is fixed that gate goes red and the
skip and the gate are deleted together. A skip nobody watches fail is a skip
that has quietly become a hole.

## Three of the eight are questions for you, not bugs

- **RED-5** a bool byte outside `{0, 1}` — §3.4 gives `bool` a `C` of 1 and says
  nothing about the other 254 values, and the read is a copy, so the byte lands
  in the caller's `bool` and every read of it is undefined. The present byte
  takes the same path. One ruling closes both.
- **RED-6 / RED-7** an enum ordinal past the last variant, and a union tag past
  the last arm. The COMPILED plan resolves them to `None`; the IDENTITY plan
  copies the byte. Both are pinned on the PROPERTY and not the answer — **the
  two paths must agree** — which survives either ruling and is red today.

## Also open, and worth a look

- **GAP-16/17**, the STREAM and MESSAGE carriers. §3.4's framing table has three
  and every fixture here is a FILE. The largest thing left.
- **GAP-18**, a caller capacity below the file's record count. §3.4 states
  nothing; the reference refuses `batch_too_large` by precedent from §3.3 and
  this PR pins that — but §3.3 also requires the reader to hand back the count
  it was short by, and `FixedLoad` has no way to.
- **A `was =` chain.** `ir.Field.WasName` is a single name, so a field renamed
  twice carries the second name's hash and a first-generation record no longer
  resolves. Not fixed-form-specific, but this is the form where losing a field
  costs a whole positional slot.

## Gates

`make tables-fixedform` green with the reds named and printed; `make
tables-fixedform-corpus` green, oracle byte-identical to the pin and to a second
run; `go build`, `go vet`, `go test ./...` clean; `golangci-lint` 0 issues.


gafferongames:
### Review of PR #840: fixed form: the coverage matrix and the fixtures that fill it

**Head SHA**: `c3b7f60be8982a0b91547bed1cd36ab8b273ebce`
**Score**: 7/10
**Verdict**: **HOLD**

---

### Summary & Merge Verification

While the technical work, test fixtures, cross-language byte oracles, and negative controls in this PR are of exceptional quality, a **HOLD** is placed because the branch does not cleanly 3-way merge into `origin/main`:

* **3-way merge test**:
  `git merge-tree --write-tree origin/main c3b7f60be8982a0b91547bed1cd36ab8b273ebce`
  * **Exit code**: `1` (Merge conflict)
  * **Conflicts detected across 19 files**:
    * `compiler/tableselixir_test.go`
    * `compiler/tablesjs_test.go`
    * `generated/bench/paired/c/FixedTableTable.h`
    * `generated/bench/paired/cpp/FixedTableTable.h`
    * `generated/bench/tables/c/BenchTableTable.h`
    * `generated/bench/tables/cpp/BenchTableTable.h`
    * `internal/codegen/elixirtable/elixirtable.go`
    * `internal/codegen/gotable/gotable.go`
    * `internal/codegen/jstable/jstable.go`
    * `internal/tablenames/elixir.go`
    * `make/c.mk`
    * `make/cs.mk`
    * `make/elixir.mk`
    * `make/java.mk`
    * `make/negative-controls.json`
    * `make/rust.mk`
    * `testdata/golden/tables/blockhome/FrameTable.h`
    * `testdata/golden/tables/examples-c/TablesTable.h`
    * `testdata/golden/tables/examples/TablesTable.h`
  *(Note: PR #840 targets base branch `fixed-table-form`, but `origin/fixed-table-form` has also diverged with content conflicts in `Makefile`, `internal/codegen/cpptable/messagecodec.go`, and `test/tables/fixedform_main.cpp`.)*

A rebase / conflict reconciliation is necessary before this PR can merge.

---

### Verification Evidence

1. **Static Analysis & Linting**:
   * `gofmt -l .`: Clean (0 files formatted).
   * `go vet ./...`: Clean (0 issues).

2. **Fixed Form Test Suites & Oracles**:
   * `make tables-fixedform`: **PASS**. Both standard and AddressSanitizer test binaries (`schema_test_fixedform`, `schema_test_fixedform_asan`, `schema_test_fixedform_runcopy`, `schema_test_fixedform_runcopy_asan`) passed cleanly. Fuzz verification (378 bytes x 8 bits — 2376 refused by name, 0 malformed, 648 records read, 0 divergences) passed. The two known open reds (`optional/absent-payload-residue-is-copied` and `bool-domain/a-byte-outside-0-and-1-lands-in-the-caller-s-bool`) are accurately tracked and named.
   * `make tables-fixedform-corpus`: **PASS**. Fixed-form byte oracle generator produced 34 test cases byte-identical to `testdata/conformance/tables/fixedform/records.dump` and verified identical across two consecutive runs.
   * `make tables-fixed-properties`: **PASS**. Both baseline and AddressSanitizer builds passed with the 4 known property reds explicitly named and tracked.
   * `make tables-serialize-include-gate && make build/schema_test_tables && ./build/schema_test_tables`: **PASS**. Evaluated cleanly, including all generated JSON source units.
   * Go package unit tests (`compiler`, `internal/codegen/{cpptable,cstable,ctable,darttable,elixirtable,gotable,jstable,rusttable}`, `tools/fixedtwin`, `bench/paired`): **PASS**. (`javatable` fails only due to missing Java runtime in this local environment).

3. **Negative Controls (Verification of Test Teeth)**:
   * `make tables-fixedform-run-copy-negative-control`: **PASS**. Sabotaged the run copy bounds anchor back to `n - 32` via `tools/sabotage`; both C and C++ test drivers cleanly aborted with `Abort trap: 6` via AddressSanitizer detecting `heap-buffer-overflow`.
   * `tables-serialize-include-gate` Negative Control: **PASS**. Injected an unincluded `serialize::` symbol into `build/tables-generated/km1/KM1Table.h`. The gate immediately failed with exit status `2`, proving that self-sufficient header inclusion enforcement is effective.
   * `go test -v ./tools/negativecontrols`: **PASS** (all 234 negative controls properly registered and verified).

---

### Conclusion

The changes in commit `c3b7f60be8982a0b91547bed1cd36ab8b273ebce` are technically sound, robustly covered by oracles and negative controls, and pass all targeted test suites. Once the merge conflicts against `origin/main` (or the base branch) are reconciled and clean tree merge is restored, this will be ready for approval.

READ-COST: reader=Emma model=gemini-2.5-pro seconds=825 tokens=unknown plan=max

gafferongames:
### Audit & Verification: HOLD (7/10)

- **Head SHA**: `c3b7f60be8982a0b91547bed1cd36ab8b273ebce`
- **Target Base**: `origin/fixed-table-form`
- **Merge Status**: Conflicts in 3 files against `origin/fixed-table-form`:
  - `Makefile` (structural conflict with recent versioning additions)
  - `internal/codegen/cpptable/messagecodec.go` (comment conflict)
  - `test/tables/fixedform_main.cpp` (header includes and test driver registrations conflict with recent versioning cases)

#### Verification Evidence
1. **Implementation & Tests**:
   - Coverage matrix fixtures (24 test schemas) and negative controls pass cleanly when run on branch.
   - `make tables-fixedform`: PASS.
   - `make tables-fixedform-corpus`: PASS.
   - Negative controls (`tables-fixedform-run-copy-negative-control`, `tables-serialize-include-gate`) pass with teeth.
2. **Hold Reason**:
   - Requires rebase and reconciliation of test driver registrations and `Makefile` targets against latest `origin/fixed-table-form`.

rowan-claude:
CLOSE who=rowan rule=future: schema#1856

PR #841, opened 2026-09-10
elixirtable: the clamps ride the decode, not the plan

gafferongames:
## The finding this answers

PR #835's Elixir runner leg, on the loaded Studio: form 3 **save 4,454 ns/record, round trip 28,843 ns**. Of a 22,191 ns whole-file load, `FixedRuntime.run/4` — the plan interpreter — was **17,056 ns (77%)**, the generated decode (one binary pattern match) 3,259 ns, `:persistent_term.get/2` 24 ns.

The cause was the identity plan, not the interpreter. For the paired bench's root it was **260 entries** — 104 `copy`, **144 `clamp`**, 8 `guard`, 2 `count`, 2 `text` — and **250 of them had `src == dst`**. Every `clamp` broke the coalesced run on both sides of itself, so a record whose whole body is one contiguous move arrived as 260 separate moves plus a `:lists.keysort/2` and a 260-piece iolist. Elixir was the only backend emitting the ranged-integer clamp op at all; the C++ reference's identity plan carries `copy`, `count` and `text` and nothing else.

## Rowan's ruling, and what it means here

The ranged clamp on load **stays** — the read side always checks. But on the **identity path** it is straight-line code in the generated decode after the copy, never plan entries. Plan entries carry the clamp only on a **compiled** plan, where the interpreter runs anyway.

So:

* **The projection clamps.** `min(max(v, lo), hi)` sits inside the one binary pattern match, for a ranged integer, for a counted array's count, and for a text length. Fixed-point bounds are scaled by `2^F` before they are spelled, because a field's declared bounds are in value units and its image is the raw scaled integer (`fixedRangeOf`, unchanged and now read by both sides).
* **A `_fixed_clamped` pass counts.** One integer per record — how many bounds the record's values crossed — added to §4's counter by `FixedRuntime.clamped/2` in a single map update, and not at all when it is zero. Its head is **derived from the projection's own match segments**, with every binding it does not read turned into a skip and neighbouring skips merged, so a field's offset is a fact stated once and the two passes cannot drift.
* **A compiled plan is untouched.** `FixedRuntime.compile/5` still emits `clamp`, `count`, `text`, `widen` and `ordinal`. Because a clamp is idempotent, the projection's own clamp is then a no-op over an image already in range and counts nothing twice.
* **The union arm's guard goes too.** An arm's bytes are its own bytes whether or not the tag names them, and the projection reads the arm the TAG names and no other, so on this path the guard has nothing to decide. The `case tag` in `_fixed_clamped` is what the guard used to be.

What is left is `[{:copy, 0, 0, body_bytes}]` — and `assemble/2` already answers that by handing the body sub-binary back as the image without copying it.

**Why the count is a second pass and not a threaded accumulator:** a BEAM function returns two things only as a tuple, so threading would put one on the heap at every nested type and every element of every list — 136 tuples per bench record to carry a number that is almost always zero. The counting pass allocates nothing. Measured alternatives, same corpus, same machine: `for … reduce:` over the run **2,440 ns** and `Enum.sum(for …)` **1,682 ns** against the tail-recursive `R.clamps_each/4`'s **1,530 ns**.

## Plan entry count

| | before | after |
|---|---:|---:|
| paired bench root (`FixedTable`) | **260** (104 copy, 144 clamp, 8 guard, 2 count, 2 text) | **1** — `{:copy, 0, 0, 1236}` |
| every other fixed root in the leg (FX1, FX2, FE1, FE2, P1, P3, Pack, Keyed, Tables, Ranges — 25 roots) | 2 … 260 | **1**, all of them |

## Before / after

`go run ./bench/paired -mode fast -fast-rounds 3 -langs elixir`, medians of three rounds, run back to back on the same loaded Studio; the "before" is `rowan/elixir-fixed-form`'s tip `8c10bdc3` measured here, not quoted from #835. Non-certified fast diagnostic, no quiet-window claim.

| Wire | Path | before µs/op | after µs/op | |
|---|---|---:|---:|---|
| table | write | 4.487 (4.432–4.735) | **4.454** (4.447–4.526) | unchanged — the writer was not touched |
| table | round_trip | 29.453 (29.132–31.589) | **12.187** (12.173–12.293) | **2.42× faster** |
| packet | write | 2.270 | 2.337 | untouched leg, noise |
| packet | round_trip | 4.748 | 4.878 | untouched leg, noise |

### The load split, reproduced

Same script both sides — `:timer.tc` over the 64-record fixed corpus, 400 repetitions after warmup, ns/record:

| | before | after |
|---|---:|---:|
| `FixedRuntime.run/4` (the plan) | **22,415** | **21** |
| the generated projection (`_fixed_decode`) | 2,443 | 2,462 |
| the counting pass (`_fixed_clamped`) | — | 1,528 |
| whole-file load | **26,266** | **6,484** |

The plan interpreter's 77% is gone: one entry, and `assemble/2`'s single-run fast path returns the body sub-binary itself. The projection did not get slower — the clamps it gained are two BIF comparisons apiece. The counting pass is the new second-largest item at ~24% of the load, and it is where a follow-on would go: a **columnar** count (one filtered comprehension per BOUND over a whole run, rather than one call per element) measured **530 ns** in place of 1,530 ns for the bench's stat array. It is not in this PR because it only applies cleanly to some element shapes and this change is already the whole of the finding.

## Nothing on the wire moved

* `make tables-elixir-fixed-form` — FX1/FX2 (the field-edit pair), FE1/FE2 (the ordinal slide), P1/P3, KeyedConfig, PackConfig: every root read, saved back, and required byte for byte against the C++ reference's corpus.
* `make tables-elixir-fixed-form-negative-control` — one byte off the write template still reds the reference byte match.
* `make tables-elixir-paired-gate` — `OK (gate only, corpus_id f27a68d241e5be42)`, the same corpus id as before the change.
* `make tables-elixir-fixed-bench` — 64/64 records agree with the packet wire, 80915 bytes identical to the reference.
* `make test-elixir` in full, including the fuzz leg (20000 mutants) and the block-lead gate.

## The check this earns its keep by

The leg's `compiled_is_identity` now also requires that **a plan compiled from MY OWN layout and the identity plan count the same `clamped` over the same body** — the identity path checking in the projection, a compiled plan checking in the interpreter, two spellings of one rule. It runs over the corpus bodies **and over a body of nothing but `0xFF`**, where every declared bound is crossed at once, with a third assertion that the hostile body really did cross one; a corpus whose values are all in range would let both sides count nothing and call it agreement.

`TestIdentityPlanOfEverythingIsOneRun` states the fold in Go: a table with a ranged integer, a fixed-point range, an enum, a `string(N)`, a `bytes(N)`, an array of ranged structs, a counted array, an array of ranged scalars, a union and a nested type — one entry, `{:copy, 0, 0, body_bytes}`, and `fixedClampsType` still true so the checks did not simply vanish.

## Gates

* `go test ./internal/... ./ir/...` — green.
* `mix format --check-formatted` over every generated file and the leg's scripts — green. The emitter learned two formatter shapes it had not needed before: a `key:` whose value is a CALL hangs to its own line (a struct literal does not), and a comprehension breaks after its generator's comma.
* `elixirc --warnings-as-errors` over every generated unit — green.
* Regeneration byte-stable twice, and identical to the committed `generated/bench/paired/elixir/`.


gafferongames:
Closed by Rowan via Glenn's account: superseded. Its first commit a0125959 is the merge-base of rowan/elixir-fast (#850), which then deleted fixedclamp.go and folded the clamps into the identity decode; its second, 057350e7 (LIVE-only counters), was redone by Johnny on rowan/elixir-fixed-form as 99b2f47e on the current shape. The Elixir gate at 99b2f47e passes the slack and absent-optional counter probes (exit 0). The remaining intent, strict identity==compiled on the pack probe, lands on #850 once it merges #835's tip.

PR #844, opened 2026-09-10
tables: C and C++ fixed runtimes agree after the named twin map

gafferongames:
`make tables-fixed-twin` emits the paired C and C++ Fixed Table headers, normalises the fixed runtimes through the token map in `bench/paired/TWIN.md`, and diffs the rest. Any leftover line that is not one of the three named differences is a failure.

The three named remaining differences:

1. RAII guard vs inc/dec
2. the two 128-bit store functions
3. snake_case behind SCHEMA_UNUSED, the `as.` union prefix

A planted extra enumerator (`kTableFixedCopy = 99`) is the negative control; it reds.

The paired runner takes `--twin-tolerance` (default 10) so C and C++ rows of the same wire and path within that percent do not fail the twin check. No new wire. No changed bytes.

Johnny Grok.


gafferongames:
recheck

PR #846, opened 2026-09-10
the three properties, and the KNOWN-RED list they print

gafferongames:
Glenn asked whether nine ports could have found tonight's bugs cheaper. This harness is the answer.

`make tables-fixed-properties` builds the C++ binary beside `test/tables/fixedform_properties.cpp` twice (plain and ASan+UBSan) and runs:

- **P1** identity plan == compiled plan on every field and every counter
- **P2** write-read-write is byte-identical
- **P3** every byte of every record, mutated to `{00,01,02,7f,80,ff}`, both paths: every landed field is within its bound or the read is a named refusal, counters moved, under ASan+UBSan

A **KNOWN-RED** list is printed by name on a green run. A listed case that starts passing turns the run red (delete it from `known_red[]` as part of landing the fix). A new red that is not on the list fails the target.

`SCHEMA_FIXEDFORM_FAULT=1` on the sanitized twin requires the run-copy 17..31 overflow inside `TableFixedRun`. The day that overflow is gone, that gate goes red with this one.

## What it found on the tip (`b8cbcd26`, merged through `db52975f`)

Seven named reds, all still red:

| key | property |
|---|---|
| `text-content/identity/invalid-utf8-is-not-malformed` | P3 |
| `run-copy/identity/a-17-to-31-byte-run-clobbers-its-neighbours` | P3 / ASan FAULT |
| `optional/absent-payload-residue-is-copied` | P3 |
| `bool-domain/a-byte-outside-0-and-1-lands-in-the-caller-s-bool` | P3 |
| `ordinal-bound/paths-disagree/enum-ordinal-past-the-last-variant` | P1 on mutated ordinals / P3 |
| `ordinal-bound/paths-disagree/union-tag-past-the-last-arm` | P1 on mutated tags / P3 |
| `kind-mismatch/compiled/a-moved-kind-is-decoded-anyway` | dedicated compiled read of Scalars as Scalars2 |

Already green on this tip, so **not** on the list (both-ends would have failed if they were):

- clamp-op / tightened bounds (a17e1b0c)
- arg-lane (8c15973d)
- bytes(N) destination row (79e34542)
- text/array slack (860d9f6f)
- absent-optional **write** zeros (db52975f) — the remaining red is **read** of a peer's residue

Johnny Grok.


gafferongames:
recheck

PR #854, opened 2026-09-10
ctable, cpptable: one identity path — copy, scatter, hole prefill

gafferongames:
`docs/SPEC-TABLES.md` §3.4 identity read, **C and C++**, one branch.

Parents (not merged, not closed):

- `origin/rowan/cpp-c-identity-fast` at **`02d7de5ceacad9ca418ce6ca25e98384c5c2739a`** — C++ identity read is one copy plus a straight-line scatter; hole prefill is cover minus what the plan lands (empty on identity).
- `origin/johnny/c-identity-fast` at **`c4ca09b547499eca4054dae3ecaaa38260928273`** / **#838** — C sibling: identity tests hold Index and the flat dest's size; gcc array-bounds on a stack Flat.

Head **`cb88c0d17f4d6e33c025602ada731d22d273fede`**. Off Rowan's C++ tip. C identity is the same shape as C++, not #838's copies-only rewrite.

Johnny Grok.

---

Glenn: prefill the bytes the plan does not write. Identity hole list is empty; no Reset of the whole image on the identity path. Packed layout is not forced.

## What this is

One identity path in both languages:

1. **One copy of the body**, then a **straight-line scatter** of the identity plan (count and text stay plan ops, unrolled). A stranger's layout still spends the interpreter.
2. **Prefill only unwritten ranges.** `TableFixedIdentityCoverage` minus what the plan lands. Empty on identity, so the identity read writes no byte twice. Compiled path Reset is of the default image, once per read, only where `fill_count > 0`.
3. **Where storage image is the wire image** (`ir.TableFixedIdentityFlat` — the coalescer folded every leaf into one unguarded COPY of the whole body from offset zero): **one memcpy, empty scatter**, none emitted.

Clamps stay in the scatter / after the copy, as Rowan's branch already does. Live count only, slack never. Not reference-fix 21. Not a plan cache.

#838's `identity.go` (copies-only identity plan, skip-prefill flag, whole-image Reset on the compiled path) is **not** taken. Rust #837's skip-prefill-on-identity is not copied: Glenn wants hole prefill.

## Tests

- `go test ./internal/codegen/ctable/` — #838's Index / flat dest-as-pointer tests, held against this shape (gocritic offBy1; gcc `-Werror=array-bounds` on a stack Flat)
- `go test ./compiler/ -run 'CTable|TablesC|FixedRecord'` — `TableFixedFill` and the fill helpers claimed in `tablenames`
- `make tables-c-fixedform` — identity + compiled + 2536 byte-flips, ASan twin
- `make tables-fixedform` — C++ identity + 0 fuzz divergences, ASan twin
- `make tables-fixed-matched` — C and C++ both `OK (corpus_id 05a8822c4a461acb)`
- `make tables-fixed-twin` — C and C++ fixed runtimes agree after the named twin map (negative control reds). Twin-gate files are #844's (`4bbf9f21` / `a9e6da84` / `4f8f8583`) plus EntryLands pointer/ref, the same class as Apply. **#844 stays open.**
- `make shape-gate` — clean

## Clock (Studio, uncontrolled, not certified)

`bench_fixed` over the paired unit, matched binaries, same corpus id:

| leg | round_trip | derived read |
|---|---:|---:|
| cpp | 4.105 M msg/s (243.6 ns) | 6.133 M msg/s |
| c | 4.174 M msg/s (239.6 ns) | 6.490 M msg/s |

Environment-noisy versus Rowan's earlier 6.4 M sitting. Bytes match the oracle. Over packet would be broken; this sitting did not re-run packet.

## Review

**Do not merge.** Do not close #838. One independent read after it is open.


gafferongames:
recheck

PR #859, opened 2026-09-10
tables: the bounds pass is the clamp on both paths; the plan never clamps (one path, §3.4)

gafferongames:
## One codepath, matched or not

Glenn's ruling today, verbatim:

> "I wish for there to be one codepath always, whether version matches or not. This is also good because we have the one path exercised and have fewer cases to test."
> "Less is more."
> "The one path constraint is primary."

PR #856 (merge `87733fd3`) added a `kTableFixedClamp = 7` plan op, so a
**compiled** plan carried a clamp entry per ranged scalar while the identity
plan did not. A foreign read then clamped a scalar twice — once in the loop,
once in the idempotent storage pass — and the two paths differed in kind.
That is the cliff: a service on messages lives on the versioned path, and a
matched-path optimisation makes the two behave differently.

## What changed

This reverses #856's substance and nothing else.

Removed, in C++ (`internal/codegen/cpptable/fixedruntime.go`,
`fixedform.go`), C (`internal/codegen/ctable/fixedruntime.go`,
`fixedform.go`) and the IR (`ir/fixedform.go`):

- the `kTableFixedClamp` op and its `TableFixedApply` case
- `TableFixedLayBounds` / `table_fixed_lay_bounds`,
  `TableFixedPushClamp` / `table_fixed_push_clamp`, and the compiler's
  `skip_clamp` counted-array guard
- the compiler's emission of clamp entries after copy and widen
- `ClampWidth` / `ClampFlags` / `ClampLo` / `ClampHi` on `TableFixedDstSpec`
  and the four columns they printed into every generated `FixedDst` row —
  nothing else read them
- `tableFixedFillClamp`, `tableStorageRange`, `tableClampBits` in `ir`
  (the `tableStorageRange` copies in the codegen packages are older and
  untouched)
- the three names in `internal/tablenames/c.go`

The generated bounds pass (`<T>FixedClamp` / `<name>_fixed_clamp_`) stays and
is now the only clamp, after **either** plan, for every bounded field. #855's
compiled-plan cache is untouched.

The doc block in `ir/fixedform.go` now says NEITHER plan clamps, and the op-set
comment in both runtimes says the four compiled-only ops and that none of them
clamps. It also fixes a stale count that predates #856 ("the other two" →
"the other four").

## The controls

The controls that named the op are reshaped to the claim that now holds — the
PASS clamps on a compiled read:

- `test/tables/fixedform_main.cpp` §1b and §5, `test/c-tables/fixedform_fx1.c`,
  `test/c-tables/fixedform_v1.c`: compile a plan from this build's own layout,
  assert no op past `widenf`, run the loop alone and check the out-of-range
  value **stands** and counts nothing, then run `FixedClamp` and check it holds
  the value and counts the same numbers the identity path counted. The
  counted-array case still proves live count, never slack — now on both paths.
- `compiler/tablefixedclamp_test.go`: `TestCppNoPlanClamps` /
  `TestCNoPlanClamps` replace the two `IdentityPlanHasNoClampOp` tests — the
  op must not exist at all, the op set stops at `widenf`, the pass is still
  emitted, and a destination row carries no clamp ends.

## Spec

`docs/SPEC-TABLES.md` §3.4: `clamp` is gone from the op table, and "THE ONE
READER PATH: A PLAN" now states that a read is one prefill, one loop over the
plan the layout hash selected, then one bounds pass over the storage the loop
wrote — the same pass for either plan; that the identity plan carries copy,
count and text and a compiled plan adds only the foreign-layout ops; and that
no plan op clamps. The read pseudo-code now shows the pass.

## Tests

```
$ go test ./ir/... ./internal/codegen/cpptable/... ./internal/codegen/ctable/... ./compiler/...
ok  	github.com/mas-bandwidth/schema/v2/ir	(cached)
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/cpptable	(cached)
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	(cached)
ok  	github.com/mas-bandwidth/schema/v2/compiler	186.379s
```

```
$ make tables-fixedform
./build/schema_test_fixedform
FX1 FxRoot: body 64, layout 225, identity plan 6 entries
fuzz: 378 bytes x 8 bits — 2376 refused by name, 0 malformed, 648 records read, 0 divergences
fixed form: versioning conformance green
./build/schema_test_fixedform_asan
FX1 FxRoot: body 64, layout 225, identity plan 6 entries
fuzz: 378 bytes x 8 bits — 2376 refused by name, 0 malformed, 648 records read, 0 divergences
fixed form: versioning conformance green
```

```
$ make tables-c-fixedform
./build/schema_test_c_fixedform
the fixed form versions, on the C leg: the identity plan, a plan compiled from a stranger's layout, and 2536 byte flips answered inside the buffer
./build/schema_test_c_fixedform_asan
the fixed form versions, on the C leg: the identity plan, a plan compiled from a stranger's layout, and 2536 byte flips answered inside the buffer
```

The negative controls that name clamps, all green:

```
$ make tables-clamp-limits-negative-control
negative control: the storage-limit clamps back turn the build red on 19 comparisons that cannot fire
$ make tables-json-clamp-prefix-negative-control
negative control: the clamp prefix control sees the defect under the sabotage and nothing without it
$ make wide-table-clamp-negative-control
negative control (table-wstring-clamp-splits-a-pair): the gate goes red on FAILED: units_equal( out.title, out.title_length, v.expect, v.expect_count ) on vector wstring-table-clamp-never-splits-a-pair (test/wide/table_main.cpp:244)
$ make tables-maps-clamp-negative-control
negative control: clamp turns the MAP GATE red on 2 failures
$ make tables-lists-clamp-negative-control
negative control: clamp turns the LIST GATE red: 4 failures
```

The whole chain, every leg, and the byte-stability of the regeneration:

```
$ make generated-current
...
generated/ tree is current
```

## Regeneration

Every generated tree was produced by the compiler; no generated file was edited
by hand. `generated/bench/{paired,tables}/{c,cpp}` were regenerated from their
stamps, the header goldens under `testdata/golden/` by
`go test ./internal/goldens -update -run TestGolden`, and `make generated-current`
re-emits the whole tree and refuses on any byte that differs from the commit.

## One thing the reversal could not do quietly

#856's C test called `cfg_fixed_load` with seven arguments where the emitter
declares eight (the plan cache is the seventh) — `make tables-c-fixedform` could
not have compiled at `87733fd3`. The reshaped test passes `NULL` for the cache,
matching every other call in the file, so the C leg builds again.

---

Rowan via Glenn's account.


gafferongames:
recheck

gafferongames:
recheck

gafferongames:
recheck

PR #868, opened 2026-09-11
elixirtable: compare the union guard at ArgW, not one cell

gafferongames:
Elixir still matched the union guard as one cell (`<<_::binary-size(^gsrc), ^tag, _::binary>>`). `tag_bytes/2` sizes the const, not the guard compare. C++ and IR compare at ArgW (1/2/4/8, zero means one). A two-byte tag whose low byte happens to be 1 is not arm 1.

## What changed

- `tag_at` in the Elixir runtime: little-endian load at ArgW bytes, never a prefix. ArgW 0 means 1, ArgW > 8 clamps to 8, as C++.
- Guard tuple is `{:guard, gsrc, tag, argw, inner}`, stamped from the compiler the way C++ stamps `e.argw` (the writer's tag width). Flavour stays on the text op.
- `step/4` compares the guard at ArgW. Hash still chooses the plan and nothing else. Identity is one COPY of the body. `assemble/2`'s single-run clause is the skip when the one write covers the image, not a second reader. No memcpy identity. ONE PATH.
- Did not re-introduce a `{:clamp}` plan op.
- Tests: `TestFixedGuardComparedAtArgW` (generated runtime) and `test/elixir-fixedform/main.exs` (0x0101 does not run arm 1; 0x0001 does; four-byte twin).

## Gates (cheap)

- `go test ./internal/codegen/elixirtable`
- `make tables-elixir-fixed-form`

Off `origin/fixed-table-form` `2d6a4940bbe118c32f9aaa7074f670954d14b372` (Merge #839). Johnny Grok.

Still draft. Do not merge.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `johnny/elixir-argw-guard` is kept at `956e00b62e0ed9333037d8559c92a66c6e82995b` (bundle `pr-868-956e00b6.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `johnny/elixir-argw-guard@956e00b62e0ed9333037d8559c92a66c6e82995b` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #882, opened 2026-09-11
fixed form: the seven layout refusals as fixtures on JS, Rust, C and Elixir, red first

gafferongames:
## The seven hostile-layout refusals, as SHARED fixtures — RED FIRST

`docs/FIXED-FORM-ALGORITHM.md` §1.1 names seven refusals a fixed-form reader raises on a hostile
LAYOUT, before it touches a single record byte. The C++ reference asserts all seven by name
(`test/tables/fixedform_main.cpp`, `layout_validation()`), and so do C#, Java and Dart. **JS, Rust
and C asserted none of them; Elixir asserted five, one of them weakly.** A rule a port got wrong
was a rule no other leg would catch.

The house rule is that a bug found by one port becomes a fixture in the SHARED conformance set. So
these are the reference's own twelve cases — the seven rules, four in both directions, plus the
residue `layout_malformed` — ported to every leg that lacked them, **fixtures before runtime fixes**.
The reds below are not a regression. They ARE the measurement of the gap.

**This PR touches no runtime and no emitter.** Go is another worker's card; JS and Rust runtime
fixes are cards of their own. One line of wiring only: the C leg's new translation unit into
`C_FIXEDFORM_SOURCES`.

### Leg by rule

| # | rule | refusal | C++ | **C** | **Elixir** | **JS** | **Rust** |
|---|---|---|---|---|---|---|---|
| 1 | count fits the length exactly | `layout_count_mismatch` | name | PASS | PASS | **RED** | **RED** |
| 1 | a count of zero is not a layout | `layout_count_mismatch` | name | PASS | PASS | **RED** | **RED** |
| 2 | every kind in the closed set | `layout_kind_unknown` | name | PASS | PASS | **RED — DECODED** | **RED** |
| 4 | the root entry is a TABLE | `layout_kind_invalid` | name | PASS | PASS | **RED — DECODED** | **RED** |
| 3 | a size its kind does not admit | `layout_size_mismatch` | name | PASS | PASS | **RED — DECODED** | **RED** |
| 3 | a table's size is the sum | `layout_size_mismatch` | name | PASS | PASS | **RED** | **RED** |
| 5 | the tree runs out of layout | `layout_tree_unclosed` | name | PASS | PASS | **RED** | **RED** |
| 5 | the layout outlasts the tree | `layout_tree_unclosed` | name | PASS | PASS | **RED** | **RED** |
| 6 | a record size past 65536 | `layout_record_too_large` | name | PASS | PASS | **RED** | **RED** |
| 6 | a size that would wrap the sum | `layout_record_too_large` | name | PASS | PASS | **RED — DECODED** | **RED — PANIC** |
| 7 | nesting past the walk's bound | `layout_too_deep` | name | PASS | PASS | **RED — DECODED** | **RED** |
| — | the residue | `layout_malformed` | name | PASS | PASS | PASS | **RED** |

C: **12/12 PASS** — the C runtime already carried all ten layout reasons; only the fixture was
missing. Elixir: **12/12 PASS**, with `layout_record_too_large` and `layout_too_deep` newly asserted
and the weak three-name check tightened to one name. JS: **18 failures**. Rust: **17 failures**.

### Two findings worse than a wrong name

**JS ACCEPTS a hostile layout and decodes records.** Rules 2, 3, 4 and 6 are not implemented at all,
so five cases do not refuse under the wrong name — they do not refuse. The reader reports `None`,
returns a decoded record, and moves §4 counters on bytes whose layout is a forgery:

```
FAILED: RULE: a kind outside the closed set is REFUSED, not skipped — expected LayoutKindUnknown, got None and a read of 1 record(s)
FAILED: RULE: a kind outside the closed set is REFUSED, not skipped: a layout refusal sets nothing and counts nothing (unknown 2, kindMismatch 2, widened 0, clamped 0, malformed false)
FAILED: RULE: the root entry is a TABLE — expected LayoutKindInvalid, got None and a read of 1 record(s)
FAILED: RULE: a constant size its kind does not admit — expected LayoutSizeMismatch, got None and a read of 1 record(s)
FAILED: RULE: a size that would overflow the sum — expected LayoutRecordTooLarge, got None and a read of 1 record(s)
FAILED: RULE: a nesting depth past the walk's own bound — expected LayoutTooDeep, got None and a read of 0 record(s)
```

**Rust PANICS on a stranger's layout**, on the very case the reference wrote to prevent it — rule 6's
wrapping sum. §1.1 says the children's sizes are summed in 64 bits *precisely so* a `u32` that
overflows is caught rather than wrapped into a small number that then agrees with its parent:

```
thread 'main' panicked at build/tables-generated-rust/tblfx1/src/fixed_runtime.rs:743:13:
attempt to add with overflow
FAIL: RULE: a size that would overflow the sum: expected LayoutRecordTooLarge, got a PANIC on a stranger's layout
```

That is the debug build. **A release build does not panic — it wraps**, which is the accepted-forgery
path the rule exists to close. The debug panic also aborts the run, so `cargo run --release` (the
target's second line) never executed.

### The rest of the red, verbatim

Rust collapses every rule into `BlockMalformed` — and note it says *Block* where the doc says
*Layout* (`NoBlock`/`BlockMalformed` against `no_layout`/`layout_malformed`), a naming divergence
beyond the seven:

```
FAIL: RULE: the entry count fits the layout length exactly: REFUSED BY NAME — expected LayoutCountMismatch, got BlockMalformed
FAIL: RULE: an entry count of zero is not a layout: REFUSED BY NAME — expected LayoutCountMismatch, got BlockMalformed
FAIL: RULE: a kind outside the closed set is REFUSED, not skipped: REFUSED BY NAME — expected LayoutKindUnknown, got BlockMalformed
FAIL: RULE: the root entry is a TABLE: REFUSED BY NAME — expected LayoutKindInvalid, got BlockMalformed
FAIL: RULE: a constant size its kind does not admit: REFUSED BY NAME — expected LayoutSizeMismatch, got BlockMalformed
FAIL: RULE: a table's size is the sum of its fields': REFUSED BY NAME — expected LayoutSizeMismatch, got BlockMalformed
FAIL: RULE: the tree runs out of layout: REFUSED BY NAME — expected LayoutTreeUnclosed, got BlockMalformed
FAIL: RULE: the layout outlasts the tree: REFUSED BY NAME — expected LayoutTreeUnclosed, got BlockMalformed
FAIL: RULE: a record size past 65536: REFUSED BY NAME — expected LayoutRecordTooLarge, got BlockMalformed
FAIL: RULE: a nesting depth past the walk's own bound: REFUSED BY NAME — expected LayoutTooDeep, got BlockMalformed
FAIL: RULE: fewer bytes than a header is layout_malformed: REFUSED BY NAME — expected LayoutMalformed, got BlockMalformed
rust fixed form: 17 FAILED
```

JS, where a name does exist but is the wrong one:

```
FAILED: RULE: the entry count fits the layout length exactly — expected LayoutCountMismatch, got LayoutMalformed
FAILED: RULE: an entry count of zero is not a layout — expected LayoutCountMismatch, got LayoutMalformed
FAILED: RULE: a table's size is the sum of its fields' — expected LayoutSizeMismatch, got None
FAILED: RULE: the tree runs out of layout — expected LayoutTreeUnclosed, got LayoutMalformed
FAILED: RULE: the layout outlasts the tree — expected LayoutTreeUnclosed, got LayoutMalformed
FAILED: RULE: a record size past 65536 — expected LayoutRecordTooLarge, got None
```

### How the assertion is written on JS and Rust, and why

On JS and Rust the doc's names **do not exist** as enum members. Naming them directly would be one
opaque red on JS (`undefined`) and a compile error on Rust that takes the whole leg down and measures
nothing per case. So both legs assert the refusal's **own member name as a string** — JS via a
reverse map over the frozen `TableFixedRefusal`, Rust via `format!("{:?}", report.reason)`, which the
derived `Debug` makes the variant name. This is exactly as strong as comparing the enum value, it
gives each of the twelve cases its own red, and **it passes unchanged once the runtime cards add the
members** — at which point it can be tightened to the enum value itself. Flagged for a reviewer: it
is a deliberate departure from "an enum in Rust".

Every case also carries the reference's second check — a layout refusal sets nothing and counts
nothing — and each leg asserts the UNBROKEN file reads clean first, so every red is the one break and
not the file.

### Targets and times

No new targets. All four are already reached by `make test` (via `test-c`, `test-js`, `test-rust`,
`test-elixir`), which `certify.yml` runs on both OSes, so these reds are visible without wiring.
The `conformance` matrix job in `ci.yml` does *not* run them — it runs `build/conformance-*` only.

| command | cold | warm | result |
|---|---|---|---|
| `make tables-c-fixedform` | 3.86s | 0.43s | PASS (strict + ASAN) |
| `make tables-js-fixed-form` | 3.79s | 0.24s | **18 FAILED** |
| `make tables-elixir-fixed-form` | 3.03s | 2.27s | PASS |
| `make tables-rust-fixedform` | — | 0.53s | **17 FAILED** |

Nothing near the two-minute rule.

### Notes for the reviewer

- **The C leg needed its own buffer.** Case 11 is 4097 entries — 69,673 bytes of file, past
  `fixedform_main.c`'s 65,536-byte `g_buffer`. A static array in the new TU, no malloc, ASAN clean.
  `kTableFixedMaxDepth` is 64, so the walk bails at depth 65.
- **Hash-last was verified, not assumed.** Breaking bytes inside a good layout only measures the rule
  if the header's hash is checked after validation (§2 step 5). Confirmed in
  `internal/codegen/ctable/fixedform.go`; Elixir's existing `refuse` helper rebuilds the header hash
  anyway.
- **Case 12 reaches a different code path on C** than the reference's prose implies: a 2-byte layout
  trips the layout-header check (`length < 4`), not the file-length check. Same name, same reason,
  different test — worth knowing for a leg that splits the two.
- **Case 5 is pinned.** The C fixture asserts entry 1 really is a 4-byte `uint32` leaf, so the case
  cannot silently stop exercising the size rule if a field moves.
- **The Rust leg does not build from a clean clone**, before any change here:
  `build/tables-generated-rust/benchfixed/Cargo.toml` wants `serialize = { package =
  "serialize-official", path = ".../serialize.rs" }` and the sibling checkout declares `name =
  "serialize"`. CI clones that sibling at a pinned `SERIALIZE_RS_TAG`, so the pin evidently renames
  it. Worked around **only in a scratch sibling** to get the measurement; no repo file touched. If
  CI's Rust row is green on this branch, the pin is fine and only local checkouts drift.

### DIFF
--- a/make/c.mk
+++ b/make/c.mk
@@ -908,6 +908,7 @@ build/tables-generated-c-fixed/.stamp: bin/schema test/tables/FX1.schema test/ta
 	@touch $@
 
 C_FIXEDFORM_SOURCES := test/c-tables/fixedform_main.c test/c-tables/fixedform_fx1.c test/c-tables/fixedform_fx2.c \
+	test/c-tables/fixedform_layout.c \
 	test/c-tables/fixedform_v1.c test/c-tables/fixedform_v2.c \
 	test/c-tables/fixedform_ut1.c test/c-tables/fixedform_ut2.c \
 	test/c-tables/fixedform_fu1.c test/c-tables/fixedform_fu2.c
--- a/test/c-tables/fixedform.h
+++ b/test/c-tables/fixedform.h
@@ -72,6 +72,16 @@ void fixed_v1_absent_optional( void );
 void fixed_fx1_text_content( void );
 void fixed_guard_width( void );
 
+/* THE LAYOUT VALIDATION (docs/SPEC-TABLES.md §3.4,
+   docs/FIXED-FORM-ALGORITHM.md §1.1). A layout arrives from an UNTRUSTED PEER
+   and is the one structure a reader must parse before it knows anything at
+   all, so §1.1's SEVEN named refusals each get a case of their own: one
+   hostile layout per rule, each breaking EXACTLY ONE THING in a layout this
+   reader accepts, plus the residue — bytes that are not a layout at all. It is
+   the FX1 reader on a file built around FX2's layout, so the bytes arrive as a
+   parameter the way they do for fixed_fx1_read_fx2. */
+void fixed_fx1_layout_validation( const uint8_t * data, int64_t bytes );
+
 /* THE BYTE-FLIP FUZZ's reader (docs/SPEC-TABLES.md §3.4, "held by test"): it
    makes no claim about the values, only that the read answers one of the three
    ways the form allows and never leaves the buffer doing it. Every offset this
--- a/test/c-tables/fixedform_layout.c
+++ b/test/c-tables/fixedform_layout.c
@@ -1,0 +1,254 @@
+/* THE LAYOUT VALIDATION, C leg (docs/SPEC-TABLES.md §3.4,
+   docs/FIXED-FORM-ALGORITHM.md §1.1). The C++ reference is the
+   layout_validation() function of test/tables/fixedform_main.cpp and this is
+   its twin: the same twelve hostile layouts, refused under the same names.
+
+   THE LAYOUT ARRIVES FROM AN UNTRUSTED PEER. It is the one structure a reader
+   must parse before it knows anything at all, so every rule it is held to
+   refuses under ITS OWN NAME, before a single record byte is touched. A
+   VALIDATION NOBODY WATCHED FAIL IS A VALIDATION NOBODY HAS, so there is one
+   case per named rule, each taking a layout this reader ACCEPTS and breaking
+   EXACTLY ONE THING in it.
+
+   THIS IS THE FX1 READER ON A FILE BUILT AROUND FX2'S LAYOUT, which is what
+   the reference does too (its `good` is written by tblfx2 and read by tblfx1).
+   C has no namespace, so the two generations cannot share a translation unit
+   (fixedform.h says why): the FX2 bytes arrive as a parameter, exactly as they
+   do for fixed_fx1_read_fx2.
+
+   The file is the HEADER (docs/SPEC-TABLES.md §3: the form byte, seven
+   reserved zero bytes, the layout hash at 8, the body at 16), then `u32 layout
+   length, layout, records` — so the layout starts at byte 20, the entry count
+   is the four bytes there, and entry k is the seventeen bytes at 20 + 4 + 17k:
+   id (u64), kind (u8), size (u32), children (u32), every number
+   little-endian.
+
+   THE HEADER'S HASH IS NOT RECOMPUTED AFTER A BREAK, and that is deliberate
+   rather than sloppy. Changing a byte inside the layout makes the hash at
+   offset 8 no longer the hash of the layout behind it, and the algorithm doc
+   §2 step 5 says that hash is checked LAST so the layout's own rules refuse
+   first. VERIFIED on this leg: the generated load (internal/codegen/ctable's
+   fixedform.go) parses and validates the layout in the not-my-hash branch and
+   only then compares the header hash, under a comment that says exactly that.
+   So each case below refuses by its own name and never as a lying header. */
+
+#include <string.h>
+
+#include "FX1Table.h"
+#include "fixedform.h"
+
+/* the plan storage the caller owns; no layout below ever gets as far as
+   compiling, which is the point — a refusal costs no plan */
+#define PlanCapacity 1024
+static TableFixedEntry g_plan[PlanCapacity];
+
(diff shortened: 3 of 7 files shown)


gafferongames:
Superseded by the contract (bill §12.4, ALGORITHM §5.3): against a KNOWN hash the seven layout refusals collapse to layout_malformed and an unknown hash is layout_newer before any walk; each leg's versioning suite carries hash_known_bytes_differ for that. The seven refusals stay as the lock's and the oracle's validation (#909, #916). Closed with Johnny's agreement, 09:26Z.

PR #899, opened 2026-09-11
monotone law, lists: the baseline refuses what a fixed record cannot survive

gafferongames:
The LISTS half of the bill's monotone law (`docs/FIXED-FORM-BILL-READS-BACKWARD.md` §6): every list that inputs into a `fixed table` only ever grows AT THE END, and the baseline refuses any commit that is not a widening of the list it holds. `internal/baseline/monotone_lists.go`, one entry point `monotoneLists(old, new *Unit) []string`, hooked into `Diff` in one line marked `// monotone law, lists (bill §6)` so `monotoneNumbers` lands beside it without a conflict.

The shape of the check is the bill's own prefix walk: the committed list must be a PREFIX of the declared one, entry by entry, and the declaration's tail is the append. The FIRST differing entry is reported, and every refusal names the fixed table, the definition and the rule, with the remedy: `fixed <Table>: <definition>: <rule>; <remedy> (citations)`.

## Every test, by the sentence of Glenn's it holds

**"types or tables referenced in a fixed table can only have new fields added at end, not existing entries modified or removed (they can be deprecated sure)"**

- `TestAFieldRemovedFromAFixedTablesClosureIsRefused` — a field dropped from a `type` the fixed table holds by value, and one dropped from the fixed table itself.
- `TestAFieldInsertedNotAtTheEndIsRefused` — a field inserted mid-list.
- `TestFieldsReorderedAreRefused` — two fields swapped.
- `TestAFieldsKindModifiedIsRefused` — the one representative MODIFY case this file holds (`float32` to `int32`); widths, bounds, sizes, ranges and `frac` are the numbers half, next door.
- `TestAFieldDeprecatedInPlaceIsAllowed` — ALLOWED: `| deprecated` keeps the slot and every wire fact on it.
- `TestAFieldAppendedAtTheEndIsAllowed` — ALLOWED, Glenn's own example: `vec {x,y,z}` to `vec {x,y,z,w}` passes, to `vec {x,y}` refuses; and an append on the fixed table's own bottom.

**"enums can have entries added at end, no shuffling of entries for meaning, new entries at end"**

- `TestAnEnumVariantRemovedIsRefused`
- `TestAnEnumVariantInsertedMidListIsRefused`
- `TestEnumVariantsReorderedAreRefused`
- `TestAnEnumVariantRenamedWithoutWasIsRefused` — a variant rides under the hash of its wire name, so a rename with no `was` puts a name no stored value was written under at that place; `was` keeps the id and passes.
- `TestAnEnumVariantAppendedIsAllowed` — ALLOWED, and it carries the answer to the deprecation half: THIS LANGUAGE HAS NO PER-VARIANT DEPRECATION MARKER (`| deprecated` is a FIELD tag, §2.10), so an enum in a fixed closure has exactly one allowed edit — append.

**a union arm: the same five**

- `TestAUnionArmRemovedIsRefused`, `TestAUnionArmInsertedNotAtTheEndIsRefused`, `TestUnionArmsReorderedAreRefused`, `TestAUnionArmRenamedWithoutWasIsRefused`, `TestAUnionArmAppendedIsAllowed` (ALLOWED).

**"Flags can only ever have new flags added. Old flags cannot be removed"**

- `TestAFlagRemovedIsRefused`, `TestAFlagMovedIsRefused`, `TestAFlagAppendedIsAllowed` (ALLOWED). These hold the sentence WHERE THE LAW ALREADY LIVES: bit i is variant i on every wire, so this package's `flags-position` row already refuses a flag removed, inserted or moved for fixed and variable alike. `monotoneLists` deliberately adds no flags rule rather than refusing the same edit twice.

**"fixed tables must only allow other fixed tables to be included in them, and not recursively include themselves"**

- `internal/check`: `TestFixedTableClosureRefusals` already covers the first half, a row per construct — a map, an unbounded array, a pointer, a byte buffer, a guard, a nested plain `table` by value and through a bounded array, a pointer arm. NEW here: `TestAFixedTableCannotReachItself` — itself by value, itself in a fixed array, itself through a `type` (a stronger refusal: a table cannot ride in a `type` at all), itself through a pointer (refused as a pointer in a fixed closure, which is why no path exists).

**the `fixed` keyword: a different FORM, not a version**

- `TestTheFixedKeywordDroppedIsRefused` — `fixed` removed from a locked fixed table.
- `TestTheFixedKeywordAddedToALockedVariableTableIsRefused` — `fixed` added to a locked variable one.
- `TestTheProjectionRecordsTheFixedKeyword` — the format addition itself, and that a baseline written before it parses and stays silent on an untouched schema.
- `TestAVariableOnlyUnitKeepsItsOwnRules` — the law's SCOPE: the same three edits on a unit with no fixed table are judged exactly as they were.

## The one format token added

`fixed=true`, at the END of the `table` line (`table Vessel name=Ship fixed=true`), the one fact the law needs and the projection did not carry. **It bumps no rendering version**, by §18.1's own test: the token rides at the end of a line every older file still parses, and the rules read it on the COMMITTED side, so a baseline written before this rendering carries it nowhere and greets an untouched schema with no diagnostic. The cost of that choice, stated plainly: a unit whose committed baseline predates the token is unguarded against the keyword moving until its next `--update`, and `monotoneForm` therefore judges the keyword only where the committed file speaks the fact at all (one `fixed=true` anywhere). **If Glenn would rather the guard be total from the first check, the change is a rendering version bump (7 to 8) and a `--update` for every committed baseline.** A map's generated ENTRY, declared by nobody (§2.8), carries no keyword and roots none of this law.

14 committed corpus baselines are re-pinned with the token (history preserved verbatim, no history entry: the projection gained a recorded fact and no schema moved).

- `TestOneLinePerBreakNotOnePerFixedTable` — the output's BOUND: a `type`, an enum or a union is commonly in the closure of many fixed tables, and one edit inside it is ONE edit to fix, so the refusals are deduplicated by the break and the count is the count of definitions that moved, never the size of the unit.

`docs/SPEC-TABLES.md` §18.1 records the token where the rest of the table line's tokens are documented, including the bump reasoning and its price.

## Test state

`gofmt -l .` is empty. `./internal/baseline/` (whole package), `./internal/check/`, `./internal/lockfile/`, `./ir/`, `./internal/publicapi/` and `./cmd/...` are green. `./compiler/ -run Fixed` has SEVEN failures, all of them PRE-EXISTING on `rowan/bill-reads-backward` and identical there (the C++/C fixed clamp switch, the Elixir/Java/Rust fixed-form layout legs, the Java fixed-form scan, and dart's table-wire value defaults) — none of them touch the baseline.

## What a reviewer should look at first

- **The overlap with `schema.lock`.** `internal/lockfile` already holds this law for fixed tables and has tests for most of it (`TestReorderRefused`, `TestRemovalRefused`, `TestInsertInTheMiddleRefused`, `TestKindChangeRefused`, `TestDeprecateIsAllowedOnceLocked`, and the value-list rows of `lockclosure_test.go`: a reordered enum, a removed variant, a renamed variant, a reordered flags mask, a reordered union). The baseline half is a SECOND guard on a DIFFERENT artifact — both are opt-in, and a unit can carry one without the other. Whether the bill wants both is Glenn's call; this PR is the baseline half as assigned.
- `monotoneLists` takes no policy, so it has no ablation control of the kind the rest of this package keeps. One existing attribution control (`TestWasChainIsRefused`) was narrowed from "nothing reports" to "the chain refusal does not fire", because a second `was` on a FIXED table's field also hashes a new id and the fixed form's law now refuses that on its own.

### DIFF
--- a/docs/SPEC-TABLES.md
+++ b/docs/SPEC-TABLES.md
@@ -14516,6 +14516,25 @@ of the map field moves no line. Its two fields are judged as any table's are,
 which is what puts a changed key kind, a moved key bound and a retyped value
 under the verdicts of §18.2.
 
+**THE `fixed` KEYWORD RIDES AT THE END OF THE TABLE LINE, as `fixed=true`**
+(§2.2), because the two forms are under two different LIST laws and the file
+has to say which one a member is under: the id-table wire finds a field by its
+id, so a removal is absorbed and a reorder moves no byte, while a FIXED record
+is walked by offset and the order IS the contract (§2.10). The monotone law of
+the fixed form reads it — a field, variant or arm removed, inserted before the
+end, reordered or renamed without `was` inside a fixed closure is a REFUSAL
+(docs/FIXED-FORM-BILL-READS-BACKWARD.md §6) — and so is the keyword itself
+moving: a fixed table that becomes variable, or a variable one that becomes
+fixed, is A DIFFERENT FORM AND NOT A VERSION, and compaction is a new table
+under a new name. A map's generated ENTRY carries no keyword: it is declared by
+nobody (§2.8), and a map makes its holder variable, so an entry is never inside
+a fixed closure. **AND IT BUMPS NO RENDERING VERSION**, by the bump rule's own
+test: the token sits at the END of a line every older file still parses, and
+the law reads it on the COMMITTED side, so a baseline written before this
+rendering carries it nowhere and judges nothing — the price being that such a
+unit is unguarded against the keyword moving until its next `--update`, which
+is what writes the fact down.
+
 **Every line in it is a WIRE fact.** The block form's layout — the
 projection's offsets, each row's size, each pitch — is not recorded and not
 judged here (§19.3): **the baseline guards what the wire cannot report, and
--- a/examples-wide/tables.baseline
+++ b/examples-wide/tables.baseline
@@ -1,7 +1,7 @@
 schema-tables-baseline 7
 package wide
 
-table Caption
+table Caption fixed=true
     field title id=0xda31296c0c1b6029 kind=33 size=7
     field line id=0xbf4ba5ad694f5907 kind=13 type=Line
     field lines id=0x5ce3f9a9f1d5001c kind=14 elem=13 type=Line array=bounded bound=3
@@ -10,14 +10,14 @@ table Caption
 table Line
     field text id=0xfa04f4ef1995407e kind=33 size=4
 
-table Stamp
+table Stamp fixed=true
     field label id=0x39f7fcec8fcb623d kind=33 size=4
     field seq id=0x823b8a195ce2133c kind=8
 
 union Body
     arm wide id=0xa633f1f655715cca kind=33 size=4
     arm narrow id=0x96569b06f223c29a kind=12 size=4
-    arm count id=0xb1e5e28e4479a274 kind=4
+    arm tally id=0xb1e5e28e4479a274 kind=4 was=count
 
 ## history
 ### 2026-09-06 (UTC) — the wstring(N) table unit: kind 33's sites (schema#522)
--- a/internal/baseline/baseline.go
+++ b/internal/baseline/baseline.go
@@ -102,6 +102,21 @@ type Table struct {
 	// DOES NOT CARRY beyond `name=`: [Render] fills it on the live projection
 	// alone, where the chain refusal reads it (§18.2).
 	Was string
+
+	// Fixed is the `fixed` KEYWORD (docs/SPEC-TABLES.md §2.2), rendered as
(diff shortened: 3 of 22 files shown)


gafferongames:
HELD, not merged: Glenn ruled at 02:12Z that the LOCK is the monotone law's home and the baseline stays the save-game projection (bill §6). The 31 subtests here are the right cases in the wrong file; the lock child (branch rowan/lock-tests-every-row) carries every row into internal/lockfile and borrows these schemas. This PR closes when that one merges. Same for #899's baseline half; its compiler-level closure tests stay wanted.

gafferongames:
Superseded: the lock is the law's home (#909 merged); the compiler-level closure tests live in internal/check/fixedclosurerows_test.go on #909.

PR #902, opened 2026-09-11
baseline: the monotone law, numbers and kinds (bill §6)

gafferongames:
My half of the bill's monotone law (docs/FIXED-FORM-BILL-READS-BACKWARD.md §6): the
NUMBERS and the KINDS, for units whose closure holds a `fixed table`. Variable-only
units keep every §18 verdict they had.

One entry point, `monotoneNumbers(old, new *Unit) []string`, in the new
`internal/baseline/monotone_numbers.go`, returning one line per edit that is not a
widening — `fixed <Table>: <field>: <rule> (<old> -> <new>)` — called from `diff.go`'s
refusal path on ONE line marked `// monotone law, numbers (bill §6)`, so `monotoneLists`
lands beside it without a conflict.

## Forbidden, one test each, named by Glenn's sentence it holds

"same with arrays and strings etc."

- a bounded array `[..N]`'s bound SHRUNK — `an array bound only grows` (8 -> 2)
- the same bound shrunk THROUGH A CONSTANT (`const Slots` 8 -> 4) — the baseline records
  the EVALUATED value, so it is the same check, and the test goes through the constant to
  prove it
- a fixed-size array `[N]`'s N SHRUNK (4 -> 2), the same rule

"wstring/strings/bytes can be widened only, not narrowed, because a narrowed
string/array cannot read the old"

- `string(32)` -> `string(16)`, `wstring(16)` -> `wstring(8)`, `bytes(64)` -> `bytes(32)`
  — `a string/wstring/bytes capacity only grows`
- a ranged scalar's MAX moved inward (1000 -> 500) and its MIN moved inward (0 -> 10) —
  `a declared maximum/minimum only moves outward`. A range DECLARED where the field had
  none is the same edit from the kind's whole domain and refuses too

"You can take an enum and widen it, but you cannot narrow, or incompatible"

- an integer NARROWED (int32 -> int16) and a float NARROWED (float64 -> float32) —
  `only widens: a narrower int/float cannot hold every value the old writer could produce`
- SIGNEDNESS changed (int32 -> uint32) and A DIFFERENT LADDER (int32 -> float32) —
  `the int ladder and the uint/float ladder are different kinds, not widths of one`
- `bits(N)` shrunk (12 -> 10) — `a bits(N) width only grows`
- an ARRAY's ELEMENT narrowed (elem 4 -> 3) — the same rule one level in

a field's kind changed

- a scalar respelled as a `string(8)` (kind 8 -> 12) and as a NESTED TYPE (8 -> 13) —
  `a field's kind is fixed: a kind change is a different definition, not a version of one`
- a NESTED TYPE swapped for another (`Buff` -> `Debuff`): asserted over a FIXED closure
  against the EXISTING referent refusal, which already carries it
- `fixed(I,F)`'s F CHANGED, with the storage width held still (`fixed(16,16)` ->
  `fixed(24,8)`): asserted over a FIXED closure against the EXISTING `frac` row
  (§4.1, §18.1), which already carries it

an optional

- `?T` -> `T` refused — `an optional only appears: a reader with no presence companion
  cannot land a stored absence`

Every forbidden case also asserts THE GATE: the very same edit on a plain `table` draws
nothing from this law.

## Allowed, one test each

a bound GROWN through a constant (8 -> 32); a fixed-size array grown ([4] -> [8]);
`string` 32 -> 64, `wstring` 16 -> 64, `bytes` 64 -> 128; a range's ends moved OUTWARD;
a range REMOVED (the largest widening there is); int32 -> int64; float32 -> float64;
`bits(12)` -> `bits(16)`; `fixed(16,16)` -> `fixed(48,16)` (I widened, F untouched);
`T` -> `?T`.

With the law stubbed to `return nil`, all seventeen forbidden cases go red and every
widening stays green. 31 subtests, `go test ./internal/baseline/` green.

## The format token added, and why

`bits=N`, AT THE END OF THE FIELD LINE, and it is the only fact added. A `bits(N)` field
rides under the KIND its storage width gives it — bits(12) and bits(10) are both kind 7 —
so N was nowhere in this file, and "bits(N) shrunk refused" had no fact to stand on. With
the token reverted and the law in place, exactly the bits case goes red, which is the
proof it was missing.

- Old baselines still parse: the token set on a line is unordered to the parser, and a
  file that lacks this one states no N, which is no fact and judged on nothing.
- It moves NO RENDERING VERSION. §18.1's bump rule exists because an added JUDGED token
  reads as an addition on every older file and the added-token branch of `diffTokens`
  would greet an untouched schema with a diagnostic per field. This token has no row in
  `DefaultTokenPolicy` at all — the monotone law alone reads it, and an absent fact is
  silence. `tables/examples/tables.baseline` regenerates with it by `--update`, history
  intact.
- Follow-on, not done here: §18.1's token list in docs/SPEC-TABLES.md should name `bits=`
  once the bill's algorithm doc lands, so the file's columns and the spec agree.

`Table.Fixed` is the other fact, and it needs NO file token: it is the live projection's
(`Render` fills it, as it fills `Was` and `JsonKey`), and the gate reads the LIVE side on
purpose — a table that has just become fixed comes under the law at that commit, and
asking the old file what a table used to be would weaken the law on every baseline
written before the keyword. A GENERATED map entry is never fixed for the law: nothing
declares one, and the map that reaches it makes its holder variable.

## What this law does not touch, and one asymmetry left for the backport

- AN ENUM'S ORDINAL WIDTH AND A UNION'S TAG WIDTH CANNOT BE SET BY HAND IN THIS LANGUAGE
  — both derive from the variant count (an enum value rides as the u16 hash of its
  variant NAME, a union opens with its arm name hash) — so there is no edit to refuse and
  NO TEST for those two rows.
- An ENUM-KEYED array's `bound=` is its key enum's size, not a capacity anyone declared:
  it shrinks only when a variant goes, which is the enum's row of the law and is reported
  there, by name. Skipped here so it is not judged twice.
- IT ONLY EVER ADDS REFUSALS. §18's `kind` row still refuses every kind change for the
  save-game reason, including int32 -> int64, which the bill ALLOWS in a fixed closure
  (the reader lands it exactly and counts `widened`). So the allowed-widening tests assert
  that THE LAW has no complaint, not that the whole baseline passes. Loosening that row for
  a fixed closure is a decision about the whole file and belongs to the bill's §8 backport.
- §18's own range fixture (`rangeSrc`) becomes a plain `table`: the warn class is the
  VARIABLE wire's answer, where the clamp still exists and the read counts it. The fixed
  form's answer to the same edit is this law's refusal, held by its own gate.

## Green

`gofmt -l .` empty. `go vet ./internal/baseline/` clean. `go test ./internal/baseline/`
green (whole package). `go test ./...` shows 44 failures in 6 packages
(compiler, ctable, elixirtable, gotable, javatable, goldens) — IDENTICAL test names
before and after this branch's changes, i.e. pre-existing on `rowan/bill-reads-backward`.

### DIFF
--- a/internal/baseline/baseline.go
+++ b/internal/baseline/baseline.go
@@ -102,6 +102,17 @@ type Table struct {
 	// DOES NOT CARRY beyond `name=`: [Render] fills it on the live projection
 	// alone, where the chain refusal reads it (§18.2).
 	Was string
+
+	// Fixed is the DECLARED fixed class (`fixed table`, docs/SPEC-TABLES.md
+	// §2.2), and it is a fact of the LIVE projection alone: [Render] fills it,
+	// the file does not carry it, and nothing parses it. The monotone law of
+	// docs/FIXED-FORM-BILL-READS-BACKWARD.md §6 is gated on it, and the gate
+	// reads the LIVE side on purpose — a table that has just become fixed
+	// comes under the law at that commit, and one that is not fixed keeps the
+	// §18 verdicts it always had. So the committed file needs no new fact for
+	// it: asking the old file whether a table used to be fixed would only
+	// weaken the law on every baseline written before the keyword.
+	Fixed bool
 }
 
 // A Field is one field of a closure member: its declared name, its EFFECTIVE
@@ -211,7 +222,12 @@ func Render(u *ir.Unit) *Unit {
 		// spelling, and a `was` rename moves that while moving no byte — so
 		// the entry is keyed by the holder's wire id and the field's, and a
 		// rename moves nothing in this file.
-		t := Table{Name: ir.ProjectionMemberName(u, name), Declared: name, Was: st.WasName}
+		// A GENERATED MAP ENTRY is never `fixed` for the law's purposes even
+		// when its own body could be: nothing declares one, the checker answers
+		// the class from the body (§2.8), and the map that reaches it makes its
+		// HOLDER variable — so an entry is not a fixed table anyone deployed,
+		// and the monotone law has no business with its key's capacity.
+		t := Table{Name: ir.ProjectionMemberName(u, name), Declared: name, Was: st.WasName, Fixed: st.FixedDeclared && st.MapEntryOf == ""}
 		if st.MapEntryOf != "" {
 			t.Declared = t.Name
 		}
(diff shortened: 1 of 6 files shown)


gafferongames:
HELD, not merged: Glenn ruled at 02:12Z that the LOCK is the monotone law's home and the baseline stays the save-game projection (bill §6). The 31 subtests here are the right cases in the wrong file; the lock child (branch rowan/lock-tests-every-row) carries every row into internal/lockfile and borrows these schemas. This PR closes when that one merges. Same for #899's baseline half; its compiler-level closure tests stay wanted.

gafferongames:
Superseded by #909 (the lock holds the monotone law for every row).

PR #906, opened 2026-09-11
fixed form: the LIST rows' read columns on the C++ reference, red first

gafferongames:
The two READ columns of `docs/FIXED-FORM-VERSIONING-TESTS.md` — **NEW-READS-OLD** and **OLD-REFUSES-NEW** — for the eleven rows that are **LISTS**, on the C++ reference. Red first, by name. No runtime and no emitter touched.

## The rows

| row | NEW-READS-OLD | OLD-REFUSES-NEW | what the red is waiting on |
|---|---|---|---|
| `field_append` | 🟢 green | 🔴 red | `layout_newer` |
| `field_deprecate` | 🔴 red | 🟢 green (a deprecation is not newer: the old reader READS, `b` landing as written) | the layout recording `\| deprecated` |
| `field_undeprecate` | 🟢 green | 🟢 green (reads, both directions) | — |
| `enum_append` | 🟢 green | 🔴 red | `layout_newer` |
| `enum_width` (255 → 256) | 🔴 red (value exact, counter missing) | 🔴 red | `COUNT widened` on a grown ordinal; `layout_newer` |
| `union_append` | 🟢 green | 🔴 red | `layout_newer` |
| `union_arm_payload_widen` | 🔴 red (the appended field lands **1**) | 🔴 red | the prefill reaching inside an arm's payload; `layout_newer` |
| `flags_append` | 🟢 green | 🔴 red (unreachable today) | the layout recording a flags' **bit count**, THEN `layout_newer` |
| `keyed_array_enum_append` | 🟢 green | 🔴 red | `layout_newer` |
| `nested_append` | 🟢 green | 🔴 red | `layout_newer` |
| `rename_without_was` | 🟢 green | 🟢 green (reads: the rename moves no layout byte — the two layouts are byte-identical) | — |

**11 known-reds**, every one printed by name and held from both ends: a listed case that starts PASSING fails the target, so landing a fix means deleting its line from `known_red[]` (the convention is `test/tables/fixedform_properties.cpp`'s). `make tables-fixedform` and `make tables-fixedform-corpus` both run green with the eleven reds named.

## What the tests found, beyond `layout_newer`

Glenn: *"Let the tests guide us in C++ to the correct implementation."* Four of the reds are not about `layout_newer` at all, and three of them say the LAYOUT is missing something before the refusal can be written:

1. **A flags append moves no layout byte.** `ir/fixedform.go` pushes a `flags` field as a plain 8-byte scalar with `children = 0`, so `Caps {Jump,Crouch}` and `Caps {Jump,Crouch,Fly}` hash identically, the reader takes the identity plan, and the bill's *"flags: bits a prefix of the reader's"* row has nothing to read. The layout entry owes a flag count the way an enum's entry carries its variant count.
2. **`| deprecated` moves no layout byte either.** Nothing in `ir/` reads `Field.Deprecated` — it is a lock and reflection marker only — so a reader cannot drop-and-count the slot it was told to stop needing (bill §3, `unknown += 1` per plan).
3. **A grown enum ordinal counts nowhere.** The value lands exactly (width 1 → width 2), but `widened` stays 0 where §5.2 says *"wider in y: widen (COUNT widened)"*. Compare FU1/FU2's `int16 → int32`, which does count.
4. **A field appended inside a union arm's payload lands `1`.** Not its declared default (55), and not zero — a byte from elsewhere in the record. The prefill reaches a table's own appended field (FX2's `added` lands 11) and does not reach one under an arm. The assertion prints the landed value so it cannot drift.

## The schemas' shape

Two schemas per row, `test/tables/VOLD_<row>.schema` and `VNEW_<row>.schema`, differing by exactly that row's definition change and nothing else. **The same table name `Lineage` on both sides** — the two layouts are ONE LINEAGE — with only the package differing (`vold_<row>` / `vnew_<row>`), so both generations compile into one test binary. That is the FX1/FX2 precedent, and it means no row needs `Old<Row>` / `New<Row>` names.

## What is where

- `test/tables/VOLD_*.schema`, `VNEW_*.schema` — 22 schemas, the eleven lineage pairs.
- `test/tables/fixedform_dump.cpp` — one marked block writing `old_<row>.bin` and `new_<row>.bin` (22 files) into `build/fixedform-corpus`, the corpus every other leg reads.
- `test/tables/versioning_lists.cpp` — NEW (450 lines): the eleven `<row>_case()`s, the known-red list, and the refusal helper.
- `test/tables/fixedform_main.cpp` — one marked block: a declaration and `failures += versioning_lists_cases();`.
- `Makefile` — marked blocks only: the 22 schemas in the cpp generate list and the stamp's prerequisites, a `VLISTS_INCLUDES` variable, and the new TU in the plain and ASan `tables-fixedform` binaries.

### On the refusal's name

The runtime carries no `layout_newer` enumerator, so no case can spell it. Each case asserts everything the name is worth and all of it is observable today: the read is refused before any record (`n < 0`), no counter moved and nothing malformed (REFUSE is total), the destination is untouched, and the reason is **not one this build already has** — not a form byte, not one of §1.1's eight broken-layout reasons. The reference exposes no *text* for a reason, so "the refusal names the row's definition" has nowhere to land on this leg; the definition is named in the case's own message instead, which is what a leg with text will assert.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,6 +230,35 @@ define tables_generate
 	# nested TYPE the older side has no name for
 	$(1) generate --lang cpp --out $(2)/fx1 test/tables/FX1.schema
 	$(1) generate --lang cpp --out $(2)/fx2 test/tables/FX2.schema
+	# ==== BEGIN rowan/cpp-versioning-lists ====
+	# THE LIST ROWS OF THE FIXED FORM'S VERSIONING LAW
+	# (docs/FIXED-FORM-VERSIONING-TESTS.md). One pair of schemas per row,
+	# the SAME table name on both sides (one lineage) and a package each, so
+	# both generations compile into one test binary — the FX1/FX2 precedent.
+	# The reference's two read columns are test/tables/versioning_lists.cpp.
+	$(1) generate --lang cpp --out $(2)/vold_field_append test/tables/VOLD_field_append.schema
+	$(1) generate --lang cpp --out $(2)/vnew_field_append test/tables/VNEW_field_append.schema
+	$(1) generate --lang cpp --out $(2)/vold_field_deprecate test/tables/VOLD_field_deprecate.schema
+	$(1) generate --lang cpp --out $(2)/vnew_field_deprecate test/tables/VNEW_field_deprecate.schema
+	$(1) generate --lang cpp --out $(2)/vold_field_undeprecate test/tables/VOLD_field_undeprecate.schema
+	$(1) generate --lang cpp --out $(2)/vnew_field_undeprecate test/tables/VNEW_field_undeprecate.schema
+	$(1) generate --lang cpp --out $(2)/vold_enum_append test/tables/VOLD_enum_append.schema
+	$(1) generate --lang cpp --out $(2)/vnew_enum_append test/tables/VNEW_enum_append.schema
+	$(1) generate --lang cpp --out $(2)/vold_enum_width test/tables/VOLD_enum_width.schema
+	$(1) generate --lang cpp --out $(2)/vnew_enum_width test/tables/VNEW_enum_width.schema
+	$(1) generate --lang cpp --out $(2)/vold_union_append test/tables/VOLD_union_append.schema
+	$(1) generate --lang cpp --out $(2)/vnew_union_append test/tables/VNEW_union_append.schema
+	$(1) generate --lang cpp --out $(2)/vold_union_arm_payload_widen test/tables/VOLD_union_arm_payload_widen.schema
+	$(1) generate --lang cpp --out $(2)/vnew_union_arm_payload_widen test/tables/VNEW_union_arm_payload_widen.schema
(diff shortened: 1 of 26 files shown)


gafferongames:
Merged on a non-author ONE PATH. Corrections for the green phase (the branch predates §11/§12): field_deprecate/NEW-READS-OLD asserts unknown == 1 and never reads back.b; §12.3 says b == 2, unknown == 0, and the case leaves known_red[]. field_undeprecate asserts the pair reads both ways; §12.3 refuses undeprecating, so the row moves to the lock's LOCK-REFUSES set and its read corpus goes. §12.4: add per refusing row the trigger assertion VOLD hash != VNEW hash; the file comment about naming the definition is superseded. keyed_array_enum_append: give the element a nonzero default so §12.6's element-default prefill is told apart from a zero fill.

Issue #876, opened 2026-09-11
fixed form: the contract's 56 items against nine legs' fixtures: 207 gaps, twelve cards


gafferongames:
Glenn's ask, 2026-09-10 evening: "Look for other cases in fixed packet table where the tests don't check." An audit of `docs/FIXED-FORM-ALGORITHM.md`'s contract against every leg's fixtures on branch `fixed-table-form` (tip 0bcb8d60), by one reader, all nine test trees read whole. 56 contract items by nine legs: 504 cells, 207 GAP, 89 weak (golden-only or fuzz-only). The seven layout refusals are cited from the earlier verification tonight (present and asserted on C++, C#, Java, Dart, five on Elixir; absent on Go, JS, Rust as named refusals; C's runtime names them, its fixture did not until tonight's PR).

## The matrix

Cells in leg order **cpp c go cs rust js java dart elixir**. `P` asserted · `~` partial, golden-only or fuzz-only · `G` GAP.

| item | cells |
|---|---|
| L1–L7 layout refusals | `P~GPGGPP~` |
| F1 previous_form | `PGPPPPPPP` |
| F2 message_form_as_file | `PGPPPPPPP` |
| F3 newer_form | `PGPPPPPPP` |
| F4 layout_malformed, truncated | `~~GPGPGGP` |
| F5 layout_malformed, lying hash | `P~GPPPPPP` |
| F6 hash checked LAST | `PGGPGGPPP` |
| F7 malformed, under 20 bytes | `GGGGGGGGG` |
| F8 malformed, ragged tail | `G~GGPPPP~` |
| F9 batch_too_large | `GGGGPPPPG` |
| F10 no_layout | `G~PPGPGPP` |
| F11 plan_too_large | `PGPPPPPPP` |
| F12 second layout for a held hash | `GGGGGGGGG` |
| E3 widen ladders | `~~~~~~~~~` |
| E4 kind_mismatch pairs | `~~~~~~~~~` |
| E5 `?T` vs plain nesting | `PGGPPPPPP` |
| E7 variant or arm inserted mid-list | `PPGPPGPPP` |
| E9 `duplicate` never raised | `GG~GGGGG~` |
| C1 count clamp v<0 | `GGGGGPGG~` |
| C2 count clamp v>Max | `GGGGGPGG~` |
| C3 text length clamp | `~GGGGPGGP` |
| C4 text content refuses BY NAME | `~~GG~GGG~` |
[...]

PR #907, opened 2026-09-11
versioning tests, the NUMBERS: 21 rows on the C++ reference, 28 red by name

gafferongames:
The NUMBERS column of the fixed form's versioning law, on the C++ reference, tests first.

Glenn: **"Let the tests guide us in C++ to the correct implementation."** So they went in before the implementation, and **three of the reds are findings the design has to answer** rather than work waiting to be done.

Design: `docs/FIXED-FORM-VERSIONING-TESTS.md` (my columns: NEW-READS-OLD, OLD-REFUSES-NEW, plus the floor and hash table), `docs/FIXED-FORM-BILL-READS-BACKWARD.md` §2/§6a/§6b/§8a.1, `docs/FIXED-FORM-ALGORITHM.md` §5.

## The rows

| row | NEW-READS-OLD | OLD-REFUSES-NEW | note |
|---|---|---|---|
| `array_bounded_grow` | 🟢 | 🔴 `layout_newer` | count and 4 elements exact, **`[..N]` slack past the count is zeros (§12.6)**, no counter |
| `array_fixed_grow` | 🟢 | 🔴 `layout_newer` | **`[4]Vec -> [8]Vec`**, the element's defaults NONZERO (`x = 7, y = 9`): 4 exact, slots 4..7 the **element's defaults**, not zero (§12.6) |
| `array_elem_widen` | 🟢 | 🔴 `layout_newer` | -1 / INT16_MIN / INT16_MAX sign-extend; `widened` moves |
| `constant_grow` | 🟢 | 🔴 `layout_newer` | the evaluated bound; as `array_bounded_grow` on the wire |
| `string_grow` | 🟢 | 🔴 `layout_newer` | length and bytes exact, the bought capacity ZERO |
| `wstring_grow` | 🟢 | 🔴 `layout_newer` | the length in CODE UNITS, a surrogate pair counting two |
| `bytes_grow` | 🟢 | 🔴 `layout_newer` | high bytes, so a sign error shows |
| `int_widen` | 🟢 | 🔴 `layout_newer` | -1, INT16_MIN, INT16_MAX, one record each; `widened == 1` per record |
| `uint_widen` | 🟢 | 🔴 `layout_newer` | 0xFFFF ZERO-extends |
| `float_widen` | 🟢 | 🔴 `layout_newer` | a signalling NaN; the 22 payload bits BELOW the quiet bit asserted exactly; the quiet bit pinned to what the reference does TODAY in one named line (#876 to rule) |
| `range_widen` | 🟢 | 🔴 **and the hash does not move** | see finding 1; **§13 (merged) rules the hash covers a definitions digest, so the hash WILL move** — promoted to `check()` when the digest lands |
| `bits_grow` | 🟢 | 🔴 `layout_newer` | every bit of bits(8) in bits(12) |
| `fixed_I_grow` | 🔴 **finding 2** | 🔴 `layout_newer` | |
| `optional_add` | 🔴 **finding 3** | 🔴 `layout_newer` | waits on **§12.8, the `present` op** |
| `floor_at` | 🔴 the floor **declared at LINEAGE INDEX 1 on the `*_floor` chain** | — | the file reads today; there is no floor for it to sit at |
| `floor_below` | 🔴 `layout_unsupported` + the floor **at LINEAGE INDEX 1 on the `*_floor` chain** | — | the file one below the floor READS today |
| `floor_raise_live` | 🔴 the floor + the `SCHEMA_FIXED_FLOOR_TEST_HOOKS` setter | — | the setter is behind its **own** define, not `SCHEMA_HAS_FLOOR` |
| `hash_unknown` | 🔴 `layout_newer` | — | a well-formed file of another table: a hash no lineage holds |
| `hash_known_bytes_differ` | 🔴 ×7 | — | the seven §1.1 malformations under a KNOWN hash |
| `hash_identity` | 🟢 | — | `cache.compiles == 0` on the reader's own hash: no plan compiler runs |
| `lineage_merge` | 🟢 | — | both pre-merge branches read on the merged build, `from_b` by NAME and not by position |

**28 RED, 0 failures.** `make tables-fixedform` green (plain, ASan, runcopy ×2).

## The three findings

1. **`range_widen`: a range is NOT IN THE LAYOUT.** `int32 | 0..100` and `int32 | 0..200` hash **identically** (`0x54909885244699d8` both sides), so an old reader cannot tell the widened writer from itself and `layout_newer` is **unreachable for this row by hash**. Bill §6a — "any change to a definition that inputs into it changes the hash by itself" — is false for a range today. The new writer's 150 is read and silently clamped. Either the range goes into the layout (and the hash), or §6a says ranges are outside it and names what protects this row instead.
2. **`fixed_I_grow`: `fixed(4,4)` → `fixed(12,4)` is not a widening rung.** The read comes back `kind_mismatch == 1` and the raw scaled value is **lost to the declared default** — the one outcome bill §2 forbids of a widening. The rung wanted is int8 → int16 on the raw value, F equal.
3. **`optional_add`: `T` read into `?T` does not land present.** `kind_mismatch == 1`, **ABSENT**, payload defaulted. Bill §2 ("`T` where the reader has `?T` (landed present)") and algorithm §5.2 ("`?T` where x has T: present := 1, then the value") both want the other answer.

And `hash_known_bytes_differ` is what §6b buys, measured: all seven malformations against a KNOWN hash's bytes come back with their own **validation-walk** name today, where §6b wants **one** answer — `layout_malformed`, "a lie about a known version" — and no walk over a stranger's layout at all.

## The reader's four corrections (a HOLD on one pair, answered)

1. **`array_fixed_grow` could not tell the law from an accident.** `[4]int32 -> [8]int32` asserting slots 4..7 `== 0` is satisfied by a plain zero fill of the reader's struct **exactly as well as** by the ELEMENT-DEFAULT prefill bill §12.6 requires — so the case passed without the behaviour it named ever existing. The element is now a nested `Vec { x int32 = 7; y int32 = 9 }`, `[4]Vec -> [8]Vec`, and the slack slots must read **7 and 9**. No element the writer sets is ever 7 or 9. **It is GREEN on the C++ reference** — the prefill is real — so it stays a `check()`, and now it is a `check()` that means something.
2. **The floor's test-only setter is behind its OWN define, `SCHEMA_FIXED_FLOOR_TEST_HOOKS`.** Not `SCHEMA_HAS_FLOOR`: that one says a build *has* a floor, and a shipped build that has one must never compile a way to **lower** it. Only the two fixedform test rules in the Makefile pass `-DSCHEMA_FIXED_FLOOR_TEST_HOOKS` (`FIXEDFORM_TEST_HOOKS`), and the `floor_raise_live` call site needs **both** defines. **The emitter must emit `T##FixedSetFloorForTest` under that define alone.**
3. **Four waits-on texts were pointing at the wrong thing.** `range_widen` now cites **§13**: the hash covers a *definitions digest*, so the hash WILL move and the red is the emitter's today, promoted to `check()` when the digest lands. `optional_add` cites **§12.8, the `present` op**. `array_bounded_grow`'s message now reads "`[..N]` slack past the count is zeros (§12.6)", so it cannot be mistaken for the `[N]` row. `floor_at` and `floor_below` state that the floor must be **declared at lineage index 1 on the `*_floor` chain**.
4. **`OLD_REFUSES_NEW` says what it does not yet check.** The refusal says "newer" but not **which** layout the file carries. §12.4 puts the file's layout hash in the report, so the green phase adds one conjunct: the reported hash equals the NEW build's own `T##FixedHash` — which is what proves the reader refused because it read the writer's hash and not because it refuses everything. **The assertion is unchanged today.**

## The schemas' shape

One pair per row, `test/tables/VOLD_<row>.schema` / `test/tables/VNEW_<row>.schema`, differing by **exactly** that row. **The table name is the SAME on both sides** (`IntWiden`, `StringGrow`, …) in **different packages** (`vold_int_widen` / `vnew_int_widen`) — the FX1/FX2 shape — so both generations compile into one binary and no `Old<Row>`/`New<Row>` renaming was needed. Every table brackets the row's field with `lead` and `trail`, so a mislaid size moves a neighbour and the equality check sees it.

Plus two chains: `VOLD_/VMID_/VNEW_floor.schema` (three layouts, each appending one field: a lineage a floor can sit in the middle of) and `VOLD_/VBRA_/VBRB_/VNEW_lineage_merge.schema` (the §8a.1 branch case).

**All 35 units generate into ONE directory** (`build/tables-generated/vnum`): the emitter names its files after the *schema*, not the table, so 35 units share a directory without collision and the test binaries take **one** `-I` instead of 35.

## The floor API this PR assumes

It does not exist, and bill §9 leaves its syntax to Glenn. What the design implies is a floor **per table, fixed at build time**, with the lineage shipped as static data beside it. Written out in full at `floor_cases()` in `test/tables/versioning_numbers.cpp`; in a generated C++ unit, for a fixed table `T`:

```
constexpr int32_t     T##FixedLineageCount;   // how many layouts the lock knows
extern const uint64_t T##FixedLineage[];      // their hashes, OLDEST FIRST, the reader's own LAST
constexpr int32_t     T##FixedFloor;          // the lowest lineage INDEX this build accepts; 0 = all
enum reason           layout_unsupported;     // a hash IN the lineage, BELOW the floor
```

and, for `floor_raise_live` only, a test-only setter, because a floor chosen at build time cannot be raised inside one process — **emitted only under `SCHEMA_FIXED_FLOOR_TEST_HOOKS`**, which only the fixedform test rules pass:

```
#ifdef SCHEMA_FIXED_FLOOR_TEST_HOOKS
void T##FixedSetFloorForTest( int32_t index );
#endif
```

**The question this asks:** is that setter the right shape? The alternative is two BUILDS of the same unit at different floors, which the Makefile can do and which costs the test its ability to say "the file that read yesterday refuses today" in one place. Say which and the test follows.

## How red is spelled, so the target still runs

`layout_newer`, `layout_unsupported`, the floor and the lineage are not in the emitter, so **naming them in C++ would not compile**, and a target that does not build is a target nobody runs. Every waiting case goes through `red()`: it prints the case name and the fix it waits on, it is counted, and it does not fail the target. Held from both ends — **a `red()` that starts PASSING is a real FAILURE**, because a case that has gone green belongs in `check()` as part of landing the fix it names. One `-D` (`SCHEMA_HAS_LAYOUT_NEWER`, `SCHEMA_HAS_LAYOUT_UNSUPPORTED`, `SCHEMA_HAS_FLOOR`) turns each family into the real comparison it already reads as. `SCHEMA_FIXED_FLOOR_TEST_HOOKS` is the one define that is **not** a "the fix landed" switch: it says *this build is a test and may move the floor*, it is set only by the fixedform test rules, and `floor_raise_live` needs it as well as `SCHEMA_HAS_FLOOR`.

No runtime and no emitter changed. One registration call, `versioning_numbers_cases()`, from `main` in a block marked `rowan/cpp-versioning-numbers`; the Makefile edits are inside the same markers, so the sibling adding `versioning_lists.cpp` does not collide.

## A reference defect found in passing

A unit whose only wide arithmetic is a **signed `fixed(I,F)` with a negative min** emits `serialize::uint128_t` in its packet codec and **does not include `serialize.h`**, so `VOLD_fixed_I_growTable.h` does not compile on its own. Worked around with one `#include "serialize.h"` in the test; the emitter keeps the fix.

## Commands

| command | wall | result |
|---|---|---|
| `make tables-fixedform` | **11.7 s** | **exit 0** — plain, ASan, runcopy, runcopy-ASan; **28 RED named on the numbers rows, 0 failures** (plus the sibling's 11 on the lists rows) |
| `make tables-fixedform-corpus` | **2.7 s** | **exit 0** — 64 files, `old_<row>.bin` / `new_<row>.bin` per row plus the floor and branch chains, both columns' corpora in one directory |

Timed after `origin/fixed-table-form` merged in (append-merge, no rebase: #906's `versioning_lists` and this branch's `versioning_numbers` are two translation units in one binary, both marked blocks kept whole).


gafferongames:
Merged after the reader's four corrections landed (8176195a). One note for the green phase: array_fixed_grow is green because back is Reset() to defaults before FixedLoad, so it proves the load does not clobber the element defaults, not yet that the load writes them; a memset before the load separates the two (§12.6).

PR #909, opened 2026-09-11
the lock holds every row of the monotone law: LOCK-REFUSES and LOCK-ALLOWS, red first

gafferongames:
The LOCK-REFUSES and LOCK-ALLOWS columns of `docs/FIXED-FORM-VERSIONING-TESTS.md`, every row, written RED FIRST against the unchanged lock and then made green. Glenn: "We must test all the cases, never assume they work. Iterate every case we need to check, then add tests for them." The law is `docs/FIXED-FORM-BILL-READS-BACKWARD.md` §2 and §6, and its home is the lock (Glenn's ruling 2026-09-11 02:12Z: "lock.").

**Not in the first pass, and HERE NOW** (see "The three finish items" below, which
is what Glenn's HOLD asked for): the lineage, the wire hash, the §13 digest field
and the retire verbs. Still out: a declared floor syntax (§9), the digest's own
bytes (#910), and the merge case (§11.3).

## The rows

`R` = `TestLock<Row>Refuses`, `A` = `TestLock<Row>Allows`, both in `internal/lockfile/lockrows_test.go`. "red" is the state before a line of `lockfile.go`/`diff.go` moved.

| row | R red? | the rule the refusal now names | A red? |
|---|---|---|---|
| `field_append` | yes (3 cases) | `field removed`, `field inserted not at the end`, `fields reordered` | no |
| `field_deprecate` | yes | `field removed` | no |
| `field_undeprecate` | yes (flipped back per §12.3) | `undeprecated (deprecated -> live)` | — |
| `field_modify` | yes | `kind changed (int32 -> string)` | — |
| `enum_append` | yes (4 cases) | `variant reordered`, `variant removed`, `variant inserted mid-list`, `variant renamed without was` | no |
| `enum_width` | — (a width cannot be set by hand) | — | **yes** (255 → 256 variants, ordinal width 1 → 2) |
| `union_append` | yes (4 cases) | `arm reordered`, `arm removed`, `arm inserted mid-list`, `arm payload changed (Av -> Cv)` | no |
| `union_arm_payload_widen` | yes | `field removed` (named on `type Av`) | yes |
| `flags_append` | yes (2 cases) | `flag moved`, `flag removed` | no |
| `array_bounded_grow` | yes | `bound narrowed (4 -> 2)` | yes |
| `array_fixed_grow` | yes | `bound narrowed (4 -> 2)` | yes |
| `array_shape` | yes (2 cases) | `shape changed (fixed -> counted)`, `shape changed (fixed -> keyed)` | — |
| `array_elem_widen` | yes | `element narrowed (int16 -> int8)` | yes |
| `keyed_array_enum_append` | yes | `variant reordered` (the array follows) | yes |
| `constant_grow` | yes | `bound narrowed (4 -> 2)` (the EVALUATED bound) | yes |
| `string_grow` | yes | `capacity narrowed (8 -> 4)` | yes |
| `wstring_grow` | yes | `capacity narrowed (8 -> 4)` | yes |
| `bytes_grow` | yes | `capacity narrowed (8 -> 4)` | yes |
| `text_kind` | yes (2 cases) | `kind changed (string -> wstring)`, `kind changed (string -> array)` | — |
| `int_widen` | yes (3 cases) | `narrowed (int16 -> int8)`, `signedness changed (int16 -> uint16)`, `ladder changed (int16 -> float32)` | yes (2 cases) |
| `uint_widen` | yes (2 cases) | `narrowed (uint16 -> uint8)`, `signedness changed (uint16 -> int16)` | yes |
| `float_widen` | yes (2 cases) | `narrowed (float64 -> float32)`, `ladder changed (float32 -> int32)` | yes |
| `range_widen` | yes (2 cases) | `range narrowed ([0, 100] -> [0, 50])`, `range narrowed ([0, 100] -> [10, 100])` | yes (2 cases: wider, and unranged) |
| `range_added` | yes | `range added where none was (unranged -> [0, 100])` | — |
| `bits_grow` | yes | `bits narrowed (8 -> 4)` | yes |
| `fixed_I_grow` | yes (2 cases) | `F changed (4 -> 8)`, `I narrowed (12 -> 4)` | yes |
| `optional_add` | yes | `optional removed` | yes |
| `nested_append` | yes | `field removed` (named on `type Vec`) | yes |
| `default_change` | yes | `default changed (1 -> 2)` | — |
| `keyword_change` | yes (2 cases) | `fixed removed`, `fixed added` | — |
| `rename_without_was` | yes | `renamed without was` | no (a `was` rename moves no byte of the lock) |
| `closure_plain_table` | no — already refused by `check_fixed.go`; named by row now | `by value` / "is a table, not a wire type" | — |
| `closure_variable_kind` | no — already refused; named by row now | `is a pointer`, `is a map`, `is an unbounded array` | — |
| `closure_self` | no — already refused by every path | the table that reaches itself | — |

The three closure rows are compiler-level and live in `internal/check/fixedclosurerows_test.go` (`TestLockClosurePlainTableRefuses`, `TestLockClosureVariableKindRefuses`, `TestLockClosureSelfRefuses`), eleven sub-cases, all green before the change: the compiler already held both halves of §2a, and what was missing was a test per ROW of the design.

## The gap list — what the lock did not do

Every `Refuses` row above was red, all 45 sub-cases: the lock refused the edit (it refuses EVERY edit to a locked entry) but no refusal named the RULE or both values. And eleven `Allows` rows were red because the lock refused the WIDENING too, which the bill does not: "Newer versions of the fixed table should be able to read OLD versions."

So the implementation is one mechanism, `internal/lockfile/monotone.go`, the same shape as the list refusals already there: one comparison per recorded fact, in the order that names the change best, the first difference is the finding, and a widening comes back as a change the RECORD must carry — the refusal that names `schema lock` — instead of a break.

## The format tokens added (rendering version 5 → 6)

Five facts the width carries only as a total and a refusal has to be able to name, each a new token at the END of an entry line so a line is read left to right the way it grew:

- `bound=N` — an array's declared element count, the EVALUATED N behind `[..N]`
- `shape=fixed|counted|list|keyed` — the array form, so `[4]T` → `[..4]T` is a shape change and not a width change
- `cap=N` — a `string(N)` / `wstring(N)` / `bytes(N)` capacity in its own units
- `bits=N` — a `bits(N)` width (two bits() widths inside one storage kind are otherwise identical in the lock)
- `ibits=N` — a fixed-point field's I (the kind carries I + F rounded to a storage width, so I alone was not recorded)

The fifteen committed corpus locks are rewritten under v6; `TestCorpusLocksAreCurrent` and `TestCorpusPassesItsOwnLock` are green. (The v5 → v6 bump still took the delete-and-rewrite sentence; from v7 on it is §11.8's SALVAGE, stated and tested below, and that is the rule for every bump after this one.)

## Three old tests now read the other way

Each encoded the law BEFORE the ruling, and each is updated with the bill cited, its narrowing twin added beside it:

- `TestWidenRefused` → `TestWidenTakenAsAChangeToRecord` + `TestNarrowRefused` (a capacity grown is a read; cut is a break)
- `TestWidenScalarRefused` → `TestWidenScalarTakenAsAChangeToRecord` + `TestNarrowScalarRefused` (`uint8` → `uint32` is a read)
- `TestUnDeprecateRefused` → `TestUnDeprecateAllowed` and back again: §12.3 restored the original answer, so this one is `TestUnDeprecateRefused` as it always was (commit `04392d2` below)
- `TestNestedRecordAppendSlidesTheHolder`: one finding now, the nested record's own appended entry — the holder's width is that append's consequence, and the block that changed is where a person fixes or accepts it
- `TestLockRefusesToWriteABreak` now offers `schema lock` a narrowing, since a widening is no longer a break
- the closure table's "a widened integer range" row became "a narrowed integer range"; its `?` row moved out to `TestLockOptionalAdd{Refuses,Allows}`, because the fixture declares the plain side and `T` → `?T` is the widening

## Green (the first pass)

`gofmt -l .` empty. `go test ./internal/lockfile/ ./internal/check/` green (54 row tests with 34 named sub-cases, plus the package's own).

---

# The three finish items, and the lineage (steps 1 and 2 of #898)

Glenn's HOLD on this PR named what had to land before merge: the undeprecation
row flipped back to REFUSED, the lineage section with the wire hash and the §13
digest, and the retire verbs — plus §11.8's salvage path stated instead of a
delete. All four are here, one commit each, red first, on top of an append-merge
of `origin/fixed-table-form`.

## 1. Deprecation is ONE WAY (bill §12.3) — `04392d2`

This branch read the docs before §12 landed and took §2's old row ("undeprecating
is allowed"). §12.3 corrects it and restores the lock's own first answer, so
`TestLockFieldUndeprecateAllows` → `TestLockFieldUndeprecateRefuses` and
`TestUnDeprecateAllowed` → `TestUnDeprecateRefused` (its original name), both red
first. The refusal names the rule:

    undeprecated (deprecated -> live) — deprecation is ONE WAY: every writer
    since the slot was deprecated wrote the default into it, so what would come
    back is not data

`schema lock` refuses it too, under `Appendable`. SPEC §21.1's table row is
corrected to match the bill.

## 2. The lineage per fixed table (rendering version 6 → 7) — `6fc143e`

One `lineage` line per layout, inside the fixed table's block, after the field
entries, OLDEST FIRST and the current layout LAST. The lock's new section shape,
five lines from a unit with two layouts and one of them retired:

```
fixed table Row layout=0x6d12719672a1e7d7
    field a id=0xaf63dc4c8601ec8c kind=4 width=4 default=0
    field b id=0xaf63df4c8601f1a5 kind=4 width=4 default=0
    lineage wire=0x4842a4b029549be5 record=4 bytes=02000000fb06c9fe19e113a00d04000000010000008cec01864cdc63af040400000000000000 retired reason=the 0.3 client is gone; nothing live writes this layout
    lineage wire=0x0d5f6af3564b4019 record=8 bytes=03000000fb06c9fe19e113a00d08000000020000008cec01864cdc63af040400000000000000a5f101864cdf63af040400000000000000
```

(real output: `Row` with one field, then two, the first layout retired. The
reason is the rest of its line, so it holds a sentence with spaces in it.)

Per entry, exactly what §5.2's table of lock facts asks for: the **WIRE hash**
(`ir.TableFixedLayoutHash`, the number the file header and every record carry,
not the lock's own `layout=` text projection — §11.7's named gap), the **layout
bytes** verbatim (LOAD memcmps them; it never parses a stranger's layout), the
**§13 definitions digest**, the **record body size**, a **retired mark** and a
**reason**.

**On the §13 digest:** #910 is not merged, so `ir.TableFixedLayoutHash` still
takes only the layout bytes. The entry records the digest as a fact of its own
and the hash is computed over `layout ++ digest` — which IS §5.2's `HASH`, and
while the digest is empty it is byte for byte the number every backend emits
today (§13: "an empty digest leaves the hash equal to a hash of the layout bytes
alone"). When ir grows `DIGEST(T)`, one line of `lineage.go` moves, and every
table carrying a range, a flags mask, a `bits(N)`, a `fixed(I,F)` or a
reader-side limit earns one new lineage entry — the true record of what happened:
its files' hashes moved.

`schema lock` appends an entry when the hash has MOVED and nothing otherwise
(§11.6: one entry per layout, not one per commit — a deprecation rewrites the
lock and moves no layout byte, so it adds nothing). Nothing is ever removed and
no committed entry is rewritten, so a retired mark survives every later append.
A **v6 lock SALVAGES into v7** (§11.8): its single layout becomes the first
lineage entry, its field lines come through unchanged, and "delete it and write
it again" is gone from here on, because deleting this file now means wiping a
fleet's lineage. Older than v6 still takes the delete sentence — there is nothing
in it a lineage can be made of.
`TestALockFromTheRenderingJustBeforeThisOneSaysTheSameThing` becomes
`...Salvages`, which is that ruling as a test.

The lineage findings are held back in `Diff` and reported only when nothing else
refused: a root's layout hash moves BECAUSE a field or a nested type moved, so
the lineage is always that change's consequence, never its account. Alone it
catches a record that is itself wrong — a hand-truncated lineage, a hand-edited
`bytes=` (the wire hash is fnv1a64 over those bytes, so the line is one statement
made twice), or a lock from a rendering that had none.

**Two things came with the walk:**

- `ir.TableFixedWalkRoot` **panicked on a payload-free union arm** (a nil arm
  field — `tables/messages`' `ack`), which is why the lock could not walk its own
  corpus. A payload-free arm is now an entry with no bytes, kind 32, exactly as
  an enum's variants are in that walk. No golden moved: every table it affects
  had no layout at all before, because the walk died on it.
- §11.6's **long-lineage advisory** (32 entries), printed by `schema lock` where
  the person who moved the file sees it, naming both remedies. Never a refusal.

The fifteen committed corpus locks are rewritten under v7.

## 3. The retire verbs — `2207235`

    schema lock --retire Table@0x<hash> --reason "..."
    schema lock --retire Table        --reason "..."

The entry form (§11.4) marks one lineage entry retired and the entry STAYS: what
the mark buys is that COMPILE emits the retired hashes beside the supported ones,
so LOAD says `layout_unsupported` (upgrade the client) rather than `layout_newer`
(ship the reader) — two answers pointing the operator in opposite directions, and
both have to stay available. The CURRENT layout is never retired, and a hash
nothing locked is refused by name.

The table form (§11.5) marks the whole table retired. Its block, its entries and
its lineage stay; what the mark buys is that the DECLARATION may then be dropped,
and that the "fixed removed" refusal applies only to a table nobody retired —
which that refusal now says, naming the verb. When the declaration goes,
`schema lock` carries the retired block forward along with the `type`, enum,
flags and union blocks nothing live reaches any more because of it, so the record
of what shipped stays whole and self-consistent.

**Both forms require `--reason`**, and that is the point: every other write this
file takes is an append, which declares nothing, while a retirement declares that
no reader will serve a layout again — the sentence the tables baseline keeps for
the same reason (SPEC §18.4). Retiring also requires the lock to be CURRENT,
since a statement about a record nobody has is worth nothing.

## 4. The accessor COMPILE reads — `abdc443`

**For Johnny (#910):** the lock now provides the lineage, and `cpptable` is NOT
touched here. The three names to swap the filename convention for:

```go
lock, ok, err := lockfile.Open(paths)              // locate + parse, ok=false when the unit has no lock
for i, e := range lockfile.Lineage(lock, "ToolMessage") {
    // e.Wire    -> R.lineage[i], the only fact a file is matched on
    // e.Layout  -> R.known[i].layout, what LOAD memcmps and PLAN walks
    // e.Digest  -> the hash's second input; it never rides the wire
    // e.Record  -> R.known[i].record_bytes, never taken from the file
    // e.Retired -> the hash emitted for layout_unsupported
}
floor := lockfile.Floor(lock, "ToolMessage")       // §5.2's one number
```

`internal/codegen/cpptable/lineage.go` reads its lineage from sibling schema
files by filename convention (`VOLD_`/`VNEW_`/`VMID_`/`VBRA_`/`VBRB_` and the
numbered evolution sets) — an interim its own comment names as one ("Emma's lock
lineage is not on the tip"). `TestLineageAccessorIsWhatCompileReads` is the
contract that swap is made against: three layouts oldest first, the first
retired, every entry's bytes hashing to its recorded wire hash, the record size
present, the floor at 1.

## Green

`gofmt -l .` empty. `go test ./internal/lockfile/ ./internal/check/` green.
`go test ./compiler/ -run 'Lock|Fixed'` green. `go test ./ir/` and the four
fixed-form backends (`cpptable`, `ctable`, `rusttable`, `elixirtable`) green
after the payload-free-arm fix.

## Left

- the **floor's syntax** is Glenn's to name (§9): the lock records retired
  entries and `Floor()` derives the one number §5.2 wants, and no per-table floor
  is declared in the schema yet;
- the **§13 digest itself** (#910's `DIGEST(T)` in ir) — the lineage records the
  field and the hash binds it; today it is empty;
- the **merge case** (§11.3: a merge commit takes both parents' locks as `old`,
  and the lineage is their union) — `schema lock` here takes one `old`;
- `cpptable`'s swap to these accessors, per the split above.


gafferongames:
Handed to Emma (steps 1 and 2 of #898). Before merge: undeprecation flips back to REFUSED per bill §12.3 (this branch read the docs before §12 landed); the lineage section, the wire hash, the §13 digest and the retire verbs come on top; a second mouth. The v5 to v6 bump wants the salvage path of §11.8 stated, not a delete.

gafferongames:
Merged on a non-author ONE PATH (abdc4433). Named follow-up from the read: nothing binds the SET of lineage lines; a hand that deletes an older lineage line is accepted by Check and by schema lock, which is the fleet-wide deletion bill §11.8 forbids. Add a lineage=0x… roll-up (or entry count) on the table line so a deleted lineage line refuses the way an edited entry does, and report it unconditionally rather than held back. Also: rowAllows asserts the hash moves, the one-entry-per-widening is proven generically in lineage_test.go; fine.

PR #910, opened 2026-09-11
C++ green: LOAD by hash, digest in the hash, the named reds

gafferongames:
ONE PATH. Hash chooses. No second reader. Still draft. Did not merge.

**SHA** `2320852ec8c4af6f180bb329af13db1387bb7a5f` on `johnny/cpp-versioning-green` off live tip `216beff417b67b37abbc5c8a7c7d3f7f14522768` (Merge #907).

## The path

COMPILE bakes known hashes and layout bytes (sibling fixture pairs; Emma's lock lineage is not on this tip). LOAD looks the **header** hash up in that set:

- hash not in known → `layout_newer` (the file's hash, nothing else)
- hash in lineage below the floor → `layout_unsupported`
- known hash, layout bytes differ → `layout_malformed`
- known hash, bytes match → run the plan (identity is a PLAN walked by the same loop)

A stranger's layout is never parsed. Older plans are compiled from the **trusted** baked layout bytes, cached by hash.

`TableFixedLayoutHash(layout, digest)` (bill §13). Digest is ranges, flags bit count + names, bits(N), fixed I/F. Never on the wire. Empty digest leaves the hash unmoved. Layout format unchanged.

Did not take Emma's lock (#909) or #886. Nothing on any other leg.

`make tables-fixedform` green (Debug + ASan, including versioning lists and numbers).

### Lists (`known_red[]` emptied)

| red | status |
|---|---|
| field_append/OLD-REFUSES-NEW | green — `layout_newer` + hash |
| enum_append/OLD-REFUSES-NEW | green |
| enum_width/OLD-REFUSES-NEW | green |
| union_append/OLD-REFUSES-NEW | green |
| union_arm_payload_widen/OLD-REFUSES-NEW | green |
| keyed_array_enum_append/OLD-REFUSES-NEW | green |
| nested_append/OLD-REFUSES-NEW | green |
| flags_append/OLD-REFUSES-NEW | green — digest records flags bit count, hashes differ, then `layout_newer` |
| field_deprecate/NEW-READS-OLD | **corrected** — a,b,c exact, no counter (bill §12.3). Did not implement skip-and-count |
| enum_width/NEW-READS-OLD | green — widen, `widened` += 1 |
| union_arm_payload_widen/NEW-READS-OLD | green — Reset of each arm so prefill has the arm's fresh image |

### Numbers (`red()` promoted to `check()`)

| red | status |
|---|---|
| range_widen hash | green — digest moves the hash |
| flags/bits/fixed I,F in the hash | green — digest |
| fixed(4,4)→fixed(12,4) | green — widening rung, not kind_mismatch |
| T into ?T | green — `present` op |
| seven §1.1 against a known hash | green — one name `layout_malformed` |
| floor API | green — `T##FixedLineage[]` oldest first, `T##FixedFloor`, setter under `SCHEMA_FIXED_FLOOR_TEST_HOOKS` |
| array_fixed_grow memset | kept (#907) |
| every OLD-REFUSES-NEW | green — `layout_newer` + the file's hash |

### Leftover

- field_undeprecate lock refusal is Emma's. Untouched.
- Plans for older hashes are compiled at first load from trusted baked bytes, not laid down as static op arrays. LOAD still never parses the file's layout.
- Flags bit count is in the digest, not as layout children (layout format unchanged per §13; a U64 leaf cannot carry children without breaking parse of trusted bytes).

Johnny Grok.


gafferongames:
HOLD after a seven-check read at c581cb47 (details on the table, rowan note of 03:50Z): (1) the §13 digest is computed on the C++ leg only, every other leg passes nil, so cross-leg hashes and the pinned corpus bytes diverge; compute the digest inside ir.TableFixedLayoutHash and re-pin once for all legs. (2) tables-fixed-properties is 24 red on P1: the cross-schema pairs are not lineages under the filename convention, and the reverse direction is now a layout_newer refusal by contract. (3) §12.5 (the plan carries the writer's bounds; per-plan clamp; the hostile fixtures) is not implemented. (4) memset in array_fixed_grow and a nonzero element default in keyed_array_enum_append still owed. (5) floor as index vs the retired set; first-load compile vs static plans; malformed-before-hash ordering; digest holes (float ranges, bounds past int64, nested dedupe, ArrayBound, Type.Elem).

gafferongames:
Re-read at a3fbb47d: corrections 1 and 2 hold. HOLD stays for: (1) testdata/golden/** not re-pinned after the digest moved every hash (make update-goldens for every leg); (2) rebase onto 0587c006; (3) gcc-only compile errors in emitted C++ (MessagesTable.h:4131,4139,12271 array-from-initializer-list; StreamTable.h:6484 reference init from TableRef), reproduce on Space (x64 gcc). Base-side reds listed on the table.

gafferongames:
Third read at c75225a6: mechanicals hold; HOLD for (1) tools/fixedtwin TestPairedHeadersAgree, new red of this PR: the twin map learns the C++-only rows as owed by the C leg (ruling, C++ first); (2) versioning_lists.cpp:727 memset over non-trivial Lineage, gcc -Werror=class-memaccess on Space. Base-side reds taken separately by a child on the tip.

PR #911, opened 2026-09-11
§5 rewritten FROM the green C++: hash selects, nothing parses a stranger

gafferongames:
Third step of Glenn's order: tests red in C++, the C++ made green by them, **then the pseudocode rewritten FROM the green C++**, then every other leg from the pseudocode. Branched from `johnny/cpp-versioning-green` at `c581cb47` (#910).

**Second pass, `91c935aa`.** #910 has MERGED into `fixed-table-form`, and #909 (the lock's lineage) with it; this branch has append-merged `origin/fixed-table-form` and §5 is re-read against what is on the branch rather than against the tip it was written from. Everything below the first section is the original write-up; the second section and the divergence list are the current ones.

§5 is rewritten so a port written from §5 alone, by someone who has never opened the reference, lands the same bytes, the same refusal names and the same counters on every fixture in `test/tables/versioning_lists.cpp` and `versioning_numbers.cpp`. Docs only — no code touched, nothing merged.

## Where §5 changed

**§5 intro** — says the hash selects, that nothing parses a stranger's layout on any path, and that §5 supersedes §2's load table at steps 3 to 5.

**§5.1 BASELINE** — kept as the bill's law. Three corrections: the `fixed(I,F)` rung is named in `LADDER` (the green C++ admits kinds 20..24 and 25..29); "fixed removed" refuses only for an UNRETIRED table (bill §11.5); the rename-through-`was` and the deprecation are stated as NOT versions — the hashes are equal and both sides take the identity plan, which is what `rename_without_was`, `field_deprecate` and `field_undeprecate` assert.

**§5.2 COMPILE** — rewritten and much larger:
- a table of **what the lock must provide per fixed table**: the lineage oldest first, and per entry the wire hash, the layout bytes verbatim, the definitions digest, the record body size, the retired mark and its reason.
- the **static data** COMPILE emits: `R.lineage[]` (hashes, oldest first, the reader's own last), `R.known[]` (hash, layout bytes, layout length, record size), `R.plans[]`, `R.floor`, the identity plan with its split and the type's cover.
- **the floor is one index** — `1 +` the highest retired index — and why the two names point the operator in opposite directions.
- **`HASH(layout_bytes, digest)`** as pseudocode: fnv1a64 over the layout bytes, then over the digest.
- **the definitions digest's own byte layout**, as a table: `'R'` + min/max i64 LE for a range, `'B'` + N for `bits(N)`, `'X'` + I + F + signed for fixed-point, `'F'` + wire bit count + `'f'` + `fnv1a64(name)` per flag, `'L'` + the limit for a reader-side limit, recursion for a nested type (once per name) and for each union arm. An empty digest leaves the hash the layout's alone.
- **`PLAN`, `MATCH` and `EMIT` as three procedures**, with the two passes (unguarded, then guarded, counting nothing the second time), the coalescing rule inside each half, the fills as cover-minus-landed, and the per-kind op for every kind: the `present` op for `T` into `?T`, the optional's present copy, the array's count and `min(their_n, my_n)` elements with the rest left to the element-default prefill, the keyed array matched by the key's id, the union's unguarded `None` const first and then the reader's ordinal under the writer's tag offset at the writer's tag WIDTH, the enum's `widen` when the ordinal grew and its remap table otherwise, text at `min(mine, theirs) - 4`, and the leaf ladder.
- the entry bound (`src + size`, `+4` for text, and `guard + the guard's width`, all inside the WRITER's record size) and the plan's carrying of the writer's bounds for the hostile pass.

**§5.3 LOAD** — rewritten as **eleven numbered steps in the green C++'s own order**: the 20-byte floor; the form byte by direction; `20 + L`; **the header's hash taken as given** (step 4 — nothing is recomputed from the wire, the digest not being on it); the lineage scan; the floor; the layout byte comparison; the plan selection; the ragged tail; the batch capacity; then the per-record loop — record hash, prefill, run, bounds. Plus **one table of every condition, its refusal name and what the refusal reports**. The vacuous `record_bytes <= 8` clause is dropped (bill §12.9). Stated: the seven §1.1 malformations under a KNOWN hash all come back as one name, `layout_malformed`.

**§5.4 (new) The counters, per op and per condition** — `unknown` and `kind_mismatch` at COMPILE once per peer; `widened` per widen entry per record (the grown enum ordinal width included); `clamped` for `count`, `text` and each bounds-pass field; nothing for `copy`, `const`, `present`, `ordinal`; and **every counter at zero for a clean NEW-READS-OLD of an appended field, variant, arm, flag or keyed slot**. A bool or present byte that is not 0/1 normalises and counts nothing.

**§5.5 The closure** — unchanged (was §5.4).

**§5.6 What is retired** — adds the run-time layout walk, and the recompute of the header hash from the layout (with §2's step 5).

**§5.7 The tests** — names the fixtures that now exist, row by row, and the two columns every row owes (NEW-READS-OLD exact + §5.4's counters; OLD-REFUSES-NEW `layout_newer` with the FILE's hash, no counter, a destination still fresh), the no-second-column rows, and the floor's three cases.

**§5.8 (new) What the green reference owes this section** — the fourteen divergences below, so no port copies them.

**Outside §5, minimally:** §1's hash sentence now names the digest as the hash's second input; §2's load table steps 3, 4 and 5 are corrected (step 5 marked RETIRED by §5.3); §4.2 points at §5.2 for where a plan comes from.

## What the re-read changed (`91c935aa`)

**§5.2 — what the lock NOW provides.** #909 landed `internal/lockfile/lineage.go`, so the row that said "the tip's lock holds none of it" is gone. A backend reads three calls: `lockfile.Open(paths)` (ok false for a never-locked unit is not an error), `lockfile.Lineage(lock, T)` (oldest first, current last, nil for a name the lock carries as a nested `type`), `lockfile.Floor(lock, T)` (one past the highest `Retired`). A `LineageEntry` is `Wire` / `Layout` / `Digest` / `Record` / `Retired` / `Reason`. The line `lineage wire=0x… record=<n> bytes=<hex> [digest=<hex>] [retired] [reason=…]` is **one statement made twice** — the parse recomputes the wire hash over `bytes` then `digest` and refuses a line whose halves disagree. The SET is bound by `lineage=0x…` on the `fixed table` line (#912), whose remedy is version control and never "write it again". `--retire` and the pre-lineage salvage are named. §5.1's implementation is `internal/lockfile/monotone.go`.

**§5.2 — the reference's lineage is the INTERIM, said so on the page.** `cpptable/lineage.go`'s filename convention plus `ir.TableFixedFixtureLineage` goes away the day COMPILE reads `lockfile.Lineage` and `lockfile.Floor`.

**§5.2 — the hash takes the SCHEMA.** `ir.TableFixedLayoutHash(layout, st)` computes the §13 digest at the hash site, so there is no argument for a leg to omit; a lineage entry, the one caller with no live schema, stores the two runs and hashes the concatenation. Also: a payload-free union arm has a layout entry and NO bytes (kind 32), because the arm list is what the layout is compared by.

**§5.2 — the prefill's image.** The reader's own fresh value, with a union reset **arm by arm at each arm's overlay storage** and only then the tag to `None` — the #910 emitter fix, stated as the contract it is. And the test of it is a POISON (`0x5A` through a byte pointer), not a `Reset`: a zeroed destination passes either way.

**§5.4 — a clamp that cannot fire is not emitted, and nothing moves.** An ordinal whose extent fills its storage width (255 in a byte, 65535 in two) has no read-side clamp; the elided check never clamped.

**§5.7 — the properties gate's pairs.** One pair run per lineage pair (SEVEN of them: `FX1`/`FX2`, `P1`/`P3`, `FN1`/`FN2`, `FM1`/`FM2`, `V1`/`V2`, `UT1`/`UT2`, `Scalars`/`Scalars2`), forward compiled through the lineage and **reverse refusing `layout_newer` by name** — the old contract's both-ways compiled read is retired, not bent. Resolved through `ir.TableFixedFixtureLineage`. The gate's record check is the reader's own compiled hash constant, never `hash_of(layout)`: the digest is not on the wire. `keyed_array_enum_append`'s element is `fixed table Qty { n int32 = 7 }`, so the appended key's slot proves the prefill rather than a coincidence of zeros.

**§5.7 — the C leg's port list.** `tools/fixedtwin` records the C++-only §5 rows as OWED by the C leg; §5.7 now carries them as the port list they are: the `present` op (enumerator, Apply case, `T` into `?T`), the `fixed(I,F)` ladder rungs `20..24`/`25..29` and `SignedKind` over them, the grown ordinal width as `widen`, `TableFixedKnownLayout`, and `layout_newer`/`layout_unsupported`/the floor in the per-type codec. Not divergences — this page's port list for C.

## Divergences that STILL STAND (§5.8)

Re-read row by row against the merged branch; fourteen remain, each with the bill's ruling kept and marked **the reference owes this**.

1. the lineage is read from sibling schema files by filename convention plus a fixture map — it owes `lockfile.Lineage`
2. the floor is hard-coded to one index for one test package — it owes `lockfile.Floor`
3. an older hash's plan is compiled at FIRST LOAD from the trusted bytes and cached by hash — `COMPILE` owes it at build time
4. the `count` bound, the ranges, the variant and arm counts are the READER's (`e.size = my_n`) — the bounds are the WRITER's per plan
5. `arg` is still a `uint8_t` lane, so a union past 255 arms cannot express its guard value (the tag is read at its own width and the remap table is `uint16_t`, so only this lane is short)
6. a nested union's arm entries carry the inner tag's guard only
7. the `ordinal` op reads through a 32-bit temporary
8. ill-formed text zeroes the field, restores the default and counts `malformed` — fix 11 owes a refusal by name
9. the prefill runs BEFORE the per-record hash check, so `no_layout` has written the caller's storage
10. the digest carries only INTEGER ranges — a float range and a reader-side limit are still out, and a flags type is re-emitted once per naming field. **An INTEROP BREAK, not a missed refusal: it moves the hash, so the reference owes it BEFORE any leg ports**
11. the `unknown` census counts once per ELEMENT of an array of tables
12. a forged ordinal counts nothing on a compiled plan while the identity plan counts `clamped`
13. an entry past the writer's declared record comes back as `layout_malformed`; `layout_record_too_large` is wired only to the 65536 bound and to a zero root size
14. **new** — the enum remap table is capped at 255 entries, so a writer's 256th variant and beyond remap to `None` silently

**Closed since the first pass:** the `BASELINE` row — the monotone law is `internal/lockfile/monotone.go`, the lock holds a lineage with a retired mark and a reason, and `schema lock --retire T@0x<hash> --reason "…"` exists. What remains of it is rows 1 and 2: the BACKEND not yet reading what the lock now holds. The digest-argument hazard is closed by `TableFixedLayoutHash(layout, st)`.

## The porter's read: ten things §5 could not yet hand a leg (`55d7852c`)

A porter read §5 and found it ACCURATE and INCOMPLETE — a leg written from §5 alone cannot produce the reference's bytes. Ten additions, each verified against the merged C++ (`internal/codegen/cpptable/fixedruntime.go`, `fixedform.go`, `lineage.go`; `ir/fixedform.go`) and cited once.

1. **The KIND CODES, the whole closed set**, as a table in §1 with a pointer from §5.2 — `i8..i64` = `2..5`, `u8..u64` = `6..9`, `f32`/`f64` = `10`/`11`, signed `fixed(I,F)` `20..24` and unsigned `25..29`, `12` string, `13` table, `14` array, `15` union, `16` keyed, `17` pointer, `18`/`19` the 128s, `30` enum, `32` no-payload, `33` wstring, `35` optional, with `31` and `34` marked OUT of the set. A port needs the numbers, not the set's shape.
2. **The AUX LANE.** Every plan entry names two destinations, `dst` and its companion `aux`, and which one an op writes is CONTRACT. A table per kind: the optional's present byte, the counted array's count word, `bytes(N)`'s length, text's buffer and the union's tag all land at `aux` — and text and `bytes(N)` are the two rows the other way round (fix 10). The `count` op's destination is named. How a port derives `aux` from its OWN storage, never from the wire.
3. **TEXT in full.** The span is BYTES, the clamp cap is `span / unit` (unit `2` for `wstring`), the op copies the WHOLE SPAN and then stores the terminator at the USED length — and on a poisoned destination a port that copies `v * unit` carries `0x5A` back out into the saved file.
4. **A widen's sign is the LADDER's.** The ladder widen sign-extends per `TableFixedSignedKind`; the same-kind widen and the enum widen ZERO-extend. A port that infers the sign from the kind's signedness agrees by accident today and diverges the first day a signed kind admits two widths.
5. **The guard width and the remap table.** `argw` clamps to `1..8` at both ends and `0` reads as `1`; the remap table carries its LENGTH in slot `0` and is read as `raw != 0 && raw <= table[0]`; the 255-entry cap is a DIVERGENCE (new §5.8 row 14); and the two-pass split gates the POOL allocations, not only the entries.
6. **LOAD step 9 also guards `record_bytes <= 8`**, and the report is TWO ANSWERS — `refused` plus `reason`, or `malformed`, never both — asserted jointly by every fixture. A table says what each refusal sets.
7. **The digest is THE CONTRACT.** §5.2's table is bill §13: every range including float, every reader-side limit at tag `'L'`, and named types once. The reference at this head emits `'R'` for `HasIntRange` alone, no `'L'`, and does not dedupe flags (`seen` covers structs only). **That is an INTEROP BREAK and not a missed refusal** — a leg implementing the table gets a different wire hash — so the reference **owes it BEFORE any leg ports**: landing it moves every hash once more and re-pins the corpus. Also: the order inside a field is `'R'`, then the kind tag, then the reference; `WireBits == 0` falls back to `len(Variants)`.
8. **§5.7.** Seven properties pairs, `UT1`/`UT2` among them (`ir.TableFixedFixtureLineage`). `union_arm_payload_widen` is an APPENDED FIELD INSIDE AN ARM, not a widen, and counts `0`. No fixture exercises a nonzero `unknown` — owed. The stale comment at `versioning_numbers.cpp:1001` describes the reader §5.3 retired: noted for Johnny.
9. **§5.4 reconciled with §5.8 row 12.** A forged ordinal's remap to `None` — one past the WRITER's variant count — is the BOUNDS PASS's `clamped`, on the compiled plan exactly as on the identity one. The §5.4 row is fixed.
10. **The trailing "(b)" line was FALSE** and is withdrawn: the reference DOES cache a compiled plan by hash (`TableFixedPlanCache`, the caller's storage, sixty-four slots). What remains is §5.8 row 3 — nothing should compile at run time at all, and then there is no cache to keep.

Docs only. No code touched, nothing merged.


gafferongames:
Merges after #910 (the C++ it describes) lands; the fourteen divergences in §5.8 are handed to Johnny on the table as the rest of the green phase.

PR #916, opened 2026-09-11
C++ digest rows (float 'R', 'L', flags by name); COMPILE from lock; owed 6 7 8 10 13; §12.5

gafferongames:
The digest missing rows first, then COMPILE from the lock, then the paused legal-file wrongs, then §12.5 hostile fixtures. Still draft. Johnny does not merge.

HOLD leftovers of this PR landed at `946040196ceb3d5aa10fe4a882671b747585fe99`. Still draft.

## Digest (bill §13 / §5.8 row 10)

The reference emitted `'R'` only for `HasIntRange` and no `'L'`. A leg that implements the digest table gets a different wire hash from C++ for any table with a float range, and the two refuse each other `layout_newer`.

Walk declared field order, each named type once:

| fact | bytes |
|---|---|
| every range (int, float, fixed-point) | `'R'` + min/max i64 LE (IEEE-754 bits for a float) |
| `bits(N)` | `'B'` + N u32 LE |
| `fixed(I,F)` | `'X'` + I, F, signed |
| flags | `'F'` once by name |
| reader-side limit | `'L'` + u64 LE — **RESERVED**: no table spelling exists; `--fixed-record-limit` is outside the law. Pin: `TestTableFixedDefinitionsDigestLReservedUntilALimitExists`. |

Structs, flags and unions share one `seen` map keyed by bare name.

The lock records that digest on each lineage entry. A lock written with an empty digest is rewritten **in place** (same layout bytes, not a new version). Corpus locks re-pinned; `BenchTable` Mixed hash moves with the float range.

Fixture nits in the same commit: `hash_known_bytes_differ`'s seven land `layout_malformed`; properties pairs are **seven with UT1/UT2**.

## COMPILE from the lock

`cpptable` reads `lockfile.Open` / `Lineage` / `Floor`. Fixture lineages are a test-only map (`fixtureLineage` / `fixtureFloor`). **`loadFixturePeers` is gated on `lock == nil`**: a locked unit with a sibling the convention can name must not append a non-oldest-first entry past the identity.

## Owed 6, 7, 8, 10, 13

Each named. 7, 8, 10, 13 now have red-first fixtures in `test/tables`.

- **6** — `arg` is full width (`uint64_t`), remap not capped at 255 (bill §12.7)
- **7** — FH1/FH2: nested union answers to the outer tag; inner tag consts stay on the inner tag
- **8** — a width-8 ordinal read whole through a 64-bit temporary
- **10** — `no_layout` writes nothing; poison the caller's storage first
- **13** — FE1/FE2: a forged ordinal COUNT clamped on the compiled plan

C is not ported. Twin map names these rows OWED.

## §12.5 hostile fixtures

The `count` op's bound is the **writer's** `their_n`. Hostile cases:

- `array_bounded_grow` (forged count 7 against old `[..4]`) — already green
- `enum_append` (ordinal 4 against the old enum's 3 variants) — already green
- `union_append` (tag past the old arm count) — added
- `range_widen` (scalar past the old range) — added

Both new cases clamp and count against the writer's bound the plan carries.

Did not take 9, 12, 14. Did not take #911 as implementer (append-merged the tip after it landed). Did not take #914 as implementer (kept our Mixed hash). Did not port C / legs. Did not merge. Did not undraft.

## Gates

- `go test ./internal/goldens/` green after re-pin
- `make tables-go-fixed-form` green
- `make tables-fixedform` green (clang, ASan+UBSan)
- gcc 13.3 on Space (`spacegame`): compile + run green (`-Wno-error=class-memaccess`)
- `go test ./ir/ ./internal/codegen/cpptable/ ./tools/fixedtwin/` green

Did not run nine-language. Did not nova-bus send.


gafferongames:
HOLD at aacb5d7f: substance confirmed (digest rows, COMPILE from the lock, owed 6/7/8/10/13, §12.5 count bound). Owed: (1) the full re-pin (testdata goldens, bench/paired/corpus, every generated stamp); (2) red-first fixtures for 7, 8, 10, 13; (3) hostile fixtures for the tag and the range rows; (4) gate loadFixturePeers on lock == nil (cpptable.go:1512); (5) pin 'L' or mark it reserved; one shared seen map by bare name. Details on the table.

PR #917, opened 2026-09-11
c leg: the fixed table reads backward (§5), step 1 and the twin's four rows

gafferongames:
The C leg's fixed form reads BACKWARD. §5 of `docs/FIXED-FORM-ALGORITHM.md` was
implemented from the page: `internal/codegen/cpptable` and the C++ runtime were
never opened, except `tools/fixedtwin`'s own canon (the twin map, which §5.7 names as
this leg's port list) and the merged Go leg as a worked example. §5.7's checklist is
now STEPS 1, 2, 3, 5 and 6; what step 3 still owes is the TIMING of the plans, named
below and nowhere else.

## The rows

`make tables-c-versioning` — the corpus is the C++ reference's own bytes
(`make tables-fixedform-corpus`), a probe unit GENERATED per row per column,
compiled with `$(CC)` and run. Each side is its own binary: C has no namespace, so
two generations of one table name have no spelling inside one translation unit.

|---|---|---|
| `array_bounded_grow` | 🟢 | 🟢 (was 🔴) |
| `array_elem_widen` | 🟢 | 🟢 (was 🔴) |
| `array_fixed_grow` | 🟢 | 🟢 (was 🔴) |
| `bits_grow` | 🟢 | 🟢 (was 🔴) |
| `bytes_grow` | 🟢 | 🟢 (was 🔴) |
| `constant_grow` | 🟢 | 🟢 (was 🔴) |
| `enum_append` | 🟢 | 🟢 (was 🔴) |
| `enum_width` | 🟢 | 🟢 (was 🔴) |
| `field_append` | 🟢 | 🟢 (was 🔴) |
| `field_deprecate` | 🟢 | 🟢 (same hash, no second column) |
| `field_undeprecate` | 🟢 | 🟢 (same hash) |
| `fixed_I_grow` | 🟢 | 🟢 (was 🔴) |
| `flags_append` | 🟢 | 🟢 (was 🔴) |
| `float_widen` | 🟢 | 🟢 (was 🔴) |
| `int_widen` | 🟢 | 🟢 (was 🔴) |
| `keyed_array_enum_append` | 🟢 | 🟢 (was 🔴) |
| `nested_append` | 🟢 | 🟢 (was 🔴) |
| `optional_add` | 🟢 | 🟢 (was 🔴) |
| `range_widen` | 🟢 | 🟢 (was 🔴) |
| `rename_without_was` | 🟢 | 🟢 (same hash) |
| `string_grow` | 🟢 | 🟢 (was 🔴) |
| `uint_widen` | 🟢 | 🟢 (was 🔴) |
| `union_append` | 🟢 | 🟢 (was 🔴) |
| `union_arm_payload_widen` | 🟢 | 🟢 (was 🔴) |
| `wstring_grow` | 🟢 | 🟢 (was 🔴) |

| the floor and the hash | |
|---|---|
| `floor_at` | 🟢 |
| `floor_below` | 🟢 (was 🔴) — `layout_unsupported`, the file's hash reported |
| `floor_raise_live` | 🟢 (was 🔴) — the floor raised by one, yesterday's file refuses |
| `hash_unknown` | 🟢 (was 🔴) — `layout_newer`, on the FILE'S hash alone |
| `hash_known_bytes_differ` | 🟢 (was 🔴) — ONE name, `layout_malformed`, not §1.1's seven |
| `hash_identity` | 🟢 |
| `lineage_merge` (two files) | 🟢, and **no longer by accident**: both pre-merge files resolve through the lineage the build laid down, not through a run-time compile |

**25 of 25 NEW-READS-OLD green. 25 of 25 OLD-REFUSES-NEW green (22 `layout_newer`,
3 with no second column). Floor 3 of 3, hash 3 of 3, `lineage_merge` green for the
right reason.** Every red of the previous head was the same red and it is closed:
the lineage, the floor and `layout_newer` exist, and this leg no longer compiles a
stranger's layout to read a file.

**The twin map closes on ALL FIVE of its rows.** `struct TableFixedKnownLayout` is in
the C runtime, member for member — `hash`, `layout`, `layout_bytes`, `record_bytes` —
and in the same place of the runtime as the C++'s, so the gate holds the two legs to
it and strips nothing. `TestOwedCLegStripped` is now `TestOwedCLegClosed` plus
`TestOwedCLegRenamedMemberReds`, its negative control: one member renamed and the
twin must go red.

**The mirror opened as the owed list closed.** §5.2 lays every older plan down off the
load path and the reference still compiles one at first load into a caller-supplied
cache (§5.8 row 3), so three C runtime rows have no C++ twin YET — the hash select,
the refusal that carries the file's hash, and one lineage entry's plan — and the map
strips them FROM THE OTHER SIDE, named, until the reference lands row 3.

## What is retired, by name

§5.6 retires reading a layout the lock has never seen, and §5.7 step 5 says skip the
tests that asserted it WITH THE SENTENCE THAT SAYS WHERE THE COVERAGE IS OWED, and
never delete them. ELEVEN checks of `make tables-c-fixedform` handed one generation's
file to another generation built with NO LINEAGE: the plan path (`fixed_fx2_read_fx1`,
twice — the second time after the byte-flip fuzz), the caller-supplied plan cache, the
bytes-row control, the forward read, §1.1's seven rules at read time
(`fixed_fx1_layout_validation`), the V1/V2 and FU1/FU2 pairs, the guard and flavour
lanes, TEXT UNDER THE SECOND ARM, and the hostile pass reached through a compiled
plan. C has no `t.Skip`, so the skip is a printed line at the CALL SITE naming the
function, §5.6 and where it is owed again — the lineage harness for the plan path, the
LOCK's validation of what it records for §1.1 — and the suite prints the count. The
functions stay in the tree. The fuzz is unchanged: every flip is answered by name
inside the buffer, which is what the sanitized twin of that binary watches.

Green: `gofmt -l .` empty, `go test ./internal/codegen/ctable/ ./tools/fixedtwin/`,
`make tables-c-fixedform` (with its ASan twin), `make tables-c-versioning`,
`make tables-fixed-twin` with its negative control. `make tables-fixed-matched` does
not LINK in this bench's tree — `_serialize_uint128_make`, a symbol the rule's
`$(CC)` line does not compile in — which is upstream of this branch and names no C
code this PR wrote; it is NOT verified green here and is reported as such.

## What §5 did not tell me

Items 1, 2, 4 and 6 of the previous head are ANSWERED by the work and restated here
as what the page should say; 3, 5 and 7 stand; 8 to 11 are new.

1. **THE TWO NEW REFUSAL NAMES HAVE NO NUMBERS, AND THE PILOT LEG HAS NONE EITHER.**
   The instruction to "pick the numbers the Go leg uses" has no source: Go spells a
   reason as a STRING, so no integer exists to agree with. What settles it is the C
   runtime's own sentence about the ten names before these two — **the VALUES are this
   leg's own and the NAMES are the reference's** — so this leg takes 18 and 19 in the
   ORDER §5.3 names them, `layout_newer` then `layout_unsupported`. §5 owes that
   sentence and the ORDER, not two integers; a leg that reads §5 alone still invents
   both.
2. **THE REPORT NOW CARRIES `layout_hash`**, a `uint64_t` last on `TableReport`, zero
   on every other path. §5.3 requires the two layout refusals to report the file's
   hash and never says where it lands, so the field name was invented here as it was
   on the Go leg (`LayoutHash`). §5.2's table should name it.
3. **THE C SIGNATURE CARRIES A PLAN CACHE AND §5 RETIRES THE CACHE, NOT THE
   ARGUMENT.** Ruled by keeping the signature: `plan` and `plan_capacity` stay a
   CAPACITY DECLARATION (§5.9 #5) and `cache` is `(void)`-cast with the reason in
   the comment. §5.6 retires mechanisms, not API — stated, because a leg whose API
   carries such an argument needs the ruling and §5 still does not give it.
4. **§5.7'S POISON IS NAMED FOR C++ STORAGE, AND THE BYTE COMPARISON NEEDS THE
   PADDING SETTLED.** A generated struct's padding is not a value `reset` covers, so
   two fresh C values differ in their slack and `memcmp` of a refused-into destination
   against a fresh one compares that slack. The probe now zeroes both values before it
   resets them — the same reason the generated load memsets its default image before
   resetting it. §5 should say the OLD-REFUSES-NEW check is about VALUES.
5. **NOTHING SAYS WHERE A LEG'S PROBE UNITS LIVE.** Unchanged: `cRunVersionProbe` in
   `internal/codegen/ctable/fixedversioning_test.go`, one generated probe per row per
   column, `$(CC)` and `-I` for `serialize.c`.
6. **THE TWIN'S FIFTH ROW IS CLOSED, AND §5.2 STILL OWES THE FOUR MEMBER NAMES.**
   `hash`, `layout`, `layout_bytes`, `record_bytes`, in that order, with the byte
   length riding BESIDE the pointer as an `int64_t` — taken from the twin's own
   fixture, not from the reference. §5.2's table should name them the way §4.1 names
   the plan entry's lanes, because the gate is what forced them and a leg writing the
   struct from §5 alone still lands a different spelling.
7. **§5.9 #3 ALLOWS "PACKAGE INITIALIZATION" AND C HAS NONE.** Stands, and item 8 is
   what it costs.
8. **§5.9 #3'S TWO CONFORMING SHAPES BOTH EXCLUDE C, AND THE PAGE DOES NOT NOTICE.**
   C has no package init and no `constexpr` runner, so "emitted as source" is the only
   shape left — and emitting a plan AS SOURCE needs a COMPILE in the GENERATOR's
   language, which this repository has only as emitted runtime text (the Go leg's
   inside a Go string, this leg's inside C). So this leg builds each older plan ONCE
   PER PROCESS **from THE LOCK'S OWN BYTES** into static storage, off the record loop,
   at the first load that selects a non-identity entry. What §5.8 row 3 forbids is
   gone either way — nothing parses a FILE's layout, and there is no cache a load can
   miss — and what this leg OWES is the TIMING ("off every load path" is the
   contract's word) and the thread safety of that one flag. The real remedy is a
   GENERATOR-SIDE COMPILE shared by every leg with neither init nor constexpr, which
   is a tool §5 does not name and no leg can write alone.
9. **§5.9 #4'S "SIZED BY CONSTRUCTION" HAS NO ANSWER WITHOUT THAT TOOL.** The static
   plan storage per lineage entry is sized by the GENERATOR from the reader's leaf and
   layout-entry counts, and an entry whose plan exceeds it records `plan_too_large` ON
   THAT ENTRY, which LOAD then reports by name — §5.9 #4's contract, reached by a
   guess about the stride rather than by construction.
10. **A LEG AHEAD OF THE REFERENCE NEEDS ITS OWN STRIP LIST, AND §5.7 NAMES ONLY THE
    OWED DIRECTION.** The twin fails on ANY difference, so the day a leg implements a
    §5 row the reference still owes, the gate goes red on the leg that obeyed the page.
    Three such rows are stripped here, named with §5.8 row 3. §5.7 should say that the
    twin's map has two sides.
11. **§5.7 STEP 5 ASSUMES A TEST FRAMEWORK WITH A SKIP.** The C suite is a `main` that
    counts failures, so "skip by name" is a printed line at the call site and a
    printed count. Worth a sentence, because the rule that matters — never delete a
    retired test — is easy to obey in Go and easy to get wrong in C.

## Owed, named rather than faked

The timing of the plans (item 8) and their thread safety; the stride guess (item 9).
The reds §5.7 says to leave red and name, all four still red and unchanged: the
hostile pass under the WRITER's bounds (§5.8 row 4), a nonzero `unknown` (no fixture
exercises it), the remap table past 255 entries (row 14), and the clamp count on a
forged ordinal under a compiled plan (row 12). `make tables-fixed-matched`'s link
failure above is not mine but it is not green either, and a red nobody wrote down is a
red nobody owes.

Merged #925 and wired the C leg into the compiler's lineage path (4e093ae3): `GenerateLineage` is
target_c.go's body, `Generate` the nil shim, `cTableLineage` says the lock's entries in ctable's
spelling, and `c` joins `fixedLineageShipTargets` — the ship test now proves the emitted C source
carries the OLDER layout's hash beside the current one (handing ctable nil fails on exactly that
line, so the gate is not vacuous). generated/ does not move: the examples corpus carries no lock.


gafferongames:
Read at 891b2153: ONE PATH once the four C View headers are regenerated (pushed by Rowan's hand). Owed, named for the follow-up: the once-flag on the lazy older-plan build is a plain static set before the loop (a racing thread can read a zeroed plan); the 0x5A poison is absent from the C harness so array_fixed_grow and keyed_array_enum_append assert only n and counters; bench/paired/TWIN.md still lists the five rows as owed and none of the three AHEAD strips. Pre-existing on the base: tables-fixed-matched does not link (make/c.mk:75-81 omits serialize.c from the paired bench link line). C's enum values differ from C++ by design (names are the contract; the twin polices names).

PR #918, opened 2026-09-11
the Rust leg reads backward: §5 from the pseudocode alone

gafferongames:
The Rust leg reads backward: `docs/FIXED-FORM-ALGORITHM.md` §5 ported **from the pseudocode alone**, with `internal/codegen/cpptable` and the C++ runtime unopened. The worked example beside it was the merged Go pilot (#914), never the reference.

## The rows, both columns

`make tables-rust-versioning` builds the C++ reference's own byte oracle (`make tables-fixedform-corpus`) and runs 57 probes — one cargo workspace of generated probe crates, one `cargo test`, 51 s warm. NEW-READS-OLD reads `old_<row>.bin` on the newer reader, whose lineage is the older unit's locked entry; OLD-REFUSES-NEW hands `new_<row>.bin` to the older reader alone.

| row | NEW-READS-OLD | OLD-REFUSES-NEW | what the row owes |
|---|---|---|---|
| `array_bounded_grow` | 🟢 | 🟢 | append: every counter zero |
| `array_elem_widen` | 🟢 | 🟢 | `widened` nonzero, the rest zero |
| `array_fixed_grow` | 🟢 | 🟢 | POISONED dest; slots 4..7 are the element's own defaults `x=7, y=9` |
| `bits_grow` | 🟢 | 🟢 | `widened` nonzero |
| `bytes_grow` | 🟢 | 🟢 | append |
| `constant_grow` | 🟢 | 🟢 | append (the lock records the EVALUATED bound) |
| `enum_append` | 🟢 | 🟢 | append; ordinals equal |
| `enum_width` | 🟢 | 🟢 | a grown ordinal width is a `widen`, unsigned — **this row was red and found the missing rung** |
| `field_append` | 🟢 | 🟢 | x,y,z exact; `w` its declared default |
| `field_deprecate` | 🟢 | 🟢 (reads) | same hash: no second column, it READS both ways |
| `field_undeprecate` | 🟢 | 🟢 (reads) | same hash |
| `fixed_I_grow` | 🟢 | 🟢 | **was red: the fixed-point ladders `20..24`/`25..29` were not in this leg** |
| `flags_append` | 🟢 | 🟢 | mask equal |
| `float_widen` | 🟢 | 🟢 | `widened` nonzero |
| `int_widen` | 🟢 | 🟢 | −1, MIN, MAX sign-extended; `lead`/`trail` unmoved |
| `keyed_array_enum_append` | 🟢 | 🟢 | POISONED dest; the appended key's slot reads `n = 7` |
| `nested_append` | 🟢 | 🟢 | append inside `Root` |
| `optional_add` | 🟢 | 🟢 | **was red: `T` into `?T` had no `present` op on this leg** |
| `range_widen` | 🟢 | 🟢 | 100 lands, `clamped == 0` |
| `rename_without_was` | 🟢 | 🟢 (reads) | same hash — the test that the hash, and only the hash, is the version |
| `string_grow` | 🟢 | 🟢 | length and bytes exact, slack zero |
| `uint_widen` | 🟢 | 🟢 | zero-extended |
| `union_append` | 🟢 | 🟢 | an `a` record lands `a`; tag width equal |
| `union_arm_payload_widen` | 🟢 | 🟢 | an APPEND despite its name: `0, 0` |
| `wstring_grow` | 🟢 | 🟢 | as string, in code units |
| `floor_at` / `floor_below` / `floor_raise_live` | 🟢 🟢 🟢 | — | at the floor reads; one below is `layout_unsupported` WITH the hash; the floor raised by one makes yesterday's file refuse |
| `hash_unknown` / `hash_known_bytes_differ` / `hash_identity` | 🟢 🟢 🟢 | — | `layout_newer` on the file's hash alone; a known hash with one byte moved is ONE name, `layout_malformed`; the own hash takes the identity plan |
| `lineage_merge` | 🟢 | — | both pre-merge files read on the merged build |

**Red, named rather than faked** (§5.7's rule, and none of them is a row above):

- **the hostile pass under the WRITER's bounds** (§5.8 row 4), HALF CLOSED: the `count` op's bound is the WRITER's now, carried by the plan — a count forged past what that peer could have written clamps and counts, and a count the reader merely grew room for does not. The RANGES, the variant counts and the arm counts are still the reader's, because they are emitted straight-line from this build's own schema and carrying a peer's would want per-entry clamp data. That half stays owed.
- **a nonzero `unknown`** (and a nonzero `kind_mismatch`): the census is carried and landed once per returning read, and nothing asserts a nonzero value — see the finding below, which is that on a LAWFUL lineage there is nothing to assert.
- **the forged ordinal's `clamped` under a compiled plan** (§5.8 row 12): this leg counts it IN THE `ordinal` OP rather than in the bounds pass, which §5.4 says is the wrong place (and counts once, not twice). Moving it needs the bounds pass to count an ordinal of `0`, which is its own change; a green fixture (`the_enum_extent`) asserts the current number.
- **the two-pass split**: this leg compiles in ONE pass with a merged coalesce and no `split`/`cover` in the plan. No fixture distinguishes it; §5.2's reasons for the split (the pool spent by a discarded pass, coalescing never across the split) do not bite a single-pass compiler, but it is not what §5.2 says.
- **the lock is still not read by the toolchain** (§5.8 rows 1–2): `GenerateLineage` is the entry point and the harness plays the lock, but `./bin/schema generate` hands no lineage, so a CLI-generated crate has a lineage of one. Every file of another generation refuses `layout_newer` on that path — which is why §5.6's retirements below are the shape they are.

## The shape this leg chose

- **"Build time" (§5.9 #3): built once at first use from the LOCK'S BYTES.** The plans live in a `LazyLock` beside the lineage, one per entry, in storage the build sized by construction; the walk happens once per process, off every load path, over bytes the lock recorded and never a file's. Nothing on the load path compiles, nothing parses a layout, and a load cannot fail for want of a plan. Emitting every plan as source is the other conforming shape and would want PLAN reimplemented in the Go emitter; this leg kept one plan compiler.
- **The static data**: `<T>_FIXED_LINEAGE` (per entry: the wire hash, the layout bytes verbatim, the record size from the lock, the retired mark), `<T>_FIXED_FLOOR`, `<T>_FIXED_OWN`, and `<T>_FIXED_PLAN_CAP` / `_REMAP_CAP` as the DECLARED capacity — an entry whose plan exceeds it records `plan_too_large` on that entry (§5.9 #4).
- **LOAD is §5.3's eleven steps in order**: the header's hash taken as given, the lineage select, the floor, the byte comparison, the record size from the lock, the per-record hash BEFORE the prefill, the census once after the loop on a read that returns (§5.9 #6). `layout_newer` reports the file's hash and nothing else; `layout_unsupported` reports it too (§5.9 #7).
- **Two of §5.8's rows closed on this leg while the fixtures were green.** Row 14: the enum remap table is as long as the WRITER's variant count — reserved in the plan's pool and filled in place, not built in a capped `[u16; 256]` temporary, because a cap remaps a writer's 256th variant to `None` as though this reader did not name it, and a silent wrong value is worse than a refusal (a table the pool cannot hold is `plan_too_large` by name). Row 4's count half: the `count` op's bound is the writer's.
- **EMIT gained what the fixtures found missing**: the `present` constant for `T` into `?T`, the fixed-point ladders and a signed fixed-point's sign, and a grown enum ordinal width as an unsigned `widen` that counts `widened`. The `arg` lane is `u64` — no byte lane anywhere.
- **The floor's comparison is not emitted where the floor is 0**, the way §5.4 does not emit a clamp that cannot fire: a warnings-as-errors consumer (the clippy gate) refuses `found < 0`.

## What §5.6 retired, by name

`make tables-rust-fixedform` is green in BOTH modes again, and 22 cases in `test/rust-fixedform/src/main.rs` now print one greppable `SKIP …` line each, naming §5.6 and naming where the coverage went. Nothing was deleted; every skipped case keeps its code and keeps compiling.

- **Cross-schema reads through the load** — `an_older_writer`, `a_newer_writer`, `an_optional`, both halves of `the_ordinal_slide`, `the_text_under_an_arm`'s two compiled halves, `the_enum_extent`'s compiled half, `the_union_controls`' compiled forged-tag half, `the_prefill_holes`' compiled missing-field block, and the `plan_too_large` control — point at the LINEAGE HARNESS. Where a case has an identity half and a compiled half, the identity half still runs, still asserts the bool/present `!= 0` ruling, and still watches a forged tag land `None` for one `clamped`.
- **§1.1's twelve cases inside `the_layout_validation`** point at the LOCK's validation of what it records, plus the oracle's validation of the corpus. Ten of them say a §1.1 malformation under a KNOWN hash is one name now; the three hand-built-layout ones say a hand-built layout's hash is in no lineage entry, so the read refuses `layout_newer` before any rule of the layout's own could fire.
- **KEPT, because §5.6 keeps it**: one KNOWN-hash malformed case, built as §5.7 step 5 describes — the fixture written and read by the SAME build so its hash resolves in its own lineage, one byte of the layout behind the header flipped, the answer `layout_malformed`, refused, not malformed, no counter moved. Watched red with the break removed, then restored.
- **Two controls got stronger, not weaker**: a flipped header hash now owes `layout_newer` with the FILE'S hash on the report (§5.3's own row), and a broken entry count on the build's own file keeps a known hash and so owes `layout_malformed`.

**A stale pin this work found**: `the_bench_corpus` asserted `FIXED_TABLE_FIXED_HASH == 0x32f1c4a302a224eb`, while `bench/paired/corpus/bench_fixed.bin` carries `0x98d3af4e8ceacd29` at byte 8 and so does every other leg's generated output — the hash moved when the digest went inside `ir.TableFixedLayoutHash` (`a3fbb47d`, which re-pinned the corpus and the other legs and missed this one line). So `make tables-rust-fixedform` was ALREADY RED on `fixed-table-form` before this branch. Re-pinned to the corpus's number, and called out here because re-pinning a byte gate is the kind of edit that deserves a reader.

## What §5 did not tell me

1. **The `present` constant's guard.** §5.2's EMIT says `emit present … under guard/arg`; §5.7's C-leg row calls the same constant **UNGUARDED**. Guarded is the only reading that cannot set a present byte for an arm the tag never named, so I took it — but the two sentences disagree and a leg must choose.
2. **How an allocation-free leg sizes a static plan.** §5.9 #4 says the build sizes it "by construction" and calls discovery shape; with no allocator and a static array there is no growing, so the cap is a FORMULA a leg invents. Mine is `body + 2·entries + 256` for the plan and `1024 + 32·entries` for the remap pool. §5 gives no bound, and a too-small one is a silent `plan_too_large` on a row that should read.
3. **What happens to a leg's OTHER caller buffer.** §5.9 #5 rules on the plan slice; this leg's signature also takes `remap: &mut [u16]`, which has no owner once the plans are the build's. I kept it in the signature and stopped writing through it. Whether such a lane should be dropped, or held to a capacity of its own, §5 does not say.
4. **The per-record refusal's name in a leg whose vocabulary is "block".** §5.3 names `no_layout`; this leg carries form 1's words — `no_block`, `block_malformed` — and `layout_malformed` beside them. No fixture asserts the first two, so I kept the leg's names; the rename is a whole-leg edit and would move a dozen tests.
5. **A lineage entry whose RECORD SIZE is wrong.** §5.9 #8 rules on an entry that is not a parseable layout (a build failure, or `layout_malformed` where the lineage arrives at run time). A recorded record size of `8` or less is not that case: §5.3 step 9 makes it `malformed` with no name, which reads a lock bug as a wire event.
6. **Where a fixture's counters come from when the op already counts.** §5.4 moves the forged-ordinal `clamped` to the bounds pass "on BOTH plans"; §5.8 row 12 lists the reference's miss. Neither says what a leg does that ALREADY counts in the op with a green fixture asserting the number — whether the fixture or the op is the thing to move.
7. **The corpus's own file names past `old_`/`new_`.** §5.7 step 1 names the pair; the floor rows need `mid_floor.bin` and `lineage_merge` needs `a_`/`b_`, which only the dump states.
8. **This leg's report has no `duplicate` counter** (§5.4 lists six; the form raises five, and Rust's report carries four plus the flags). The probes assert the four that exist.
9. **How to prove "REFUSE wrote no destination byte"** in a language whose generated `Row` derives neither `PartialEq` nor `Debug`: the probes compare a byte view, which is also how the poison is laid. §5.7 says "through a byte pointer" for the poison and nothing about the comparison.
10. **THE CENSUS CANNOT BE NONZERO ON A LAWFUL LINEAGE, and §5.7's owed row cannot be written.** §5.7 says a nonzero `unknown` is proved by "a row whose OLD schema carries a field the new one dropped, over an array of tables", and calls the row owed. But §5.1 refuses a dropped field ("field removed"), refuses a rename without `was`, and keeps a deprecated field's slot — so a newer reader names every field of every writer in its own lineage, and `unknown` is structurally 0. The same argument kills `kind_mismatch`: a kind that moved off every ladder is refused at commit. A nonzero census therefore needs either a lock that is NOT monotone — a build bug, not a wire event — or a peer outside the lineage, which is now `layout_newer` before a plan exists. So §5.9 #6's "the census lands once after the loop" is machinery no lawful read can move, and the owed row wants a DELIBERATELY UNLAWFUL lineage entry handed in by a test, which is the only shape that can exercise it. Worth a ruling on the page rather than a row on a list.
11. **A pre-existing red the probes had to route around** (not §5's): with the `block` accelerator on, a `wstring(8)` table's block projection fails two §19.3 const asserts (`size_of == 56`, `offset_of!(trail) == 52`) — `probe_wstring_grow_*` compiles only with `default = []`. That belongs to `tables-rust-features`, and it wants its own card.

## Gates

`gofmt -l .` empty · `go vet ./...` clean · `go test ./internal/codegen/rusttable/` green · `make tables-rust-versioning` green (57/57) · `make tables-rust-fixedform` green, debug and release · `make tables-rust-fixed-matched` green · `make tables-rust-clippy` green. One flake seen once and named: under `go test ./...` run beside two concurrent `cargo` gates, `array_elem_widen` failed with its message clipped away; it passes alone and passes under `go test ./internal/...`, and the harness's clipping is fixed in its own commit so the next one says what happened. `tables-rust-versioning` hangs off `test-rust` and rides the `go-test` job of CI beside the Go leg's, with the `../serialize.rs` checkout the probe crates need.


gafferongames:
Read at b86c6b51: ONE PATH in substance (hash selects from static data; byte compare; LayoutNewer with the hash only; plans in a LazyLock from the lock's bytes; rows 14 and half of 4 closed; the re-pin of the_bench_corpus is right and the target was red on the base; array_fixed_grow asserts both the lead/trail bracket and the element defaults 7/9). HOLD for two CI reds of the PR's own: lint (two unconvert findings) and the versioning step (probable rustc floor on the runner: edition 2024 / resolver 3 / is_multiple_of); a child is fixing both. Doc note for §5: the present op lands under the SAME guard as its payload; at top level that guard is empty.

Issue #33, opened 2026-08-14
tooling: both open family PRs display a stale red check that would now pass, and nothing will re-trigger them


rowan-claude:
Severity: **low** · Kind: tooling · Call: ours-to-fix · Affected repos: serialize.go (#24), serialize.rs (#29), plus the spec-sync job design in all four vendoring repos

Filed on schema as the estate hub; the re-runs happen in the two PR repos.

## What is wrong

Both open PRs in the family — mas-bandwidth/serialize.go#24 and mas-bandwidth/serialize.rs#29 — display a red, required-looking "STANDARD.md matches upstream" check that would now pass, and nothing will re-trigger it without a push. The spec-sync job fetches upstream main at run time, so its verdict is a function of WHEN it ran rather than of the branch's content. (The register's own "four PRs open" count was also stale — it is two.)

## How it manifests

Anyone triaging by check status — including a future agent working from this register — reads "blocked" where the truth is "green on a re-run". These are the two PRs that between them close the family's last live wire divergence and the last two stale STANDARD.md copies.

## Evidence

> VERIFIED via our authenticated GitHub CLI wrapper pr list --state open across all six repos: exactly TWO PRs are open, not four — serialize.go#24 (branch degenerate-range-compat-and-docs, mergeable MERGEABLE, failing check 'STANDARD.md matches upstream') and serialize.rs#29 (branch docs-state-anchors, mergeable UNKNOWN, same failing check). serialize#45, serialize.c#1 and serialize.rs#31 have merged. #24's red is provably stale: its newest run 31781432311 is from 2026-08-14T07:49:30Z — before serialize PR #43 merged at 21:08 — and its branch blob is now correct: `git show degenerate-range-compat-and-docs:STANDARD.md | shasum -a 256` = b66c378bcb958980e9d3b5784fdd728dbbf5cafabb0a9db5b23aa53eea31a3bf, 433 lines, identical to serialize origin/main's.

Re-verified 2026-08-15 at filing time: still exactly two open PRs, #24 and #29.

## Fix direction

Push an empty commit or re-run the workflow on both branches to get an honest status before merging; rebase #29 over bbfe1de. Worth considering separately: a spec-sync job whose result depends on a remote fetched at run time cannot be trusted from a cached conclusion — it should record the upstream sha it compared against.

Not blocked on any DECISION issue.

---
Part of the 2026-08-15 estate known-issues register (index issue on mas-bandwidth/schema).

rowan-claude:
[...]

Issue #32, opened 2026-08-14
tooling: a 2.67 MB Mach-O arm64 executable is committed at the repo root, plus three small port defects to sweep with it


rowan-claude:
Severity: **low** · Kind: tooling · Call: ours-to-fix · Affected repos: schema (the binary), serialize (typo), serialize.c (narrowing), serialize.cs (duplication)

Filed on schema because the dominant item (the committed binary) lives here; the three small port defects are named so this can be swept in one pass.

## What is wrong

Four cheap, independent items:

1. **schema carries a 2.67 MB Mach-O arm64 executable at the repository root** (`tools`, blob baea000a, added by 5a6a25b — a `go build ./bench/tools/...` output; source is bench/tools/aggregate.go and relative.go). Architecture-specific, useless on x86, and silently stale the moment the Go source changes — which it has repeatedly (e57e7d2, d7e2896, 34b7ed8, 7db74ef all touch bench tooling). .gitignore covers /bin/ and /build/ but has no /tools entry.
2. **serialize misspells "endianness" in a user-facing #error**: serialize.h:126, :137 and :161 all read "#error Unknown machine endianess detected" — the string a consumer sees at the exact moment their build fails on an unrecognised target, and the string they paste into a search box (closed issue #7 is a user quoting it back).
3. **serialize.c narrows an unsigned clamp bound through a signed parameter**: serialize.c:191 passes a value clamped to 4294967040 (above INT32_MAX) as signed serialize_int32_t. No wire effect today (the span is computed in the unsigned domain), but out-of-range conversion to a signed type is implementation-defined at the library's stated C89 floor, and the signed/unsigned signature split against C++'s `bits_required( uint32_t, uint32_t )` is a latent trap for the next caller.
4. **serialize.cs duplicates the compressed_float quantization expression** at src/Serialize.cs:1314 and :3187 — the arithmetic that already drifted five ways once. serialize.c's own doctrine applies: "a formula copied three times is a formula that drifts twice." (Also noted in the serialize.cs intermediate-pinning issue.)

## Evidence

[...]

PR #920, opened 2026-09-11
java leg: §5, the fixed table reads backward (the lineage, LOAD's eleven steps, the two columns of every row)

gafferongames:
# The Java leg reads backward: §5 on the Java leg

**DRAFT.** Written FROM `docs/FIXED-FORM-ALGORITHM.md` §1 and §5, `docs/FIXED-FORM-BILL-READS-BACKWARD.md` and `docs/FIXED-FORM-VERSIONING-TESTS.md`, with `internal/codegen/cpptable` and the C++ runtime **unopened**. The Go pilot (#914), the C leg (#917) and the Rust leg (#918) were read as worked examples for SHAPE only.

Three commits, in §5.7's own order: **the fixtures red first**, then LOAD and the static data, then the retirements.

## What landed

**Step 1 — the fixtures, and they were watched failing first.** `internal/codegen/javatable/fixedversioning_test.go` reads the C++ reference's byte oracle out of `build/fixedform-corpus` (`old_<row>.bin` / `new_<row>.bin`, and the names only the dump states for the rows that are not a pair, §5.9 #28), generates **one Java probe per row per column** beside this leg's own generator (§5.9 #18), compiles it with this leg's `javac` and runs it. 23 of 25 rows and every floor and hash row were RED before a line of the reader existed; the two that were green are the two whose edit moves no layout byte.

**THE VALUE ORACLE IS THE LEG'S OWN OLDER BUILD**, never a number restated in the test. `NEW-READS-OLD` reads `old_<row>.bin` twice — with the OLD build, whose identity plan IS the writer's layout, and with the NEW build through its lineage — and every field the two builds share must land equal while every field only the newer build names must come back a fresh value's declared default. The destination is POISONED with `0x5A` through every field by reflection first (§5.7, §5.9 #17). That covers all 25 rows' values rather than the 2 a hand-written check can afford.

**Step 2 — §5.3's eleven steps, and the lineage as static data.**
- `GenerateLineage(u, lineage)` beside `Generate(u)`, the lineage keyed by table name, oldest first (§5.9 #1); `FixedLineageOf` so a test plays the lock in one line. The backend opens no file.
- `TableFixed.KnownLayout` with the four members §5.9 #19 names in that order — `hash`, `layout`, `layoutBytes`, `recordBytes` — and `known` is the lineage with the current layout LAST (§5.9 #2); `floor` is 1 + the highest retired index, derived in the emitter from the entries it emits.
- **One plan per lineage entry, built in a class initializer from THE LOCK'S BYTES** (§5.9 #3), each carrying its own remap table because the ordinal table is part of the plan. The caller's `plan` and `remap` arguments become CAPACITY DECLARATIONS, checked and never written through (§5.9 #5, #24), and they stay in the signature because §5.6 retires mechanisms and not API (§5.9 #16).
- LOAD: the framing checks, the header's hash **taken as given**, the lineage select, the floor, the byte comparison against the lock's bytes, the record size **from the lock**, the file's own tail arithmetic, the caller's capacity, and the per-record hash **before** the prefill. `layoutNewer` and `layoutUnsupported` in the order §5.3 names them (§5.9 #14), both reporting the FILE's hash on `layoutHash`, last on the report (§5.9 #15, #7). The census lands once after the loop, only on a read that returns (§5.9 #6).
- **Three rows of EMIT this leg did not have**: `T` into `?T` as the `present` op, a constant 1 at the companion lane carrying the row's own guard and ordinal (§5.9 #13); the `fixed(I,F)` ladder rungs `20..24` and `25..29` with a signed fixed-point's sign; and a grown enum ordinal width as an unsigned `widen` that counts `widened` (§5.2 EMIT kind 30, bill §12.7).

**Step 5 — seven cases retired by name**, printed at the call site with where the coverage is owed, the function left in the tree (§5.9 #23), plus `TestFixedFormFU1FU2` skipped by name in Go.

**The target and CI.** `make tables-java-versioning` depends on `tables-fixedform-corpus` and sets `SCHEMA_REQUIRE_CORPUS=1`, under which the harness's skip is a FAILURE; the step is wired into `ci.yml`'s `go-test` job beside the Go leg's, because the harness is a Go test that drives `javac`.

## The rows

🟢 green, 🔴 red, ⚪ no such column (the edit moves no layout byte and no digest byte, so both sides take the identity plan and it reads BOTH ways — §5.7).

|---|---|---|
| `field_append` | 🟢 | 🟢 |
| `field_deprecate` | 🟢 | ⚪ |
| `field_undeprecate` | 🟢 | ⚪ |
| `enum_append` | 🟢 | 🟢 |
| `enum_width` | 🟢 | 🟢 |
| `union_append` | 🟢 | 🟢 |
| `union_arm_payload_widen` | 🟢 | 🟢 |
| `flags_append` | 🟢 | 🟢 |
| `keyed_array_enum_append` | 🟢 | 🟢 |
| `nested_append` | 🟢 | 🟢 |
| `rename_without_was` | 🟢 | ⚪ |
| `array_bounded_grow` | 🟢 | 🟢 |
| `array_elem_widen` | 🟢 | 🟢 |
| `array_fixed_grow` | 🟢 | 🟢 |
| `bits_grow` | 🟢 | 🟢 |
| `bytes_grow` | 🟢 | 🟢 |
| `constant_grow` | 🟢 | 🟢 |
| `fixed_I_grow` | 🟢 | 🟢 |
| `float_widen` | 🟢 | 🟢 |
| `int_widen` | 🟢 | 🟢 |
| `optional_add` | 🟢 | 🟢 |
| `range_widen` | 🟢 | 🟢 |
| `string_grow` | 🟢 | 🟢 |
| `uint_widen` | 🟢 | 🟢 |
| `wstring_grow` | 🟢 | 🟢 |

**Every row green, both columns.** The last one to fall was `union_arm_payload_widen`, whose appended field lives INSIDE a union arm: the prefill image's union case returned early — "a fresh union is None, tag 0, which is the zero image" — which is true of the TAG and false of every arm's own defaults, so a reader-appended field inside an arm landed 0 where its declaration said 55. The image now walks the arms in DECLARED ORDER and resets each at its own overlay storage before leaving the tag at None, which is §5.2's sentence implemented rather than paraphrased. See finding 7 for what a porter has to infer there.

| the floor and the hash | state |
|---|---|
| `floor_at` — nothing retired, all three of the lineage read | 🟢 |
| `floor_below` — entry 0 retired: the oldest refuses `layout_unsupported` with the file's hash, no counter, nothing decoded, and the file AT the raised floor still reads | 🟢 |
| `floor_raise_live` — the floor raised by one more: yesterday's file refuses today | 🟢 |
| `hash_identity` — the reader's own hash selects the identity plan | 🟢 |
| `hash_unknown` — a moved header hash is `layout_newer` with the file's hash and nothing else | 🟢 |
| `hash_known_bytes_differ` — a known hash whose layout bytes differ is `layout_malformed` | 🟢 |
| `lineage_merge` — both pre-merge writers read on the merged build | 🟢 |

## The gates

| gate | state |
|---|---|
| `gofmt -l .` | empty |
| `go test ./internal/codegen/javatable/ ./internal/goldens/` | 🟢 (7.1 s / 1.2 s) |
| the repo's lint (golangci-lint v2.12.2, modernize v0.23.0) | 0 issues (3.5 s) |
| `make tables-java-fixedform` (the byte gate) | 🟢, with the seven §5.6 retirements printed |
| `make tables-java-fixedform-negative-control` (the write template's byte) | 🟢 — still reds the reference match |
| `make tables-java-fixedform-arm-negative-control` (the arm's text lane) | 🟢 — still reds the arm's text |
| `make tables-java-versioning` | 🟢 — 25 rows, both columns, three floor rows, three hash rows, the lineage merge (9.9 s) |
| `make tables-java-versioning-negative-control` (§5.9 #11) | 🟢 — **written**, and it reds on the refusal NAME (7.0 s) |

## What §5 did not tell me

The most valuable output of porting from the page alone. Each is a place the page is silent and a porter must choose.

1. **The NEW-READS-OLD column has no VALUE ORACLE a leg can read.** §5.7 says "every old value lands exactly" and §5.9 #9 says which root to read each file with — but the corpus is bytes, and the values live in `fixedform_dump.cpp`'s source, which a leg porting from the page does not open. So every leg invents an oracle, and the Go pilot's covers 2 of its 26 rows. **The rule I took, and which §5 should state:** the leg's OWN OLDER BUILD reading `old_<row>.bin` under its identity plan IS the oracle, field by field, with the newer build's fresh value the oracle for the tail. One extra probe run per row, every row covered.
2. **The `0x5A` poison assumes a byte view of the destination.** §5.7 says "through a byte pointer"; §5.9 #17 rules on the COMPARISON in a language with no byte view but not on the POISON. The rule a porter needs: **the poison is per FIELD, through whatever assignment or reflection the language has** — it is the same test, and it is what caught `array_fixed_grow`'s element defaults here.
3. **"The same value" is undefined across two generations in a host language with no unsigned integers.** A `uint16` is a Java `short` on the older build and an `int` on the newer, so one landed value prints `-1` and `65535` and a naive comparison reds on a correct read. The rule I took: the only admitted difference is the UNSIGNED REINTERPRETATION of the narrower storage, which is exactly the ladder's zero-extension; a sign-extension where a zero-extension was owed still fails.
4. **A `was =` rename breaks every name-keyed comparison across generations.** §5.2 says the id matches by the OLD name, and the VALUE surface is spelled with the NEW one. §5.7 calls `rename_without_was` a both-directions row and never says how a harness pairs `b` with `a was = b`. The harness now translates the newer build's names through the rename map — which is the same sentence §5.1 already writes about ids, read from the test's side.
5. **THE PRESENT COMPANION IS THE ONE NEWER-ONLY FIELD THAT IS NOT A DEFAULT.** `optional_add`'s reader gains a present byte, and it owes `1` and not its fresh value's `false` (bill §12.8). §5.7's "the reader's tail is its declared default" is true of every other appended thing and false of this one, and the page never draws the line.
6. **Java's 64 KB class-initializer limit is named as a cost, never as a shape.** Bill §11 says legs carry layout bytes "as a resource or a string constant where the language needs it"; §5.2 says emit the lineage's bytes, sizes, plans and floor as static data; neither says which, and the limit is on the whole initializer rather than on one constant. **What this leg does:** every OLDER lineage entry's layout rides as an octal-escaped STRING decoded once (`TableFixed.bytes(String…)`), never as a `byte[]` literal, because an array initializer is code and a lineage multiplies `4 + 17E` bytes by its entry count. **And the `int[] dest` lane has no such fallback today** — one per reader rather than per entry, so it is not yet the binding constraint, but it is the next one and it wants the same chunking.
7. **"Arm by arm" is a mechanism only once you work out that the arms OVERWRITE each other.** §5.2 says the prefill image is reset "arm by arm, at each arm's own overlay storage, and only then the tag to None", and warns that an arm whose defaults the image does not carry becomes a silent zero. What it does not say is the consequence that makes the sentence implementable in a FLAT image: the arms overlay, so declared order decides and a later arm overwrites an earlier one's bytes where the two meet — an arm's default survives wherever no later arm covers it, and a default under a byte two arms both claim is a hazard no flat image can avoid. A porter reading the hazard sentence as a WARNING (as this leg's author first did) looks for a per-arm mechanism that is not there and leaves the row red; a porter reading it as an INSTRUCTION writes four lines. The page should say which, and should say what happens to the byte two arms claim.

8. **A leg's NEGATIVE CONTROL can be pinned to a case §5.6 retires.** §5.7 step 5 rules on the retired test and says nothing about the control that watched it. Two of this leg's three controls survived only by luck — the arm-text control's case reads through `parse`/`compile`/`run` directly rather than through `load`, so §5.3's refusal does not reach it — and the `plan_too_large` control did NOT survive: with plans static and a stranger's file refused before any plan is selected, a one-entry plan slice cannot reach the check, so the control moves to the lineage harness with the coverage. **The rule §5 owes: a control moves with the case it watches, and a leg says in its PR which of its controls moved.**
9. **Nothing in the tree CALLS the entry point §5.9 #1 names.** `compiler/target_java.go` and `compiler/target_go.go` both call the plain `Generate`, so `lockfile.Open`/`Lineage`/`Floor` reach no backend on ANY leg: a leg can port §5 whole and still not ship a lineage in a real build, only in a test that plays the lock. §5.8 rows 1 and 2 hold the REFERENCE to reading the lock; the CALLER's three lines are nobody's named work item.

## Owed, named rather than faked

- **The versioning gate's negative control is no longer owed.** `make tables-java-versioning-negative-control` plants the WRONG REFUSAL NAME for a file outside the lineage — `noLayout` where §5.3 says `layoutNewer` — through a `go test -overlay`, and requires the suite to go red and to print which name it wanted (`OLD-REFUSES-NEW: refused=true reason=noLayout, want a refusal named layoutNewer`). The name is the whole of §5.3's answer and every counter still reads refused under the sabotage. It is the retired `plan_too_large` control's replacement: **the control moved here with its case**, which is the rule item 8 above asks §5 to state.
- **The FU1/FU2 and UT pairs as lineage pairs** (§5.7's seven pairs), which is where `textUnderAnArm`'s and `TestFixedFormFU1FU2`'s coverage is owed.
- **The hostile pass under the WRITER's bounds** (§5.8 row 4), the **remap table past 255 entries** (row 14, and this leg's table is a `short[]` like the reference's), and the **clamp count on a forged ordinal under a compiled plan** (row 12): all three are the reference's debts, reproduced here, left red and named rather than matched or faked.
- **`arg` is still a BYTE lane.** #876 card 14's specific bug is not present on this leg — `arg` is the arm ordinal alone and the text flavour rides in its own `meta` byte — but the lane is a byte, which §5.2 and bill §12.7 refuse, and widening it to full width is not in this PR.
- **The `generated` tree is re-rendered in this PR.** `generated/bench/paired/java/FixedTableFixed.java` carried the layout-bytes-only known hash; #916 filled the §13 definitions digest, which moves the wire hash without moving a layout byte, so the committed file is regenerated (`0x98d3af4e8ceacd29` → `0x6237c1dc195f9ec9`) along with `TableFixed.java`. Nothing else under `generated/` carries this leg's fixed runtime.

## The one residual, and it is now bounded rather than named

**THE LINEAGE PLANS NO LONGER ALLOCATE FOR A ROOM NOTHING NEEDS.** `lineagePlans` grew and retried from 256 by fours and gave up at `1 << 18`, which put an `Entry[262144]` plus a `short[524288]` inside a CLASS INITIALIZER — and an `OutOfMemoryError` under `<clinit>` is not a caught error: the JVM turns it into an `ExceptionInInitializerError` and POISONS the class for the life of the loader, so one oversized pair would take every later read of a perfectly good file down with it. It was the one way this leg's load path could fail that no refusal name covered.

Every entry a compile writes is either one of MY entries taking a value or one of THEIRS being stepped over, and coalescing only ever shortens the run, so

    room = me.count + theirs.count + 1

bounds the plan for that pair — **both counts the lock's own**, read out of the layout bytes the lock recorded, which makes the size a generate-time number and not a search (§5.9 #21: a stated formula). One allocation, no cap, no loop. A compile that still does not fit would be the FORMULA being wrong rather than a layout being large, and it is refused by the same `planTooLarge` name the cap used, so nothing about the load path's answers moved — and the cap's own negative control, which now has nothing left to red, moved to the refusal-name control above.

**The driver's lineage now reaches this leg (#925, §5.9 #1).** `target_java.go` answers `GenerateLineage`, `java` joins `fixedLineageShipTargets`, and the ship test goes red without it. One fact is not a copy of the Go leg: the lock records a BODY size and §5.2's `record_bytes` is the WHOLE record, `8 + C(root)`, so the eight bytes are added in `javaTableLineage` exactly as the C++ reference adds them off the same lock — handed through raw it read every record of a LOCKED unit eight bytes short and `make tables-java-fixedform` went red in 25 cases. `compiler/target_go.go`'s `goTableLineage` passes the lock's number through unchanged and looks to have the same defect, unseen because the Go fixed-form gate's fixtures carry no lock.

---

## Since the cold read

A cold read of this leg against `docs/FIXED-FORM-ALGORITHM.md` §5 found seven items. Six are in `da4abf32`; the
state of each is below, and what is NOT done is named rather than quietly dropped.

**1. THE GUARD'S WIDTH — 🟢 green, was WRONG BYTES and hostile.** The guard was compared at ONE BYTE
(`fixedruntime.go:407`), the arm ordinal rode a `byte` lane (`:282`), and the union stamped `(byte)(j+1)` with no
width at all (`:956-959`). §5.2 "THE GUARD'S WIDTH": `argw` is the guard's width in bytes, `1..8`, zero reading as
one, and the tag is compared WHOLE. A union now stamps its own tag width onto the tag entry and onto every entry
beneath it, restoring the outer width on the way out so a nested union leaks nothing upward; `arg` is a `long`,
because a byte lane could hold neither an arm past 127 nor what a two-byte tag carries (§5.8 row 5: no byte lane);
`coalesce` will not fold two entries of different widths. **Red first**: a hand-built layout whose union tag is TWO
bytes, read with a forged `0x0101`, landed arm 1's tag and its eight payload bytes before the fix and lands nothing
after it — the prefill's `None`.

**2. THE UNSIGNED ORDINAL — 🟢 green on the bytes and the throw; ⚪ the counter's PLACE is NOT done.** The ordinal
op read an eight-byte ordinal as a SIGNED long (`:439-452`), so `0x8000000000000000` passed `raw <= n`, truncated to
`0` under `(int) raw` and returned the remap table's own LENGTH WORD as a variant, and `0xFFFFFFFFFFFFFFFF`
truncated to `-1` and indexed BEFORE the table — an `ArrayIndexOutOfBoundsException` escaping a load that was asked
a question. `Long.compareUnsigned` is the fix: every value past the writer's table length, the negative ones
included, lands `None` and throws nothing. **Red first**: the forged pair landed `2` and then threw; they now land
`None`, `clamped == 1`, nothing malformed.
**What is NOT done is §5.9 #27**: the forged ordinal's `clamped` is still counted IN THE OP and not in the bounds
pass. Moving it is not a deletion — on a COMPILED plan the op lands `0`, which the emitted bounds pass reads as
`None` and correctly does not clamp, so the count can only move once the plan carries THE WRITER's variant bound
into that pass (§5.2's "the plan also carries the WRITER's bounds for the hostile pass"). Removing the op's count
without that makes `clamped` `0` on the compiled plan and `1` on the identity one — the disagreement #27 exists to
prevent — so it is left in one place, named here, and owed as its own change.

**3. THE REMAP TABLE'S LENGTH — 🟢 green, was WRONG BYTES.** The table was capped at 255 (`:973`,
`Math.min(theirVariants, 255L)`), so a writer's 256th variant on silently landed `None` as though this reader did
not name it — a wrong value and not a version difference (§5.2; §5.8 row 14 holds the reference to the cap, not this
page). It is now as long as the WRITER's variant count. **Red first**: 300 variants, variant 299 landed `0` before
and lands `299` after, with `clamped == 0`.

**4. THE REPORT CONTRACT — 🟢 green on the bound and the lane; ⚪ no probe, and why.** An entry past the writer's
record was a per-record `malformed` set AFTER values had landed (`run`'s re-checks against `srcLength`, `:400-477`),
so a report could carry `n >= 1` and `malformed` at once, which §5.3's joint table does not allow. Every entry is
now bounded ONCE, at compile, against the writer's OWN declared record — `src + the op's reach`, and `guard + argw`
too — and one entry past it refuses the plan WHOLE on that lineage entry's own lane by the name
`layoutRecordTooLarge` (§5.9 #36, #45: two failure codes and not one). **THE REACH IS THE OP'S AND NEVER THE `size`
LANE'S**: for `count` that lane is THIS READER's bound, which a grown array makes larger than the writer's whole
record while the entry still reads four bytes, so a bound reading `size` for every op would refuse every lawful
array that grew. The run loop's own checks stay as defence.
**No probe lands the refusal**, and the reason is the parse: a table's size must equal the SUM of its children and
every offset a compile walks is derived from that walk, so on a layout that PARSES no entry can reach past the root
— the lane is unreachable-by-construction today rather than dead. "A clean read never sets `malformed`" is held by
`tables-java-fixedform` and both columns of every §5.7 row.

**5. THE VERSIONING HARNESS — 🔴 NOT DONE.** `fixedversioning_test.go`'s oracle is still the leg's OWN OLDER BUILD
(`:24-31`, `:572-585`) and it still softens three ways (`:631 continue`, `:709 sameNumber`, `:641 bytes: prefix`),
which §5.9 #32/#37 rule out: the values ASSERTED must be `build/fixedform-corpus/manifest.txt`'s, the older build at
most a second column. Nothing in this change touched it. Also still owed with it: `widened == exactly 4` on
`array_elem_widen` (`:761-763`, §5.9 #33 — the row asserts only `!= 0` today), `present == 1` on `optional_add`,
`wstring_grow` marked ⚪ UNPORTED by name citing `compiler/widetext.go`'s refusal rather than generated around
(§5.9 #44), and `hash_identity` asserted by shape. This is the largest item of the seven and it is the one left
whole.

**6. THE CONTROL NOBODY RUNS — 🟢 green.** `tables-java-versioning-negative-control` was written, registered in
`make/negative-controls.json` and hung off NO recipe, so only the negative-controls matrix ever ran it and `make`
itself could not. It now sits in `tables-java-release` beside the byte gate's own control — and **not** in
`test-java`, because it rebuilds the compiler over a sabotaged emitter and runs both columns of every §5.7 row
behind it, which is the expensive half `tables-java-release` exists to hold (`make/java.mk:505-513`: `make test` sits
at fourteen minutes against a fifteen-minute timeout). CI's java group already enumerates it —
`go run ./tools/negativecontrols targets java` lists it, `check` passes, and `matrix` includes the `java` row — so
the CI half needed nothing. It reds for the right reason: `reason=noLayout, want a refusal named layoutNewer`.

**7. THE TAIL — partly.** 🟢 the false comment at `fixedruntime.go:1060-1063` ("the tag is unguarded and lands" —
the tag entry is a `const` GUARDED on the writer's own tag, so the tag's storage is a hole like every other arm
byte). 🟢 `Header.declaredHash` deleted with its stale comment (`:207`/`:235`): it held a copy of the hash nothing
ever read, and the recompute it was named for is retired by §5.6.
🔴 **NOT DONE**: the unknown census still counts per field per ELEMENT rather than once per field per peer
(`:849` reached from `:911-935`); `plan_too_large` from the leg's own declared cap on the IDENTITY lane and for a
null plan (`fixedform.go:1631`, §5.9 #45); `fu1`/`fu2` still unread in `Main.java` (`:938-947` reads six of nine,
and `fxw` stays ⚪ with the refusal).

### Gates, on this head

| gate | time | state |
|---|---|---|
| `go test ./internal/codegen/javatable/ ./compiler/ -run 'Fixed\|Lineage\|Java'` | 18.5 s | 🟢 |
| `make tables-java-fixedform tables-java-versioning tables-java-versioning-negative-control` | 30.3 s (corpus built first, `SCHEMA_REQUIRE_CORPUS=1`) | 🟢 |
| `make tables-java-compile` (`javac -Xlint:all -Werror`) | 0.8 s | 🟢 |
| the new hostile driver, compiled `-Xlint:all -Werror` | 0.8 s | 🟢, and red on each of the three sabotages |
| `gofmt -l .` | — | empty |
| regen sweep (`rm -f bin/schema`) | — | `generated/bench/paired/java/TableFixed.java` re-pinned, 156/28, the same hunks as the emitter's |


gafferongames:
Read note for the merge: this branch's javaTableLineage adds 8 to the lock's Record (the lock stores the body size; §5.2's record_bytes is the whole record). A compiler-level fix (branch rowan/lineage-record-size) makes FixedLineage carry the whole record for every target; when it merges, drop the +8 here or the Java entry will be eight over. Go and C had the eight-short bug from #925's helper; the new ship test catches it per target.

Issue #31, opened 2026-08-14
harness: the estate's largest Go codebase has no Go linting at all, while serialize.go runs six quality gates


rowan-claude:
Severity: **medium** · Kind: harness · Call: ours-to-fix

## What is wrong

The estate's largest Go codebase has no Go linting at all — no vet, no golangci-lint, no gofmt check, no govulncheck, no tidy check — while serialize.go runs six separate quality gates. There is also no Windows CI leg, so the compiler that generates C++ for a Windows-targeting game engine is never itself run on Windows.

## How it manifests

The repo with five code-generating backends, an IR that four backends render from, and the register's largest cluster of correctness findings has the weakest static analysis in the estate — none. `go vet` alone catches printf-verb mismatches across an emitter codebase built almost entirely out of format strings, which is exactly where the C-backend naming and screaming-snake defects live.

## Evidence

> VERIFIED: `grep -niE 'go vet|golangci|gofmt|govulncheck|staticcheck|go mod tidy' schema/.github/workflows/ci.yml schema/Makefile` returns nothing. schema's entire ci.yml is one job (`test`) on `os: [ ubuntu-latest, macos-latest ]` running `make test` and `go test ./internal/fuzz/`. CONTRAST serialize.go/.github/workflows/ci.yml: `go vet ./...`, `go test -short -race -shuffle=on ./...`, `go mod tidy -diff`, golangci-lint pinned at v2.12.2, `modernize`, and a separate `vuln` job — plus a windows-latest test leg and a `go: 1.23.x` floor leg.

## Fix direction

Lift serialize.go's lint job verbatim into schema's ci.yml — it is a self-contained job with no repo-specific configuration — and add a windows-latest entry to the test matrix (the `make test` leg needs the five sibling clones, so a vet/lint-only Windows job is the cheap first step).

Not blocked on any DECISION issue.

---
Part of the 2026-08-15 estate known-issues register (index issue on mas-bandwidth/schema).

Issue #30, opened 2026-08-14
harness: 130 generated files are committed and never once proven to match the current generator's output


rowan-claude:
Severity: **medium** · Kind: harness · Call: ours-to-fix

## What is wrong

130 generated files are committed under generated/ and nothing ever proves they match what the current generator emits. CI regenerates the committed tree into the workspace, runs against it, and discards the workspace — the files in git are never compared to the generator's current output.

## How it manifests

A backend change that alters emitted text produces a green build and a stale tree, and nobody learns until someone reads a file. That matters more here than in most repos: the committed generated/ tree is what this register treats as ground truth — the C-backend naming finding is receipted against generated/c/ObjectsWire.h:298 vs generated/rust/src/objects.rs:470 vs generated/cpp/ObjectsWire.h, and the C++ constant-folding finding against generated/cpp/Constants.h:21. Those receipts are only sound while the tree is current, and nothing enforces that.

## Evidence

> VERIFIED: `git ls-files generated | wc -l` = 130 tracked files under generated/ (only the .stamp files are gitignored — .gitignore lines 17-24). `grep -niE 'git diff|git status|--exit-code' .github/workflows/ci.yml Makefile` returns NOTHING. The Makefile's generated/*/.stamp rules (Makefile:40, :47, :51, :56, :61, :74, :82, :89) run `./bin/schema generate ... --out generated/...`, overwriting the tracked tree in place, and `test:` (Makefile:180) depends on those stamps. CI's only steps are `make test RUSTUP_BIN=/usr/bin` and `go test ./internal/fuzz/`.

## Fix direction

One step after `make test` — `git diff --exit-code generated/` — with a message pointing at the regeneration command. It also converts every generated-output change into a reviewable diff, which is the cheapest possible guard for the emitter-naming and folding classes already in the register.

Not blocked on any DECISION issue.

---
Part of the 2026-08-15 estate known-issues register (index issue on mas-bandwidth/schema).

Issue #29, opened 2026-08-14
correctness: C++ data-header constant folding depends on schema declaration order


rowan-claude:
Severity: **medium** · Kind: correctness · Call: ours-to-fix

## What is wrong

C++ data-header constant folding depends on schema declaration order. emissionOrder builds its dependency graph from Struct/Enum/Flags/Object only — no `*ir.Const` case — and its edges only from named field-type references, so const→const expression dependencies contribute no edge and constants stay in declaration order. A const that references a later const folds to the literal instead of rendering symbolically.

## How it manifests

schema declarations are order-free (SPEC §4.2), but the C++ data header is not a function of the schema's meaning — it is a function of the order the author happened to write the constants in. Values are identical either way (folding is arbitrary-precision), so there is no wire or value defect; what is lost silently and with no diagnostic is the symbolic link that makes the generated header maintainable. It is also a cross-target inconsistency: Go, Rust and C# render symbolically regardless of order, and C always folds — C++ is the only target whose output text moves with source order. Observable today: reorder the two declarations in examples/Constants.schema and `MaxPositionUnits = MaxWorldMeters * PositionUnits` becomes the folded literal.

## Evidence

> VERIFIED READ. internal/codegen/cpp/cpp.go:284 emitDataFile resets g.emitted to empty; cpp.go:872 renderable() returns g.emitted[e.Name] for a same-file identifier ('same file: only if already emitted above'); cpp.go:824-832 renderInt folds to the literal when !renderable. cpp.go:383-416 emissionOrder builds its byName index from Struct/Enum/Flags/Object only — no *ir.Const case — and its edges only from f.Type.Kind == ir.TNamed field references, so const→const expression dependencies contribute no edge and constants stay in declaration order. Contrast: the wire header pre-marks every same-file const emitted (cpp.go:341-345) so it never folds. Observable today at generated/cpp/Constants.h:21 `inline constexpr int64_t MaxPositionUnits = MaxWorldMeters * PositionUnits;` — reorder the two declarations in examples/Constants.schema and that line becomes the folded literal.

## Fix direction

Add const declarations and their expression dependencies to emissionOrder's graph so a forward-referencing const is emitted after its dependency and renders symbolically.

Not blocked on any DECISION issue.

---
Part of the 2026-08-15 estate known-issues register (index issue on mas-bandwidth/schema).

Issue #24, opened 2026-08-14
correctness: two fixed-point rounding rules ship, and three places in the source assert they are the same rule


rowan-claude:
Severity: **high** · Kind: correctness · Call: ours-to-fix

## What is wrong

schema ships two fixed-point rounding rules, and three places in the source assert they are the same rule. The emitters emit the pure shift `( raw + half ) >> drop` (ties toward +infinity, matching SPEC.md:1360), while the data compiler rounds half-away-from-zero — and ir.go:225-226, encode.go:375-377 and pack_test.go:212-213 each claim the two are one rule. Both rules are independently in the spec (SPEC.md:1358-1361 vs :1385-1386); the false claims are in the code comments.

## How it manifests

Nuance verified rather than assumed: internal/pack does not implement object views, so today no single value passes through both rules and no currently-pinned byte is broken. The LIVE defect is the documentation — and it is live regardless of how the unification fork (#17) is settled: the next person who makes pack emit shallow views, or who writes an app-side decimal-to-raw bridge against the IR comment, will pin the wrong bytes on negative ties, and the comments will tell them they are right. ir.go is the source of truth four backends render from.

## Evidence

> VERIFIED READ. internal/ir/ir.go:225-226 says rule 2b Quantize is 'a round-to-nearest (half away from zero) arithmetic shift'. The emitters actually emit `( raw + half ) >> drop` — internal/codegen/cpp/objects.go:245-246, and the same in golang/objects.go and rust/objects.go — whose own comment at cpp/objects.go:241 says 'ties toward +infinity', matching SPEC.md:1360 ('ties resolve toward +infinity'). Meanwhile internal/pack/encode.go:375-377 rounds half-away-from-zero (ratRoundHalfAway) and its comment claims it is 'the family's rounding convention, the one the generated shallow-narrowing Quantize uses'. internal/pack/pack_test.go:212-213 repeats the false claim. Pinned behaviour of the +inf rule: test/go-ludicrous/main.go:213 '-1.5 eighths ties toward +inf to -1 (half-away would say -2)'. Both rules are independently in the spec: SPEC.md:1358-1361 vs SPEC.md:1385-1386.

## Fix direction

[...]

PR #923, opened 2026-09-11
The Dart leg reads backward: §5 from the page, red first

gafferongames:
The Dart leg reads BACKWARD: `docs/FIXED-FORM-ALGORITHM.md` §5 implemented from the page, with `internal/codegen/cpptable` and the C++ runtime unopened. The fourth leg after the Go pilot (#914), C (#917) and Rust (#918); the Go leg was read as a worked example, the C++ was not.

**THE SHAPE THIS LEG TOOK** (§5.9 #3 asks a leg to say which): **plans built LAZILY, ONCE PER ISOLATE, from the lock's own bytes.** A Dart top-level `final` is initialised on first read and never again, and statics are per-isolate, so `<table>FixedLineagePlans` is built off every load path, with no lock and no cache a load can miss: the timing is "the first load that touches the list", the thread safety is the isolate (no two isolates share the list, so there is no flag to race), and a plan that will not build is a lane carrying its own refusal name rather than a throw at the first file that selects it. Nothing on the load path compiles, and nothing on it parses a layout a FILE carried.

## The rows

`make tables-dart-versioning` — 25 row pairs × 2 columns, the three floor rows, the three hash rows, the branch merge. The probes are GENERATED, one per row per COLUMN (§5.9 #18), run with the pinned Dart SDK. Every row was RED before the reader existed (commit 1: `FixedLineageEntry`, `FixedLineageOf`, `GenerateLineage` undefined).

|---|---|---|
| `array_bounded_grow` | 🟢 | 🟢 |
| `array_elem_widen` | 🟢 `widened` | 🟢 |
| `array_fixed_grow` | 🟢 poisoned `0x5A` | 🟢 |
| `bits_grow` | 🟢 `widened` | 🟢 |
| `bytes_grow` | 🟢 | 🟢 |
| `constant_grow` | 🟢 | 🟢 |
| `enum_append` | 🟢 | 🟢 |
| `enum_width` | 🟢 `widened` | 🟢 |
| `field_append` | 🟢 x,y,z exact, w = 77 | 🟢 |
| `field_deprecate` | 🟢 | 🟢 reads both ways (one hash) |
| `field_undeprecate` | 🟢 | 🟢 reads both ways (one hash) |
| `fixed_I_grow` | 🟢 `widened` | 🟢 |
| `flags_append` | 🟢 | 🟢 |
| `float_widen` | 🟢 `widened` | 🟢 |
| `int_widen` | 🟢 −1, MIN, MAX; neighbours intact | 🟢 |
| `keyed_array_enum_append` | 🟢 poisoned `0x5A` | 🟢 |
| `nested_append` | 🟢 | 🟢 |
| `optional_add` | 🟢 | 🟢 |
| `range_widen` | 🟢 `clamped == 0` | 🟢 |
| `rename_without_was` | 🟢 | 🟢 reads both ways (one hash) |
| `string_grow` | 🟢 | 🟢 |
| `uint_widen` | 🟢 `widened` | 🟢 |
| `union_append` | 🟢 | 🟢 |
| `union_arm_payload_widen` | 🟢 an APPEND: `0, 0` | 🟢 |
| `wstring_grow` | 🟢 | 🟢 |
| `floor_at` / `floor_below` / `floor_raise_live` | 🟢 / 🟢 `layout_unsupported` + the file's hash / 🟢 | — |
| `hash_unknown` / `hash_known_bytes_differ` / `hash_identity` | 🟢 `layout_newer` + the file's hash / 🟢 `layout_malformed` / 🟢 | — |
| `lineage_merge` | 🟢 both pre-merge files read on the merged build | — |

The two poison rows assert the appended slots carry THE ELEMENT'S OWN DEFAULTS (`Vec { x = 7, y = 9 }` past the old bound, `Qty { n = 7 }` in the slot the appended key opened), and the poison is proved to bite: switching the prefill off through `go test -overlay` reds both rows with `0x5A5A5A5A` in the slots.

Green: `gofmt -l .` empty, `go test ./internal/codegen/darttable/`, `make tables-dart-fixed-form`, `make tables-dart-fixed-form-negative-control`, `make tables-dart-versioning`, `make tables-dart-versioning-negative-control` (new, §5.9 #11), `make tables-dart-clean`, `make tables-dart-names-negative-control`.

## What landed

- **LOAD, §5.3's eleven steps in order**: framing, the header's hash TAKEN AS GIVEN, the lineage select, the floor, the layout held to a BYTE COMPARISON against the bytes the lock recorded, the per-record hash BEFORE the prefill. `layout_newer` (15) and `layout_unsupported` (16) in the order §5.3 names them — the names are the contract, the integers are this leg's (§5.9 #14) — both filling `report.layoutHash`, the LAST member of the report and zero on every other path (§5.9 #15, #7).
- **The static data**: `GenerateLineage(u, lineage)` beside `Generate` (§5.9 #1); `<table>FixedKnown` oldest first with the build's own layout appended last (§5.9 #2), entries spelled `TableFixedKnownLayout(hash, layout, layoutBytes, recordBytes)` (§5.9 #19); `<table>FixedFloor` = 1 + the highest retired index; one plan per entry from the lock's bytes.
- **Three §5 rows the compiler owed**, each found by a red fixture, not by reading the C++: `T` into `?T` lands `present`, a constant `1` at the aux lane carrying the row's own guard and ordinal (§5.9 #13); the ladder's fixed-point runs `20..24` / `25..29` with a signed fixed-point sign-extending; a grown enum ordinal width is a `widen`, unsigned, counting `widened` (bill §12.7).
- **The count op's bound is the WRITER's** (§5.2's EMIT, bill §12.5) — the count half of §5.8 row 4, closed on this leg.
- **The census lands once**, after the record loop, on a read that returns (§5.9 #6); it is correctly zero on every lawful row (§5.9 #30).
- **Eight tests retired by §5.6, skipped BY NAME and never deleted.** This suite is a `main` with no skip verb, so each is a printed line at the CALL SITE with a count at the end (§5.9 #23): the four cross-schema compiled reads (`fxCase`, `fuCase`, `vCase`, `pCase`), `layoutValidation`'s run-time walk of §1.1's seven rules, the header-hash recompute, the load-path compile, and `plan_too_large` over a stranger's file. `no_layout` KEEPS its assertion and moves to a file this reader's lineage holds, because the select now runs first. THREE OF THE EIGHT WERE NOT FUNCTIONS: they were inline blocks with no name to keep, so they are deleted and their replacement is named in their place (`hash_known_bytes_differ`, `hash_unknown`) — recorded here for the merger rather than left as a silent subtraction from the count.
- **The API is not what §5.6 retires** (§5.9 #16, #24): `TableFixedPlanCache` and the plan's `theirs` / `mine` / `hash` / `ready` / `recordBytes` lanes stay, with the reason at the declaration; the plan's entry capacity stays a CAPACITY DECLARATION, checked against the selected entry's count and never written through (§5.9 #5).

## What §5 did not tell me

1. **Which scratch is the BUILD's and which stays the CALLER's, when a leg's plan object bundles them.** §5.9 #5 rules on the plan's entries and #24 on "a second scratch lane", but this leg's `TableFixedPlan` carries the entries, the REMAP POOL, the record IMAGE, the prefill range list and a float-conversion lane in one object. The entries and the remap pool had to become the build's (per lineage entry, static) because `aux` is an index INTO the pool — §5.2 says `aux` is "the table's BYTE OFFSET from the plan's base", which silently assumes entries and pool share one allocation; where they do not, a lane's entries are meaningless without the pool that travels with them. The image and the conv lane stayed the caller's. The page should say: **whatever an `ordinal` entry's `aux` indexes is part of the plan and moves into static data with it.**
2. **A single failure code cannot separate a lock bug from a capacity miss.** §5.2's PLAN refuses `layout_malformed` when either layout fails to parse and `plan_too_large` when the entries do not fit; §5.9 #4 and #8 rule on each. This leg's compiler returns one `-1` for "my own layout did not parse" and for "the pool or the entries overflowed", so a build that cannot parse THE READER'S OWN layout grows to the cap and then records `plan_too_large` — the wrong name for a reader-side bug. §5 does not say how a leg whose compiler has one failure code should tell them apart (re-parse its own layout up front, or carry two codes). Owed, and named here rather than papered over.
3. **Where the `0x5A` POISON goes when the destination is not a byte image.** §5.7 poisons "the destination through a byte pointer" and §5.9 #17 says the refusal column compares VALUES, not slack — but on this leg the load's destination is a generated OBJECT and the prefill lands in the plan's record IMAGE, which is a different thing. The poison can only go through the image. The page should name the image as the poison's target for legs where the prefill's destination and the caller's value are two objects.
4. **Whether a plan is built for entries BELOW the floor.** §5.2 emits one plan per lineage entry and the floor is an index cut at LOAD, so a retired entry's plan is never reachable. Building it is wasted build work; skipping it is unobservable. Unstated; this leg builds all of them, so raising the floor changes nothing but the answer.
5. **A row's VALUES are dump-only, the way §5.9 #28 made the file names dump-only.** `int_widen`'s bracketing `lead`/`trail` are `0xAAAAAAAA`/`0xBBBBBBBB` on the wire while the SCHEMA's defaults are `1` and `2` — I wrote the schema's numbers into the fixture and got a red that looked like a misplaced offset. §5.7 says "a `lead` and a `trail` bracketing the row so a mislaid size moves a neighbour" and nothing more, so a leg asserting values has to read `test/tables/fixedform_dump.cpp` — the C++ the porting rule says not to open. **A small per-row manifest beside the corpus — root table, record count, the values the dump set — would remove the last place a port has to read the reference.**
6. **Which table a corpus file's root is, per row, is derivable but not written down.** §5.9 #9 gives the rule ("the outer table, the one no other fixed table names by value") and I implemented the heuristic; the manifest above would state it instead.
7. **What `report.layoutHash` should do on a leg whose report ALREADY carried a hash member.** This leg's report has `hash`, the record's layout hash, for triaging `no_layout`. §5.9 #29 covers a report with FEWER counters than §5.4 lists, but not a member that already means almost what the new one means. I kept both and split their jobs: `hash` is set once the layout is accepted, `layoutHash` only by the two layout refusals.

## Left RED, named rather than faked

- **§5.8 row 4's remainder**: the ranges, the variant counts and the arm counts are still the READER's, because this leg holds them in the generated straight-line bounds pass over its own declaration rather than in a plan entry. The COUNT half is closed (above).
- **§5.8 row 12**: a forged ordinal remapped to `None` counts `clamped` on the identity path and not on a compiled one.
- **§5.8 row 14**: the remap table is still capped at 255 entries, so a writer's 256th variant and beyond remap to `None` as though the reader did not name them.
- **`plan_too_large` has no fixture on this leg**: the name is wired against the selected lineage entry's own count, but reaching it wants a LOCKED lineage for `test/tables`, which the generated probe libraries do not have. The negative control watches `layout_newer` instead.
- **§5.8 row 10, the digest, is WAITED ON and not matched** (§5.9 #12): every hash goes through `ir.TableFixedLayoutHash`, so this leg's numbers move with the reference's the day the row lands.
- **A red this leg did not write** (§5.9 #31): `go test ./internal/codegen/javatable` fails on this bench with "Unable to locate a Java Runtime" — no JDK on the host, nothing to do with this diff, and it fails identically on `fixed-table-form`.


gafferongames:
Merged on a non-author ONE PATH after the two fixes. The one red, pins, was the job's 10-minute timeout (update-goldens' closing go test ./... outgrew it with four versioning suites); #917 raises pins and go-test to 15 minutes, and the real remedy, dropping the duplicate go test ./... from update-goldens, is carded.

PR #924, opened 2026-09-11
corpus: the dump writes a MANIFEST beside the bytes, so a port never opens it

gafferongames:
Four legs ported from `docs/FIXED-FORM-ALGORITHM.md` §5 — Go (#914), C (#917), Rust (#918) and Dart — each opened `test/tables/fixedform_dump.cpp`, the one file §5.7 forbids, for the same four facts: what values a row's records actually carry (`int_widen`'s `lead`/`trail` are `0xAAAAAAAA`/`0xBBBBBBBB` on the wire while the schema declares 1 and 2), which table is a row's root where a schema declares two, how many records a file has, and the names of the rows that are not a pair. A rule four legs in a row had to break is not the legs' bug — so the emitter says what it wrote.

**The manifest.** `make tables-fixedform-corpus` now writes `build/fixedform-corpus/manifest.txt` beside the bytes, one plain-text line per corpus file, nothing to parse it with:

```
file=<name> row=<row> side=old|new|mid|a|b|none root=<Table> records=<n> values=<field>=<value>[,<field>=<value>...]
```

Three of the sixty-six lines:

```
file=fx1.bin row=fx1 side=none root=FxRoot records=2 values=r0.keep=4242,r0.narrow=40000,r0.renamed=321,r0.gone=654,r0.nested.a=111,r0.nested.b=222,r0.label="fx1",r0.label_length=3,r0.marks[0]=101,r0.marks[1]=202,r0.marks_count=2,r0.blob[0]=222,r0.blob[1]=173,r0.blob[2]=190,r0.blob[3]=239,r0.blob_length=4,r1.keep=1,r1.narrow=2,r1.renamed=3,r1.gone=4,r1.nested.a=5,r1.nested.b=6,r1.label_length=0,r1.marks_count=0,r1.blob_length=0
file=old_int_widen.bin row=int_widen side=old root=IntWiden records=3 values=r0.lead=2863311530,r0.v=-1,r0.trail=3149642683,r1.lead=2863311530,r1.v=-32768,r1.trail=3149642683,r2.lead=2863311530,r2.v=32767,r2.trail=3149642683
file=old_float_widen.bin row=float_widen side=old root=FloatWiden records=1 values=r0.lead=2863311530,r0.v=nan|0x7F8ABCDE,r0.trail=3149642683
```

**It cannot drift.** `row` and `side` come from the FILE NAME, `root` from the record's own TYPE (`typeid`, so a renamed table renames its line), `records` from the vector the dump saved, and every value is stored through one helper that performs the assignment AND records the line from the same expression — `MS` / `MSI` / `MSE` / `MSEI` / `MSTR` / `MSTRI` / `MWCPY` (the `I` pair take the SUBSCRIPT the call site is looping over, so the path holds the index the bytes hold and never the variable's name). A changed value changes both or neither. Records are prefixed `r<i>.`, nested fields dotted, array and keyed slots carry the index the bytes carry (`r0.slots[1].n=101`), text is quoted (`u"…"` wide, `\uXXXX` for anything not printable ASCII), a float is given as digits AND bits because a signalling NaN's payload is the value. `values=` is the LAST field and **runs to the end of the line** — a quoted value may carry spaces and the commas inside one are escaped (`\,`), so split the head on spaces and take the rest whole. `root=` is **not** unique across rows (a lineage pair's two sides share one table name by construction, and FU1's and FU2's is `FuRoot`); `row=` is the per-file name.

**The check.** `manifest_case` in `test/tables/fixedform_main.cpp`: every `.bin` in the corpus has a line, every line is shaped, one line per file and no line without a file, and every `root=` names a table the generator actually emitted (the `<Table>FixedSave` the dump calls, found in the generated trees). `tables-fixedform` takes `build/fixedform-corpus/.stamp` as a prerequisite so the case always has something to hold; it skips with a printed NOTE if the corpus is absent. Nothing under `build/` is committed.

**#895's FU rows fold in.** The tip added `fu1_file()`/`fu2_file()` — the oracle bytes for the two widening rungs — with the assignments written by hand, which is the one way `manifest.txt` could go quiet while the corpus grew. Both writers now assign through the same macros, so `fu1.bin` and `fu2.bin` carry manifest lines with `root=FuRoot` and records 3 and 2. The bytes are unchanged: `fu_oracle_case` reads both files back and asserts the values, and it is green. `tables-fixedform` passes the corpus directory to both binaries (#895) and `manifest_case` takes the same path instead of assuming `build/fixedform-corpus`.

**Documented** in §5.7 step 1 (the format, and that the declared default is not the value on the wire) and §5.9 as ruling **#32**: a port asserts the manifest's values, never reads the dump.

**Gates**

| gate | result |
|---|---|
| `make tables-fixedform-corpus` | exit 0, 66 files + 66 manifest lines, 2.3 s |
| `make tables-fixedform` | exit 0, `manifest: 66 lines, 66 corpus files, 23 distinct roots, all generated`, 13.4 s |
| `make tables-go-versioning` (spot check) | exit 0, 15.8 s, unchanged |

The summary line says `all generated` only when the root check passed, and counts DISTINCT roots, since several rows share one.


gafferongames:
Merged on a non-author ONE PATH. Nits for a follow-up, from the read: the doc lists MS/MSI/MSE/MSTR/MWCPY but MSEI and MSTRI exist too; the narrow quoter emits \xXX where the doc says \uXXXX (latent); the format line should say values= runs to end of line (quoted values carry spaces: a port must not split on whitespace); root= is not unique across rows (fx1/fx2 both FxRoot; row= disambiguates); the summary printf says 'all generated' even in a run that FAILed on a root; manifest_case is a no-op on Windows.

PR #925, opened 2026-09-11
the compiler ships the lock's lineage: one read, every backend that takes it

gafferongames:
#920 found the hole: **nothing in the tree called the entry point §5.9 #1 names.** Every table backend that grew `GenerateLineage` grew it for its own versioning harness, and `compiler/target_*.go` all called the plain `Generate`. So a unit whose committed lock held five layouts generated a reader for ONE of them, on every leg, and no test anywhere said so. The lock was right, the emitters were right, the driver dropped the lineage on the floor.

## Red first

`compiler/fixedlineageship_test.go` builds the fixture the way an operator does — `schema lock`, append a field (§5.1's widening), `schema lock` again — checks the premise against `lockfile.Lineage` (two entries, oldest first, floor 0), then asserts on the bytes `Compiler.Generate` returns: the OLDER layout's wire hash must be in the emitted source beside the current one.

Before the change:

```
--- FAIL: TestFixedLineageEntryPointIsReachedFromTheDriver
    cpp does not take the lineage from the driver (§5.9 #1: the caller makes lockfile.Open/Lineage/Floor …)
    go  does not take the lineage from the driver …
    rust does not take the lineage from the driver …
--- FAIL: TestFixedLineageIsShippedByEveryTargetThatTakesIt
    go:   the emitted source carries no entry for the OLDER layout 0x4842a4b029549be5 — the driver handed
          the backend no lineage, so a build of this unit reads only its own records (§5.9 #1)
    rust: the emitted source carries no entry for the OLDER layout 0x4842a4b029549be5 — …
```

**`cpp` was green on the lineage assertion and red only on the entry point**, because it opens the lock itself — §5.8 row 1's interim. That makes it the control: the one leg that already shipped a real lineage proves the fixture measures what it claims to.

The second test, `TestFixedLineageWithoutALockIsOneEntry`, is the other half of §5.9 #1: an unlocked unit still carries its own layout, and the shared lock read must neither error nor move a byte.

## The one place the lock is read

`compiler/lineage.go` — `lockfile.Open` beside the unit's schema files, then `lockfile.Lineage` and `lockfile.Floor` per fixed table, translating the lock's wire name to the declared name the emitters hold.

- **no lock hands nothing**: a unit that was never locked promises nothing and keeps its lineage of one;
- **a lock that will not PARSE is an error**, because the alternative is exactly this PR's defect — a shorter lineage and no sentence;
- the floor comes from `lockfile.Floor` and is **checked against the entries' own retired marks**, one statement made twice, the way the lock's own lines are.

`lineageGenerator` is §5.9 #1's second entry point as the driver sees it; `Compiler.Generate` calls it where a target has one and plain `Generate` where it does not. `Compiler.Generate`'s doc now says it reads that one file.

## Each target's call, after this PR

| target | table backend call |
|---|---|
| `go` | `gotable.GenerateLineage(u, goTableLineage(u, lineage))` |
| `rust` | `rusttable.GenerateLineage(u, rustTableLineage(u, lineage))` |
| `cpp` | `cpptable.GenerateLineage(u, lineage.Lock())` |
| `c`, `cs`, `dart`, `elixir`, `java`, `js` | `…table.Generate(u)` — **no second entry point on this branch** |

The six on plain `Generate` are not a regression: their backends export no `GenerateLineage` on `fixed-table-form`. #917 (c), #920 (java), #922 (js) and #923 (dart) each grow one on their own branch, and the day one merges its target is three lines — a conversion function and the two-line method — and `fixedLineageShipTargets` in the test gains a name.

## The C++ backend's own lock read (item 3)

**The OPEN moved; the interpretation did not, and here is why not.** `cpptable.Generate` is now `GenerateLineage(u, lock)` and the driver hands it the lock it already opened, so the disk is read once per generate. Its lineage ENTRIES cannot come from `lockfile.LineageEntry` yet, because they carry two facts the lock has no spelling for:

- `collectKnownRanges` reads an older declaration's integer RANGES off a peer schema **unit**, not off the lock — it is what the reference's range-narrowing guards are built from;
- `lineageFloor` falls back to `fixtureFloor`, a test-only map keyed by package, for the `VNEW_floor` fixture that has no lock at all.

Both belong to the sibling-filename convention §5.8 row 1 holds as an interim. Moving them now would mean teaching the shared path the convention, which is the opposite of retiring it. So `FixedLineage.Lock()` exists for exactly one caller and goes away with the convention.

## Gates

- `gofmt -l .` empty.
- `go test ./compiler/ -run 'Lineage|Fixed' ./internal/codegen/...` green.
- `go test ./compiler/ ./internal/lockfile/ ./internal/codegen/cpptable/ ./cmd/...` green (compiler 221s).
- **Regen sweep** — `rm -f bin/schema`, every `generated/**/.stamp` removed and remade: `git status --porcelain generated/` and `git diff --stat generated/` both **empty**. Nothing moved, and the reason is checkable: no unit under `examples/`, `examples128/` or `bench/corpus/` carries a `schema.lock`, so every committed tree's lineage is still the one entry it always was. The committed locks live under `tables/`, `examples-wide/` and `test/table-base64/`, which no committed generated tree is emitted from — the first lock to land beside a generated unit is the first time this sweep will move a file, and it will move it by ADDING older entries.

### DIFF
--- a/compiler/fixedlineageship_test.go
+++ b/compiler/fixedlineageship_test.go
@@ -1,0 +1,203 @@
+// THE COMPILER SHIPS THE LOCK'S LINEAGE (docs/FIXED-FORM-ALGORITHM.md §5.2's
+// COMPILE(lock, T), and §5.9 #1: the lineage reaches a backend AS DATA through
+// a second entry point, and the CALLER — this package — makes the lock's three
+// calls).
+//
+// The Java port (#920) found what this test is here to pin: every backend that
+// grew `GenerateLineage` grew it for its own versioning harness, and NOTHING in
+// the tree called it. `schema generate` called the plain `Generate`, so no real
+// build on any leg shipped a lineage of more than one entry, however long the
+// committed lock's lineage was. A port passing its own harness proved its
+// emitter; it proved nothing about the compiler.
+//
+// So the fixture is a REAL LOCK, written the way an operator writes it — `schema
+// lock`, a widening, `schema lock` again — and the assertion is made on the
+// bytes `Compiler.Generate` returns: the OLDER layout's wire hash is in the
+// emitted source, as static data, beside the current one. That is the only
+// statement that distinguishes a backend the driver hands the lineage to from
+// one it does not.
+package compiler
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/mas-bandwidth/schema/v2/internal/lockfile"
+	"github.com/mas-bandwidth/schema/v2/ir"
+)
+
+// fixedLineageShipTargets are the targets whose table backend takes the
+// lineage as data today: `go` and `rust` through `GenerateLineage`, and `cpp`
+// through the lock the driver now opens for it. Every other built-in target's
+// table backend has no second entry point yet (c, cs, dart, elixir, java, js);
+// their `GenerateLineage` lives on the open port branches, and the day one
+// lands its target joins this list and nothing else changes.
+var fixedLineageShipTargets = []string{"cpp", "go", "rust"}
+
+// TestFixedLineageEntryPointIsReachedFromTheDriver: the driver must HAVE the
+// second entry point for those targets. Before #921 no caller in the tree did —
+// every target answered only `Generate`, so the lock's lineage reached no
+// backend from any real build.
+func TestFixedLineageEntryPointIsReachedFromTheDriver(t *testing.T) {
+	c := New()
+	for _, target := range fixedLineageShipTargets {
+		g, ok := c.gens[target]
+		if !ok {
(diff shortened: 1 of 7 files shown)


gafferongames:
Merged on a non-author ONE PATH. For each remaining leg's merge (c, cs, dart, elixir, java, js), five mechanical edits in compiler/target_<lang>.go: rename Generate to GenerateLineage(u, opts, lineage), add the two-line Generate shim calling it with nil, change the one xtable.Generate(u) call to xtable.GenerateLineage(u, xTableLineage(u, lineage)), add the xTableLineage helper (a copy of goTableLineage with the leg's entry type), and add the target to fixedLineageShipTargets in compiler/fixedlineageship_test.go. A corrupt lock is loud on cpp/go/rust today and on each leg as it lands.

Issue #35, opened 2026-08-14
Estate known-issues register, 2026-08-15


rowan-claude:
The verified known-issues register for the estate — mas-bandwidth/{schema, serialize, serialize.c, serialize.go, serialize.rs, serialize.cs} — harvested from the estate's record by five parallel readers, deduped and verified against the actual code, extended by a completeness critic, pruned by an adversarial refuter (8 refuted items dropped before filing), and re-spot-checked against fresh origin state at filing time. 37 issues filed 2026-08-15.

Issues titled **DECISION:** are forks for the owner to settle — they state options and per-port consequences and deliberately do not recommend. Everything else is ours-to-fix. Cross-repo items were filed on the repo where the fix lands: serialize-family spec/format items on serialize (STANDARD.md's home), estate-wide process/tooling items here on schema.

**Nine further items were withheld from public filing** as internal/operational (credentials, auth expiry, operational bench details, a named external correspondent) and travel privately. This index covers only the public register.

---

## Critical — compressed_float wire divergence, released artifacts affected

serialize.go v1.8.0 and serialize.c v1.0.0 are published and carry the divergence; no released version of any port yet implements the corrected arithmetic.

- mas-bandwidth/serialize.go#25 — wire: main still fuses compressed_float into an FMA on arm64; the family's last live writer divergence (fix exists as open PR mas-bandwidth/serialize.go#24).
- mas-bandwidth/serialize.go#26 — wire: v1.8.0 is a published, immutable release carrying the arm64 FMA divergence; needs v1.8.1 plus a `retract` directive.
- mas-bandwidth/serialize.c#3 — wire: released v1.0.0 quantizes compressed_float in double; needs v1.0.1 and should not stay Latest.
- mas-bandwidth/serialize#47 — wire: the compressed_float READER is unspecified and left contractible in C++, C and Go; the repo's own wire-proving binaries build with -ffast-math.

## High

- mas-bandwidth/serialize#53 — wire: vendored STANDARD.md has inverted; serialize.go and serialize.rs main carry the superseded 392-line spec (fixes: PRs go#24, rs#29, plus a serialize.cs re-run).
- mas-bandwidth/serialize#54 — wire: degenerate fixed range; serialize.go and serialize.cs disagree with themselves across the 64/128 storage boundary (blocked on #49 for direction).
- mas-bandwidth/serialize.cs#4 — correctness: the only port that never explicitly pins the compressed_float intermediate (latent, ECMA-334 allows excess precision).
[...]

Issue #34, opened 2026-08-14
tooling: Dependabot and Actions pinning are split down the middle of the family, and the flagship is on the older side


rowan-claude:
Severity: **low** · Kind: tooling · Call: ours-to-fix · Affected repos: all six (dependabot.yml missing in serialize, serialize.c, serialize.cs, schema; checkout pinned at two generations across the family)

Filed on schema as the estate hub; the sweep touches every repo.

## What is wrong

Dependabot and GitHub Actions pinning are split down the middle of the family: `.github/dependabot.yml` exists in serialize.go and serialize.rs only, and the six repos run two generations of actions/checkout (serialize, serialize.c, schema on v4; serialize.go, serialize.rs, serialize.cs on v7) for no stated reason — with the flagship C++ repo on the older side.

## How it manifests

The four repos with no dependabot.yml get no action-version updates at all. Separately: serialize now carries a Go module (tools/conformance/go.mod) — a new dependency-capable surface in the C++ repo. The family's loudest stated value is zero third-party dependencies (repeated verbatim in serialize.cs's csproj comments and serialize.rs's empty `[dependencies]`), and the conformance tool honours it in its own header ("Standard library only -- no dependencies, ever") — so the right guard for that module is a check that keeps the dependency list empty, not just a version bumper.

## Evidence

> VERIFIED: `.github/dependabot.yml` exists in serialize.go and serialize.rs only (both `package-ecosystem: github-actions`, weekly); serialize, serialize.c, serialize.cs and schema have none. Action pinning: `grep -rhoE 'actions/checkout@v[0-9]+'` gives serialize=v4, serialize.c=v4, schema=v4, serialize.go=v7, serialize.rs=v7, serialize.cs=v7. serialize also now carries a Go module — tools/conformance/go.mod, `go 1.21`, standard-library-only — which is a new dependency-capable surface in the C++ repo and produced run 31841069820 'Graph Update: go_modules in /tools/conformance'. pipeline/QUESTIONS.md line 206 boards 'Enable Dependabot alerts on yojimbo/serialize/matchmaker/next? (Currently disabled)' as an unanswered board decision from 2026-08-07.

## Fix direction

Copy serialize.go's four-line dependabot.yml into the other four repos, bring every checkout to one version in a single commit, and answer the boarded Dependabot-alerts question while the family is being touched anyway.

Not blocked on any DECISION issue.

---
Part of the 2026-08-15 estate known-issues register (index issue on mas-bandwidth/schema).

[...]

Issue #19, opened 2026-08-14
DECISION: unsigned fixed-point spelling — derive signedness from bounds, or add ufixed(I, F)?


rowan-claude:
Severity: **low** · Kind: spec-decision · Call: **glenn-decides** — this issue states the fork and its consequences; it deliberately does not recommend.

## What needs deciding

The unsigned fixed-point spelling: derive signedness from the bounds, or add an explicit `ufixed(I, F)` type? Long-parked as SPEC.md §9 item 17 (OPEN with Glenn, 2026-08-06); it will keep resurfacing until answered, and it is cheap to answer. v1 ships `fixed(I, F)` signed only.

## Evidence

> SPEC.md:2107-2117 (§9 item 17), marked 'OPEN with Glenn (2026-08-06)', cross-referenced from the type table at SPEC.md:974 ('the unsigned spelling is OPEN, §9 q17') and from SPEC.md:1515. v1 ships fixed(I, F) signed only.

## The options and their consequences

**Option A — derive signedness from the bounds.** min >= 0 means unsigned storage; no new syntax. Consequence: a field's storage TYPE changes when a bound crosses zero.

**Option B — explicit `ufixed(I, F)`.** One more keyword; storage always manifest in the type name; matches the int/uint precedent the integer family already set.

SPEC records that the wire is range-derived either way, so nothing on the wire moves; signedness only decides which storage widths can host a given non-negative whole-unit domain — signed Q8.8 tops out at 127 units where unsigned reaches 255.

---
Part of the 2026-08-15 estate known-issues register (index issue on mas-bandwidth/schema).

rowan-claude:
Status: awaiting Glenn, boarded. Five of the register's eight DECISION forks were settled this morning (degenerate fixed, trusted write path, rounding, string, wstring); the unsigned fixed-point spelling was not among them. It stays open and is boarded for Glenn's next sitting.

rowan-claude:
[...]

PR #926, opened 2026-09-11
§5: the fourth through seventh legs' findings folded in as rulings #33 to #45

gafferongames:
Four more legs were ported from `docs/FIXED-FORM-ALGORITHM.md` §5 alone — Java (#920), C# (#921), JavaScript (#922), Dart (#923) — and each ended with a "what §5 did not tell me" section. **Every item is folded into §5 as a RULE**, checked against `docs/FIXED-FORM-BILL-READS-BACKWARD.md`, the merged C++ (`internal/codegen/cpptable`, unopened except to read), and all seven ports (the four branches, `gotable`, `ctable` on `rowan/c-reads-backward`, `rusttable`), so the rule written is the one they now share. **Docs only**, one file.

§5.9 gains **#33 to #45**, after **#32** (the corpus manifest, #924 — the block's opening paragraph says so, which is why there is no gap to explain later).

| # | the ruling | where it landed |
|---|---|---|
| **33** | **The flat-element FOLD and a widening are incompatible.** A leg folds a run only where the writer's and the reader's element widths are EQUAL; across a widen EACH ELEMENT IS ITS OWN WIDEN ENTRY, or a folded run takes an entry that widens element by element into the span its setter takes (C#'s `FlatWiden`, sign `0`/`1`/`2`). `widened` counts per ENTRY (§5.4), so a folded run is **1** and an unfolded one is `min(their_n, my_n)` — **4** on `array_elem_widen`. **The C++ reference asserts `r.widened > 0`** (`test/tables/versioning_numbers.cpp:345`), never an exact count, so **NONZERO is the number every leg asserts**; a leg pinning `4` reds the next leg that folds | §5.9 #33; pointers at §4.2's coalescing bullet, §5.2's widen paragraph, §5.4's `widened` row |
| **34** | **The static-data struct's name on the page is `TableFixedKnownLayout`** with #19's members. Two divergences admitted and named: Go's and Rust's `TableFixedKnown` is a **leg-local spelling**, allowed where no twin gate binds the leg (the gate holds the C/C++ pair member for member), and the page's name is the contract; **a managed leg carries THREE members** where the array carries its own length (C# `fixedruntime.go:141`), the same order, `layout_bytes` dropped — Java, JS and Dart kept four. The hash's place and the record size's source (the LOCK) move for nobody | §5.9 #34; pointer at §5.3 step 8 |
| **35** | **The `0x5A` poison has three conforming targets**, this being the third: a byte view; the PLAN'S RECORD IMAGE where the prefill's destination and the caller's value are two objects; or **the unit's own value surface, field by field**, where there is neither. A leg says which it laid, and proves it bites | §5.9 #35; pointers at §5.2's poison paragraph and §5.7's |
| **36** | **A lineage entry that will not build is a LANE, never a throw.** The parse failure fails the build at GENERATE naming table and hash (#8's primary answer, sharpened); a per-leg static-init failure — `ExceptionInInitializerError`, `TypeInitializationException`, Dart's lazy `final` re-evaluated, a JS module-init throw — is **unreachable by construction**: every entry carries its own refusal reason (`layout_malformed`, `plan_too_large`), reported only if a file selects it | §5.9 #36; pointers in §5.2's PLAN and a new §5.3 refusal row |
| **37** | **The NEW-READS-OLD value oracle is the corpus MANIFEST** (#32), never the leg's own older build — that is a good cross-check and, as an oracle, the leg agreeing with itself. The manifest also settles the record count, the row's root (#9) and the non-pair names (#28) | §5.9 #37; pointers at §5.7's two columns and step 1 |
| **38** | **"Arm by arm" in the prefill is an INSTRUCTION**: in one flat image the arms OVERWRITE each other in DECLARED ORDER at their overlay storage, and the tag lands `None` last — four lines. The only thing the overlay cannot hold is a default under a byte two arms both claim, and that is the case wanting per-arm images or a guarded prefill entry | §5.9 #38; pointer in §5.2's prefill paragraph |
| **39** | **A negative control pinned to a case §5.6 retires moves with it, and the PR names which moved** — a control that survives by reaching its case another way survived by luck. **Every versioning gate owes its own control**: the wrong refusal name on a file outside the lineage reds every OLD-REFUSES-NEW row and nothing else | §5.9 #39; pointers at §5.7 steps 5 and 6 |
| **40** | **The present companion is the one newer-only field that is not a default**: it lands `1`, the plan's `present` constant under the row's own guard and ordinal, not the fresh value's `false` | §5.9 #40; pointer at §5.7's two columns |
| **41** | **A `was =` rename is paired by WIRE ID in any harness, never by name** — §5.1's own sentence read from the test's side; the manifest gives it for free | §5.9 #41; pointer at §5.7's two columns |
| **42** | **The split is a property of the plan's ORDER that a leg whose loop tests the guard per entry still owes**: it is what makes coalescing across the split illegal (a merged guarded/unguarded run lands an arm's bytes unconditionally), and what gates the pool | §5.9 #42; pointer in §5.2's PLAN |
| **43** | **The census counts once, in MATCH, and never again in the second pass** — "pass 2 counts nothing" governs the census LOOP as well as the entries, or every `unknown` doubles. A dead report for pass 2, or the census in pass 1 only | §5.9 #43; pointers at PLAN's pass 2 and MATCH's census line |
| **44** | **A row a leg cannot generate at all is UNPORTED, not red** — JS's `wstring_grow`, kind `33` refused in a JS table: ⚪ both columns with the blocking refusal cited, never folded into a green. A leg's row table carries three marks | §5.9 #44; a new paragraph beside §5.7's "what to leave RED" |
| **45** | **`plan_too_large` is owed by every leg**, including one whose caller-owned plan object also owns the image and the hole list: the ENTRIES are the build's, **whatever an `ordinal` entry's `aux` indexes moves into static data with them**, the image and the conversion lanes stay the caller's — and a compiler with one failure code for "my own layout did not parse" and "the pool overflowed" owes two names | §5.9 #45; pointers in §5.2's PLAN and §5.3's refusal table |

**Also**: §5.9's heading and opening paragraph now say where #32 to #45 come from, so the block reads as the fourth-through-seventh legs' the way #13 to #31 reads as the C and Rust legs'.

**One thing a reader should check.** PR #921's BODY is a copy of #920's (the C# rows and gates are edited in, the findings and the opening are Java's). The C# findings were taken from that branch's **commit messages**, which carry them in full — `157268a6` (the fold and the widening, and `widened` per entry), `fdb4765e` (one assembly per column where namespaces give two generations a spelling; the value-surface comparison as #17's third form), `2422b092` and `98d03b13` (the retirements and the controls) — and from `internal/codegen/cstable/fixedruntime.go` on the branch. The PR body is worth re-pasting.

**Not done here**: no leg code moves, nothing is merged, and #32's own text lands with #924 — this branch starts at #33 deliberately so the two do not collide.

### DIFF
--- a/docs/FIXED-FORM-ALGORITHM.md
+++ b/docs/FIXED-FORM-ALGORITHM.md
@@ -269,7 +269,9 @@ compile-time evaluation carries this form. Two shapes are permitted, both the co
 
 - **coalesced runs straight into the reader's storage.** Merge two entries only when both are `copy`,
   `guard`/`arg`/`meta` agree, and `src` and `dst` both advance by `size`. An array of a FLAT element type — one
-  whose storage image is its wire image — is ONE run however many elements it holds.
+  whose storage image is its wire image — is ONE run however many elements it holds. **The fold holds only where
+  the writer's element width and the reader's are EQUAL: across a widening there is no fold, and each element is
+  its own widen entry** (§5.9 #33).
 - **one whole-body copy into a RECORD IMAGE, then a straight-line scatter.** In the image domain the record's
   declared order IS the destination's order, so every run merges and the plan is one entry — the shape a language
   takes when its storage is not the wire's: a `string(N)` stored as `N+1` units, a union that is a real tagged enum.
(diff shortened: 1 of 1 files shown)


gafferongames:
Merged. One tightening for the next §5 pass: since #33 forbids a fold across a widen, array_elem_widen has no folded shape on any leg, so its widened count is exactly min(their_n, my_n) = 4 everywhere; the C# fixture asserts 4 (its FlatWiden counts per element) and the reference fixture may too.

PR #927, opened 2026-09-11
go fixed form: the known-layout struct is TableFixedKnownLayout

gafferongames:
The Go leg named its static known-layout struct `TableFixedKnown` where docs/FIXED-FORM-ALGORITHM.md §5.9 #19, the C++ reference and the C# leg all spell it `TableFixedKnownLayout`, so the type, its registration in internal/tablenames/go.go and the committed Go trees now carry the page's name (the per-table array `<T>FixedKnown` is unchanged — it is identical on every leg). internal/tablenames/rust.go picks up its own `TableFixedKnown` registration, which had been riding on the Go entry until this move left the Rust claim scan with an unregistered name.


gafferongames:
Merged (a rename, gates green, the Rust registration it exposed). Carded: internal/tablenames is a union registry, so cpp.go and cs.go never registered TableFixedKnownLayout and ride on dart.go's entry; each backend registers what it spells.

Issue #82, opened 2026-08-20
Additive precomputed compressed-float entry points across the family, in support of the generated path


rowan-claude:
Glenn's ruling, 2026-08-20, three beats verbatim on #28: 'Runtime side-entry point could be something that supports schema.' / 'But in addition to, not instead of, the current function.' / 'It's not either/or, it's YES to both.'

THE WORK: each runtime gains an ADDITIVE precomputed-params entry point for compressed floats — constants (step count, wire width, delta, min) computed at schema generation time and passed in — designed for the generated path to call. SerializeCompressedFloat32 stays untouched everywhere. Reference-first landing order: serialize (C++), then c, cs, go, rs mirrors. Then per-backend, the emitter adopts whichever mechanism measures best (the #79 inline fold vs the Pre call is a measurement per backend, not a transplant — gc's inline budget decides Go; the JIT verdict question decides C#; C/C++/Rust already fold the existing call to zero hot calls, so adoption there is about one-audited-home, not speed). Wire bytes identical by construction, gated by the #79 differential pattern per backend.

WHY: one audited home per runtime for the quantization arithmetic (the FMA-rounding discipline currently lives in both the Go emitter fold and the runtime), and a supported fast path for the generated code across the whole family.

Kind: family API + performance · Call: ruled by Glenn, YES to both halves · Order: after the measured levers in today's burn (#77, #62, #63, #65/#80)

rowan-claude:
The reference leg is merged: mas-bandwidth/serialize#83, on the flagship's main at 89916be1. The C++ serialize now carries the additive precomputed entry points (serialize_compressed_float_params as the reference derivation callable from tooling, serialize_compressed_float_precomputed_internal, and the house macro), with the old entry point's signature untouched and its body split statement-for-statement onto the one audited home. The identity proof is permanent rather than one-shot: a frozen verbatim copy of the pre-split body lives in the test section and a 3.5-million-check differential holds the composition to it, over 18 declaration shapes with the read side exhaustive through 16-bit widths. Three falsification directions each went red and were restored.

The adoption spec for every emitter, per field, four literals: max_integer_value = ceil(float32(float32(max-min)/res)) clamped to [1, 4294967040]; bits = bits_required(0, max_integer_value); delta = float32(max)-float32(min); and min. This matches ir.CompressedFloatParams and #79's fold constants, and the differential pins the C++ derivation against the schema table's shapes.

[...]

PR #933, opened 2026-09-11
bench/paired: the C table leg links serialize.c like the packet leg

gafferongames:
## The gap

`buildLegs` in `bench/paired/main.go` built the C **table** leg with `-I$SERIALIZE_C` and nothing else:

```
cc ... -DBENCH_MATCHED -Igenerated/bench/paired/c -I$SERIALIZE_C bench/tables/c/table_main.c -o build/paired/table-c -lm
```

while the C **packet** leg three lines below it appends `filepath.Join(serC, "serialize.c")` as a translation unit. serialize.c is a translation unit, not a header-only library like the C++ one. `table_main.c`'s form-1 arm calls into it, so the link failed:

```
  "_serialize_int128_compare", referenced from: _bench_mixed_load_body ...
  "_serialize_int128_make",    referenced from: _bench_mixed_reset, _bench_mixed_save, _fixed_load_all ...
  "_serialize_uint128_make",   referenced from: _bench_mixed_reset, _bench_mixed_save ...
ld: symbol(s) not found for architecture arm64
```

## Why it cost the whole table, not one row

The driver builds the entire sitting in ONE invocation, because `build/paired/build.json` is written whole and replaces what was there, and `-mode table` refuses a leg the manifest does not carry. So a C link error did not drop the C row — it left **no leg measured at all**. The nine-language table came back empty for a one-word omission in one leg's argument list.

## The fix

Link it the same way the packet leg does, and keep the two cases parallel. The bench measures form 1 beside form 3, so the library's objects are not optional for the table leg either — a comment at the call site says so, since the obvious-looking reading is that a table leg needs only headers.

## Gates

- `bin/schema-paired -mode table-langs` → `cpp,c,go,rust,cs,js,java,elixir` (`not measured on this tree: dart`)
- `go build ./bench/...` clean
- `go vet ./bench/...` clean
- `gofmt -l .` empty
- `bench/paired/nine.sh --rounds 1` on the Studio: `table-c` compiles and links; the sitting proceeds past the C leg.

No results are committed: the sitting's `NINE.md` / `nine.json` stay out of the ledger.

## Noted, not fixed here

The C **packet** leg then fails separately on this bench against `/Users/glenn/rowan-working/serialize.c`, which sits at **v1.2.0** while `origin/main` is **v1.11.0**. The tree's C emitter adopted the runtime's precomputed compressed-float entry points (`b53c1b5c`, issue #82), and the v1.2.0 header has no `serialize_write_compressed_float_precomputed` / `serialize_read_compressed_float_precomputed`. That is a stale bench checkout, not a defect in this tree — no local serialize.c working copy carries the symbols, and `origin/main` does. The numbers below were taken against a scratchpad clone of serialize.c at `e67dc6a` (1.11.0).


gafferongames:
## The fix is verified; three further blockers stand between it and a nine-language table

**This PR's fix is green.** `table-c` compiles, links, and gates against this sitting's corpus:

```
cc -std=c99 ... -DBENCH_MATCHED -Igenerated/bench/paired/c -I$SERIALIZE_C \
   bench/tables/c/table_main.c $SERIALIZE_C/serialize.c -o build/paired/table-c -lm
schema tables bench (c, Release)
OK (corpus_id 44c211c33e1b1939)
```

`go build ./bench/...`, `go vet ./bench/...` and `gofmt -l .` are clean. `-mode table-langs` answers `cpp,c,go,rust,cs,js,java,elixir` / `not measured on this tree: dart`.

No `NINE.md` was produced, because the sitting then stops on three things that are **not** in this diff. Recording them here so the next run does not rediscover them:

### 1. `/Users/glenn/rowan-working/serialize.c` is at v1.2.0; the tree's C emitter needs v1.11.0

The C emitter adopted the runtime's precomputed compressed-float entry points (`b53c1b5c`, issue #82). The v1.2.0 header has neither:

```
generated/bench/c/BenchWire.h:1273:11: error: call to undeclared function
  'serialize_write_compressed_float_precomputed'
```

No local serialize.c working copy carries the symbols (`serialize.c`, `serialize.c-ro`, `serialize.c-work` are all at or behind `cecaa04`); `origin/main` (`e67dc6a`, 1.11.0) does. The bench checkout wants a pull. Numbers here were taken against a scratchpad clone at `e67dc6a`.

### 2. The Elixir emitter emits code `mix format --check-formatted` rejects

`generated/bench/paired/elixir/WrapFixed.ex`, two hunks — a trailing `#` comment the formatter lifts above its statement, and a missing blank line before a `defp`:

```
229 229  |  # the loop, the same pass for either plan (docs/SPEC-TABLES.md 3.4).
230 230  |  defp fixed_table_fixed_records(stated, layout, records, report, opts) do
231     -|    hash = stated # the header's hash; digest is not on the wire
    231 +|    # the header's hash; digest is not on the wire
    232 +|    hash = stated
...
298 299  |  end
    300 +|
299 301  |  defp fixed_table_fixed_header_names_it(_stated, _layout), do: :ok
```

The emitter is meant to emit the formatter's own shape rather than be formatted afterwards, so this is a defect in the emitter, and `nine.sh` gates it one language at a time — so it halts the whole sitting.

### 3. The rust leg dirties a TRACKED generated file, so every build records `dirty: true`

`generated/bench/paired/rust/Cargo.toml` is committed with a relative dependency path and rewritten in place with the absolute `$SERIALIZE_RS`:

```diff
-serialize = { package = "serialize-official", path = "../../../../../serialize.rs" }
+serialize = { package = "serialize-official", path = "/Users/glenn/rowan-working/serialize.rs" }
```

`build.json` snapshots `Dirty` after generation (`main.go:805`), and both `-mode fast` and `-mode table` refuse a dirty checkpoint (`fast.go:286`). So **no clone can ever measure**, whatever it fixes.

This is invisible on the keeper's canonical layout, where the checkout sits five levels above `serialize.rs` and the committed relative path already resolves — the file only changes for a checkout somewhere else. It is the reason a scratchpad clone cannot produce the table even with legs 1 and 2 fixed: `nine.sh` checks the tree is clean *before* the build, and the build dirties it afterwards. Either that Cargo.toml should not be tracked, or the rust leg should pass the path by `--config`/`-p` rather than rewriting it.

PR #936, opened 2026-09-11
bench/tables/dart: the Dart leg is measured

gafferongames:
`bin/schema-paired -mode table-langs` printed eight names and `not measured on this tree: dart`, because `bench/tables/dart/` did not exist. The Dart fixed-form emitter had landed and its bench leg had not, so the Dart row could never appear in a form-3 (`bench_fixed`) round. This writes the leg and registers it.

## The leg

`bench/tables/dart/table_main.dart`, a port of the C++ reference and closest in shape to `bench/tables/js/table_main.mjs`:

- **Form 3 and only form 3.** Dart has no form-1 table wire (schema#514), so there is no `bench_table` row and none is faked.
- **The tolerant corpus is still read** — all five files, so the corpus id (§1.6) matches the one the paired driver folds — and what is checkable without a decoder is checked: the length index is 64 little-endian uint32 lengths, they sum to the concatenation, variant 0 is the pinned instance byte for byte.
- **Four gates before the clock**, the C++ leg's own, layout first.
- **1 warmup + 7 measured runs**, `--round K` drops that to one and one; the iteration count is the standard 400,000 and uniform (§2.1); the clock is a monotonic `Stopwatch` read in microseconds, the same instrument `bench/dart/main.dart` times the packet wire with; the sink observes a written byte at a rotating offset, because Dart has no empty-asm memory clobber.
- **Shape-blind**, with the exact count on `bench/tables/dart/SHAPE-GATE.allow`; `make shape-gate` is clean.

## The one way it differs from the js leg

**The timed form is the AOT executable.** `dart compile exe` compiles the generated libraries into one binary with the runner, so the leg has a build product for the driver to hash and needs none of the interpreted legs' manifest machinery, and its `linkage` column says `aot` — the spelling `bench/dart/main.dart` already uses of the packet wire, and the Dart answer to the Rust leg's `crate`.

## The registration

`dart` moves from `pendingLanguages` to `tableOnlyLanguages`: its packet leg is the type board's runner, with neither `--gate` nor `--iterations`, so it rides the table wire alone and enters no ratio, no confirmation pass and no board. `pendingLanguages` is now empty and the machinery around it stays — the next language's runner will land before its wiring does exactly as this one's did, and `absentReason`'s "no leg in this driver yet" branch is the honest answer for it then.

The build step holds the generated unit to the two gates `make/dart.mk` holds it to everywhere else: `dart analyze` over the whole unit, and `dart format --set-exit-if-changed` scoped to the `*Fixed.dart` libraries exactly as `tables-dart-fixed-form` scopes it — the packet backend's declaration headers are not format-canonical today, which is a separate emitter item and not this leg's to gate on.

## Evidence

`go build ./bench/... && go vet ./bench/...` clean, `gofmt -l .` empty, `go test ./bench/paired/` ok, `make shape-gate` clean, `dart analyze` clean on the runner and the generated unit.

    $ bin/schema-paired -mode table-langs
    cpp,c,go,rust,cs,js,java,dart,elixir

Nothing is named absent any more. `-mode build -langs dart` generates, analyzes, format-checks, `dart compile exe`s and gates the leg green.

One measured round at the standard count:

    bench_fixed        write            1.374 M msg/s     1656.4 MB/s   (min 1.374, max 1.374, spread 0.0%)
    bench_fixed        round_trip       0.846 M msg/s     1019.5 MB/s   (min 0.846, max 0.846, spread 0.0%)
    bench_fixed        read             2.199 M msg/s   (DERIVED: round-trip minus write, informational — not a measured row)
    dart,bench_fixed,write,400000,1264.296875,1,1373777,1373777,1373777,1656.40,0.00,44c211c33e1b1939,table,aot,contract,default,unknown
    dart,bench_fixed,round_trip,400000,1264.296875,1,845539,845539,845539,1019.49,0.00,44c211c33e1b1939,table,aot,contract,default,unknown

Both measured runs clear §2.1's 200 ms floor at the standard count (0.29 s and 0.47 s), so the driver's escalation never has to raise it for this leg. These are a bud-bench diagnostic on a loud machine, not a sitting.

## Where the standard was silent

`BENCH-STANDARD.md` §2.1 names one warmup per (bench, path) and its 200 ms floor, and §2.9's `--quick` note says a shortened warmup "would measure the interpreter and be wrong" — but it states no warmup floor for a JIT or AOT-with-startup leg, and no `linkage` spelling for a language whose generated code is compiled into the runner's own binary by its SDK rather than by a named compiler and flags. This leg took the full per-path iteration count as its warmup (§2.1 as written) and `aot`, the spelling the Dart packet leg already uses. Both are stated in the runner's header so a later ruling has something to correct.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/bench-tables-dart` is kept at `865d7bfeb87bb82349faccf2f1a41be86b7cf625` (bundle `pr-936-865d7bfe.bundle` in Rowan's retained store), and the reads it earned (emma) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/bench-tables-dart@865d7bfeb87bb82349faccf2f1a41be86b7cf625` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #939, opened 2026-09-11
dart leg: the lineage reaches the backend, the guard at its width, None first

gafferongames:
A cold read of the DART leg (`internal/codegen/darttable`) against
docs/FIXED-FORM-ALGORITHM.md §5. Seven items were found; four are fixed here
and three are named below, not done.

## Done

**1. THE DRIVER NEVER HANDED DART THE LINEAGE.** `compiler/target_dart.go`
called `darttable.Generate(u)` and no `GenerateLineage` existed on
`dartTarget`, so every real `schema generate --lang dart` shipped ONE known
entry and a floor of 0 — all of §5 was inert on this leg however long the
committed lock's lineage was. `GenerateLineage` is now the real entry,
`Generate` a two-line shim, and `dartTableLineage` translates the lock's
facts, the record size passed through and never recomputed.
`compiler/fixedlineageship_test.go` names `dart` in
`fixedLineageShipTargets`, spells a Dart hex constant the way `fixedDartHex`
does (no zero padding below the smi boundary), and greps the emitted
known-layout line for its record argument. Starved once, all three rows bite
by name.

**2. THE GUARD'S WIDTH (§5.2).** The read loop compared a union tag at ONE
BYTE, so a foreign two-byte tag of `0x0101` — an ordinal no arm of this build
names — fired arm 1 and ran that arm's entries over a stranger's record.
`argw` is now a lane of its own (`meta` already carries a widen's width and
sign and a text entry's flavour), clamped `1..8` with zero reading as one; the
union case stamps the writer's tag width and restores the enclosing one; the
coalescer will not join two guards of different widths; and the IDENTITY plan
carries the same lane, because this build's own records were read at one byte
too.

**5 and 6. THE ORDINAL IS UNSIGNED, AND IT COUNTS (§5.4, §5.9 #8).** A Dart
`int` is 64-bit signed, so an eight-byte ordinal of `0x8000000000000000` came
out of the load loop NEGATIVE, passed a signed `raw <= table[0]` and indexed
the remap negatively: a `RangeError` thrown out of the read loop at hostile
bytes, which the forgery oracle refuses. Any raw the writer could not have
written — past its variant count, the negative case included — is now None AND
COUNTED, so a forged ordinal clamps on the compiled plan as it already did on
the identity one. The note on the identity path that claimed its `> n` test
covered both plans is gone: on a compiled plan the `ordinal` op has already
landed the forged value as None before that line runs, so `> n` cannot trip.

**4. THE PREFILL RESETS A UNION ARM BY ARM (§5.2, §5.9 #38),** at each arm's
own overlay storage and in declared order, then the tag to None. It returned
at `*ir.Union` before. A whole-value reset leaves an APPENDED field inside
another arm holding whatever the image held last — and the image is reused
record to record.

## Not done, by name

**3. THE UNGUARDED `const 0` None ENTRY, and the TWO-PASS SPLIT** (§5.2 EMIT
kind 15, §5.2 PLAN, §5.9 #42). The compiled union still emits no unguarded
`const 0` of `my_tag` bytes first, so a record whose tag matches no arm
reports the PREVIOUS record's arm out of the reused image. It needs the split
— unguarded entries first, `split`, guarded second, coalesce never across —
which `internal/codegen/jstable` already has as `TableFixedKeeping`/`pass` and
this leg does not. Not started.

**7. THE VERSIONING HARNESS.** `build/fixedform-corpus/manifest.txt` is not yet
the value oracle for every NEW-READS-OLD row;
`internal/codegen/darttable/fixedversioning_test.go` still asserts values on 4
of 25. The `present == 1` row (§5.9 #40), `widened == exactly 4` on
`array_elem_widen` (§5.9 #33) and the 0x5A poison on `array_fixed_grow` and
`keyed_array_enum_append` (§5.9 #35) are all still owed. Not started.

**As time allowed, neither was reached:** the unknown census re-entered per
element rather than once per field per peer, and the writer's record-size bound
in `compile` refusing `layout_record_too_large`.

## State of the gates

All of these were RUN on this head, after merging `origin/fixed-table-form`
(#884, #928, #932, #934) — the ship list in `compiler/fixedlineageship_test.go`
resolved as the union WITH dart. The "unverified draft" warning that stood here
is withdrawn.

- `make build/fixedform-corpus/.stamp` with `SCHEMA_REQUIRE_CORPUS=1`, after
  `rm -f bin/schema` — GREEN, 4s.
- `make tables-dart-fixed-form` — GREEN, 11s. `dart analyze` over
  `build/dart-fixed` and `test/dart-tables/fixedform.dart` is inside it: no
  issues. `dart format --set-exit-if-changed` over the generated tree is too,
  and it caught one drift in the emitted Dart, fixed here.
- `make tables-dart-fixed-form-negative-control` — GREEN, 2s.
- `make tables-dart-versioning` — GREEN, 10s.
- `make tables-dart-versioning-negative-control` — GREEN, 2s.
- `go test ./internal/codegen/darttable/ ./compiler/ -run 'Fixed|Lineage|Dart'`
  — GREEN, 6s.
- `gofmt -l .` — empty.
- GOLDENS MOVED: NONE. The Dart leg commits no table goldens — there is no
  `update-goldens-dart` and nothing under `testdata/golden` is Dart fixed-form
  output — so the ninth lane and the changed prefill image move no pinned file.
  The leg's own gate is `dart analyze` plus `dart format` over the freshly
  generated tree, and both ran.

## The review items, after this head

- **(a) The unguarded None tag, and the two-pass split — DONE, red then green.**
  A compiled plan wrote every union tag value under a GUARD, so a record whose
  tag named no arm landed no tag at all and the decode read the PREVIOUS
  record's. The fix is the C++ reference's, which already had it: the tag's
  None goes down FIRST and UNGUARDED (§5.2 EMIT kind 15), which needs the plan
  partitioned unguarded-first — so the two-pass split comes over from JS's
  `fixedruntime.go`: `keeping` tested at the push, `plan.split` recorded, each
  half coalesced on its own and never across, and the guarded walk reporting
  into a QUIET report because the census is once per peer. The enum tests its
  half before `layRemap` spends the pool. The None entry carries the writer's
  arm count in `meta` and the read loop counts one clamp for a tag past that
  set, read UNSIGNED (§5.4 asks for the counter on both plans).
  `TestFixedCompiledPlanTagPastArmSet` is the behavioural probe: two records of
  the `union_append` row, the second the first with its tag forged to 9 past
  the writer's two arms. Without the fix it reports arm 1 — the previous
  record's — and moves no counter; with it, None and `clamped == 1`.
- **(b) The other forged-bytes probes — NOT DONE.** The 2-byte tag forged
  `0x0101` and the 8-byte ordinal `0x8000000000000000` still have only
  source-grep tests. The harness to write them is now in place and proved:
  `TestFixedCompiledPlanTagPastArmSet` forges its own bytes in the probe from
  the corpus file's own header, which is the shape those two want.
- **(c) The manifest as the value oracle — NOT DONE.** `versionRows`' `check`
  column still carries hand-written assertions on 4 rows of 25.
- **(d) The unknown census and the writer's record-size bound — NOT DONE.**


gafferongames:
**Full `go test ./...` read, after the deadline push.** `internal/codegen/darttable` and `compiler`'s lineage rows are green. The suite does fail elsewhere, and EVERY failure was checked against a clean `fixed-table-form` worktree with the same `dist/` and `serialize/` links:

- `compiler` (TestCppCollectionWidening, TestRepeatedArrayTailDefaults, TestCppVariableWideAlignment, TestTableClosureCountDefaults, TestCppTableRetainFramedDepth, TestCppTableAnnouncementDistinctQuantizedShapes, TestGoUnitViewCorpus) and `internal/codegen/golang` (TestEnumNameTakesTheEnumType): **the identical set fails on base.**
- `internal/codegen/gotable` (18 rows) and `internal/codegen/javatable` (TestIdentityCompiledBytesN): **the identical set fails on base.**
- `gotable`'s TestFixedVersioning{NewReadsOld,OldRefusesNew} SKIP on base for want of the corpus and FAIL here only because a whole-suite run builds the corpus and the probe then wants a `serialize.go` sibling this clone has no link to — an environment gap, not this change.

So: no regression from this branch, and no claim yet about the Dart toolchain gates, which are still not run. Read the gate section of the PR body as it stands.

PR #945, opened 2026-09-11
docs: one ceiling, and a name for ill-formed text

gafferongames:
Two DOC reconciliations, each settled by reading the code and running the tests.

## 1. The 65536 ceiling — what the code does

`ir.TableFixedRecordBounds` has **no refusal branch at the ceiling at all**. Past 65536 bytes of record body it appends a WARNING — naming the table, the size, the bound and "form 1" — for a DECLARED `fixed table` exactly as for one merely derived into the form, and `ir.TableFixedFormRoots` / `TableFixedEmitted` then drop the fixed form for it. The table keeps form `1`. The only thing that fails a compile over a record size is `--fixed-record-limit N`, which is **off by default (`0`)**.

`compiler/fixedrecordsize_test.go` pins every step of that: `TestFixedRecordPastTheWireCeilingLosesTheForm` (one warning, no error, no `HugeFixedLayout` emitted), `TestDeclaredFixedTablePastTheCeilingKeepsFormOneAndIsNamed` (`FixedDeclared` true, `len(errs) == 0`), `TestFixedRecordLimitRefusesTheCompile` (the flag is the gate, and off by default), and `TestTheCorpusMegabyteFixedTablesCompile` — `tables/examples`' declared `fixed table WideBlob` at 280012 bytes **generates, with the warning**.

The LEAF CAP is the bound the keyword does change: `ir.TableFixedLeafCapRefusals` refuses a DECLARED fixed table by name and the unit does not compile (`compiler/fixedleafcap_test.go`, `TestWalkedElementArrayPastTheCapIsRefusedByName`), where one merely derived is warned and keeps form `1`.

**The one true sentence:** *past 65536 bytes of record body the FIXED FORM IS NOT EMITTED and the table is NAMED, declared and derived alike — it keeps form 1 and the unit still compiles; the only record-size bound that fails a compile is `--fixed-record-limit`, off by default.*

So `docs/SPEC-TABLES.md` §3.4's "THIS IS NOT A COMPILE REFUSAL, AND THE `fixed` KEYWORD DOES NOT MAKE IT ONE" and `ir.TableFixedFormRoots`'s comment were right, and **`FIXED-FORM-ALGORITHM.md` §6 was the text that disagreed with the code.**

**Texts moved:** §6's 65536 row (the refusal claim removed, `WideBlob` and the two tests named), §6's `--fixed-record-limit` row (now says it is the only record-size compile failure), §6's LEAF CAP row and the fix-4 row (the split is the cap's OWN, not "the 65536's"), SPEC-TABLES §3.4's HELD BY TEST size-bounds bullet (said a declared table past 65536 does not compile — now says it does, and moves the keyword's split onto the leaf cap), §3.4's `--fixed-record-limit` bullet (one added sentence), §3.4's GENERATOR BOUND bullet (said "nothing in §3.4 stops such a type" where a declared fixed table is in fact refused), and two comments in `ir/fixedform.go` (`TableFixedLeafCap` said the compile fails "exactly as the record ceiling's own branch does" — there is no such branch; `TableFixedLeafCapRefusals` said "the split is the same one").

## 2. Fix 11's name — what the code does

**The packet reader has the VERDICT and no enumerator.** `internal/codegen/cpp/functions.go` returns `false` on a payload that is not well-formed UTF-8 (SPEC.md §4.7) and the Go leg returns `ErrValidation`; nothing in the packet or message runtimes carries a named refusal for malformed UTF-8 or an unpaired surrogate, and none of the refusal vocabularies (`TableMessageReason`, `TableRefuseReason`, the C `SCHEMA_TABLE_*` enum) has a value for it. So the name fix 11 owes did not exist anywhere and is this page's to give: **`text_ill_formed`, `20` in the C and C++ pair, appended after `19 layout_unsupported`** (append-only — a gap in the order beats a peer refused under a name it does not know).

**Texts moved:** §4.5's `text` row (the name, the content rules it covers, the `bytes(N)` exemption, and that the packet reader has no enumerator); §5.3's condition table last row (`the name fix 11 owes` → `text_ill_formed`, with the per-flavour conditions and the leg status); §5.3's C/C++ enumerator note (`20` appended, and why it is appended rather than slotted); §5.3's "REFUSE is total" paragraph (`text_ill_formed` is the one refusal that cannot be, since it fires in the scatter); a new §1.1 note saying why no layout rule can catch it early; SPEC-TABLES.md's `TableRefuseReason` table (a row for it) and §3.4's `text` op row (was "a violation is `malformed`"); the fix-11 row in "Reference fixes pending"; and `FIXED-FORM-VERSIONING-TESTS.md`, which gains three hostile rows — `string_grow_hostile_text` (a lone continuation byte `0x80`), `wstring_grow_hostile_text` (an unpaired high surrogate beside a well-formed pair), and `bytes_grow_hostile_text`, **the negative control that must read CLEAN** because `bytes(N)` has no content rule.

**§5.8 row 8 stays OPEN, and now says what every leg does**, because the legs do not fail the same way: C++ (`cpptable.emitFixedTextContent`), C, Rust and **Elixir** (`text_ok?`) validate the used units and report DAMAGE — zero the field, restore its default, count one `malformed`; Go, JavaScript, Dart, C# and **Java** carry the text flavour in the plan's `meta` and **never check the used units at all**, so ill-formed text lands as bytes and the read looks clean. Two of the nine named here were not in the brief: Elixir validates, Java does not.

## Where the code looked like the bug

Nowhere in task 1 — the compiler, `ir`, the tests and §3.4 agree, and §6 was simply stale. In task 2 the code is the known pending divergence fix 11 already records: four legs answer ill-formed text with damage where the ruling is a refusal, and five answer nothing at all. The second half of that — five legs with no content check on the fixed path — is a real hole that no text said out loud before this PR, and it is now §5.8 row 8's first sentence.

## Gates

`go test ./compiler/ -run 'DocExamples|FixedRecord|FixedLeaf'` ok; `gofmt -l .` empty; `go build ./...`, `go test ./ir/ ./internal/lockfile/ ./tools/fixedtwin/` all ok.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/ceiling-and-text-name` is kept at `ea80ddf4fa360380f0163eadda7379e0cc5311d1` (bundle `pr-945-ea80ddf4.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/ceiling-and-text-name@ea80ddf4fa360380f0163eadda7379e0cc5311d1` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

Issue #23, opened 2026-08-14
correctness: C backend names are absent from the claimed-name registry — a same-file object/type pair emits a duplicate C function


rowan-claude:
Severity: **high** · Kind: correctness · Call: ours-to-fix

## What is wrong

The C backend's generated names are absent from the claimed-name registry, and the C object writer builds its name from the OBJECT name while every other backend builds it from the STRUCT name. Concrete collision: object `Ship` → `write_ship_shallow`; type `ShipShallow` → `write_ship_shallow` (snake("ShipShallow") == "ship_shallow"). Both functions land in the same generated <Base>Wire.h — an accepted schema whose generated code will not compile, the exact class commit c213c4f named.

A second instance of the same hole: C constants are `#define screaming(name)` while Rust uses ir.RustConstName, and the two snake functions disagree on consecutive capitals — `MaxHP` becomes `MAX_H_P` in C and `MAX_HP` in Rust — and only the Rust spelling is registered.

## How it manifests

A schema author can write a legal schema (an object and a type whose names collide only in C's snake_case) that passes every check and produces C output that fails to compile. SPEC.md:1119-1125 makes claimed generated names a language rule; the checker implements it for four of five targets.

## Evidence

> VERIFIED READ. C object writer: internal/codegen/c/objects.go:192 emits `write_%s_shallow` from snake(d.Name) — the OBJECT name. C struct writer: internal/codegen/c/c.go:482 emits `write_%s` from snake(st.Name). Concrete collision: object Ship → write_ship_shallow; type ShipShallow → write_ship_shallow (snake("ShipShallow") == "ship_shallow"). The registry (internal/check/check.go:1897-1919, 1932-1946) registers only Pascal forms and Rust forms; there is no C-form registration anywhere in check.go. Both functions land in the same generated <Base>Wire.h (c.go:462-470 emits Struct and Object functions from one g.file.Decls loop). Divergence confirmed in committed output: generated/c/ObjectsWire.h:298 `write_ship_shallow` vs generated/rust/src/objects.rs:470 `write_ship_data_shallow` vs generated/cpp/ObjectsWire.h `WriteShipData_Shallow`.

## Fix direction

Pick one C spelling (emitting `write_ship_data_shallow` is a generated-source golden re-pin with no wire movement), promote the C snake/screaming functions into internal/ir so the checker and the emitter cannot drift, and register the C forms in addStructSymbols and the ObjectDecl arm.

Not blocked on any DECISION issue.

---
[...]

PR #952, opened 2026-09-11
c leg: the versioning gate's control, the manifest oracle, and the refusals by name

gafferongames:
A cold read of the C leg against docs/FIXED-FORM-ALGORITHM.md §5 and
docs/FIXED-FORM-VERSIONING-TESTS.md found nine test gaps. **Eight are closed
here; item 6 is not started and is stated as owed.** Tests and make only — no
runtime and no emitter behaviour changes (`internal/codegen/ctable/fixedruntime.go`
and `internal/codegen/cpptable` are untouched; the twinned runtimes are another
line's).

1. **`tables-c-versioning` had no negative control — GREEN.** §5.9 #39's shape:
   the name a file OUTSIDE the lineage owes (`layout_newer`) is swapped for the
   one a file BELOW THE FLOOR owes (`layout_unsupported`), through
   `go test -overlay`, and every OLD-REFUSES-NEW row must go red naming it and
   nothing else. Registered in `make/negative-controls.json` — there is **no `c`
   group** in that file, so it gets its own `c-versioning` group beside the other
   compile-heavy `base-c-*` jobs, with the `why` line saying why;
   `go run ./tools/negativecontrols check` is green. `ci.yml` enumerates nothing:
   the negative-controls matrix is built from the JSON, so no workflow edit is
   owed. Also wired into `test-c`.
2. **The NEW-READS-OLD value oracle was hard-coded — GREEN.** §5.9 #32/#37: the
   oracle is `build/fixedform-corpus/manifest.txt` and nothing else. The harness
   reads it (there was no manifest reader on the Rust leg to copy — Elixir's is
   the only one in the tree and it reads `lead`/`trail` alone), and every one of
   the 27 rows now asserts the record COUNT and every value the file carries, at
   the path the manifest holds it at. Two facts stay on the row because they are
   the READER's declaration and not a value: a renamed field's spelling (§5.9
   #41, paired by wire id) and an enum-keyed slot's STORAGE index (§2.4, key k at
   index k - 1). The two hand-written oracles the manifest replaces are deleted;
   what survives is the appended tail's declared default and the prefill over
   §5.7's poison, which no manifest line holds.
3. **`array_elem_widen` asserted `widened != 0` — GREEN.** §5.9 #33: a widened
   run is never folded, so it owes EXACTLY `min(their_n, my_n)` = 4.
4. **`floor_below` and `floor_raise_live` never asserted `layout_hash` — GREEN.**
   §5.9 #7: both named layout refusals report the file's hash.
5. **Seven §5.3 conditions had no C test — SIX GREEN, ONE KNOWN-RED.** One probe
   each in `test/c-tables/fixedform_refusals.c`, each asserting the NAME from a
   forged file, each holding §5.3's "REFUSE is total" against a 0xCD-stamped
   destination: `no_layout`, `batch_too_large`, `previous_form`,
   `message_form_as_file`, `newer_form`, and the nameless ragged tail
   (`malformed` true, `refused` false, reason untouched, returns -1) all fire.
   **`plan_too_large` is printed KNOWN-RED by name**: the capacity check sits
   under the not-my-hash branch, and FX1's generated load emits no such branch at
   all (one known layout, its own), so an identity read never reaches it. The
   probe is in the tree and the line says where the row is owed again.
6. **NOT DONE — the hostile column on the C versioning harness.**
   `test/c-tables/fixedform_main.c` still skips `fixed_fx2_bounds` and points at
   a lineage harness with no hostile column, so the per-row
   `<row>_hostile_case()` of docs/FIXED-FORM-VERSIONING-TESTS.md is still
   unreached through a compiled plan. This is the one item the window did not
   reach; it is owed, and rows 5/6/7/14 (a runtime divergence another line is
   fixing) and row 4 (the reader's bounds) will want KNOWN-RED marks by printed
   name when it lands.
7. **`fixedform_layout.c`'s header asserted a retired reader — GREEN.** It said
   the header's hash is checked LAST so a broken layout byte refuses by the
   layout's own rule; that was the run-time walk §5.6 retired. §5.3 looks the
   hash up FIRST. The words are fixed, the function stays and stays compiling,
   which is §5.9 #23.
8. **The C leg had no layout-against-ir gate — GREEN.**
   `TestCFixedFormLayoutIsIrsByteForByte` beside the Rust one (it is the only
   other leg carrying this test today), comparing the emitted layout bytes and
   hash against `ir.TableFixedLayoutBytes`/`Hash` per root, with no other backend
   in the room.
9. **`fixedform_fx1.c:291` PINS §5.8 row 8 — GREEN, not flipped.** The
   assertion now says so: ill-formed text in the used units owes a refusal by
   name under §5.3, the green reference zeroes the field and sets `malformed`
   instead, and what these three lines hold is the REFERENCE's divergence and not
   the page's rule. They flip the day fix 11's name lands, and not before.

## Gates

| gate | state | time |
|---|---|---|
| `make tables-c-fixedform` | green, 1 KNOWN-RED printed (`plan_too_large`) | 37 s |
| `make tables-c-versioning` | green, 27 rows × 2 columns | 23 s |
| `make tables-c-versioning-negative-control` | green (reds on the swapped name, and on nothing else) | 11 s |
| `go run ./tools/negativecontrols check` | green, 238 controls grouped | 1 s |
| `go test ./internal/codegen/ctable/ ./tools/negativecontrols/ -run 'C\|Fixed\|Lineage'` | green | 21 s |
| `go test ./compiler/ -run 'C\|Fixed\|Lineage'` | the new test green; `TestTableClosureCountDefaults` and `TestGoUnitViewCorpus` red on this bench for a missing sibling `serialize.go` / `serialize.cs` checkout, identically before and after this branch | 262 s |
| `gofmt -l .` | empty | — |


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/c-versioning-test-gaps` is kept at `64ec9083508b0e71188fda24417078059eb7e8ea` (bundle `pr-952-64ec9083.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/c-versioning-test-gaps@64ec9083508b0e71188fda24417078059eb7e8ea` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

Issue #26, opened 2026-08-14
spec: quantize units are never validated as exactly representable, three backends emit a contractable expression, and the C leg omits -ffp-contract=off


rowan-claude:
Severity: **medium** · Kind: spec-gap · Call: ours-to-fix

## What is wrong

Three related gaps around the fixed-point quantize arithmetic:

1. Nothing validates that a declared quantize unit is exactly representable (or a power of two) — no exactness check exists anywhere in internal/schema or internal/check.
2. Three backends (C, C++, Go) emit the quantize as one contractible `a * b + 0.5` expression.
3. The C conformance leg omits the `-ffp-contract=off` that SPEC.md:1931 states as a blanket rule for conformance builds — the C++ leg has it, the C rule does not.

## How it manifests

This is the hazard the family just removed from compressed_float, one level up, and it cannot fire today purely by luck of the corpus: every declared unit is 1024 (2^10), so the product is an exponent adjustment and is exact, making a fused multiply-add and a rounded one agree. The moment anyone writes `quantize = 1000`, the C, C++ and Go backends can contract while C# and Rust cannot — five languages, different bytes, no diagnostic — and the gate SPEC.md names as the catch is not armed on the one native leg whose generated packer emits the contractable expression.

## Evidence

> VERIFIED READ. internal/codegen/c/objects.go:338, internal/codegen/cpp/objects.go:260 and internal/codegen/golang/objects.go:231 each emit the quantize as one `a * b + 0.5` expression — observable at generated/c/Objects.h:170 `double quantized_value = floor( (double) input->position.x * (double) 1024 + 0.5 );`. No exactness or power-of-two validation exists anywhere in internal/schema or internal/check. Every declared unit today is 1024 (examples/Constants.schema). Flags: Makefile:5 `CXXFLAGS ?= -std=c++17 -Wall -Wextra -Werror -ffp-contract=off`; the C rule at Makefile:125-127 compiles with `-std=c99 -Wall -Wextra -Werror -Wtype-limits $(C_TAUTOLOGICAL) -O2` and NO fp-contract flag — while SPEC.md:1931 states the flag as a blanket rule for conformance builds.

## Fix direction

[...]

PR #953, opened 2026-09-11
fixed form: depth past the bound is named, never dropped

gafferongames:
**THE BOUND IS 64 — the LAYOUT's own**, and the compiler now holds the same number the wire does.

A red-team pass found two numbers for one fact: `ir.TableFixedSupported` refused a closure past **16**, while §5.2 refuses `layout_malformed` past **64**, §6 and `docs/SPEC-TABLES.md` §3.4 state *"the walk's nesting cap is a SMALL STATED CONSTANT: 64 nested bodies"*, and every runtime holds the wire to 64 (`kTableFixedMaxDepth`, `tableFixedMaxDepth`). So a DECLARED `fixed table` nested 17 deep lost the fixed form **with nothing said** — at 17, at 101, at any depth — against §4.5 fix 4's *"A REFUSAL BY NAME, NEVER A SILENT DROP"*.

**Why 64 and not 16.** The 16 has no reason written beside it and none in its history: `git log -S 'depth > 16'` names one commit, f144cd9b, the commit that brought `ir/fixedform.go` into being, whose message says nothing about a depth at all. A compile bound **below** the wire's is a compiler that drops a form the wire would have carried, so the two agree now: `ir.TableFixedMaxDepth = 64`, asked of the **layout's** nesting (`ir.TableFixedLayoutDepth`, read from the layout bytes the way §5.2's EMIT counts them, root at depth 0) rather than of the struct closure, because the layout is where the wire's answer lives and is what a lock holds for a layout that shipped.

**The treatment past it is the 65536 ceiling's, for the ceiling's own stated reason.** On the tip today the ceiling *warns* — `ir.TableFixedRecordBounds` (ir/fixedform.go), naming the table and the size: *"THE FIXED FORM IS NOT EMITTED FOR IT … it keeps form 1, and the `fixed` keyword still buys it the CLASS"* — because 65536 is the form's bound and not the class's. A nesting past 64 is the same kind of thing: §3.4 calls `layout_too_deep` *"a bound on the WALK and not on the wire"*. So `ir.TableFixedDepthBounds` names the table and its depth, always on, and `ir.TableFixedEmitted` drops the form for it; a DECLARED table is named the more loudly (the keyword is a request, and a request the form cannot serve is said out loud). Nothing is silent either way — the surprise fix 4 forbids is silence, not form 1.

[...]

PR #957, opened 2026-09-11
seams: the form byte first, the hash as computed, the reserved bytes, the bound record

gafferongames:
A seams read found five defects Rowan rules now. Each lands across the DOCS and, where code moves, the **C++ REFERENCE first** and then every leg whose emitted load carried the same line.

### 1. The form byte is read before the length

`docs/FIXED-FORM-ALGORITHM.md` §5.3 ordered "fewer than 20 bytes: `malformed`" ahead of the form byte. A **committed form-1 file is ten bytes** and a **form-2 batch three**, so SPEC-TABLES §3.4's promise — "a fixed reader given `1` answers `previous_form`, given `2` answers `message_form_as_file`" — was false for every real file of either form: they all answered `malformed`.

New order: a file with **no first byte** is `malformed`; otherwise **the form byte earns its name first**; then the twenty-byte minimum. §5.3's steps 1 and 2 swap (step numbers 3.. are untouched, so no cross-reference in the tree goes stale), §2's load table likewise, and SPEC-TABLES §3.4 states the same order beside the header picture.

Code: the C++ reference's emitted load and all eight other legs. **The C# outlier is folded in too** — `data[0] < Form → previous_form` named form byte **0** `previous_form`; it is now §5.3's own rule, `1` previous_form, `2` message_form_as_file, OTHERWISE newer_form, the same as the other eight.

### 2. §3.4 stated the wrong hash

§3.4 said the hash is `fnv1a64` over the layout's bytes "exactly as written". The bill §13 and `ir/fixedform.go`'s `TableFixedLayoutHash` fold the **definitions digest** in after the last layout byte. §3.4 declares itself the law, so it now states the hash as the code computes it, cites bill §13 and §5.2, notes that an empty digest leaves the hash equal to the layout bytes alone, and gives the reason (`int32 | 0..100` and `| 0..200` would otherwise hash identically). §21 names the digest where it names the hash.

§3.4's two "**a header whose hash is not the hash of the layout behind it is refused**" mandates were retired by §5.3 — no runtime recomputes a hash, the digest not being on the wire — and are rewritten to §5.3's rule: **byte-compare against the lock's bytes for a known hash, `layout_newer` otherwise**.

### 3. A float range's digest bytes

§5.2's contract table said min and max ride "each as an i64 LE" for every range. The code writes `math.Float64bits` for a float bound (`ir/fixedform.go`). A port that followed the words would hash `0.5` and `0.25` the same — both truncate to `0`. The table now splits the three: integer bounds as i64 LE, **float bounds as the IEEE-754 bits of the `float64` value, LE**, fixed-point as its raw i64. `ir/` pins one float range's digest against a **literal**.

### 4. Reserved header bytes 1..7 refuse nonzero

Both header statements said "reserved, zero" and **no leg checked**. A reader now refuses a nonzero reserved byte as `malformed`, so the bytes stay usable when the cook and the block form join the registry under the same header. Doc in §5.3 (step 2's second half) and §3.4; code on the reference and every leg's load.

### 5. The lock binds `record=` to the entry's layout

`internal/lockfile/lineage.go`'s `parseLineage` recomputed the wire hash over bytes+digest and refused a disagreement, but `record` was **only range-checked**. §5.9 #26 rules a record size that is not what the entry's layout accounts for a **LOCK BUG, and the build fails**. Entry 0 of the layout bytes carries the root's constant size, so `parseLineage` now refuses a `record=` that is not that size, **with both values in the sentence** — never routed into §5.3's arithmetic over a file's tail. `lockrows_test.go` hand-edits a lineage line: red then green.

Also §5.2 now says the plan cap is over **SUPPORTED entries only** (§5.1's `plan cap` row): a retired entry's plan is not built and cannot fail the build.

### Per-leg and per-gate state

| leg / gate | state |
|---|---|
| C++ REFERENCE (`cpptable`) | code landed |
| C twin (`ctable`) | code landed |
| Go (`gotable`) | code landed |
| Rust (`rusttable`) | code landed |
| JS (`jstable`) | code landed |
| Dart (`darttable`) | code landed |
| Java (`javatable`) | code landed; **gate cannot run on this bench** — no Java runtime installed (`javac: Unable to locate a Java Runtime`), which is why `TestIdentityCompiledBytesN` is red here and was red before this branch |
| C# (`cstable`) | code landed, plus the `previous_form` outlier fixed |
| Elixir (`elixirtable`) | code landed |
| `go test ./ir/ ./internal/lockfile/ ./compiler/ -run 'Fixed\|Digest\|Lock\|Lineage\|Paired\|DocExamples'` | green |
| `go build ./...`, `gofmt -l .` | green, empty |
| `make tables-fixedform` | **green** — reference and C twin, plain and under ASAN; the byte-flip fuzz answers 2968 by name, 56 `malformed`, **0 divergences** |
| `TestPairedHeadersAgree` (tools/fixedtwin) | green |

**The per-leg fixed-form suite cases this branch owes** — a ten-byte form-1 file and a three-byte form-2 batch answering by NAME, a zero-byte file `malformed`, a form byte `0` answering `newer_form`, and a file with byte 3 set answering `malformed` — are **not yet written on every leg's probe harness**; the reference and the C twin come first and the rest follow. That is the one piece of this PR still open, which is why it is a draft.

**A DeepSeek read** of this branch, verified by triage, found five seams where the rule and the code stating it had come apart, and `532ae711` closes them: the Elixir reserved-byte refusal was UNREACHABLE (the success clause bound bytes 1..7 to a wildcard and matched first — it now spells the seven literal zeros, and the reserved clause stands before the length clause); §5.9 #26's refusal named neither the table nor the entry's hash and now names both; `record=0` parsed and is now refused as the lower rung under the entry-0 comparison, never over it, because a fieldless fixed table's body is honestly zero; SPEC-TABLES §3 still said the seven reserved bytes are "read by nobody" against §3.4's refusal; and the COMPILE pseudocode built a plan for every entry while the prose says a retired entry's plan is never built — the floor is computed first now and the loop skips below it. **The five suite cases this PR owed are written on the ELIXIR leg** (the ten-byte form-1 file, the three-byte form-2 batch, zero bytes, form byte `0`, and byte 3 set), red before the fix and green after; the other six legs still owe them.

---
**Rebase note, 2026-09-13 (Rowan):** rebased onto fixed-table-form at 8ea5ed8e; head 31e3bae8. Two lane facts followed, neither a textual conflict: (1) `internal/lockfile/lineage.go`: the lane's `ValidateLayout` owns §1.1's seven rules, so §5.9 #26's `record=` comparison is now a three-arm switch (the RANGE rung, the rules, the entry-0 comparison); (2) `ir/fixedform_hash_test.go` now pins `'Q'` beside every `'R'`. The nine-leg load move required re-pinning 42 golden Table sources and 12 generated files; every hunk is the same three changes and no wire moved. Local gates on 31e3bae8: gofmt, vet, tidy clean; compiler/ir/lockfile tests ok; `make tables-fixedform` exit 0 (2968 refused by name, 56 malformed, 0 divergences); block-zero-cost 123 byte-identical. Fast CI 34766919279 green, 13 jobs.

**Corrections to the owed-probe text above:** of the five cases on nine legs, five are present on Elixir and one on Go (form byte 0 → newer_form); **39 are still owed**, and the sentence "the other six legs" should read **eight**: neither the reference nor the C twin carries any of the five. No leg but Elixir probes a nonzero reserved byte.


gafferongames:
Read the nine changed loader paths at `31e3bae834712cdeb332d972b651656f2f397ec4` against algorithm §5.3 steps 1–3: absent first byte is malformed; an available non-3 form byte receives its named refusal before the form-3 minimum-length check; nonzero reserved bytes are rejected before layout decoding. The Elixir successful pattern now requires seven zeros, so the reserved-byte refusal is reachable. C# now classifies 0 as newer_form rather than previous_form.

Read the lineage size comparison and its ordering below validated layout shape, plus the literal floating-bound digest test. **One small documentation inconsistency remains:** §5.2 says Q only accompanies compressed floats, while `tableFixedDigestField` emits Q for every `HasFloatRange`, and this PR's new uncompressed-range test explicitly expects Q plus eight zero bytes.

Prepared the complete three-line repair as **`085b27d5`**, on `codex/seams-digest-contract-20260913`, based on this exact head. Please cherry-pick it into the existing PR and check the wording; no wire/code change is needed.

Local validation: 18 named tests passed, 2.47 seconds including the command:

```sh
go test ./ir/ ./internal/lockfile/ ./tools/fixedtwin/ -run 'TableFixedDefinitionsDigest|LockLineageRecord|RecordedLayoutRule|PairedHeadersAgree' -count=1 -json
```

All 14 non-skipped CI checks also passed at the reviewed head. **I found no additional blocker in the changed loader/lock code during this scoped read.** Landing these repairs can proceed through Rowan's existing lane with the wording aligned. This is not a claim that the missing 39 per-language probes passed: retain them as required unfinished Fixed Tables work, with explicit ownership and proof before declaring the feature complete.

rowan-claude:
Rebased again after #953 landed on the lane (9785a76c): one conflict, `internal/lockfile/lockrows_test.go`, both sides appending test blocks (the lane's depth-bound tests, this PR's record-size test); both kept whole. Stella's digest-doc repair (085b27d5) is carried. Local: gofmt, vet, `go test ./internal/lockfile/ ./compiler/ ./ir/` green. Lands when this head's run is green; the read on record is Stella's at 31e3bae8 (loader and lock code unchanged since, only the rebase seam in a test file).

PR #958, opened 2026-09-11
packet wire: the compressed float's index clamps to the step count on every leg

gafferongames:
## The hole

docs/SPEC.md §4.3 names `serialize_compressed_float` as the wire oracle for a
quantized `float32` — the `ceil`, the `+0.5f` rounding, the clamp — but it never
stated the **integer** clamp the reference performs *after* the rounding and the
floor. The C++ reference has it (`if ( integerValue > max_integer_value )`), and
so do the C, Rust, C# and JS runtimes. Four legs fold the arithmetic at
generation time and folded it **without the clamp**:

| leg | site |
|---|---|
| Go | `internal/codegen/golang/functions.go` (`emitWriteCompressedFold`) **and** `internal/codegen/golang/flat.go` (`flatCompressedPiece`) |
| Java | `internal/codegen/java/functions.go` (`emitWriteCompressedFloat`) |
| Dart | `internal/codegen/dart/functions.go` (`emitWriteCompressedFloat`) |
| JS flat | `internal/codegen/js/flat.go` (`emitWriteCompressedFloat`) |

Elixir already emitted `min(integer, miv)` and called it "the normative integer
clamp" — it was the only folding leg that did.

## The tie

For a step count in `[2^23, 2^24)` the float32 ulp at the top of the range is
exactly 1, so `fround(steps + 0.5f)` is a **tie** and rounds half-to-even **UP**:

```
8388609  -> 8388610        12582913 -> 12582914
8388611  -> 8388612        16777215 -> 16777216
```

The writer emits `steps + 1` and every reader's headroom refusal (`v > steps`)
rejects it — a leg writing bytes its own reader will not read, and bytes the
other five legs do not write. Outside `[2^23, 2^24)` the clamp changes no byte,
so it is free everywhere else.

## What changed

1. **docs/SPEC.md §4.3** — the `float32 | min, max, resolution` wire row now
   states the clamp as normative (`index = min(index, steps)`), and a new bullet
   states the boundary case with the triple, the tie, and the four values.
2. **The four emitters** emit the clamp the way Elixir does (five sites — the Go
   leg has both a functions form and a flat form).
3. **The vector**: `examples/Wire.schema` gains `CompressedCeiling` —
   `ceiling float32 | min = 0, max = 8388609, resolution = 1`. That derives
   `steps = 8388609` (odd, in `[2^23, 2^24)`) on a 24-bit wire;
   `ir.ValidCompressedFloatParams` accepts it. Writing exactly `max` normalizes
   to `1.0f`, scales to `8388609.0f`, and `fround(8388609.0f + 0.5f)` is
   `8388610.0f`.
4. **All nine legs** write it against one golden —
   `testdata/wire/compressed_ceiling.bin`, `01 00 80`, index 8388609 — which is
   the cross-language byte-identity gate the existing `compressed_probe` vector
   uses. Each leg also asserts the read returns **exactly** `max`, so an
   unclamped index fails the read before the bytes are even compared. The JS leg
   carries it through `flatCross` too, so the flat form is held to the same
   bytes. The C++ and C legs additionally assert the tie itself, so the vector
   cannot quietly stop discriminating.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/packet-cfloat-max-clamp` is kept at `457d7a71fd50dd662825c4027514f9c5f247f9c4` (bundle `pr-958-457d7a71.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/packet-cfloat-max-clamp@457d7a71fd50dd662825c4027514f9c5f247f9c4` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #959, opened 2026-09-11
spec: the read buffer's 8 bytes of slack are the design

gafferongames:
The packet reader's slack past the payload was nine per-target facts in §6.3 and a per-language list in USAGE.md — which called it "the one thing that differs per language" — with nothing normative in §4 or §5 saying what the CALLER owes. So the question kept being asked.

Glenn's ruling, 2026-09-11, verbatim:

> The design intent is to allow fast implementations by making this 8 bytes past the payload enabling 64bit reads. This is intentional. It is not required that implementations do this, but the standard allows them to do it if they want, to get the fastest, correct implementation. Please mark this in the spec because I have answered this already.

It is law now, stated once, in **SPEC.md §4.3** beside the framing sentence:

> **A caller hands a reader a buffer with AT LEAST 8 BYTES READABLE PAST THE PAYLOAD'S LOGICAL LENGTH.**

An implementation MAY read up to 8 bytes past the payload — that is what lets it load 64 bits at a time — and MUST NOT depend on more. One that reads nothing past the payload is conforming too: the contract is a permission to the reader and an obligation on the caller. The slack's contents are never interpreted, so a read's verdict and its bits consumed are identical whatever those bytes hold — the slack is allocation, never wire, and `Write` still returns exactly `ceil(bits/8)`.

Under that law:

- **§5** separates a short buffer from a malformed payload. "No target traps, panics or aborts on a malformed payload" is a guarantee about BYTES; a buffer shorter than the contract is a CALLER error, and a reader permitted to load 64 bits at a time cannot detect that the last load was outside the allocation — so a target fails however that target fails on a caller error.
- **§6.3's** buffer row is retitled *what it READS* and framed as **what each target reads today, not nine contracts**. Go's fast-path 7 stays as the fact it is, with the contract stated at 8 like the others. The JavaScript row says a buffer short of the contract throws a `RangeError` off the DataView — a caller error, not a malformed payload.
- **USAGE.md** (~783) stops teaching the nine numbers as the contract and points at §4.3; the numbers stay as a record, flagged as one a target may change without the caller changing anything.
- One code comment, in `internal/codegen/js/flat.go`, says `FLAT_READ_SLACK = 8` is the spec's contract and not that tier's own number.

**Gate.** The repo has no table-of-contents or anchor test for SPEC sections — `compiler.TestDocExamplesParse` covers fenced examples only — but it does have page-sentence gates in the style of `TestPagesPlaceTheVoidArmOutsideTheTableClosureClass`. This adds one: `compiler.TestPagesStateTheReadBufferContractAsLaw` holds the law's halves in §4, §5's caller-error clarification, §6.3-as-a-record, USAGE.md's pointer, and the four stale per-target spellings.

`go test ./compiler/ -run 'DocExamples|ReadBufferContract' ./internal/ci/` green; `gofmt -l .` empty. Docs plus one comment plus one test — no generated bytes move.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/spec-read-slack-is-the-design` is kept at `edbaa93926cfa315bcbb03c25d95e164d53aa358` (bundle `pr-959-edbaa939.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/spec-read-slack-is-the-design@edbaa93926cfa315bcbb03c25d95e164d53aa358` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #961, opened 2026-09-11
packet wire: a non-finite value at a compressed float is a write-contract violation on every tier

gafferongames:
A quantized `float32` has no bit pattern on the wire — it has a step index — and NaN and ±Inf name no step. A spec read found that writing one was **undecided by SPEC** and decided eight different ways by the runtimes, with Go deciding nothing at all.

## The rule (SPEC §4.3, new paragraph)

A non-finite value at a compressed float is a **write-contract violation**, exactly like an int outside its range or a count outside `[A, B]` (§4.6). It takes §5's tiers with **no exception**: the seven debug-only targets assert it, the two every-build targets refuse it in every build.

The paragraph also states the **release outcome**, which was the part nobody had written down. In a release build of the seven the write PROCEEDS, and what it writes is defined rather than unspecified: the value goes through the quantizer's saturating clamp, whose form is normative for exactly this reason — written `if !(n >= 0) n = 0; else if !(n <= 1) n = 1;`, the `!>=` first arm so that **NaN lands on step 0, `min`'s index**, never on the undefined float-to-unsigned cast a plain `n < 0` test would let it reach. `-Inf` lands on step 0 too; `+Inf` lands on the top step, `max`'s index, by the ordering the second arm already applies to any value above the range. So the bytes are deterministic, identical in all seven, the index is always a legal one, and the read returns `min` (or `max` for `+Inf`) rather than failing. The read side is untouched.

**One deviation from the brief, deliberate:** the brief asked for clamp-to-min for `+Inf` as well. That cannot be had. A finiteness test that survives release IS a release check, which §5 forbids outright ("No runtime should ever promise to keep checks in writing packets (asserts) in release build"), and it would also move the C++ reference's bytes — the family's byte oracle. `+Inf` therefore saturates to `max`'s index, which is equally deterministic and equally a legal index. The SPEC paragraph says so explicitly rather than leaving it to be rediscovered.

## Per-leg state, before and after

| leg | before | after |
|---|---|---|
| **Go** | **nothing.** NaN clamped to step 0 silently, in any build (`internal/codegen/golang/functions.go` ~534) | `if x-x != 0 { return serialize.ErrValueOutOfRange }`, **always on**, in both the object-tier fold and the flat run's guard. The clamp stays below it — the bytes must not depend on the tier |
| C++ | `serialize_assert( value - value == 0.0f )` at write intake in `serialize_compressed_float_precomputed_internal`; release clamps | unchanged — already the rule |
| C | `serialize_assert( serialize_float_is_finite( value ) )` in the same entry point; release clamps | unchanged — already the rule |
| Rust | `debug_assert!(value.is_finite())`; release clamps | unchanged — already the rule |
| C# | `Debug.Assert(float.IsFinite(value))` in `QuantizeCompressedFloat`, the shared stream+batch write intake; release clamps | unchanged — already the rule |
| Java | `assert angle - angle == 0.0f` in the `checkWrite` predicate. **Checked as asked: `x - x == 0` covers NaN AND both infinities (Inf - Inf is NaN), not only NaN** | unchanged — already the rule |
| Dart | `assert(x.isFinite)` | unchanged — already the rule |
| JS | checked flat writer `if (!Number.isFinite(x)) return -1`; production holds nothing. **Checked as asked: production does NOT clamp "to step 0 via NaN arithmetic" — `Math.fround(NaN)` reaches the `!(n >= 0.0)` arm and lands on step 0, which IS min's index**, the same path every other release build takes | unchanged — already the rule |
| Elixir | `raise ArgumentError, "a compressed float writes a finite number"` in `cf_quantize`, always on | unchanged — already the rule |

So the rule cost one leg a change and eight legs a read that confirmed them.

## Tests

`compiler/nonfinitecompressed_test.go` pins both halves across all nine targets:

1. **the contract is held in each target's own tier idiom and nowhere else** — the assert text for Java, Dart and Elixir, the runtime call site for C, C++, C# and Rust, the every-build refusal for Go, and for JS the checked/production **fork** (`!Number.isFinite(x)` present in `writeTFlatChecked`, absent from `writeTFlatProduction`), plus the inverse claim that no debug-only target carries an every-build refusal release cannot drop;
2. **the release outcome is the same in all nine** — the clamp's `!>=` first arm pinned as text wherever generated code carries the clamp itself, so a regression that rewrites it as `n < 0` (and sends a release NaN to the cast) is named.

Both go red if the Go guard or any leg's assert is removed.

## Gates

- `gofmt -l .` empty.
- `go test ./internal/codegen/golang/ ./internal/check/ ./ir/ ./internal/goldens/` green.
- `go test ./compiler/ -run 'NonFinite|TheClamp'` green.
- Regen sweep (`rm -f bin/schema`, rebuild, every `generate --out generated/` invocation re-run) + `go test ./internal/goldens -update`: **only Go files moved**, as the change predicts — `generated/go/Wire.go`, `generated/bench/go/Bench.go`, `generated/bench/go/realworld/RealWorld.go`, `generated/bench/paired/go/Bench.go`, `testdata/golden/go/Wire.go`. Hunks read: one refusal per compressed-float field, ahead of the fused run, above the unchanged clamp.

## Not runnable in this bench session — named, not hidden

- **The cross-language packet conformance/byte-identity gate (`make conformance`) was not run here**, and the release-outcome vector was NOT added to it. The gate needs every leg's toolchain plus the C++ reference binaries built; the local `serialize.c` checkout is two releases behind (no `_precomputed` entry point at all) and had to be replaced with a fresh `origin/main` clone to read the C leg. The byte claim this PR makes about release builds — NaN and −Inf to step 0, +Inf to the top step, identical in seven legs — is stated in SPEC and pinned structurally by claim 2 above, but it is **not yet pinned as bytes**. That vector, and the per-leg debug-build NaN/+Inf runtime cases (including the C# packet suite's Debug and Release arms), are the follow-up.
- No merge, and the branch is draft for that reason.

## Follow-up a70d992 — carries #958, and the witness a DeepSeek read found missing

**This branch now carries #958** (`rowan/packet-cfloat-max-clamp`, merged in at `4a984b0`) and **must merge after it**. #958 is the integer clamp after rounding on Go, Java, Dart and JS-flat; without it this branch's Go release path had `uint32(float32(normalizedValue*steps) + 0.5)` with no ceiling, so at `steps = 16777215` a value at max wrote `16777216`, masked to `0`, and read back as **min, silently**.

A **DeepSeek read** then found three things missing, and `a70d992` fixes them:

- **The flat tier was never witnessed.** Go folds the quantizer twice — per field in the object tier (`functions.go`) and again inside the flat word codec (`flat.go`) — and each copy carries its own every-build refusal. `compressedUnit` was one field, and `worthFlattening` refuses a run of fewer than two pieces, so the gate only ever reached the object tier: `flat.go`'s refusal could have been deleted and every test stayed green. `compressedFlatUnit` is the two-field twin, and claim 1 asserts the refusal in **both** tiers. Negative control: emptying `flatCompressedPiece`'s guard reds four lines.
- **Claim 2 read Go and Elixir as clampers.** They are not — their refusal is every-build, so nothing non-finite reaches their clamp, and pinning its `!>=` form for them claimed the clamp is what saves them. They move out of `clampForm` into `refusesBeforeTheClamp`, which asserts the refusal stands **before** the clamp in both of Go's tiers and inside Elixir's `cf_quantize`. `everyBuild` is cross-checked against both maps, so no target can sit in neither.
- **Seven, not eight.** The comment above `emitWriteCompressedFold` said the clamp is what "the other eight targets leave standing in release". Elixir is not one of them.

The merge also left `generated/go/Wire.go` and its golden stale — #958 adds `CompressedCeiling`, this branch adds the Go refusal, and neither side's generated output carried the other's. Regenerated off a rebuilt `bin/schema` and re-pinned; the only bytes that moved are the three-line refusal. `go test ./compiler/ -run 'NonFinite|TheClamp|Ceiling'`, `go test ./internal/codegen/golang/ ./internal/goldens/` and `gofmt -l .` all green.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/packet-nonfinite-cfloat` is kept at `f31d151529eb1fa1bef21c44fbb7c80cfecb7837` (bundle `pr-961-f31d1515.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/packet-nonfinite-cfloat@f31d151529eb1fa1bef21c44fbb7c80cfecb7837` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #962, opened 2026-09-11
fixed form: a NaN in a bounded float lands min and counts, on every leg

gafferongames:
A red team (PR #955, `TestFixedCFloatNaNSurvivesTheBoundsPass`) found that the FIXED form's read-side bounds pass (docs/SPEC-TABLES.md §3.4, §5.3 step 11) let a **NaN in a bounded float land whole and count nothing**. IEEE makes every ordered comparison against a NaN false, so `v < lo` and `v > hi` are **both** false — and `-inf` clamps and counts correctly under the very same code, which is what kept it quiet.

**The rule (Rowan's):** a NaN in a bounded float lands `min` and counts `clamped == 1`, exactly as `-inf` does. `-0.0` against a `min` of `+0.0` is **in range** (it compares equal) and lands as written, sign bit and all, counting nothing.

**The fix is the low test, and only the low test:** `!(v >= lo)` is true for a NaN — quiet (`0x7FC00000`) and signalling (`0x7F800001`) alike, the pass compares and never inspects bits — and for every value under the minimum, and false for `-0.0` against `+0.0`. The high end stays the plain `v > hi`: the low test already caught the NaN. A bounded float is a **compressed** float on this wire (`min`, `max` and `resolution` all three together, SPEC §4.6) and rides in the fixed form as the float it is, so the two are one case.

## Per leg

| leg | state | what changed |
|---|---|---|
| **C++ reference** (`internal/codegen/cpptable/fixedform.go`) | **green** | new `fixedClampFloat`: `clamped += (int) ( !( v >= lo ) ) \| (int) ( v > hi );` then the same select. The integer `fixedClampBoth` is untouched |
| **C twin** (`internal/codegen/ctable/fixedform.go`) | **green** | the same text byte-for-byte, through the existing `fixedClampBoth` with a negated low test — `tools/fixedtwin` pairs |
| **Go** (`internal/codegen/gotable/fixedform.go`) | **green** | `if !(v >= lo) { … } else if v > hi { … }` |
| **Rust** (`internal/codegen/rusttable/fixedform.go`) | **green** | new `emitFixedClampFloat`. **`f32::clamp` PROPAGATES a NaN** rather than pinning it, so the intrinsic is exactly what a bounded float may not use; the count stays a non-short-circuiting `\|` of the two ends |
| **Dart** (`darttable/fixeddart.go`) | **nothing to do, named** | `emitClamp` covers `TInt`/`TFixed`/`TBits` only — no float range is clamped in the fixed form at all |
| **JavaScript** (`jstable/fixedjs.go`) | **nothing to do, named** | `emitDecodeBounds` runs off `ir.TableRawRange`, which is integers only |
| **Java** (`javatable/fixedform.go`) | **nothing to do, named** | no float-range clamp emitted |
| **Elixir** (`elixirtable/`) | **nothing to do, named** | no float-range clamp emitted |
| **C#** (`cstable/`) | **not done, named** | the fixed-form bounds pass of **#941 is not on `fixed-table-form`** — `cstable/fixedform.go` carries no float clamp. `cstable/wire.go`'s `ClampRaw` is the **form-1** wire and is out of this change's scope. When #941 lands, its float clamp needs the same negated low test |

**The four legs with no fixed-form float-range bounds at all is a separate gap and a bigger one** — a bounded float there is never held to its declared range, NaN or not. Named here, not fixed here.

## Docs

- **`docs/SPEC-TABLES.md` §3.4** — the bounds-pass section gains the rule beside the plan-op table: the negated low test, the NaN landing `min` and counting one as `-inf` does, and `-0.0` in range.
- **`docs/FIXED-FORM-ALGORITHM.md` §4.6** — a new bullet beside "A RANGED SCALAR", including the note that a leg whose `clamp`/`min`/`max` intrinsic propagates a NaN may not use it.
- **`docs/FIXED-FORM-ALGORITHM.md` §5.4** — the bounds-pass counter row names the NaN and the `-0.0` control.
- **`docs/FIXED-FORM-VERSIONING-TESTS.md`** — two hostile rows: **`float_nan_bounded`** (a forged `0x7FC00000` and a forged `0x7F800001` in a `[0,1]` float, read on **both plans**, → the bits of `min`, `clamped == 1` exactly, never `>= 1`) and **`float_negzero_bounded`** (a forged `0x80000000` → lands `0x80000000`, `clamped == 0` — the negative control, and the reason the test is `!(v >= min)` and not `v != v || v < min`).

## Tests

`compiler/tablefixednanbounds_test.go` — one compressed-float fixed table, generated for all four legs, asserting the NaN-aware low test is emitted **and that the naked `<` is gone** (the shape the red team's read walked through), plus that Rust does not reach for `clamp`.

## Gates

- `go test ./compiler/ ./internal/codegen/... -run 'Fixed|NaN|Bounds|Paired|Clamp'` — **green**
- `gofmt -l .` — **empty**
- regen sweep (`rm -f bin/schema`, `make update-goldens`) — the only emitted-output hunk is `generated/bench/tables/cpp/BenchTableTable.h`, four bounded floats (`server_time`, `aim_x/y/z`), read and exactly the two expected lines each. **No golden under `testdata/golden/tables/` carries a bounded float**, so none moved.
- `make tables-fixedform tables-c-fixedform` — **green** (exit 0, both legs; the C leg's 11 §5.6 retirements and 2536 byte flips as before)
- `TestPairedHeadersAgree` — **green**
- `go test ./internal/codegen/javatable/` fails on this bench with `Unable to locate a Java Runtime` — **environment, not this change**; `TestIdentityCompiledBytesN` needs `javac`
- **Owed:** the reference's `test/tables` fixtures carry no compressed float, so a **runtime** hostile case (forged `0x7FC00000` / `0x7F800001` → the bits of `min`, `clamped == 1`, on both plans; forged `0x80000000` → `0x80000000`, `clamped == 0`) is not yet on the corpus. The two doc rows above are its specification, and the emitted-text tests are what stands in the meantime.

Draft, no merge.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/fixed-nan-bounds` is kept at `de454d130898b47a3518b34d11dc685e8337005d` (bundle `pr-962-de454d13.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on #941. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/fixed-nan-bounds@de454d130898b47a3518b34d11dc685e8337005d` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #963, opened 2026-09-11
packet wire: no contraction, enforced per language

gafferongames:
**Ruling:** "We should enforce the equivalent of `-ffp-contract=off` per-language, if possible."

## The live finding first

`make/c.mk:14` and `:16` built the C packet-text driver — debug and release — **without** `-ffp-contract=off`, twenty lines above the C corpus test that has carried the flag all along. Apple clang 21 on arm64 contracts **by default** for C, so that was not a slow build, it was a build that writes a different quantize index.

Measured on this Studio (arm64, Apple clang 21.0.0), the fold `normalized * steps + 0.5f` over `[0, 10]` at resolution `0.01`:

| value | two roundings (the law) | fused, i.e. no flag | 
| --- | --- | --- |
| `0.005` | **1** | **0** |

Same divergence under `-ffp-contract=fast` and under no flag at all; `-ffp-contract=off` gives 1.

Ten compile lines were unflagged: `make/c.mk` :14 :16 :24 :1019 :1021 :1024 :1034 :1104, `make/checks/table-base64.mk` :12 :16, plus `bench/run.sh`'s C flags. All now carry it.

**Whether a leg's bytes actually differed before the change:** not established end to end. The C packet-text schema (`test/packet-text/Narrow.schema`) has no float field, so that unflagged driver never exercised the fold — the bug was latent, not live on the wire. The fold's real cross-language gate is `examples/Wire.schema`'s `CompressedProbe` against `testdata/wire/compressed_probe.bin`, and every build in that path already had the flag. So: a real wrong-bytes hazard in the build system, no leg caught writing wrong bytes today. Saying otherwise would overstate it.

## What the spec did not say

Contraction is not a C and C++ problem with a C and C++ remedy. It is a permission each language either grants or withholds, and the discipline for withholding it was living in emitter comments — Go's `float32()` wrap, JS's `Math.fround`, Dart's `_fround`, Elixir's Veltkamp/Dekker `fr/1`, the `ctable`/`cpptable` `FORCE_ROUND` barriers. A comment is not a gate.

## §7.2 as written

**Floating-point discipline.** Strict IEEE-754 arithmetic is the normative
wire for compressed floats. The quantize fold is `normalized * steps + 0.5`
with TWO roundings, and a compiler that contracts the multiply and the add
into one fused multiply-add keeps the extra precision and writes a DIFFERENT
step index. The golden-wire corpus pins compressed-float values chosen to
DISCRIMINATE — values where a fused or double-precision build produces a
different index — with the discrimination property itself asserted in the C
and C++ legs, so a vector cannot quietly stop discriminating.

Contraction is not a C and C++ problem with a C and C++ remedy; it is a
permission each language either grants or withholds, and every leg states
which. **No leg may emit a contracted product.** The equivalent of
`-ffp-contract=off` per language:

| Language | Mechanism | Where enforced | How tested |
| --- | --- | --- | --- |
| C | `-ffp-contract=off` on every compile of generated or conformance sources. clang's DEFAULT for C contracts, so an unflagged build is a WRONG-BYTES build, not a slow one. | `make/c.mk`, `make/checks/*.mk`, `bench/run.sh`, `bench/tables/*/leg` — every `$(CC)` line | `TestNoContractionFlagOnEveryCompile` greps every compile line in `Makefile`, `make/` and `bench/` for the flag; the C leg asserts the discrimination property directly |
| C++ | `-ffp-contract=off` in `CXXFLAGS` and on every literal `$(CXX)` line | `Makefile:33`, `make/*.mk`, `make/checks/*.mk`, `bench/run.sh` | same grep test; the C++ oracle asserts `plain != fma` on the pinned probe |
| C, C++ (the message codec, where the arithmetic IS emitted) | Two barriers, because the consumer's flags are not ours. (1) The fold goes through a NAMED `float` object — `float scaled = normalized * count;` then `scaled + 0.5f` — and assignment to a `float` lvalue forces the intermediate rounding. (2) An empty-`__asm__` `FORCE_ROUND` on `scaled` pins the value into a float register, with a `volatile float` fallback off gcc/clang. | `internal/codegen/ctable/message.go`, `internal/codegen/cpptable/message.go` | `TestNoContractionInEmittedCode` requires BOTH the two-statement form and the barrier macro; the form was checked on arm64 clang under `-ffp-contract=fast` (index 1, correct; one expression gives 0) |
| Go | The Go spec permits fusing `x*y+z` unless an explicit conversion intervenes, and arm64 takes the permission. Every emitted product is wrapped: `float32(normalizedValue*steps) + 0.5`. | `internal/codegen/golang/functions.go`, `internal/codegen/golang/flat.go` | `TestNoContractionInEmittedCode` requires the `float32(` wrap on every emitted quantize product |
| Rust | rustc NEVER contracts; a fused product requires an explicit `f32::mul_add`. Nothing to turn off; the rule is that no emitter may reach for it. | `internal/codegen/rust`, `internal/codegen/rusttable` | `TestNoFusedMultiplyAddIntrinsics` greps the emitters for `mul_add` |
| Java | JLS 15.18.2 FORBIDS contraction, and since JDK 17 (JEP 306) strict semantics are the only semantics. Java is the one inlining leg that therefore needs NO wrap: it emits the bare `n * 200.0f + 0.5f` on purpose. A fused product would require `Math.fma`. | `--release 17` on the Java leg; `internal/codegen/java/functions.go` | `TestNoFusedMultiplyAddIntrinsics` greps for `Math.fma` — the absence IS the enforcement, so the grep is the whole gate |
| C# | RyuJIT does not contract; a fused product requires `Math.FusedMultiplyAdd`. The message codec additionally narrows through a `float`-typed local, the C and C++ twins' shape, in FIVE near-duplicate copies. | `internal/codegen/csharp`, `internal/codegen/cstable/{message,regionmessagewrite,messageread,regionmessage,retainmessage}.go` | `TestNoFusedMultiplyAddIntrinsics` greps for `FusedMultiplyAdd`; `TestNoContractionInEmittedCode` requires the `float`-typed local in all five copies |
| JavaScript | The language has no FMA and no `float` type. The discipline is the opposite problem — doubles must be narrowed — so EVERY intermediate is wrapped in `Math.fround`. | `internal/codegen/js/flat.go` | `TestNoContractionInEmittedCode` requires every emitted float32 intermediate inside `Math.fround(` |
| Dart | The VM and AOT compilers do not fuse; `double` is the only float type, so the float32 discipline is an explicit `_fround` helper on every intermediate. | `internal/codegen/dart/functions.go` | `TestNoContractionInEmittedCode` requires the `_fround` wrap; the Dart leg runs the pinned probe compiled AOT on arm64 |
| Elixir | BEAM floats are doubles and the VM does not fuse. float32 rounding is emulated by the Veltkamp/Dekker `fr/1` helper, applied to every intermediate. | `internal/codegen/elixir/functions.go` | `TestNoContractionInEmittedCode` requires the `fr(` wrap |

The flag and the source-level form are belt and braces on purpose: the flag
governs THIS tree's builds, and the named-`float`-object form governs the
CONSUMER's, whose flags are theirs. A generated header cannot reach into a
consumer's build; it can only be written so that no build can contract it.

CI needs every target toolchain for gates 3–5 and 7; that cost is accepted —
it is the product's central claim.

### 7.3 The corpus discipline

The examples corpus (`examples/`) is the language's proving ground: a
construct earns its place by a realistic example needing it, and a construct
no example needs is a construct the language may not need. The corpus must
always compile under the spec as written (`make check`), and it feeds the
golden pins of §7.2 — so the language's own examples are also the compiler's
first test suite.

### 7.4 schemafmt — the one style

gofmt's philosophy: one style, no options. Built as the parser's first
consumer, and run by `schema fmt`, which is the only command that writes a
schema file. Rules:

1. **Indent: 4 spaces** per block level, never tabs. **Braces are Allman**
   (§4.1): a multi-line block's `{` on its own line at the construct's
   depth, `else` alone between its braces; a one-line `{ ... }` group stays
   whole. One field or declaration per line.
2. **Alignment groups**: within a contiguous run of fields — broken by blank
   lines and by comment lines — pad names to align the type column, and pad
   past the longest DEFINITION (the type plus any `= default`) to align the
   `|` qualification column. The same rule aligns `=` in const runs. Single
   space minimum between columns. **A `///` DOC comment does not break the
   group** (§4.1): it belongs to the field under it rather than standing between two
   fields, and a run whose every field is documented would otherwise become a
   run of one-field groups and lose the column the rule exists to produce.
   Any other comment line breaks the group as before.
3. **Qualifications**: `| min = 0, max = MaxHealth` — a space each side of
   `|`, spaces around `=`, comma-space between entries. Valueless markers
   first, then valued keys — the PARSER accepts any order and this rule is
   where the one order comes from (§4.2). A trailing `//` comment survives
   after the section.
3b. **Migration is a one-shot mode, not the formatter's default**:
   `schema fmt -migrate` additionally accepts the two RETIRED spellings —
   the trailing `[ ... ]` attribute block (with its default after the
   attributes) and the `[<= N]` bound — and re-emits the file in the
   CURRENT canonical form: qualifiers moved right of `|`, the default moved
   before the pipe, a qualified declaration's body brace moved to the next
   line, then the ordinary rules above. An attribute block wrapped across
   lines (the old grammar allowed it) is refused with "unwrap the attribute
   block first" — a wrapped block can carry interior comments that have no
   home right of a pipe. Plain `schema fmt` refuses retired spellings
   exactly as the compiler does.
4. **Expressions**: single spaces around binary operators.
5. **Enums**: the variant list is one line while it fits (a qualified enum
   is two lines by grammar — the list is the measured unit); the wrap
   trigger is decided at the first multi-line instance (no line-length
   limit exists). **A list in which any variant carries a qualification or a
   doc comment is one variant per line with NO commas**, the separator being
   the newline (§4.1), and the `|` column is aligned across the list by rule
   2. That is the one form in which the comma and the pipe would compete for
   the end of a line, and the formatter answers it the same way at every
   list rather than mixing the two spellings inside one. **A MIXED list — a
   comma after one variant and a newline after the next — is accepted by the
   parser and normalized here**, to this rule when any variant is qualified
   or documented and to the comma form otherwise. Two separators is what
   admitting the newline cost, and the formatter is where that cost is paid
   rather than in a diagnostic.
6. **switch/case** (reserved for `switch`'s return): `case` at the same
   indent as its `switch`; a single-item case body inline after the label,
   bodies column-aligned across the cases of one switch; multi-item bodies on
   following lines, one level in.
7. **Blank lines**: collapsed to one; preserved as group separators.
8. **Comments preserved, never reflowed** — file-header block, section
   dividers (`// ---- name ----`), and doc comments stay attached to what
   they precede.
9. **schemafmt never reorders declarations** — a formatter formats; it does
   not move code. (Declaration order at file scope carries no wire meaning
   (§3.1), so the aspect layout (§2) stays a convention. The order of
   variants INSIDE an enum, a `flags` or a union is the opposite — declared
   order is the wire, an enum's ordinal and a `flags` bit position and a
   union's tag alike (§4.2, §4.8) — and that is the second reason a
   formatter never sorts.)

## 8. Repository layout

The public Go API is `compiler/` and `ir/`; everything under `internal/` is
implementation, with no compatibility promise (VERSIONING.md).

```
cmd/schema/            the CLI — a client of the public API, and nothing more
compiler/              PUBLIC: the driver — load, generate, format, and
                       the generator registration interface
ir/                    PUBLIC: the lowered form; the wire shape projection
                       (§3.1); the derived parameters the backends share
internal/scanner/      tokens, positions
internal/parser/       AST construction, error recovery
internal/ast/
internal/check/        resolver, constant folding, shape checks, dominance rule,
                       the protocol id
internal/format/       schemafmt
internal/codegen/      c/  cpp/  csharp/  dart/  elixir/  golang/  java/  js/
                       rust/ — registered on the driver through the public
                       generator interface; ctable/, cpptable/, cstable/,
                       darttable/, elixirtable/, gotable/, javatable/,
                       jstable/ and rusttable/ are the table emitters, one
                       per backend, all nine (SPEC-TABLES.md)
internal/fuzz/         compiler fuzzing (gate 6)
internal/publicapi/    the acceptance gate: an external module, public API only
examples/              the corpus — always compiles under this spec as written
examples128/           the fixed-point + 128-bit corpus
testdata/              golden generated source, golden ids, golden wire bytes
test/                  the per-language wire test legs
bench/                 the benchmarking program (see bench/BENCH-STANDARD.md)
notes/                 extracted runtime API contracts (design inputs, non-normative)
SPEC.md                this document
```

License: AGPL-3.0, with an explicit additional permission for generated
output — see LICENSE.

## 9. Open questions, gathered

Rows keep their numbers permanently — code and corpus cite them as `§9 qN`.
Every row to date is settled, deferred with its design banked, or discarded.

1. ~~Strings as byte strings~~ — settled: §4.7. One shape for
   `string`/`bytes`; every backend composes the wire from primitives; the
   UTF-8 encoding check and the interior-null check are generated-code
   validation in every target.
2. ~~Storage-type overrides~~ — settled by the integer family: storage is
   declared by the type name (`thrust int8 | min = 0, max = 100`); no
   override mechanism exists.
3. ~~Wide strings and relative integers~~ — wide strings are settled:
   `wstring(N)` is §4.12, its wire the classic `serialize_wstring` and its
   reader rules uniform across the nine targets. Relative integers are out
   of scope by decision (§4.10): the construct belongs to the delta layer,
   which is replicant's and serialize.pro's.
4. ~~`schemafmt` timing~~ — settled: built early, as the parser's first
   consumer; rules in §7.4.
5. ~~Doc comments~~ — settled: §4.1, and OPT-IN. A `///` block binds to the
   declaration, field, variant or arm below it and rides the IR verbatim into
   the `doc` descriptor column (SPEC-TABLES.md §8) and into line comments in
   the generated code; a plain `//` block above the same item stays an
   ordinary comment and reaches nothing. No id moves either way.
   `| doc = "..."` is refused by name (§4.11) — one text, one spelling.
6. ~~A root/packet marker~~ — discarded.
7. ~~Const expressions over enum counts~~ — settled: `E.Max` (§4.2);
   `const NumTeams = Team.Max`. `len(Team)` was declined: it has three
   plausible meanings, and every one sizes enum-indexed tables wrong under
   `max` headroom — the max is the true primitive.
8. ~~Platform-conditional constants~~ — settled: constants are
   platform-uniform (§4.2); platform-varying tuning stays in application
   code.
9. ~~Explicit enum variant values, flag enums, the separator~~ — settled:
   canonical enums only (§4.2) — explicit and sparse values are declined
   permanently; `flags` is supported (§4.2); variants are comma-separated.
10. ~~Sentinel-terminated collections~~ — deferred to the delta pass,
    which designs all three measured terminator idioms at once
    (bool-continuation, sentinel value, stop action). Readers are
    structurally oblivious to packet splits, so the language owes readers
    only a terminated-stream loop.
11. **Enum subranges, and the enum-as-index pattern** — deferred; the design
    is banked here for its return, likely with the claiming/index pass. Two
    kinds of enum-driven value are both required: an enum that can carry null
    (`None` — the plain `enum`), and an index over an enum's valid values
    that excludes it (laser index, missile index). The banked `enum_index`
    design: a type-level declaration; wire = value − 1 in [0, max − 1];
    storage keeps 0 as an unset sentinel that is never wire-legal. The v1
    consequence, stated plainly: a kind field is a plain `enum`, so a
    `[1, max]` create wire is not expressible — one unused wire value is
    spent per kind field. The derived-range rule stays normative (§4.2); the
    general `index(Set)` feature stays deferred with `index` informally
    reserved (§1).
12. ~~`int` → `int32`?~~ — settled by the integer family: bare `int` is
    gone; every integer names its width, and `int`/`uint` are reserved purely
    to give a helpful diagnostic.
13. ~~Interpolation policy~~ — settled: no interpolation generation;
    interpolation stays hand-written until a claiming pass assigns per-tag
    actions (§4.2, Tags).
14. ~~The replication-policy boundary~~ — discarded: no send-scheduling
    knobs (priority/TTL/coherence) exist in this architecture, and no policy
    attributes ever will. schema fully owns serialization; nothing else was
    ever in scope.
15. ~~The engine interop surface~~ / 16. ~~The engine adoption asks~~ —
    discarded; externally-derived interop and adoption material is out of
    this repo.
17. ~~The unsigned fixed-point spelling~~ — settled: the explicit
    `ufixed(I, F)` keyword (§4.3, §4.6), storage always manifest in the type
    name — the integer family's own int/uint precedent. Deriving signedness
    from `min >= 0` was declined because a field's storage type would
    silently change when a bound crossed zero.
consumer's build; it can only be written so that no build can contract it.

## The gate

`internal/codegen/fpcontract`, four tests, read as source so they cost nothing and hold on the platforms whose toolchain is not installed here:

- **`TestNoContractionFlagOnEveryCompile`** — every C/C++ compile line in `Makefile`, `make/` and `bench/` carries the flag. Logical lines are joined so a flag on a continuation counts, comments quoting flags are skipped, and the test fails if fewer than twenty compile lines are found: a drifted glob is a vacuous gate. Negative control run by hand — removing the flag from `make/c.mk:14` reds it by file and line.
- **`TestNoFusedMultiplyAddIntrinsics`** — no emitter writes `mul_add`, `Math.fma` or `FusedMultiplyAdd`. For Rust and Java the absence IS the enforcement, so this grep is the whole gate.
- **`TestNoContractionInEmittedCode`** — the per-leg wrap is still there: `float32(`, `Math.fround(`, `_fround(`, `fr(`, the named `float scaled` object in C, C++ and all **five** C# message-codec copies, and both `FORCE_ROUND` barriers.
- **`TestTheDeterminismCaseStillDiscriminates`** — the witnesses must still disagree fused vs unfused, or the pinned vectors have quietly stopped proving anything.

## The determinism case

Go's comment named `0.005` over `[0, 10]` at `0.01`. Two more, found by an **exhaustive** sweep of every float32 in each range rather than by sampling:

| value | range @ resolution | plain | fused |
| --- | --- | --- | --- |
| `0.005` | `[0, 10]` @ `0.01` | 1 | 0 |
| `0.049999997` | `[0, 100]` @ `0.1` | 1 | 0 |
| `0.000499999966` | `[0, 1]` @ `0.001` | 1 | 0 |

Worth recording why these are pinned by hand: the sweep found only **two** divergent float32 in the whole of `[0, 10]` at `0.01`. Near-ties are rare enough that a sampling generator would essentially never find one.

## Not done, named

The three values are gated in Go, not yet in the nine-leg cross-language byte-identity gate. That gate's values are **hardcoded and duplicated in all nine drivers** (`test/main.cpp:619`, `test/c/main.c:457`, `test/go/main.go:249`, `test/rust/src/main.rs:341`, `test/cs/src/Program.cs:564`, `test/java/Main.java:260`, `test/js/main.mjs:459` **and** `:939`, `test/dart/main.dart:446`, `test/elixir/suite.exs:254`) with no shared vector file, and extending `CompressedProbe` means re-pinning `testdata/wire/compressed_probe.bin` across all nine. That is a clean follow-up, not something to rush inside this window. The existing probe already pins `0.005` and asserts the discrimination property in the C and C++ legs.

Also worth a separate look: `elixir/functions.go:2687` — `fr/1`'s Veltkamp/Dekker fast path is itself a multiply feeding a subtraction. Safe on the BEAM, which does not fuse, but it is the one place in the estate where the *rounding barrier* has a shape a contracting compiler could fuse.

## Gates

- `go test ./internal/codegen/... ./compiler/ -run 'Contract|FMA|Fused|Determin|Packet'` — green
- `gofmt -l .` — empty
- No regen sweep: no emitter output changed, so no golden moved. The `#pragma STDC FP_CONTRACT OFF` idea was dropped deliberately — the packet-form C and C++ emitters delegate the arithmetic to the runtime (`serialize_compressed_float_precomputed`) and emit no float expression at all, so a pragma in those headers would guard nothing, and the table-form headers that DO emit the arithmetic already carry the stronger `FORCE_ROUND` barrier. Verified the pragma works on clang (it beats `-ffp-contract=fast` on the command line) should a future emitter inline the fold the way Go and JS do.
- Not run here: the nine-leg `make test` legs. Out of window.

- 10f15baf, after **a DeepSeek read**: all four gates here could not go red. `TestNoContractionInEmittedCode` was a file-wide token grep (ten unrelated `float32(` in `golang/functions.go` kept it green with the wrap deleted from the product), `gotable/message.go` quantizes at runtime in Go and no test read it, the flag sweep globbed non-recursively and only at `-std=c` lines so `bench/tools/*.sh` went unscanned (22 compile lines scanned before, 204 now; three were unflagged and are flagged here), and the FMA gate labelled `Math.fma` in `internal/codegen/js` as a Java bug. Every gate is now anchored on the named function that emits its fold; each was proved red by removing one wrap, and one flag, by hand.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/fp-contract-off-nine` is kept at `03ccaaaebeb6dc743da42e43e2eb48a126a6dad5` (bundle `pr-963-03ccaaae.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/fp-contract-off-nine@03ccaaaebeb6dc743da42e43e2eb48a126a6dad5` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #964, opened 2026-09-11
packet wire: NaN patterns pinned on the narrowing legs, measure_* named, §6.1 true, one JS step count

gafferongames:
Four holes from a packet-wire spec read, none of which needed a decision.

**1. Signalling NaN is pinned on the three narrowing legs.** SPEC §4.3 says a
backend holding a `float32` in a wider cell carries the pattern bit for bit and
never through a float conversion. The only pin was on the TABLE wire in C++
(`testdata/wire/tables/floats_nan.bin`). On the PACKET wire the three
wider-cell legs are Dart, JS and Elixir — all three narrow in software, and
none had a NaN test (Elixir had one quiet NaN). Each leg now writes `Input`'s
five bare `float32` fields with the patterns a conversion moves — a signalling
NaN `0x7F800001`, a quiet NaN with payload `0x7FC00001`, a negative NaN
`0xFFC00000`, `-0.0`, and a finite control — and reads the bits back **off the
wire**, not through the reader: `Input`'s floats are the first 160 bits,
byte-aligned and LSB-first, so field *i* is bytes `[4i, 4i+4)` little-endian. A
reader round-trip alone would not catch a quieted sNaN, because the quieting
happens on the write. **All three survive**, both JS tiers included. Dart's
`_float32BitsFromDouble` comment now states the rule rather than reading as an
exception to it — the hardware narrowing of a non-NaN *is* the bit-for-bit
carry, exactly because float32 ⊂ float64.

**2. `measure_*` is named.** §6.1 said "There is no generated measure
function." Java, Dart and Elixir each emit one per type, and `docs/USAGE.md`
documents Elixir's. The three agree on semantics — exact wire **bits**, trusted
like the writer — so §6.1 now says which three emit it, what it returns, that
the bytes are `(measure + 7) / 8`, and that it is not part of the family
surface. `internal/check` claims `Measure<Name>`, `measure<Name>` and
`measure_<name>` in all nine targets, so a name three targets take cannot be
taken by a schema compiled for any target.

**3. §6.1's signatures are true.** It stated `Write(buffer, object) ->
bytesWritten` and `Read(buffer, object) -> ok/error`; no leg has either. The
nine actual shapes are now one table — C++ `bool`, C `int` 1/0, Go `error`,
Rust `Result`, C# `bool` + latched `Error`, JS `bool` + `stream.error`, Dart
`int` bytes, Java `int` bytes, Elixir `{:ok, value} | :error` with a binary out
of write — with the reason only three report a size: they own the buffer and
have no stream object to ask afterwards. §4.3's "`Write` returns exactly that"
of `ceil(bits/8)` was false on seven legs; it now says what is true, that this
is the wire's logical length, and points at the table.

**4. One JS step count — and it was a wire bug, reproduced.** The two
derivations are the same algorithm (`fround(delta/res)`, clamp to
`[1, 4294967040]`, `ceil`), so they already agree on the count; what diverged
was the **normative integer clamp after the floor** (`serialize.h
serialize_compressed_float_internal`, serialize#88, schema#109). `serialize.js`
applies it; `internal/codegen/js/flat.go` did not. Once the count reaches `2^23`
the float32 ulp at the top of the range is 1, so the rounded sum can land one
past the count. Reproduced at 8388609 steps with the value at max:
`fround(8388609 + 0.5)` is 8388610 under round-to-even, the floor keeps it, and
**the flat tier wrote `0x800002` where the runtime tier wrote `0x800001` — bytes
the runtime reader itself refuses.** The flat tier now clamps on the same side
of the floor, emitted only at counts `>= 2^23` so nothing in real use pays an
instruction. Regenerating `generated/js` and `generated/js-ludicrous` moves no
byte, which is why every gate was green. `compiler/packetjsstepclamp_test.go`
pins both halves: the clamp present in both write bodies at 8388609 steps,
absent at 1000, and the runtime tier still passing the raw triple so the two
never both fold it. Unifying further — having the runtime take a count instead
of the triple — would need a `serialize.js` API change in a sibling repo, and
is not needed now that the two agree byte for byte.

**Known red, by name: Dart and Java carry the same missing clamp.**
`internal/codegen/dart/functions.go` and `internal/codegen/java/functions.go`
fold the same count the flat JS tier does and omit the same clamp; Elixir's
`cf_quantize` has it. One line each, and each wants its own gate, so they are
left for a follow-up rather than pushed through this child's window.

Gates: `test-js` packet subset (`test/js/main.mjs`, both node modes) green;
`test-dart` packet subset (`test/dart/main.dart`, `dart analyze`, `dart format
--set-exit-if-changed`) green; `test-elixir` packet subset
(`test/elixir/main.exs`) green. `go test ./internal/check/ ./compiler/ -run
'Claim|Doc|Name|Packet'` green; `go test ./internal/check/` green; `gofmt -l .`
empty. The full `make test-js test-dart test-elixir` legs were not run here —
they pull in the table accelerators, fuzzers and sabotage-based negative
controls, well past this child's window.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/packet-snan-docs-jssteps` is kept at `1f84e9d60bcb4ae9a29de4f0dfe4ec558f511f39` (bundle `pr-964-1f84e9d6.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/packet-snan-docs-jssteps@1f84e9d60bcb4ae9a29de4f0dfe4ec558f511f39` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #965, opened 2026-09-11
packet wire: write-side checks are debug only on every leg, and no write unwinds

gafferongames:
A nine-leg read of SPEC §5's tiers against the emitters found four write paths that survive a release build, against §5's own sentence: "No target panics and none throws: Elixir's raise is the only unwinding path in the nine."

### The four fixes

1. **Dart wstring throws → `assert`.** `internal/codegen/dart/wstring.go` emitted `throw ArgumentError('wstring length')` and `throw ArgumentError('wstring null unit')` — alive without `--enable-asserts`, where every other Dart write contract is `assert(...)`. Both are now `assert(...)`.
2. **Java wstring throws → the dormant `checkWrite<Name>` predicate.** `internal/codegen/java/wstring.go` emitted `throw new IllegalArgumentException(...)` for the same two rules. The length range was already asserted in `checkWrite<Name>`, so the emitter now emits nothing; the interior-null rule joins it there (`internal/codegen/java/functions.go`, `emitCheckScalar`).
3. **Rust's write no longer panics in release.** `&value.text[..value.text_length as usize]` (`internal/codegen/rust/functions.go`) panics on a negative or oversized length once `debug_assert!` is compiled out. The used length is clamped into `[0, N]` once and that length is what rides the wire.
4. **C#'s write no longer throws in release.** `value.Text.AsSpan(0, value.TextLength)` (`internal/codegen/csharp/functions.go`) throws `ArgumentOutOfRangeException` once `Debug.Assert` is gone. Same clamp, same single use for both the length field and the payload slice.

Fixes 1 and 2 would have opened the same hole they closed: the Dart `Uint16List(N)` and the Java `char[N]` wstring buffers hold exactly N units, so a length of N+1 would have become a `RangeError` / `ArrayIndexOutOfBoundsException` in a build without asserts. Both legs clamp too, so all four write the clamped length.

Also in scope: Rust and C# gained the write-side interior-null debug check for `string(N)` that cpp and c have carried in `serialize_assert`, so §4.7's "writes assert per §5" is now true on those legs.

### SPEC

`docs/SPEC.md` §5 gains the rule the clamp follows: a debug-only check obliges the release path to be total, so in a release build a write with an out-of-contract length writes the clamped length — the bytes of the clamped write, byte for byte, never a trap — while the debug assert on the declared contract is unchanged and fires first.

### Gates

- `go test ./internal/codegen/{dart,java,rust,csharp}/ ./compiler/ -run 'Write|Throw|Tier|Packet|Golden'` — green.
- New generator tests `internal/codegen/{dart,java}/write_throw_test.go` grep every emitted write body for `throw` and fail. **Red on the tip's emitters**, green here.
- Goldens re-pinned through `go test ./internal/goldens -update -run TestGolden`: `packet-wide/{dart,java,cs,rust}`, `rust/*`, `cs/*`, `wide/cs`. Every hunk read.
- `gofmt -l .` empty.
- `test/packet-wide/dart/main.dart` now probes the contracts only under `--enable-asserts` (the compiled-exe run of the same file has no asserts) and, without them, checks that the length-N+1 write equals the clamped write bit for bit.

### Not yet run in this window

The dart/java/rust/cs packet make gates (`test-dart`, `test-java`, `test-rust`, `test-cs` packet subsets) and the C# Release runtime test of the clamp were not run here — toolchain builds exceed the window. Draft until they are green.

Follow-up in `fb344666`: a DeepSeek read of this PR against SPEC §5 found the fix half-applied — the Rust and C# `wstring(N)` writers still unwound in release, and C and C++ clamped nowhere for `string(N)`, `bytes(N)` or `wstring(N)`. All four now clamp the used length into `[0, N]`, with a generator test per leg (each red on the emitter this PR left behind), and §5's sentence now names all nine truthfully.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/packet-write-tiers` is kept at `8ea96d45b4b84312a154146acef4ad0606dc7165` (bundle `pr-965-8ea96d45.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/packet-write-tiers@8ea96d45b4b84312a154146acef4ad0606dc7165` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #968, opened 2026-09-11
fixed form: the guard chain, nested unions without a bound

gafferongames:
**Stacks on #951 and must merge after it.** #951 landed the SECOND guard lane (guard2/arg2/argw2) for arms nested one deep and refused a third nesting by name. This replaces that lane.

GLENN'S RULING: *"We should support infinite nested unions, or we should not support them at all."* A lane per level is neither — it makes the admissible schemas depend on how many lanes an entry happens to carry. The choice here is **infinite**.

## The shape

A plan entry carries `guards` — a BYTE OFFSET into the plan's guard pool — and `gcount`, the chain's length; `gcount == 0` is an entry that runs unconditionally, which is nearly all of a plan. The pool holds `(guard, arg_lo, arg_hi, argw)` links, **outermost union first**, and the run tests them in order and skips the entry on the first mismatch — so a forged inner tag under an unselected outer arm lands nothing, at every depth.

The identity walk (`ir/fixedform.go`, the one walk both emitters lay down) and both compiled walks push one link per nested union on the way into an arm and pop it on the way out. A union's own TAG entry, and the compiled path's `None` const, carry the chain of the ENCLOSING unions — all the outer tags, each at its own width — because nothing about the byte that says which arm is live may be conditional on what it says. The only depth bound anywhere is the layout walk's 64 (`ir.TableFixedMaxDepth`), never a union count, and a chain that does not fit the pool refuses `plan_too_large` BY NAME (§5.9 #5) rather than drop a condition.

## The cost

**The row got smaller: 56 bytes to 28.** The chain replaced `(guard, arg, argw)` twice over — five lanes and two 64-bit ordinals gone, four bytes and a count in their place. In the pool a link is **16 bytes**, laid **once per arm** and not once per entry (every entry under one arm names the same bytes); the identity pool goes further and SHARES, so a chain that is already a run of links in the pool is not laid twice and an arm's chain and the chains of the arms nested inside it are the same outer bytes. A type with no union carries a one-link placeholder pool so the array is not zero-sized, with `GuardCount = 0`.

## Tests

`TestFixedNestedUnionArmAnswersToBothTags` now walks a tower of unions at depth 1, 2, 3 and 4 on the one identity walk both twins lay down: every chain is outermost-first, every link a distinct tag at its own width, the pool holds exactly the chain the entry points at, and the deepest chain is one link per level. **The three-deep case reds on #951** — the two-lane walk panics "three nested union tags on one entry" — and reads exactly here. `guard_chain_depth_case` on both runtimes is the runtime half: a chain of three rides when all three links hold, lands nothing when the OUTER tag is forged, lands nothing when the MIDDLE one is (the condition a two-lane shape dropped), and a fourth link costs one more pool link and nothing else.

## Gates

`go test ./ir/ ./compiler/ -run 'Paired|Twin|Cpp|Fixed|Guard'`, `make tables-fixedform`, `make tables-c-fixedform`, `make tables-c-versioning`, `make tables-fixed-twin` and its negative control, regen sweep with goldens re-pinned for the C and C++ roots, `gofmt -l .` empty.

Docs: §4.1 (the plan's shape and the chain), §5.2 (THE GUARD'S WIDTH, and PLAN/EMIT as a chain rather than a lane), §5.8 row 6 (landed, chain), §5.9 #49 (the ruling and the choice).


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/twin-guard-chain` is kept at `05fdc9034d100bf857a104d05ff7440b3be8a76b` (bundle `pr-968-05fdc903.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/twin-guard-chain@05fdc9034d100bf857a104d05ff7440b3be8a76b` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

Issue #27, opened 2026-08-14
spec: the round = nearest|up|down attribute is parsed, validated, and never affects a byte


rowan-claude:
Severity: **medium** · Kind: spec-gap · Call: ours-to-fix

## What is wrong

The `round = nearest|up|down` attribute is parsed, validated, and then never affects a byte. Every surviving use in the codebase is a comment or a debug dump; the generated code emits no rounding at all (the projected field lives in the wire-integer domain in both the Shallow and Interpolate views, so the float-to-step conversion is the application's job).

## How it manifests

A schema author writing `round = up` gets `nearest` with no diagnostic — and the SPEC's stated rationale is that ceil is load-bearing ("any positive health must quantize alive"), so the silent default is exactly the wrong one for the case the attribute was added for. SPEC.md:1324 does mark `round` PROPOSED, but the checker accepts it today without saying so.

## Evidence

> VERIFIED READ. SPEC.md:1385-1386 — 'The optional `round = nearest|up|down` key (default `nearest`, half away from zero) selects the write rounding'. Accepted and validated: internal/check/check.go:495, :1161, :1213. Every surviving use is a comment or a debug dump: `grep -rn '\.Round\b' internal/` returns only internal/codegen/golang/golang.go:514-515 and :563-564, rust/rust.go:670-671 and :718-719, cpp/cpp.go, csharp/csharp.go (all building a `note` string) and internal/ir/projection.go:188-189. Live example: examples/Objects.schema:43 declares `health float32 [interpolate, min = 0, max = MaxHealth, resolution = 1, round = up]`; generated/cpp/Objects.h:150 says 'round up' in a comment and QuantizeShip emits nothing but `output.health = input.health;`.

## Fix direction

Either implement it, or state in §4.8 that `round` is advisory metadata carried to the generated comment for the application to honour — and say so in a diagnostic at the point of acceptance, not only in the spec.

Not blocked, but adjacent: DECISION #17 concerns the default rounding rule's identity (SPEC.md:1385-1386 is a shared receipt); settling that fork first would avoid documenting `round`'s default twice.

---
Part of the 2026-08-15 estate known-issues register (index issue on mas-bandwidth/schema).

PR #969, opened 2026-09-11
fixed form: the ordinal op lands the raw value, the bounds pass counts, on every leg

gafferongames:
**RULING (Rowan, final, §5.9 #27).** The place is kept and made possible. A pass over STORAGE cannot tell a forged `None` from a real one — an unset union writes that same zero lawfully on every clean record — so a counter in the pass needs **the raw value to survive the loop**:

- the `ordinal` op **lands the raw value**: remapped through the writer's table while the raw is inside it, and **the raw itself, unremapped, once it is past the writer's variant count**;
- the union tag's `None` entry **lands the writer's raw tag**, not a zero — the arms overwrite it under the writer's own tag, so the raw stands exactly where the tag names no arm;
- the **bounds pass** (§4.6, §5.3 step 11, over storage) clamps any ordinal or tag past **this reader's** extent to `None` and counts `clamped`, on both plans;
- **no op counts and no op clamps.**

The lock is what makes landing the raw safe: widths only grow (§5.1's LADDER), so a raw the writer's storage could hold fits the storage this reader declared.

### Per-leg state

| leg | state |
|---|---|
| docs | **GREEN** — FIXED-FORM-ALGORITHM §5.2 lane rows, §4.5 scatter rows (`ordinal`, `const`), §4.6, §5.4's two counter rows, §5.9 #27 restated with the reason; SPEC-TABLES §3.4's op table and the bounds-pass paragraph |
| cpp (reference) | **GREEN, builds** — the op lands raw and no longer counts; the `const` op reads the record under `meta = kTableFixedConstRawTag`; the **`dstsize` double meaning is removed** (the writer's arm count is gone — the op has nothing left to clamp), so `size` is the source's width and `dstsize` the destination's, the pair `ordinal` spells; the prefill's `lands` row for `const` follows |
| c (twin) | **GREEN, builds** — paired with cpp line for line, same lanes, same `const` op. The C leg keeps its owed 32-bit ordinal temporary (§5.8 row 7), untouched |
| go | **ordinal op landed**; `fixedguard_test.go` re-pinned off the op per #27 |
| rust | **ordinal op landed**; the count is gone from the op |
| java | **ordinal op landed**; the count is gone from the op |
| cs | **ordinal op landed**; the count is gone from the op. Merge order: **PR #941 first** — this branch is cut from the tip and the cstable site at ~1161 is the tip's, not #941's; whoever merges second re-applies the same three-line shape |
| dart | **ordinal op landed** (it counted nothing before, it landed a zero — the landing was the fix) |
| js | **ordinal op landed** — the op is in `jstable/fixedruntime.go`, not `fixedjs.go`; it counted nothing before either |
| elixir | **ordinal op landed**; both `ordinal/3` clauses return the raw |

### NOT DONE — owed, named here rather than left unsaid

- **Gates.** Pushed at the deadline on a green `gofmt -l .` (silent) and a green `go build ./...`. The `make` fixed-form gates per leg, `TestPairedHeadersAgree`, the compiler/codegen `Fixed|Ordinal|Clamp|Paired` run, and **regen + goldens have NOT been run** — the `const` entry's lanes moved, so **goldens and the paired headers will need regenerating** and the twin map re-checked. This is a DRAFT for exactly that reason.
- **The union-tag half is the reference's only.** `cpp` + `c` land the writer's raw tag on the `None` entry; the other eight legs' `None`/const entries still land a zero, so a forged tag on their compiled plans still counts nothing. Each needs the `meta`/`dstsize` pair its plan lanes can express, or to say it cannot.
- **Tests per leg were not written.** The two fixtures the ruling asks for — a forged ordinal past the writer's count on BOTH plans (`None`, `clamped == 1`), and a forged union tag on the compiled plan — are owed on every leg; only go's existing pin was moved.
- **A band the pass cannot see, for Rowan.** Where the reader knows MORE variants than the writer, a raw between the writer's count and the reader's extent now lands as one of the reader's own variants and the pass counts nothing — it is in range for the reader. That follows from the ruling as written (the pass tests the READER's extent); flagging it because it is the one case where landing the raw is weaker than remapping to `None`.
- Two stale leg comments found in passing and left alone: `rusttable/fixedruntime.go` ~1339 (a 256th variant "remapped to None") and `darttable/fixeddart.go` ~833 ("where the plan's own remap lands it too"). Also: `darttable/fixeddart.go:835` clamps the bounds pass on `len(r.Variants)` where js uses headroom-aware `r.Max` — it matters more now that the raw reaches the pass. And the C# `Ordinal` op reads sizes 1/2/4 only, so a size-8 ordinal lands 0.

---

## CORRECTION (Rowan, 2026-09-11) — THE PASS CLAMPS AGAINST THE WRITER'S COUNT, OUT OF THE PLAN

`7c168d32`. The band flagged above is **closed**, and the ruling is amended where it said the pass tests "this
reader's extent":

> **the bounds pass clamps an ordinal — and a union tag — against the WRITER's variant count (arm count) CARRIED BY
> THE PLAN for that entry.** On the identity plan the writer is the reader, so the two bounds coincide.

§5.2 already said the plan carries it: *"the plan also carries the WRITER's bounds for the hostile pass (bill
§12.5): that peer's count bound, variant count, arm count and range."* **No new lane was needed:**

- an **`ordinal`**'s remap table's **LENGTH WORD** is the writer's variant count (`table[0]`);
- the raw-tag **`None` const's `aux`** is the writer's **ARM count** — the lane the constant vacated when the
  entry began landing the record's own tag, which is the fact that needed it;
- the raw the pass judges is read back **OUT OF THE RECORD**, not out of storage: storage holds the **REMAPPED**
  ordinal, the reader's own number for the writer's variant, which is lawfully past the writer's count whenever a
  variant moved or was appended;
- a **GUARDED** entry is re-tested against the writer's tag exactly as the loop tested it, so an arm that did not
  run is never held to a bound it never landed.

**The identity plan is UNAFFECTED, and that is not an exception** to "the same pass on either plan": an identity
plan carries only `copy`, `count` and `text` (§5.4), its ordinals ride inside plain copy runs, and its writer IS
its reader — the generated straight-line pass over storage clamps at an extent that is the writer's too. So the
plan-carried half runs on the compiled plan only, at zero cost on the hot path.

This closes the **variant/arm half of §5.8 row 4**.

### Per-leg state of the CORRECTION (the band case)

| leg | band case | note |
|---|---|---|
| cpp (reference) | **GREEN** | `TableFixedClampPlanBounds` in `cpptable/fixedruntime.go`, called from the generated `Load` under `hash != <T>FixedHash`; the `None` const's `aux` carries the writer's arm count again |
| c (twin) | **GREEN** | `table_fixed_clamp_plan_bounds`, paired line for line — `tables-fixed-twin` green, `TestPairedHeadersAgree` green on **regenerated** `generated/bench/paired/{c,cpp}` |
| go | **NOT DONE** | the leg is green as #969 left it, but its pin is the READER-extent case (raw `9` past the reader's two variants); the band is owed |
| rust, java, cs, dart, js, elixir | **NOT DONE** | owed: the same writer-bounds pass, plus the union-tag half those legs already owed |

### Gates RUN (the ones #969 skipped)

| gate | result | time |
|---|---|---|
| `make tables-fixedform tables-c-fixedform` | **GREEN** — `0 RED, 0 failure(s)` on both, asan included | 1:14 (74s wall, 23s cpu) |
| `make tables-fixed-twin` (+ its negative control) | **GREEN** — "C and C++ fixed runtimes agree after the named twin map" | 20s |
| `gofmt -l .` | silent | — |
| `go build ./...` | green | — |
| regen | `rm -f bin/schema`; `generated/bench/paired/{c,cpp}` stamps dropped and re-generated, goldens re-pinned and the hunks read (the `const` entry's lanes moved) | — |
| `go test ./compiler/ ./internal/codegen/... -run 'Fixed\|Ordinal\|Clamp\|Paired'` | **one red, and it is NOT this change**: `rusttable TestFixedVersioning/array_bounded_grow` failed on a MISSING CORPUS FILE (`build/fixedform-corpus/old_array_bounded_grow.bin`) because the `make` gate was regenerating that corpus in the same tree at the same time. **Re-run serially: `ok github.com/mas-bandwidth/schema/v2/internal/codegen/rusttable 39.864s`** — the red was the race, not the change. Every other package green. Named here so it is not carried quietly (§5.9 #31); the lesson is that a `make` gate and a `go test` must not share one tree's corpus | 3:38 |
| the other seven legs' fixed-form / versioning gates | **NOT RUN** — out of window | — |

### Fixtures

- the **owed-8** probe was **RE-PINNED rather than the op bent back** (§5.9 #27's own rule): the width-8 ordinal of
  `2^32` lands **RAW** and the op counts nothing; the pass clamps it and counts **once** — the same proof of the
  64-bit temporary, moved to the place the contract puts it.
- a new **BAND probe** (`band_forged_ordinal_writer_bound_case`): writer **two** variants, reader **five**, raw
  **three** → the op lands it raw and counts `0`; the pass lands `None` and counts `1`. A raw **inside** the
  writer's set remaps to the reader's **fourth** ordinal and counts nothing, even though `4` is itself past the
  writer's count — which is why the raw is read from the RECORD. And the forged union **TAG** is the same row,
  through the raw-tag const's `aux`.
- docs: §4.6's new bullet, §5.4's bounds-pass row, §5.9 #27's restatement, §5.8 row 4 and the owed list,
  SPEC-TABLES §3.4, and `FIXED-FORM-VERSIONING-TESTS.md`'s `forged_ordinal_both_plans` gains the band case.

### RANGES from the lock: **OWED, and it is not small**

`cpptable` has `TableFixedKnownRange` and fills it on the **FIXTURE** path only — `collectKnownRanges` reads the
peer's own schema UNIT. On the **LOCK** path it passes `nil`, and that cannot be fixed here: a
`lockfile.Lineage` entry carries `Wire`, `Layout` and `Record` and **no declared bounds at all**. Carrying the
writer's ranges from the lock needs the **lock format** to record a range per field first. Named as owed in
§5.8 row 4 and in the owed list, not faked.

### Merge order

**#941 (cs) first** — unchanged from above; this branch is cut from the tip.


gafferongames:
Deadline note, after the push: the Go leg's re-pinned fixture is GREEN — `TestFixedForgedOrdinalLandsRawAndTheBoundsPassClamps` passes, asserting the op lands the raw 9 and counts nothing, then `HostFixedClamp` clamps it to None and counts `clamped == 1`. The whole-read number did not change (1 on both lanes); what moved is where it is asserted (at the op: 1 -> 0). The other `gotable` failures in that run are environmental and pre-existing: `SCHEMA_REQUIRE_CORPUS` is set and `build/fixedform-corpus` does not exist in a fresh clone, which needs `make`. Everything else in the NOT DONE list above stands.

rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/ordinal-lands-raw` is kept at `21071a6ab5cadef3723e6482cd84dce3621ae32a` (bundle `pr-969-21071a6a.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/ordinal-lands-raw@21071a6ab5cadef3723e6482cd84dce3621ae32a` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #971, opened 2026-09-11
fixed form: the hash pre-pass, so REFUSE is total on a file of many records

gafferongames:
## The ruling

§5.3's per-record hash check sat **inside the landing loop**, first thing. For a file of ONE record that is
the same answer. For a file of MANY it is not: a forged record at index 7 of 64 was found only after records
0 through 6 had been prefilled, run and bounded — the caller's storage was written, the counters had moved,
and **"REFUSE is total" plus §5.3's joint-answer table were false sentences about the very read they
describe** (C++ reference `internal/codegen/cpptable/fixedform.go` ~571; every leg mirrored it).

The fix is the **order**, not the check. A new **step 10b** walks the whole tail once, comparing each
record's leading eight bytes to the selected hash and refusing `no_layout` on the first that differs; **step
11 now has no hash check at all**. A file is ONE layout by construction, so the second walk only ever
changes the answer on a corrupt or hostile file.

## The doc step choice: 10b, and nothing is renumbered

The brief offered `8b` or a fold into step 9. It is **`10b`**, for three reasons:

- it needs `n`, which step 10 computes;
- it must sit **after** step 10 so `batch_too_large` still answers first — that is the precedence every leg
  already shipped, and moving it would be a second, unasked-for behaviour change;
- `10b` avoids renumbering exactly as `8b` would. **Steps 1 through 11 keep their names**, so PR #957's
  seams pass and its ~20 cross-references are untouched.

## Per leg

| leg | emitted load | gate | state |
|---|---|---|---|
| **C++ (the reference)** | step 10b pre-pass added, in-loop check removed | `make tables-fixedform` | **GREEN** — and the new fixture is **RED without the change** (3 assertions fail), green with it |
| **C (the twin)** | identical, `scan` declared C89-style at the top | `make tables-c-fixedform` | **GREEN** |
| **Go** | pre-pass over the tail slice, in-loop check removed | `make tables-go-fixed-form` | **GREEN** — layout and the whole 80915-byte file still match the C++ reference byte for byte |
| **Rust** | pre-pass over `rest`, `NoBlock` by name | `make tables-rust-fixedform` | **GREEN** |
| **Dart** | pre-pass over the `ByteData` view | `make tables-dart-fixed-form` | **GREEN** |
| **JS** | pre-pass on the two 32-bit halves | `make tables-js-fixed-form` | **GREEN** |
| **Java** | pre-pass over `8 + bodySize` | `make tables-java-fixedform` | **RED, AND RED AT BASE TOO** — see below |
| **C#** | pre-pass over a `ReadOnlySpan` walk | `make tables-cs-leg` | **GREEN** — 61s |
| **Elixir** | **unchanged** — `_fixed_split` already IS the pre-pass: it walks the whole tail and holds every record's hash before the `with` that runs the plan can open. Confirmed by reading, left alone, and its comments now name step 10b as what that split is | `make tables-elixir-fixed-form` | **GREEN** — "the Elixir leg matches the C++ reference" |

### Java is red on a check this PR does not touch

`test/java-fixedform/src/Main.java:693`, case 2b — "a header that names a layout it does not carry" — fails.
**It fails identically with `internal/codegen/javatable/fixedform.go` reverted to the base commit**, so it is
a pre-existing red on `fixed-table-form`'s tip (199fa5c3) and not a regression here. The Java emitter's own
change is in; the gate cannot go green until 2b is fixed, which belongs to another PR.

## Gates run

- `make tables-fixedform` — green, 21s
- `make tables-c-fixedform` — green, 4.3s
- `make tables-go-fixed-form` / `tables-rust-fixedform` 9.1s / `tables-dart-fixed-form` 6.5s /
  `tables-js-fixed-form` 2.5s / `tables-cs-leg` 61s / `tables-elixir-fixed-form` 2.7s — all green
- `make tables-java-fixedform` — red on case 2b, **red at base too**
- `go test ./compiler/ ./internal/codegen/... -run 'Fixed|NoLayout|Prepass|Paired'` — **green, 60s wall**
  (every leg's `fixedversioning_test.go` probe harness runs its leg's real runtime against the regenerated
  load, `TestPairedHeadersAgree` included)
- `gofmt -l .` — empty
- regen sweep: `rm -f bin/schema` before every build; goldens re-pinned and the hunks read

**One bench note, not a code problem:** the child kit's `env.sh` exports `SERIALIZE_RS` and `SERIALIZE_GO` as
absolute paths while `make/rust.mk` and the Go rules prepend `../../../`, which yields a doubled path and a
Cargo/go.mod that cannot resolve. The Rust gate needs `SERIALIZE_RS=../serialize.rs`; a polluted
`generated/bench/paired/go/go.mod` was reverted and is **not** in this PR.

## Tests

- `test/tables/versioning_lists.cpp` — new `refuse_writes_nothing_case()`: 64 records from the OLD build,
  **record 7's hash word inverted**, the header's hash untouched, read by the NEW build so a compiled plan
  with a **nonzero** fill (`Vec.w = 88`) is in hand, over storage poisoned `0x5A` in every byte of all 64
  rows. Asserts `n == -1`, `reason == no_layout`, `malformed` false, `layout_hash` untouched, every counter
  `0`, **and every poisoned byte still `0x5A`** — plus controls that record 0 and the header are good, so the
  row cannot pass vacuously.
- `test/tables/fixedform_dump.cpp` — writes the row's bytes as `nolayout_nested_append.bin`, the same
  64-record, record-7 file, so every leg has one fixture. The manifest is **67 rows and 67 files**.

## Still owed

Each other leg's probe harness (`internal/codegen/*/fixedversioning_test.go`) should gain the same
64-record/record-7 case against the new corpus file. Their existing harnesses all pass, and the corpus file
is now there for them; the seven per-language probe bodies did not fit this window.

## Docs

`FIXED-FORM-ALGORITHM.md` §5.3's pseudocode, its condition table, its "REFUSE is total" paragraph and §5.9
row 8; §5.4 gains the row saying no counter moves on any refusal by name; `SPEC-TABLES.md` §3.4's load
description and its `no_layout` registry row; `FIXED-FORM-VERSIONING-TESTS.md`'s `refuse_writes_nothing` row
becomes the 64-record, record-7 file with the poison asserted byte for byte.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/refuse-is-total-prepass` is kept at `7729b196494ff82f6062df6446303ebfcab45a3f` (bundle `pr-971-7729b196.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/refuse-is-total-prepass@7729b196494ff82f6062df6446303ebfcab45a3f` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #972, opened 2026-09-11
lock: the two-parent run at a merge commit

gafferongames:
**RULING (Rowan, final).** The bill's §8a.1/§11.3 name-subset clause is IMPLEMENTED AS A TWO-PARENT LOCK RUN, never as a run-time rule: reads already pair fields by wire id (algorithm §5.9 #41), so a merged layout reads both parents' files through the plans COMPILE lays down. What a merge needs is a lock run that knows it has two `old`s.

The flag's spelling, once per parent:

```
schema lock --parent <A.lock> --parent <B.lock>
```

It holds the merged declaration to EACH parent by wire id: every parent field present (deprecated in place allowed), kind unchanged, widen only (the ordinary monotone rules per field), and the merged record's NEW fields ANYWHERE in the order — position-free at a merge, since neither parent's layout can be a prefix of a merge that took both branches' appends. **Within ONE branch the strict prefix rule stays exactly as today.** The merged lock's lineage is both parents' entries concatenated in commit order, deduped by wire hash, then the merged layout's own entry; retirements from both parents stand, so the floor is the max of the two.

`--parent` is a flag and not a detection: the lock a parent committed lives in that parent's tree, so auto-detection would mean a `git show` per parent on every `schema lock`, in a compiler that otherwise never shells out. A merge nobody declared is held to the ordinary prefix rule, which is the safe answer.

## Landed

- `internal/lockfile/merge.go` — `Merge`, `diffAgainstParents`/`diffParent`, `diffTableAtMerge` (the one changed clause), `mergeLineageParents`, `carryRetiredParents`, beside `diff.go`'s single-parent `Diff`
- `compiler/lock.go` — `MergeSchemaLockAtMerge`
- `cmd/schema/main.go` — the repeatable `--parent` flag (`parentLocks`)
- docs: bill §8a.1 and §11.3 say "at a merge, through the two-parent lock run"; `FIXED-FORM-ALGORITHM.md` §5.1's "at a merge commit, run once per parent lock" made precise (and WIDENS' table clause); `FIXED-FORM-VERSIONING-TESTS.md`'s `lineage_merge` row written out as five halves

## The five rows

| row | the case | state |
|---|---|---|
| **(a) LOCK-ALLOWS** | A appends x, B appends y, M has both — in either interleaving → allowed against both parents; the merged lock holds all four fields and is current afterwards. Red half: the ordinary single-parent rule refuses that same merge against exactly ONE side, naming the order | green |
| **(b) LOCK-REFUSES** | the merge drops y → refused, naming **y** and the parent **B** | green |
| **(c) LOCK-REFUSES** | the merge changes **x**'s kind (`int32` → `float32`) → refused, naming x and the parent **A**; only the ORDER is position-free at a merge | green |
| **(d) THE RECORD** | the merged lineage holds A's entries, B's new one and the merged layout's own (shared base deduped by wire hash), the `lineage=` roll-up recomputed, the floor = max of the parents', A's retirement and its reason standing | green |
| **(e) NEW-READS-OLD** | build A, B, M; write under A and B; read both under M — each branch's own field exact, the other branch's its declared default, all counters zero. Self-contained Go probe, no C++ corpus | green |

Plus (e)'s companion: a single-branch insertion with no `--parent` is the refusal it has always been (`TestLockSingleBranchInsertionStillRefused`).

Tests: `TestLockLineageMerge*` and `TestLockSingleBranchInsertionStillRefused` in `internal/lockfile/lockrows_test.go`; `TestFixedRedTeamMergeReadsBothParents` in `internal/codegen/gotable/fixedredteam_test.go`.

A value list is NOT loosened at a merge: a variant, a flags bit and an arm have no id on the wire — the stored number IS the position — so §5.1's prefix sentence stands for those and the run keeps refusing two branches that both append to one enum. Said in the code and in the doc.

## Gates

- `go test ./internal/lockfile/ ./compiler/ ./internal/codegen/gotable/ -run 'Lock|Merge|Lineage|Fixed'` — green
- `go test ./cmd/...` — green; `go vet` clean; `gofmt -l .` prints nothing
- CLI end to end: lock A, lock B, merge, second run idempotent, `schema check` current on the merged tree

Note: under `SCHEMA_REQUIRE_CORPUS=1` the four pre-existing `TestFixedVersioning*` suites fail in a fresh clone because `build/fixedform-corpus` has not been built (`make tables-fixedform-corpus`); unrelated to this change, and the new read probe needs no corpus.


gafferongames:
**CI triage — every red row is pre-existing infrastructure, none of them this change.**

| row | state | why |
|---|---|---|
| `fixed form + versioning (go)` | **pass** | the leg this change's read probe lives on |
| `fixed form + versioning` (dart, js, cs, elixir) | pass | |
| `fixed form + versioning` (rust, c, cpp, java) | fail | also red on #973, which touches none of this code; #974 ("every leg row carries its toolchain and the corpus") is the fix |
| `go-test-touched` | fail | 51 failures, **not one of them a row of this PR** — all `TestCpp*` / generated-runtime cases failing on `replacement directory /home/runner/work/schema/serialize.go does not exist`, i.e. the sibling runtime checkouts are not on the runner. Identically red on #973 |
| `plan`, `block-gate`, `lint-fast`, `caa` | pass | |

Locally, with the sibling checkouts present: `go test ./internal/lockfile/ ./compiler/ ./internal/codegen/gotable/ -run 'Lock|Merge|Lineage|Fixed'` and `go test ./cmd/...` are green, `go vet` is clean and `gofmt -l .` prints nothing. Holding as draft until the leg rows are green on their own.

rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/lock-two-parent-merge` is kept at `c52b5267902a49948539fa8c7cb42f874d60d9db` (bundle `pr-972-c52b5267.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/lock-two-parent-merge@c52b5267902a49948539fa8c7cb42f874d60d9db` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #977, opened 2026-09-11
ci-fast: an empty plan skips the leg rows; go-test-touched has its siblings

gafferongames:
Two fast-lane defects seen on real pull-request runs. One file, `.github/workflows/ci-fast.yml`, and no per-leg case entry is touched — **the sibling branch `rowan/ci-fast-rust-row` is fixing the leg-gate rows' toolchain setup in the same file**, so the two diffs are deliberately disjoint and either order of merge is a clean rebase.

## 1. An empty selection dispatched a leg row with no leg (PR #933)

A bench-only diff touches no leg, and the matrix still ran one row — `fixed form + versioning ()` — which died on this file's own refusal:

```
::error::no fixed-form gate is registered for leg ''
```

The cause is a platform difference in `xargs`, not in the selection logic: **POSIX and GNU xargs run the command once even with no input** (BSD/macOS xargs does not), so `printf '' | xargs -n1` prints one blank line on the ubuntu runner, `sort -u` keeps it, and the awk that builds the array emits a single empty element. Verified on x86-64 Linux before the change:

```
$ printf "%s" "" | xargs -n1 | od -c
0000000  \n
$ printf "%s" "" | xargs -n1 | sort -u | awk 'BEGIN{printf "["} ...'
[""]
```

The splitting is now an unquoted expansion with `awk NF` dropping blank records — with no input at all, awk's `BEGIN`/`END` still print `[]`, which is what `leg-gate`'s existing `if: needs.plan.outputs.legs != '[]'` tests. The same `xargs` is gone from the packages trim for the same reason. And an empty element that survives anyway is now a named error **in `plan`**, not a mysterious matrix row nine steps later.

### The plan step run locally, verbatim, over a faked `origin/base...HEAD`

The step's shell was extracted from the YAML unmodified and run in a throwaway worktree against a fake base ref (only `git fetch` is shimmed, since there is no remote to talk to locally):

| case | files | `broad` | `packages` | `legs` |
| --- | --- | --- | --- | --- |
| bench-only | `bench/run.sh`, `bench/BENCH-STANDARD.md` | `no` | *(empty)* | `[]` |
| bench-only, with Go | `bench/paired/main.go`, `bench/run.sh` | `no` | `./bench/paired` | `[]` |
| one leg | `internal/codegen/rusttable/fixedform.go` | `no` | `./internal/codegen/rusttable` | `["rust"]` |
| `ir/` broad | `ir/unit.go` | `yes` | `./...` | `["c","cpp","cs","dart","elixir","go","java","js","rust"]` |

So a bench diff now runs `lint-fast`, `block-gate` and `go-test-touched` and **no leg row**; a bench diff with no Go file in it skips `go-test-touched` too, because there is no package to test. `ir/` still gets all nine legs — the conservative direction is unchanged.

## 2. `go-test-touched` ran without the siblings the probes link (PR #955)

```
--- FAIL: TestArrayLengthCacheWire ...
FAIL probe [setup failed]
```

The emitter tests are not unit tests: `internal/codegen/gotable`'s helper writes a probe module into a temp dir whose `go.mod` `replace`s the runtime to `../../../../serialize.go`, then compiles it (`internal/codegen/ctable`, `cstable` and `rust` do the same with `serialize.c`, `serialize.cs`, `serialize.rs`, and `compiler/` and `bench/paired` reach the same directory). `ci-full.yml`'s `go-test` clones those siblings before it runs a single test for exactly this reason; the fast lane did not.

It does now: five shallow clones in parallel, at the pins this file's `env:` block already carries. **All five, always, and measured** — about 1 s locally for five ~1 MB repositories against this lane's own 25-35 s job floor — so selecting the subset a package happens to need would cost more reading than it saves running. `../serialize` is in the list on its own account: `internal/fuzz`'s compile check *skips itself* without it, which is the quiet green `ci-full.yml`'s block names.

`rowan/unit-tests-two-minutes` is not on origin yet, so the `SCHEMA_SLOW` alternative was not available; at 1-3 s the clones are well inside the 30 s the alternative was meant to beat.

## Gates

* `go test ./internal/ci/` — green (`ok ... 6.356s`), the gate that reads every `uses:` pin in this file.
* `gofmt -l .` — clean; no Go source changed.
* The four plan runs above.

Note that this diff touches `.github/workflows/ci-fast.yml`, which **is** one of the broad paths, so this PR's own fast-lane run takes the broad selection: all nine leg rows and `go test -short ./...` — which is the widest possible exercise of the sibling fix.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/ci-fast-plan-fixes` is kept at `818d20f417ed0082534e255cf3bd7f762b6965f6` (bundle `pr-977-818d20f4.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/ci-fast-plan-fixes@818d20f417ed0082534e255cf3bd7f762b6965f6` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #979, opened 2026-09-11
USAGE: the fixed table, version to version, as a stranger walks it

gafferongames:
`docs/USAGE.md` gains one section for **users** of the fixed table, written as a walk a stranger can follow with only this page. It sits in the Tables chapter, right after *The tables baseline* and before *Going wide*, and is linked from the Tables sub-list in the table of contents.

Every command in the walk was run against a scratch unit; every refusal sentence is copied from what the tool printed.

### The section's headings

`### The fixed table, version to version`

1. **Declare one, generate, write a file.** — the `fixed` keyword, the `ShipConfigFixed*` surface, a real file, and `schema check <file>` answering `FORM 3 the fixed form`.
2. **Commit the lock.** — `schema lock`, its idempotence, and one paragraph on what the lineage records: the layout bytes, the definitions digest, the record body size, and the wire hash that **is** the version. Plus what `layout=` and `lineage=` are for.
3. **Widen lawfully.** — six widenings in one edit (a field appended with a default, an enum variant appended, an array bound grown, a string grown, a union arm appended, a field deprecated in place), read off the lock's own diff, and the teaching point that all six are **one** lineage entry. Then **And six it refuses, each by name** — reorder, remove, narrow, change a kind, change a default, undeprecate, verbatim.
4. **Ship it: the reader goes out before the writer.** — one plan per lineage entry as static data, the deploy rule, a new reader on an old file, and what the report shows (`n`, `widened`, `clamped`, `unknown`, `kind_mismatch`). **And a refusal is not one of those events** — REFUSE is total.
5. **An OLD reader meets a NEW file.** — `layout_newer` carrying the file's hash and nothing else, and what an operator does with the hash: look it up on a `lineage wire=` line in the lock of the generation that wrote it.
6. **Retire a layout when nothing writes it any more.** — `schema lock --retire <Table>@<hash> --reason`, nothing removed, the floor, `layout_unsupported` to a client, and the mirrored deploy order. Plus the `--reason` and unknown-hash refusals, and the whole-table form.
7. **The refusal names, and what to do about each.** — one user-facing table: `previous_form`, `message_form_as_file`, `newer_form`, `layout_newer`, `layout_unsupported`, `layout_malformed`, `no_layout`, `batch_too_large`, `plan_too_large`, `layout_record_too_large`, each with *what happened* and *what to do*; then `malformed` set apart, because it is a report flag and not a refusal.
8. **The two rules that keep it safe.** — newer reads older, never forward; widening with default values is what saves us.

### Gates

- `go test ./compiler/ -run DocExamples` — green. `compiler/docs_test.go` now lists the section's declaration block (`fixed table ShipConfig`), so the example cannot go stale in either direction.
- `gofmt -l .` — silent.

### Where the tool and the docs disagree

Reported rather than smoothed; the section follows the tool, and none of these is written into the page as settled.

1. **`SPEC-TABLES.md` §21.2 vs the tool and `FIXED-FORM-ALGORITHM.md` §5.4 — a deprecated field.** §21.2 says "A field the reader deprecated: dropped, counted once under `unknown` per plan." The tool does neither: a deprecated field keeps its id in the lock, so the compiled plan matches it by id and **copies** the slot, and `unknown` stays `0`. Demonstrated — a v1 writer wrote `legacy = 77`, a v1.2 reader that deprecated `legacy` read back `legacy=77` with `unknown=0`. §5.4 agrees with the tool ("once per writer field no reader field names"), so §21.2 is the outlier.
2. **`FIXED-FORM-ALGORITHM.md` §5.1 names one failure where the tool prints three.** §5.1's `WIDENS` has `or FAIL "kind changed"`. The tool distinguishes `narrowed (int32 -> int16)`, `ladder changed (float32 -> int32)` and `kind changed (float32 -> string)` on the same line shape.
3. **`schema lock --retire` is not idempotent and contradicts its own API doc.** `compiler/lock.go`'s `RetireSchemaLock` documents `rewrote == false` for an already-retired target, and `cmd/schema/main.go` has the `"%s is already retired in %s"` branch for it. Retiring the same hash a second time with a different `--reason` silently **overwrites the reason** and reports `retired ...` again.
4. **`--all-below` does not exist.** No flag, test, or doc mentions it anywhere in the tree. Retirement is one entry per command; the floor sits above the newest retired entry. The section says so.
5. **`text_ill_formed` does not exist** as a refusal name — nothing in the tree names it. The refusal vocabulary is the `TableMessageReason` enum in `internal/codegen/cpptable/cpptable.go`; the section's table uses that vocabulary and adds `layout_record_too_large`, which the generated readers do raise.
6. **`malformed` is not a refusal name.** It is a `bool` on `TableReport`, set on framing damage. The section presents it that way rather than as a row of the refusal table.
7. **A payload-free union arm in a fixed closure silently drops the whole fixed surface** in the Go leg — no error from `schema generate`, and `<T>FixedSave` / `<T>FixedLoad` / `<T>FixedHash` are simply absent. Consistent with SPEC-TABLES §11 making payload-free table-side arms C++ only, but the failure mode is silence where every other refused table-side use is a named refusal. Not a doc issue; flagging it as a compiler one.

Also noted, not a disagreement: the lock's `lineage ... record=N` is the record **body** size (66 for the v1.1 layout) where the generated `TableFixedKnownLayout.Record` is the whole record, body plus the repeated 8-byte hash (74). The section says "record body size", which is the lock's own word.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/usage-fixed-table-walk` is kept at `8f265c57b86b510c68d162ca48c3e9375a66abb7` (bundle `pr-979-8f265c57.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/usage-fixed-table-walk@8f265c57b86b510c68d162ca48c3e9375a66abb7` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #974, opened 2026-09-11
ci-fast: every leg row carries its toolchain and the corpus

gafferongames:
## The finding: (a) — the lane's own bug, not PR #949

CI fast run [34600421804](https://github.com/mas-bandwidth/schema/actions/runs/34600421804) on `rowan/rust-union-prefill-writer-cap`: the job `fixed form + versioning (rust)` failed at

```
go: go.mod requires go >= 1.26 (running go 1.24.13; GOTOOLCHAIN=local)
make: *** [make/rust.mk:309: tables-rust-versioning] Error 1
```

**The `::error::no fixed-form gate is registered for leg 'rust'` line in the step log is a red herring** — it is the shell echoing the case statement's unreachable `*)` arm while it prints the recipe, not a gate that ran. The case statement has always carried a rust entry.

**The real cause is the override beside it.** The row ran

```
make tables-rust-fixedform tables-rust-versioning RUSTUP_BIN=/usr/bin
```

`RUSTUP_BIN=/usr/bin` is copied out of `test/conformance/rust/ci.json`, where it is correct: every target that row names is a `cargo build`, and the image's cargo is the one it wants. But make prepends `$(RUSTUP_BIN)` to PATH around the whole recipe, and `tables-rust-versioning` (make/rust.mk:309) is **a Go test that shells out to cargo**. So `/usr/bin` went in front of the Go 1.26 `actions/setup-go` had just installed, the image's own go 1.24.13 answered, and the gate died before asserting one row of §5.

PR #949 is not implicated by this job.

## Per-leg audit of leg-gate's case against ci-full.yml and the conformance registry

| leg | gates it runs | toolchain in the row | corpus | verdict |
|---|---|---|---|---|
| cpp | `tables-fixedform` | `../serialize` (cloned every row), the image's g++ | target depends on `build/fixedform-corpus/.stamp` | correct, no change |
| go | `tables-go-fixed-form tables-go-versioning` | `../serialize` + `../serialize.go`, setup-go | `tables-go-versioning: tables-fixedform-corpus` | correct, no change |
| c | `tables-c-fixedform tables-c-versioning` | `../serialize.c`, the image's cc | `tables-c-versioning: tables-fixedform-corpus` | correct, no change |
| **rust** | `tables-rust-fixedform tables-rust-versioning` | `../serialize.rs`, `dtolnay/rust-toolchain` | both targets depend on the corpus | **FIXED: `RUSTUP_BIN=/usr/bin` dropped** |
[...]

PR #980, opened 2026-09-11
ci: nothing runs on a PR; the fast lane on push to our branches, the full lane where a merge lands

gafferongames:
**The owner's ruling, 2026-09-11, verbatim:**

> "Here is another option to speed up, check in to branches without PRs. Only run tests when merging into main, local runners elsewhere."

This pull request is the shape of that ruling. It also carries the merges of #975 (the fast lane on the six studio self-hosted runners) and #974 (the rust row stops shadowing its own Go toolchain), resolved in favour of both.

## What changes

**The pull request stays the READ surface and starts nothing.** The diff, the body and the owed lists live here; Emma, Johnny and DeepSeek read them. No workflow of ours is triggered by a `pull_request` event any more, so an open pull request costs zero hosted minutes and zero queue.

**`ci-fast.yml` — the push to one of our branches.** Trigger: `push` to `rowan/**`, `johnny/**`, `emma/**`, `freddy/**`, `stella/**`, plus `workflow_dispatch`. No `pull_request` trigger at all. Every job stays `runs-on: [self-hosted, studio]` and never a hosted runner, so a push to a branch of ours gets a second-opinion fast run on our own hardware for free. The fork guard becomes structural rather than a job-level `if:` — the only trigger is a push to a branch *in this repository*, so no fork's code can reach the pool.

**`ci-full.yml` — where a merge lands.**

| event | runners |
| --- | --- |
| push to `fixed-table-form` | the studio pool (self-hosted) |
| push to `main` | hosted `ubuntu-latest` — the one hosted run per main merge |
| nightly schedule | hosted |
| `gh workflow run ci-full.yml --ref <branch>` | the studio pool |
| `full-ci` label | hosted, and now on `labeled` **only** — an explicit ask, not a plain pull-request trigger |

Three rows cannot move and say so in place: `big-endian` (s390x cross gcc and `qemu-user` at pinned Ubuntu versions), `windows` and `msvc`. The pool is macOS and has none of those.

Two portability fixes for the pool: `$(nproc)` becomes `$(getconf _NPROCESSORS_ONLN)` (three rows), and `erlef/setup-beam` — which has no macOS support — is gated on `runner.environment == 'github-hosted'`, where `dist/` already carries the BEAM on PATH.

**The merge condition** is the local gate (`merge-lane.sh run --local-gates`) plus a read. The lane already supports a pull request with NO hosted checks at all: `pr_checks` reports `g0/p0/r0` when `gh pr checks` returns nothing, and `one_pass`'s `(( C_PENDING > 0 || C_GREEN == 0 ))` branch does not wait when a green gate exists for the PR's *current* head. No lane change was needed; a hosted RED still stops a PR, as before.

**Not touched:** `cla.yml` (the Contributor Assignment Agreement, `pull_request_target`). It is not a test and not ours to change — it is the one check that still appears on a pull request.

**`docs/CONTRIBUTING.md`** states the loop in five lines — local gate, push the branch, open the PR as the record, a read from another line, the lane merges on local green + read, the full lane runs on the merge — and says hosted CI runs only on main and nightly, citing the ruling with its date.

## Gates

- `go test ./internal/ci/` green (pins, and the Go-shadowing gate).
- `gofmt -l .` prints nothing.
- Measured wall times are in the comment below.

---

## The studio pool's three defects, found by #980's own first dispatch

The first full-lane dispatch on the pool (run 34603066277) went **52 of 72 jobs
red**, and the fast lane's `go-test-touched` went red with it (run 34603030435).
None of the three causes was the diff; all three were the pool.

### (1) The sibling clones were not idempotent on a warm workspace

A self-hosted runner's workspace **parent** persists between runs, and the
sibling runtimes the Makefiles reach for as `../serialize*` live there.
`actions/checkout` cleans the repository directory; it does not touch the
parent. So the second run of any job that clones a sibling dies by name:

```
fatal: destination path 'serialize' already exists and is not an empty directory
```

Every clone step in **both** workflows now `rm -rf` the target and re-clones at
the pinned tag — ten sites in `ci-full.yml`, and the fast lane's `leg-gate`
already had it. The rust versioning row's `if [ ! -d ../serialize.rs ]` guard
goes with them: it did not fail, it **reused** whatever the previous run left,
which is how a stale sibling outlives a pin bump. The repository checkout itself
is untouched and still cleans on every run — nothing in either file sets
`clean: false`, and nothing may.

### (2) The cs rows — the pin was never the problem

The decision asked for was install-10.0.400 or move-the-pin-to-10.0.401.
Neither: `.github/dotnet-version` holds the **band** `10.0`, and the Studio has
`10.0.400` **and** `10.0.401` under `~/.dotnet`, both inside the band, both
complete, both working (`dotnet --version` under a `global.json` pinning either
answers). **The pin and the Studio already match.**

The defect is that `ci-full.yml`'s four `setup-dotnet` steps were **not gated to
hosted runners**, unlike `ci-fast.yml`'s. All six studio runners share one
`$HOME`, so the pool ran six concurrent installs of the same SDK into the same
`~/.dotnet`, and a job read one of them half-written:

```
A fatal error was encountered. The library 'libhostpolicy.dylib' required to
execute the application was not found in '/Users/glenn/.dotnet/sdk/10.0.401/'.
```

— every `TestFixedVersioning` row failing on a probe that never built. All four
steps are now `if: runner.environment == 'github-hosted'`, so the pool installs
nothing and uses the SDK already on its PATH at `~/.local/bin/dotnet`. No SDK
was installed and the pin did not move; matching the pin is what this does.

### (3) `go-test-touched` had no sibling checkout at all

This is the `internal/codegen/gotable` FAIL at 190 s. A broad diff hands the job
`./...`, which reaches every leg's table codegen test, and those generate a probe
whose `go.mod` **replaces** `github.com/mas-bandwidth/serialize.go` with
`../serialize.go`. With no sibling the gate does not skip — it fails by name:

```
Probe.go:12:2: github.com/mas-bandwidth/serialize.go@v0.0.0: replacement
directory /Users/glenn/actions-runner-3/_work/schema/serialize.go does not exist
```

It now clones **all six** siblings, idempotently, for the same reason the
negative-control matrix clones all six: the job's package list is not known when
the step is written.

### (4) The java row — noted, not chased

`REFUSED BY NAME` on the java row is the tip's pre-§5 Java leg. **PR #920 carries
the fix.** Nothing here touches it, and it is the one row expected red until #920
lands.

### The pool's operating notes, in `docs/CONTRIBUTING.md`

A new **The studio pool** subsection under the gates: the six runners and their
directories (`~/actions-runner`, `~/actions-runner-2..6`), `./svc.sh stop|start`
per directory and the rule never to touch them during a benchmark window, the
warm-workspace rule with **both** ways it was earned (a bare clone fails by name;
a `[ ! -d ]` guard quietly reuses a stale sibling), why a toolchain install is
not a per-job step on a pool that shares one `$HOME`, and a table of where each
toolchain lives on the Studio.

### Gates

- `go test ./internal/ci/` green.
- `gofmt -l .` prints nothing.

### DIFF
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -410,6 +410,28 @@ jobs:
       # THE TWO GATES. The target names are the tree's own — the same ones
       # `make test` and ci-full.yml run — so a leg that renames a gate renames
       # it in one place and this row follows.
+      #
+      # AND THE OVERRIDES ARE THE ONES ci-full.yml PASSES, NOT THE ONES
+      # test/conformance/<leg>/ci.json PASSES, and the difference cost this
+      # lane its first red (run 34600421804, the rust row on PR #949): the
+      # rust row here was written with `RUSTUP_BIN=/usr/bin`, which is what the
+      # conformance registry hands the CONFORMANCE leg — correct there, where
+      # every target it names is a `cargo build`. But `tables-rust-versioning`
+      # is A GO TEST that shells out to cargo (make/rust.mk:309), and make
+      # prepends $(RUSTUP_BIN) to PATH around the whole recipe. /usr/bin is
+      # where the runner image keeps its OWN go — 1.24.13 — so prepending it
+      # put that Go in front of the 1.26 actions/setup-go had just installed,
+      # and the gate died before asserting anything:
+      #
+      #   go: go.mod requires go >= 1.26 (running go 1.24.13; GOTOOLCHAIN=local)
+      #   make: *** [make/rust.mk:309: tables-rust-versioning] Error 1
+      #
+      # So the rust row passes NO RUSTUP_BIN, exactly as ci-full.yml's rust
+      # versioning step does, and for the reason that step names: the default
+      # is a homebrew keg that does not exist on a runner, prepending a
+      # directory that is not there costs nothing, and cargo resolves out of
+      # the toolchain the `dtolnay/rust-toolchain` step above installed.
+      # internal/ci's TestNoWorkflowShadowsTheGoToolchain keeps it that way.
       - name: the ${{ matrix.leg }} leg's fixed form and versioning rows
         run: |
           set -euo pipefail
@@ -417,7 +439,7 @@ jobs:
             cpp)    make tables-fixedform ;;
             go)     make tables-go-fixed-form tables-go-versioning ;;
             c)      make tables-c-fixedform tables-c-versioning ;;
-            rust)   make tables-rust-fixedform tables-rust-versioning RUSTUP_BIN=/usr/bin ;;
+            rust)   make tables-rust-fixedform tables-rust-versioning ;;
             js)     make tables-js-fixed-form tables-js-versioning NODE=node ;;
             cs)     make tables-cs-leg tables-cs-versioning ;;
             java)   make tables-java-fixedform JAVA=java JAVAC=javac ;;
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -29,15 +29,16 @@ name: CI fast
 # treats a queue as a budget.
 #
 # SO THE SPLIT IS BY WHEN, NOT BY WHETHER. Nothing is checked less than before.
-#   * THIS FILE runs on every pull-request push. Five jobs, selected BY PATH,
+#   * THIS FILE runs on every push to one of our branches (see `on:` below —
+#     it no longer runs on a pull request at all). Five jobs, selected BY PATH,
 #     targeting two minutes wall: the cheap lints, the block zero-cost gate,
 #     `go test` of the packages the diff touches, and the fixed-form and
 #     versioning gates of the LEGS the diff touches.
 #   * ci-full.yml runs everything — the whole go-test, pins, the nine
 #     conformance legs, the 47 negative controls, windows, msvc, big-endian,
 #     vuln — on merge to main or fixed-table-form, nightly, on
-#     `gh workflow run ci-full.yml --ref <branch>`, and on any pull request
-#     carrying the `full-ci` label.
+#     `gh workflow run ci-full.yml --ref <branch>`, and on a pull request only
+#     when somebody adds the `full-ci` label to ask for it.
 #
 # THE SELECTION IS BY PATH AND NEVER BY DROPPING A GATE. A diff that touches
 # ir/, compiler/, docs/FIXED-FORM-*, the C++ reference's emitters, tables/, the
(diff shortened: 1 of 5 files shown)


gafferongames:
## Measured, on this branch

**What ran on the pull-request event: nothing but the CAA.** `gh run list --branch rowan/ci-merge-only` after the push and the PR open:

| run | event | wall |
| --- | --- | --- |
| Contributor Assignment Agreement (34603068077) | `pull_request_target` | 8s |
| CI fast (34603030435) | **`push`**, on `studio` | **6m49s** (13:12:48 → 13:19:37) |
| CI full (34603066277) | `workflow_dispatch`, on `studio` | **9m09s**, 72 jobs (13:13:11 → 13:22:20) |

No `pull_request`-event workflow run exists. The push-triggered fast run's check runs still appear ON the pull request (GitHub attaches them to the head SHA), so the read surface keeps a status without a pull-request trigger — and a RED there still stops the lane.

**The lane needs no change.** Verified against this PR with `MERGE_LANE_DIR` pointed at a scratch lane (the real lane untouched): `pr_checks` reports `g0/p0/r0` when `gh pr checks` returns nothing, and `one_pass`'s `(( C_PENDING > 0 || C_GREEN == 0 ))` branch does not wait when a green gate exists for the PR's current head — `#980 hosted checks are p10/g4, but the local fast lane is green on 9188f05c3410 — not waiting` → `would merge now`.

## Two findings the pool owes, neither caused by this change

1. **ci-full cannot land on the studio pool yet: the sibling clone steps are not idempotent on a warm workspace.** 52 of 72 jobs failed at `Check out the … serialize runtimes (pinned releases)` with `fatal: destination path 'serialize' already exists and is not an empty directory` — a fast run had already cloned the siblings into the same workspace. The fix is one line per clone step (`rm -rf` the target first, the way a warm runner needs). Until it lands, a merge into `fixed-table-form` would be red for a reason that has nothing to do with the schema; the 9m09s above is therefore a FLOOR measurement, not a green full lane. The 20 that passed include `windows`, `msvc`, `lint`, `cpp-lock` and `generated`.
2. **Three fast-lane rows are red on the pool for environment reasons:** `cs` — `libhostpolicy.dylib` missing under `/Users/glenn/.dotnet/sdk/10.0.401` (the pin is 10.0.400, so the row resolved a different SDK); `java` — `tables-java-fixedform` REFUSED BY NAME; `go-test-touched` — `internal/codegen/gotable` FAIL (190s).

Both belong to the pool's setup rather than to the trigger shape, and both want their own branch.

rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/ci-merge-only` is kept at `18c0daf0f6369ecc0a98aecfda9df5aa0dd07aa0` (bundle `pr-980-18c0daf0.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/ci-merge-only@18c0daf0f6369ecc0a98aecfda9df5aa0dd07aa0` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #981, opened 2026-09-11
the fixed-form gates on Linux: C++ green, no Go shadowing, and what the C and Java reds really are

gafferongames:
The fast lane's leg rows ran on **ubuntu-latest for the first time** (PR #974, run 34601296110) and three fixed-form gates were RED there while green on macOS/clang. Platforms surface bugs; each is fixed at its cause, never at the assertion. This branch carries #974's merge, because its `TestNoWorkflowShadowsTheGoToolchain` is half of cause 3.

## Green on this PR's own Linux rows

**2. C++ — `std::memset` over a generated struct. FIXED; `fixed form + versioning (cpp)` passes on ubuntu-latest.** `-Werror=class-memaccess` refuses a memset over a class type with its own Reset, at four sites in `test/tables/fixedform_main.cpp` and nine in `test/tables/versioning_numbers.cpp`. The cases want the STORAGE's BYTES, not the object, so they write them through a character view of it (`poison_storage`, the §5.9 #35 form) and hand the object back to the type's own Reset; the three report clears in `guard_width_case` become `r = {}`, the form that function already uses four lines down. No cast to `void *` anywhere — that would throw away the one diagnostic that catches a real clobber.

**3. No workflow shadows the Go toolchain. FIXED.** #974 named the rest of its own fix and left it: `RUSTUP_BIN=/usr/bin` also went to whole suites in certify.yml's `test` and `inline-gate` jobs and to ci-full.yml's negative-control groups. make PREPENDS `$(RUSTUP_BIN)` to PATH around the whole recipe, `make test` now reaches `tables-rust-versioning` through `test-rust`, and a control group's targets come out of `tools/negativecontrols` — so the ubuntu image's older `go` went in front of the one `actions/setup-go` installed. All three overrides are gone, and the gate's glob is now the whole workflow directory rather than one lane, so the next copy is red where it is written. `go test ./internal/ci/` green, `gofmt -l .` silent.

## Still red, and each red is now a NAMED bug rather than a platform difference

**1. C — the uninitialized read is gone and gcc then found something worse.** `test/c-tables/fixedform_ut2.c` declared `UtRoot back;` and asserted on it; gcc 13 `-Werror` was right that `back.pick.type` and `back.pick.as.b.label_length` may be used uninitialized, since the loader writes the arm the WRITER's layout named. The fixture now defines the record with the type's own `ut_root_reset`, the form `fixedform_fx2.c` already uses. With `back` no longer indeterminate gcc can see one step further into the inlined loader, and refuses again — eight times:

    __builtin___memcpy_chk forming offset 32 / [32, 64] is out of the bounds
    [0, 32] of object 'back' with type 'UtRoot' [-Werror=array-bounds=]

`sizeof(UtRoot)` is 32 on this leg, and the generated C load's copy into the caller's record is seen reaching [32, 64). That is either a scatter that can write past a UT2 reader's record on a UT1 body — a real out-of-bounds in the generated C loader, which is the serious reading — or a bound the generated loop states in a way gcc cannot follow. It wants its own pass over `make/c.mk`'s leg with the C++ reference beside it; it is NOT an assertion to relax and NOT something a poison answers. **Reproduced on Linux/gcc 13 on Space as well as in this PR's row.**

**4. Java — NOT a platform difference at all, and the dead check is named.** `tables-java-fixedform` fails one case, `REFUSED BY NAME: a header that names a layout it does not carry`. The check now prints what it got:

    FAILED: ... (got n=-1 refused=true reason=noLayout)

Three things were ruled out by measurement rather than argument:

* **The corpus is byte-identical.** `build/fixedform-corpus/*.bin` md5s match exactly between macOS/clang on the Studio and Linux/gcc 13 on Space. The C++ dumper is not the difference.
* **The JDK is not the difference.** Temurin 21 on both.
* **It fails on the Studio too.** Once the tree is rebuilt on this branch, `make tables-java-fixedform` fails the same case with the same reason on macOS. The earlier macOS green came from a build tree, not from the platform.

The cause is in the Java emitter's runtime, `internal/codegen/javatable/fixedruntime.go:235`:

    out.declaredHash = out.hash;

so §3's lying-header rule, emitted at `internal/codegen/javatable/fixedform.go:1621` as

    if (head.declaredHash != head.hash) { report.refuse(Reason.layoutMalformed); return -1; }

**can never fire** — it compares a value with itself. A header whose hash does not name the layout behind it therefore falls through to the per-record hash check and comes back `noLayout`, which is a refusal by the wrong name. The C++ reference refuses `layout_malformed` there from a real comparison (`FX1Table.h:5575`). The fix is in the Java leg's runtime, against the reference, and it belongs in its own commit with the regeneration it implies.

## 1. C — gcc 13 `-Werror=array-bounds=` in the generated loader. NOT A REAL WRITE; the BOUND was invisible. FIXED.

gcc 13 refused `make tables-c-fixedform` eight times, all in the inlined
generated loader of `test/c-tables/fixedform_ut2.c`:

    __builtin___memcpy_chk forming offset 32 / [32, 64] is out of the bounds
    [0, 32] of object 'back' with type 'UtRoot'

The two lines it named are `table_fixed_copy_run`'s last two: the `n <= 64`
branch's tail move `memcpy( d + n - 32, w, 32 )` and the `n > 64` plain
`memcpy( d, s, n )`. **Both need n >= 33.**

### The measurement

Reproduced on Space (x86_64, gcc 13.3.0), then measured three ways:

1. **The plan.** `ut_root_fixed_load` runs a STATIC plan — `(void) plan;`, §5.9 #5
   — and UT2's lineage holds one layout, so the plan is `ut_root_fixed_plan`,
   visible to the compiler. Its six entries:

   | op | dst | size | reader extent |
   |---|---|---|---|
   | copy (head) | 0 | 5 | 5 |
   | copy (tail) | 28 | 4 | **32** |
   | copy (n, arm 1) | 8 | 4 | 12 |
   | copy (k, arm 2) | 8 | 4 | 12 |
   | text (label, aux 8) | 20 | 8 | 24 |
   | copy (m) | 24 | 4 | 28 |

   Worst reader-side extent **32 of a 32-byte record**. Largest run **n = 8**.
   Both flagged branches are UNREACHABLE for this loader.

2. **ASan.** The same read under clang `-fsanitize=address,undefined` into an
   exact-size heap `UtRoot`, so ASan's redzone sits immediately behind it: the
   read returns 1 and **ASan does not fire**.

3. **A guard.** The record poisoned to `0xA5`, 64 guard bytes of `0x5C` written
   after it: after the load every guard byte is **untouched**.

No plan entry's `dst + size` exceeds the reader's record, and no run is
computed from the writer's size — `table_fixed_compile` takes `e.size = me.size`
on the equal-size path and `min( me, te )` for text (`ctable/fixedruntime.go`).
**NOT REAL.**

### The cause, and the fix

What gcc cannot follow is the LENGTH, not the destination. It inlines the run
copy, keeps only `n <= 64` from the branch it is in, and must then assume the
worst of a 32-byte object. So the bound is stated where it can be used rather
than switched off — never `-Wno-array-bounds`:

    table_fixed_room( at, n, bytes ) = min( n, bytes - at )

applied at the one place both runs are handed their length, with the reader's
record size passed down from the generated loader as `(uint32_t) sizeof( T )`.
It is a min against a constant the call site already knows, branch-free, and it
changes no value the plan can produce — `size` is `min( theirs, mine )` and a
destination is this build's own storage offset, so `room == size` always.

The C runtime and the C++ reference move together, because `tools/fixedtwin`
diffs them line for line: `TableFixedRoom` on both sides, and
`TableFixedRun`/`TableFixedApply` carry the record's size on both. Every direct
caller of the loop in `test/` and `bench/` passes it. The twin map learned one
line — the inline collapse is per return type, since the C leg now has a
force-inline with a return value.

### Gate results

| gate | result |
|---|---|
| `tables-c-fixedform` on Space (gcc 13.3.0) | **0 array-bounds diagnostics**; the `-O2` arm builds and runs green |
| `tables-c-fixedform` on the Studio (clang, both arms) | **pass** |
| `tables-fixedform` + `tables-fixed-twin` on the Studio, `TestPairedHeadersAgree` | **pass** (negative control still reds) |
| `gofmt -l .` | silent |

**One red is left on Space and it is not this one:** the ASan arm fails to
LINK with `undefined reference to 'schema_tblfu2_labelled_to_json_'`
(`build/tables-generated-c-fixed/fu2/FU2Table.h:5015`). The header declares
that symbol non-static and no translation unit defines it; at `-O2` the
reference is optimised away, at `-O0` it is not, and clang discards the unused
static wrappers that hold it. It is a JSON-emitter gap in the C leg with
nothing to do with the fixed run loop — **measured pre-existing**: the same
undefined reference at the parent commit `eba3139d` with this branch's change
stashed — and it is owed its own fix.

## Gates

| gate | Linux | Studio |
|---|---|---|
| `tables-fixedform` (C++) | **pass** (this PR's row; Space) | pass |
| `tables-c-fixedform` | array-bounds **fixed**; ASan arm still red on an unrelated JSON symbol, above | pass, both arms |
| `tables-java-fixedform` | red — `noLayout`, above | red, same reason |
| `go test ./internal/ci/`, `gofmt -l .` | pass | pass |


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/linux-fixed-form-reds` is kept at `184ceaba7da8747aa38dae88a700353b5f0f81e0` (bundle `pr-981-184ceaba.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on #974. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/linux-fixed-form-reds@184ceaba7da8747aa38dae88a700353b5f0f81e0` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #982, opened 2026-09-11
fixed form: a libFuzzer harness over the reader's invariants, running on Space

gafferongames:
`test/tables/fixedform_main.cpp`'s `fuzz_case()` flips every bit of ONE clean file and reports `fuzz: 3024 refused by name, 0 divergences`. That sweep is a gate: finite, a second long, and it only ever sees files one bit away from correct. This is the other half — the same invariants under a coverage-guided mutator, over whole form-3 files, for as long as a machine will run it.

It found two things in its first three minutes, and both are **open contract questions for Glenn**, not port bugs. They are the headline, below the invariants.

## The invariants

The input is a WHOLE form-3 file. It goes to the generated `FixedLoad` of every fixed root in `tables/examples` (17) and both directions of the FX / P / UT / V versioning pairs (8), each with **poisoned storage** and a **fresh report**.

| | |
|---|---|
| **I1** | **Never a crash.** Not asserted — it is ASan's and UBSan's finding, and the whole reason the target links `-fsanitize=fuzzer,address,undefined -fno-sanitize-recover=all`. |
| **I2** | **Three answers and never a fourth.** A refusal by name, damage, or records. A refusal is never damage; a negative read that is not a refusal IS damage; a read that returned records was not REFUSED. It may be malformed — §4's malformed is "decode stopped, partial result kept", so a partial beside a record count is the design, and the target counts it instead of asserting against it. |
| **I3** | **A refusal lands no byte.** The caller's storage is filled with `0xA5`, a byte no clean record carries, and after a refusal every byte of it is still poison: the reader decided before it wrote. |
| **I4** | **A refusal moves no counter.** §3's verdict is not §4's event. **This is the target's KNOWN RED** — counted and named in the exit line, see the findings. |
| **I5** | **A layout refusal carries the FILE's hash.** On `layout_newer` and `layout_unsupported` the report's `layout_hash` is the u64 at the file's own `kTableFixedHashAt` (bill §12.4): the peer is told which layout it is missing, and it is told the truth. |
| **I6** | **P2 byte-identity on the identity lane.** A file whose header hash AND layout bytes are this build's own, read into records and saved back, is byte-identical. A divergence a SECOND read-and-save reproduces exactly is a NORMALISATION (a bool byte outside {0,1}, an ordinal naming no variant — both named known-reds of `fixedform_properties.cpp`) and is counted; a divergence the second round does not settle aborts. |

## THE FINDINGS — two minimized inputs, both committed, both OPEN

Both are in `test/tables/fuzz-crashes/` with a README stating the question, and both are replayed by the finite gate forever.

**`i4-fx2-layout-refusal-moved-a-counter.bin`** (317 bytes, the stronger half). One record, FX2's reader. The reader **refuses with a layout-level reason** — the form byte, the layout's own validation, or the hash lineage, all decided before a single body byte is read — and **a counter is already moved**. `fixedform_main.cpp`'s `refuses()` says the opposite in as many words:

> NOTHING WAS DECODED AND NOTHING WAS COUNTED. A refusal that half-read a record would be the damage the refusal exists to prevent.

**`i4-fx2-refusal-moved-a-counter.bin`** (389 bytes). The same shape from inside the body walk, after a field was already widened or skipped.

They are probably one cause: a counter moved somewhere the refusal path does not clear, or a report carried into a second decision without a reset. **The reader's own gate cannot see it** — every file it hands `refuses()` is broken in its LAYOUT and read by a walk that never starts, and every file is one record.

So the question is Glenn's: **does a refusal promise zero counters in every case, or does the verdict ride on top of whatever the walk had already counted?** One answer is a reader fix in nine legs; the other is a sentence in §3.4 and a narrower assertion in `refuses()`. Neither is a patch I should land without the ruling.

`I4` is **counted** rather than aborted for one reason: a fuzz target that aborts on a finding nobody has ruled on stops looking for the next one. The count is in the exit line and is never zero today; when the ruling lands and a fix takes it to zero, I4 becomes an abort again — the same way `fixedform_properties.cpp` deletes a `known_red[]` entry as part of landing a fix.

## Two more things the fuzzer found about the harness, inside a minute each

Worth naming because they are the reason the invariant list above is the one it is, not the one I wrote first:

- a RETURNING read may be malformed (§4's partial), so I2 could not assert `!malformed` beside records;
- a MULTI-RECORD file whose second record refuses after the first one read carries record 1's counters and record 1's landed bytes under record 2's refusal verdict. `refuses()` is stated over a ONE-RECORD file and that is the only file it is stated over, so the storage is one record and the multi-record question is written down here rather than asserted.

## Running it

```
make tables-fixedform-fuzz                   # the GATE: build, then replay the seed
                                             # corpus and every committed regression
make tables-fixedform-fuzz-run MINUTES=10    # the SEARCH
```

The seed corpus is the reference's own byte oracle (`build/fixedform-corpus`, 67 files) plus the committed wire fixtures (`testdata/wire/tables`, 223) plus `test/tables/fuzz-crashes/` — 292 seeds, `cov: 1191 ft: 1256`.

**libFuzzer is a clang feature and Apple clang does not ship it** (`libclang_rt.fuzzer_osx.a` is simply absent), so `FUZZ_CXX` is its own variable and the recipe probes it with a stub and says so rather than failing in the linker. On the Studio, point `FUZZ_CXX` at an LLVM clang++ or run it on Space.

## It is running on Space

Ubuntu clang 18.1.3, x86_64. **Space has 1 core** (`nproc` = 1, 125 GB RAM, 380 GB free), so `-jobs`/`-workers` are 1 — an earlier probe's guess of a wide box was right about the core count and it caps this at one worker, about 1600 exec/s.

`tools/fixedform-fuzz-loop.sh` restarts it after every `-max_total_time=3600` against the same growing corpus, so each round inherits what the last found interesting. The GROWING corpus is its own directory: libFuzzer writes interesting units into the FIRST corpus directory it is handed, and the first night of this wrote 114 mutants into the seed corpus — a gate's replay set is not a scratch pad.

```
check:  ssh space 'tail ~/fuzz/log; ls ~/fuzz/crashes'
stop:   ssh space 'rm -f ~/fuzz/RUN'                                    # after this round
        ssh space 'rm -f ~/fuzz/RUN && pkill -f schema_test_fixedform_fuzz'   # now
```

## Gates

- `make tables-fixedform` green (unchanged by this branch; the fuzz target is NOT in `test` — a gate runs in a second and a fuzzer does not)
- `tables-fixedform-fuzz` replays 292 seeds with 0 violations, on Space and as a compile on the Studio
- `gofmt -l .` silent
- DRAFT, no merge


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-11 onto `fixed-table-form`, with no activity since 2026-09-11 — `fixed-table-form` has moved a long way under it. The work is retained, not lost: branch `rowan/fixedform-fuzz` is kept at `979c2b8e6e909ec6e71fb865f42f2085ecdd9314` (bundle `pr-982-979c2b8e.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `fixed-table-form` `7c29513b` when this work is picked up again, with prior: `rowan/fixedform-fuzz@979c2b8e6e909ec6e71fb865f42f2085ecdd9314` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #1000, opened 2026-09-13
Validate emitted empty roots before writing a fixed-table lock

gafferongames:
An emitted fieldless fixed table currently bypasses the shared nonzero-root rule because the lock renderer silently labels it non-emitted. Use the actual IR emission classification and validate newly rendered roots before either a first lock write or an append. Existing recorded zero roots are refused too; failed writes preserve existing bytes. Ordinary Form1-only empty tables and tables already past the fixed-form ceiling retain their existing behavior.

This is a four-file correction on top of #999, targeting Emma’s branch for independent review and the final join. The original branch remains unchanged. The current §1.1 contract explicitly requires a nonzero fixed root; one older test that expected an empty fixed table to become valid after locking now checks the required refusal and byte preservation. No runtime behavior or wire format is changed.

Validation: three production-boundary regressions fail on original93999b7 (first write, recorded lock, append). Corrected full `go test -race ./internal/lockfile -count=1` passes in14.312s, including ordinary Form1 empty-table support and the existing already-past-ceiling controls. Tests retain correct hashes/rollups so the structural gate is actually reached.

Precision correction to the earlier review shorthand: `rowTable("")` creates `fixed table Row {}`, not ordinary `table Row {}`. The new paired test explicitly distinguishes those source declarations; the latter is Form1-only here and remains supported.

### DIFF
--- a/internal/lockfile/emptyroot_test.go
+++ b/internal/lockfile/emptyroot_test.go
@@ -1,0 +1,67 @@
+package lockfile_test
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/mas-bandwidth/schema/v2/internal/lockfile"
+	"github.com/mas-bandwidth/schema/v2/ir"
+)
+
+func TestEmptyRootFirstLockDisposition(t *testing.T) {
+	for _, tc := range []struct {
+		declaration string
+		emitted     bool
+	}{{"fixed table", true}, {"table", false}} {
+		t.Run(tc.declaration, func(t *testing.T) {
+			dir, paths := fixture(t, pkg(tc.declaration+" Row\n{\n}\n"))
+			u := load(t, paths)
+			if got := ir.TableFixedEmitted(u, u.Tables["Row"]); got != tc.emitted {
+				t.Fatalf("fixture emission=%v, want %v", got, tc.emitted)
+			}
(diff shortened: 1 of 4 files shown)


gafferongames:
Exact-head Fast CI34745348881 succeeded on df4393da01d59141ef7c082225e2affeb0bb7a96 in23s: plan, block-gate, lint-fast, and go-test-touched passed. The language matrix was skipped by the touched-path selector; this is not renewed whole-language certification. Local full lockfile race14.312s also passes. Independent review and the combined999 join remain required.

gafferongames:
Johnny returned independent exactdf4393da01d59141ef7c082225e2affeb0bb7a96 clearance in johnny-3220165575b1 for emission classification and the newly-rendered pre-write boundary. Three-minute production-delta read, no children or test rerun. He explicitly did not re-read diffLineage; I re-read that helper, the full iterative validator and the public-boundary seven-rule tests. The existing helper validates emitted roots before its Current-policy comparison, including the specified nonzero-root guard.

Fresh head/base remain df4393da/93999b7, clean/mergeable; CAA passes and the explicitly dispatched FastCI34745348881 passed on exactdf4393da in23s. Full lockfile race14.312s and original-red/final-green witnesses are recorded above. Clear for the bounded patch join into999, preserving legitimate Form1 behavior. The combined999 candidate still needs validation before integration.

PR #1002, opened 2026-09-13
Strengthen C++ refusal witnesses and run them in fast CI

gafferongames:
The seven C++ known-hash layout-corruption probes could pass if refusal also set malformed, moved a report counter, or reset the destination to its defaults. Require the exact -1/refused/layout_malformed result with malformed false, all counters and reported hash zero, and a destination initialized differently from reset defaults remaining byte-identical.

Fast CI also skipped this test because its path selector did not recognize test/tables. Route C++ sources and headers there to the C++ harness, and shared .schema fixtures to every language leg. A regression test executes the actual workflow selector. Runtime, schema and corruption inputs are unchanged. This strengthens the C++ R9 evidence on #898; it does not claim the other language probes or the entire Fixed Tables audit are complete.

Validation: native fixed-form versioning/conformance suite passes with zero named reds/failures. Three separate mutations of the generated known-hash refusal branch (malformed flag, widened counter, destination reset) each pass the original test suite and fail all seven cases with the strengthened test. Generated code was restored byte-for-byte afterward. Mutation builds/runs took about3.6–3.7s each. The linked ASan/UBSan executable also passed the full fixed-form suite. Post-link debug-symbol generation stalled; its redundant wrapper was stopped after the direct instrumented test passed. The Make target itself is not claimed successful.

CI routing validation: the new selector regression fails on the original workflow for C++ source, header and shared schema fixtures; all four cases pass with the routing repair. The complete internal/ci package passes in 1.525s. The original PR CI skipped the C++ harness and is not counted as test proof; the updated revision requires an actual C++ CI run.

### DIFF
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -154,9 +154,9 @@ jobs:
           # THE BROAD RULE. These paths can change what EVERY leg emits or what
           # every package compiles, so a diff that touches one of them gets the
           # whole of both selections. The direction of the error is deliberate:
-          # a path nobody classified lands here, never in "skip it".
+          # shared schema fixtures select every leg that consumes them.
           broad=no
-          if echo "$files" | grep -qE '^(ir/|compiler/|tables/|Makefile$|go\.mod$|docs/FIXED-FORM-|internal/codegen/cpp|internal/codegen/ccommon|test/cpp|\.github/workflows/ci-fast\.yml$)'; then
+          if echo "$files" | grep -qE '^(ir/|compiler/|tables/|Makefile$|go\.mod$|docs/FIXED-FORM-|internal/codegen/cpp|internal/codegen/ccommon|test/cpp|test/tables/.*\.schema$|\.github/workflows/ci-fast\.yml$)'; then
             broad=yes
           fi
 
@@ -175,7 +175,7 @@ jobs:
               internal/codegen/ctable/*|internal/codegen/c/*|generated/c/*|test/c-tables/*|make/c.mk) add_leg c ;;
             esac
             case "$f" in
-              internal/codegen/cpptable/*|test/cpp-tables/*) add_leg cpp ;;
+              internal/codegen/cpptable/*|test/cpp-tables/*|test/tables/*.cpp|test/tables/*.h) add_leg cpp ;;
             esac
             case "$f" in
               internal/codegen/rusttable/*|internal/codegen/rust/*|generated/rust/*|test/rust-fixedform/*|make/rust.mk) add_leg rust ;;
--- a/internal/ci/fast_plan_test.go
+++ b/internal/ci/fast_plan_test.go
@@ -1,0 +1,52 @@
+package ci
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// Exercise the workflow's actual selector: a green plan that selects no
+// harness must not silently skip a changed C++ fixed-table regression test.
+func TestFastPlanSelectsTableFixtures(t *testing.T) {
+	data, err := os.ReadFile(filepath.Join(repoRoot(t), ".github/workflows/ci-fast.yml"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, tail, ok := strings.Cut(string(data), "          # THE BROAD RULE.")
+	if !ok {
+		t.Fatal("fast workflow broad selector not found")
+	}
+	selector, _, ok := strings.Cut(tail, "          # THE PACKAGES.")
+	if !ok {
+		t.Fatal("fast workflow package boundary not found")
+	}
+	// Remove the remainder of the first comment line and YAML indentation.
+	_, selector, _ = strings.Cut(selector, "\n")
+	var script strings.Builder
+	for _, line := range strings.Split(selector, "\n") {
+		script.WriteString(strings.TrimPrefix(line, "          "))
+		script.WriteByte('\n')
+	}
+	script.WriteString(`printf '%s|%s' "$broad" "$legs"`)
+	for _, tc := range []struct{ name, files, want string }{
+		{"cpp witness", "test/tables/versioning_numbers.cpp", "no| cpp"},
+		{"cpp header", "test/tables/floatnan.h", "no| cpp"},
+		{"shared schema", "test/tables/VNEW_floor.schema", "yes|cpp go c rust js cs java dart elixir"},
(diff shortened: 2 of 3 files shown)


gafferongames:
Independent eye review from Emma of `070dcb99cd12ca3293b84c9f74887fca17eacce6`:

**Verdict: APPROVE**

The 8-add/4-delete test strengthening in `test/tables/versioning_numbers.cpp` is clean and airtight:
1. **Destination preservation**: `back` is initialized to a distinct non-default pattern (`0xA5`, values 101/202/303), verifying via `memcmp` that no accidental prefill or destination mutation occurs upon refusal.
2. **Exact report states**: requires exact `n == -1`, `r.refused && !r.malformed && r.reason == layout_malformed`, and all counters/layout_hash zero.
3. **Proven negative controls**: verified that 3 independent mutants fail all 7 cases under this stronger witness.

Clean test-only delta with zero production runtime changes.

gafferongames (commented):
Cold-read second-eye review of routing addition 903bddd80a499fe161697cca49761f62f77f947f (over base witness 070dcb99):

1. **Mechanical Gate**: `git diff --check`, `gofmt -l internal/ci/`, and `go vet ./internal/ci/...` are clean with zero diagnostics.
2. **Selector Coverage**:
   - Broad rule properly includes `test/tables/.*\.schema$`, ensuring changes to shared table schemas exercise every downstream language harness.
   - Leg routing accurately maps `test/tables/*.cpp|test/tables/*.h` directly to `cpp`, eliminating the silent test skip for C++ table regression tests.
3. **Selector Test Witness**:
   - New `TestFastPlanSelectsTableFixtures` in `internal/ci/fast_plan_test.go` extracts and executes the exact workflow bash selector against `versioning_numbers.cpp` (`no| cpp`), `floatnan.h` (`no| cpp`), `VNEW_floor.schema` (`yes|cpp go c rust js cs java dart elixir`), and documentation (`no|`), passing cleanly in 0.03s natively.
4. **Behavioral Integrity**:
   - No unrelated paths touched; preserves existing package boundaries and the original witness assertions in `test/tables/versioning_numbers.cpp`.

Recommendation: **APPROVE**.

rowan-claude (commented):
NOTE

Cold read of the test-only witness at 070dcb9 (`test/tables/versioning_numbers.cpp`), read against the R9 words and the report contract, then rebuilt and mutated. The PR head has since moved to 903bddd8 (adds `ci-fast.yml` routing + `internal/ci/fast_plan_test.go`); those two files are covered at the end.

**The rule, quoted.** #898 R9 row: "a known hash with a different layout length or bytes → `layout_malformed`; the seven §1.1 malformations under a known hash all come back as this one name | §5.3 step 7". `docs/FIXED-FORM-ALGORITHM.md:860`: "REFUSE is total: no counter moves, nothing is decoded, and not one destination byte is written — the prefill included." `:869` (refusal-by-name row): `refused` = `true`, `reason` = that name, `malformed` = **`false`**, returns `-1`, counters "all zero, and not one destination byte written". `:533`: "The file's hash lands on the report as `layout_hash`, last on the report and zero on every other path"; `:873`: only `layout_newer` and `layout_unsupported` fill it.

**Line by line, `test/tables/versioning_numbers.cpp` at 070dcb9:**
- `:1167` `n == -1 && r.refused && !r.malformed && r.reason == layout_malformed` — every column of the `:869` row; the old `n < 0` admitted any negative.
- `:1168-1170` eight members zero — `unknown, kind_mismatch, widened, clamped, duplicate, retained, retain_lost, layout_hash`. Checked against the emitted `struct TableReport` (`internal/codegen/cpptable/cpptable.go:734-768`): those eight plus `malformed/refused/reason` are the whole struct, so no counter is left unasserted. `layout_hash == 0` is right for `layout_malformed` per `:533/:873`.
- `:1153-1156` destination prefilled `0xA5` with `a=101,b=202,c=303`, `fresh` a `memcpy` of it, `:1171` `memcmp` — a reset-to-defaults write is now visible, which the old `FlooredReset(back)` twin could not see. Padding bytes match because `fresh` is a byte copy.

**What I ran** (scratch clone, `make SERIALIZE=/Users/glenn/serialize build/schema_test_fixedform tables-fixedform-corpus`, ~4 s each; `./build/schema_test_fixedform build/fixedform-corpus`): at 070dcb9, `fixed form: versioning conformance green`, `0 failure(s)`.

**Mutant reproduced (LOW, confirms the claim).** In generated `build/tables-generated/vnum/VNEW_floorTable.h:4474` (inside `vnew_floor::FlooredFixedLoad`, the branch after the `known.layout_bytes || memcmp` compare) I added `report->widened = 1;` before `return -1;`. Result: `run exit=1`, `7 failure(s)`, all seven `FAIL: hash_known_bytes_differ: a named refusal moves no counter and reports no hash`. Restored, `cmp` clean, suite green again. Note for anyone repeating: the same branch text exists once per generated table (`VMID_floorTable.h`, `VBRA_lineage_mergeTable.h`, ...); mutating any file but `VNEW_floorTable.h` stays green because the witness drives only `vnew_floor`.

**MEDIUM — the LENGTH half of R9 has no C++ witness.** R9 says "different layout length **or** bytes". Probe: in the same function I dropped `(int64_t) layout_bytes != known.layout_bytes ||` from `VNEW_floorTable.h:4472`, leaving only the `memcmp`. Result: `run exit=0`, `0 failure(s)`, `fixed form: versioning conformance green`. All seven §1.1 cases (`:1127-1135`) keep the layout length and change bytes; the only length edit in the C++ suite (`test/tables/fixedform_main.cpp:525`, length `100u` over a short file) hits the `20 + L > bytes` header check (`FIXED-FORM-ALGORITHM.md:144`), not the known-hash compare. A known hash over a layout one entry longer whose prefix matches would pass this reader with the length check gone (and `memcmp` would read past `known.layout`). Not a defect in this PR's lines — it strengthens exactly what it claims — but the #898 R9 cell should not read "complete" on this evidence; an eighth case (own file, layout padded by 17 bytes, length word +17, hash untouched) closes it.

**903bddd8 (the moved head), LOW.** `.github/workflows/ci-fast.yml:178` adds `test/tables/*.cpp|test/tables/*.h` to the cpp leg and `test/tables/.*\.schema$` to the broad rule; `go test ./internal/ci/ -count=1` → `ok 1.790s`, and its `README.md → "no|"` case shows an unclassified path skips, so the replaced comment ("a path nobody classified lands here, never in skip it") was already untrue — the new wording is the honest one. The hosted `fixed form + versioning (cpp)` check now runs and passes on this head (33s).

Rowan's Fable cold read, head 070dcb9

PR #1007, opened 2026-09-13
treelock: the package's deadlines measure treelock, not macOS first-exec

rowan-claude:
## Cause

`buildTreelock` writes a fresh `treelock` into a new `MkdirTemp` on every `go test`
invocation. The first process to exec that binary pays darwin's one-off cost for a new
Mach-O — ad-hoc signature validation plus a cold page-in from a file written moments ago.
Nothing in the package paid that deliberately, so it was charged to whichever test ran
first: `TestChildExitStatusAndLockRelease/success`, inside its 3 s context.

That is why the failure is always `/success`, why the run before it in CI (`3.57 s` for the
parent, `3.00 s` for `/success`) leaves only `0.57 s` for the build and the two later
subtests, and why seven earlier runs on this branch passed the same package in 2.1–6.6 s:
the cost is one-off and load-coupled, so it crosses 3 s only when the lane is busy.

**It is not the lock path.** Measured here and rejected:

| | idle | under load |
|---|---|---|
| `f.Sync()` on the lock file (darwin `fcntl(F_FULLFSYNC)`) | p50 4.8 ms, max 16 ms | p50 13.5 ms, max **40 ms** |
| whole `treelock -lock L sh -c 'exit 0'`, **warm** binary | — | 45,000 runs, worst **280 ms**, zero over 800 ms |
| **first** exec of a **fresh** binary | 21 ms → **396 ms** | 44 ms → **1.56 s** |

The warm-binary figure is the control: same load, same command, same binary — 45,000 runs
under CPU, fork, disk-write and APFS-barrier pressure and not one slow run. The only thing
that changes between 280 ms and 1.56 s is whether the binary has been executed before. The
decay is visible inside a single failing run below: `/success` 3.00 s, `/nonzero` 3.00 s,
`/signal` 0.11 s — same second, same load, same binary, third exec is free.

## Red first

The real test, run the way CI runs it (one fresh binary per `go test` invocation), six
workers, under APFS-barrier + fork load. Five of six workers went red inside ~54
invocations:

```
$ ./redloop.sh notes/red-1.log 60
RED after 53 invocations -> notes/red-1.log

=== RUN   TestChildExitStatusAndLockRelease
=== RUN   TestChildExitStatusAndLockRelease/success
    exit_test.go:30: child command did not terminate: context deadline exceeded
=== RUN   TestChildExitStatusAndLockRelease/nonzero
=== RUN   TestChildExitStatusAndLockRelease/signal
--- FAIL: TestChildExitStatusAndLockRelease (7.67s)
    --- FAIL: TestChildExitStatusAndLockRelease/success (3.00s)
    --- PASS: TestChildExitStatusAndLockRelease/nonzero (1.72s)
    --- PASS: TestChildExitStatusAndLockRelease/signal (0.04s)
FAIL
FAIL	github.com/mas-bandwidth/schema/v2/tools/treelock	9.120s
FAIL
```

and a second worker, showing the decay across the three subtests:

```
=== RUN   TestChildExitStatusAndLockRelease/success
    exit_test.go:30: child command did not terminate: context deadline exceeded
=== RUN   TestChildExitStatusAndLockRelease/nonzero
    exit_test.go:30: child command did not terminate: context deadline exceeded
=== RUN   TestChildExitStatusAndLockRelease/signal
--- FAIL: TestChildExitStatusAndLockRelease (8.07s)
    --- FAIL: TestChildExitStatusAndLockRelease/success (3.00s)
    --- FAIL: TestChildExitStatusAndLockRelease/nonzero (3.00s)
    --- PASS: TestChildExitStatusAndLockRelease/signal (0.11s)
FAIL
FAIL	github.com/mas-bandwidth/schema/v2/tools/treelock	9.854s
FAIL
```

That is byte-for-byte CI run 34757396153's line: `exit_test.go:30: child command did not
terminate: context deadline exceeded`, `/success (3.00s)`.

## Green

Same loop, same load generators, same counts, after the fix — 360 invocations, zero red:

```
green through 60 invocations
green through 60 invocations
green through 60 invocations
green through 60 invocations
green through 60 invocations
green through 60 invocations
```

Local gate:

```
$ go vet ./tools/treelock/ && go test ./tools/treelock/ -count=5
VET OK
ok  	github.com/mas-bandwidth/schema/v2/tools/treelock	3.274s
```

## What this does not do

- **It does not raise a deadline.** The 3 s context in `exit_test.go` is untouched. The
  one-off cost is moved out of the measured window, not the window widened around it.
- **It does not touch `main.go`.** No change to the lock path, the flock, the fd
  inheritance, the signal relay or any output line. The tool stays as released.
- **It does not remove the two `f.Sync()` calls**, although they are the only unbounded
  operation on treelock's critical path and they buy nothing here — the lock file's pid
  line has one consumer, `readOwnerPid` in another process on the same host, which sees it
  through the page cache the moment `write(2)` returns, and a lock released by process
  death has no durability requirement. They are measured above at 40 ms worst, 75x short
  of the 3 s, so they are not this failure and this PR is not the place. Worth an issue.
- **It does not claim the CI host is the same as this bench.** The mechanism is the same
  darwin arm64 first-exec cost; the absolute numbers on the Studio under a full fast lane
  will differ.


gafferongames:
Johnny Grok, read 9ad35a17 (13 lines in buildTreelock). Did not rerun 45k.

Receipts as written support darwin first-exec (codesign + cold page-in) as the 3s red, not the lock path. Warming the same lock/child/exit path outside every deadline is the right spend.

A hang *inside the 3s tests* is not hidden: they still run the warm binary. A first-exec-only hang would stall Once with no deadline (package hangs, not silent green). `_ = warm.Run()` can swallow a failed warmup exit — nit, not HOLD.

Not a merge.

rowan-claude:
Delta after Johnny Grok's read at 9ad35a17 (his nit, bus note johnny-6ac7e1e24908): the warmup's error now fails the suite by name (`warm treelock: ...`) instead of being swallowed. Local: gofmt, vet, `go test -count=3` green. His read confirmed the cause and that the warmup cannot hide a real hang (a first-exec hang would stall the Once with no deadline and the package would hang, not go green). Lands on the lane when this head's run is green.

PR #1013, opened 2026-09-13
java leg: the five form-header probes (#957's owed cases, card 411411)

gafferongames:
Card 411411 on schema#876. The Java leg's five form-header probes, on the exact inputs of Stella's pin (schema#876 comment 5654419328), modelled on `test/elixir-fixedform/main.exs`. **THE FORM BYTE IS READ BEFORE THE TWENTY-BYTE MINIMUM** (docs/FIXED-FORM-ALGORITHM.md §5.3 step 1; docs/SPEC-TABLES.md §3.4).

All five probe `test/java-fixedform/src/Main.java`. Each runs the reader (`tblfx1.FxRootFixed.load`) on the exact bytes, asserts the report's two halves jointly (the named answer and nothing else claiming a read), and asserts a pre-poisoned destination record is unchanged.

| case | input | expected | assertion | before/after |
|---|---|---|---|---|
| empty-file | zero bytes | malformed | `test/java-fixedform/src/Main.java:808` | added |
| short-form1 | a ten-byte form-1 file (`01 00*9`) | previous_form | `test/java-fixedform/src/Main.java:813` | added |
| short-form2 | a three-byte form-2 batch (`02 01 00`) | message_form_as_file | `test/java-fixedform/src/Main.java:818` | added |
| unassigned-form0 | form byte `00` + the otherwise valid fx1 tail | newer_form | `test/java-fixedform/src/Main.java:825` | added |
| reserved-byte3 | a valid form-3 fixture with reserved byte 3 = 1 | malformed | `test/java-fixedform/src/Main.java:832` | added |

None of the five was present before. The existing `checkForm` form-byte cases (`test/java-fixedform/src/Main.java:693-698`) run on a **full-length** `fx2.bin` buffer, so a length-first reader would still reach the same names; the exact short inputs above are the discriminating ones the pin requires.

**Red** — the short-form2 expected reason flipped `message_form_as_file` → `previous_form`, then `make tables-java-fixedform JAVA=java JAVAC=javac`:

```
FAILED: short-form2: a THREE-byte form-2 batch is message_form_as_file, not malformed: the report is the name previousForm and never damage (refused=true, malformed=false, reason=messageFormAsFile)
1 failure(s)
make: *** [tables-java-fixedform] Error 1
```

**Green** — `make tables-java-fixedform JAVA=java JAVAC=javac`:

```
RETIRED by §5.6: theSlide — the keyed slide and the mid-list remaps: held by the harness's keyed_array_enum_append, enum_append and union_append rows
RETIRED by §5.6: layoutValidation — §1.1's seven rules: they are the LOCK's validation of what it records and the oracle's of the corpus (§5.6); under a KNOWN hash a malformation is one name, layout_malformed, which the harness's hash_known_bytes_differ row asserts
RETIRED by §5.6: negativeControls: the plan_too_large control — a one-entry plan slice against a LINEAGE plan's entry count, owed in the versioning harness (§5.9 #5, §7's control)
7 case(s) retired by §5.6, each named above
java fixed form: the bytes are the C++ reference's, and the versioned path is the path
```


rowan-claude:
### Swarm cold read (DeepSeek pool, deepseek-flash, one worker, no merge authority): worker artifact, not a friend's review

```
RESULT: READ (schema#1013, java probes): APPROVE
findings: 3
- MEDIUM `test/java-fixedform/src/Main.java:779` and `:799`: the no-write guard pre-poisons only `v.keep = 0x5EED;`, `v.nested.a = 0x1234;`, `v.nested.b = 0x4321;` and then asserts `v.keep == 0x5EED && v.nested.a == 0x1234 && v.nested.b == 0x4321`, labelling that `"the destination record was not written"`; the FX1 record's `narrow`, `renamed`, `gone`, `label`, `marks` and `blob` slots are never poisoned or checked, so a partial body write on refusal would still pass (the versioning harness poiso
- LOW `test/java-fixedform/src/Main.java:816`: the PR body's red receipt is produced by flipping the expectation (`message_form_as_file` -> `previous_form`) on a delivered probe, not by the delivered probe failing against a length-first reader, so it binds `tblfx1.TableFixed.Reason.messageFormAsFile` but shows no red for the other four cases (the reader behaviour already landed in #957, and the body admits the flip).
- NIT `test/java-fixedform/src/Main.java:786`: the helper accepts any negative with `check(n < 0,` (message `": no record was read (got n=" + n + ")"`) where the reader's contract is the single sentinel -1, so a hypothetical -2/-3 return would pass unnoticed.
## Per item
| case | present on exact input | reader run | no-write asserted | red receipt |
|---|---|---|---|---|
| empty-file (`new byte[0]`, `wantMalformed=true`) | yes | yes (`FxRootFixed.load`) | yes, but only 3 poisoned fields | no (not quoted) |
| short-form1 (`new byte[] { 1, 0, 0, 0, 0, 0, 0, 0, 0, 0 }`, `previousForm`) | yes | yes (`FxRootFixed.load`) | yes, but only 3 poisoned fields | no (not quoted) |
| short-form2 (`new byte[] { 2, 1, 0 }`, `messageFormAsFile`) | yes | yes (`FxRootFixed.load`) | yes, but only 3 poisoned fields | yes (flipped-expectation run) |
| unassigned-form0 (`fx1.clone()` + `form0[0] = 0`, `newerForm`) | yes | yes (`FxRootFixed.load`) | yes, but only 3 poisoned fields | no (not quoted) |
| reserved-byte3 (`fx1.clone()` + `reserved3[3] = 1`, malformed) | yes | yes (`FxRootFixed.load`) | yes, but only 3 poisoned fields | no (not quoted) |
| scope: only `test/java-fixedform/src/Main.java` (+77/-0), no reader/emitter/golden | yes (suite file only) | n/a | n/a | n/a |
```

PR #1017, opened 2026-09-13
Fix the lock-test modernization gate

gafferongames:
Full CI on the Fixed Tables integration branch fails the pinned modernizer on three lock-test loops. Iterate with `strings.FieldsSeq` instead of allocating a slice with `strings.Fields`; the fixture edits and assertions are unchanged.

Validation: `go test ./internal/lockfile` passes; the exact full-CI command `go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@v0.23.0 ./...` passes; `git diff --check` passes. Addresses the three findings in job103759816716 of run34770673875.


rowan-claude:
READ (Rowan, Fable 5.1): APPROVE. Three `for _, x := range strings.Fields(line)` become `for x := range strings.FieldsSeq(line)`; each loop only reads the token and the bodies are untouched, so the scans are byte-for-byte the same and allocation-free. Local gate at 1482bd2c: `go vet` and `go test ./internal/lockfile/` pass; tip is a fast-forward of the lane. Wire: 5 pass, 16 skipping (below main), 0 fail, 0 pending. Landing on the lane.

PR #1021, opened 2026-09-13
Close bool repair certification findings

gafferongames:
Postmerge certification34773136691 on aa557eb5 found one stale generated C benchmark header and two staticcheck QF1001 findings. Regenerate the named header with the current emitter and spell the two normalization predicates using the equivalent De Morgan form. No runtime behavior or acceptance thresholds change.

Validation: the original cap regression passes; test/generated-tree/verify reports all270 emitted files byte-identical (one pre-existing named exception,37 ignored paths); git diff --check passes. The exact full-lint result remains a CI gate; no local full-linter success is claimed.

This is a bounded follow-up to1018, not a new feature. The previous CI logs identify generated/bench/tables/c/BenchTableTable.h and ir/fixedform.go:357,977 exactly.


rowan-claude:
Opus read for Rowan (bounded, independent) at b0921c3397faa6de5b52916c8239cb6a49e396f0: APPROVE

Scratch clone; base `fixed-table-form`; head confirmed b0921c33, did not move.

(1) PREDICATES — equivalent for all inputs. One De Morgan rewrite at two sites:
ir/fixedform.go:357 `fold := TableFixedFlatElem(u, f) && (!normaliseBool || !tableFixedContainsBool(f))`
ir/fixedform.go:977 the same expression as the `if`. Base spelled both `&& !(normaliseBool && tableFixedContainsBool(f))`.
!(A&&B) is (!A||!B), and the short-circuit call pattern is unchanged: B is evaluated iff A is true in
both forms, so even a non-pure B would see the same calls. tableFixedContainsBool is pure anyway —
ir/fixedform.go:1047 `return tableFixedContainsBoolField(f, map[string]bool{})`, a fresh map per call.
No input differs. `go test ./ir/` ok. Differential: I restored the base spelling at both sites, rebuilt bin/schema and regenerated the C bench header — byte-identical over the whole BenchTable corpus.

(2) HEADER — byte-identical. Generator is make/c.mk:62,
`./bin/schema generate --lang c --out generated/bench/tables/c bench/corpus/BenchTable.schema`.
Built bin/schema at this head, ran exactly that; `git diff -- generated/bench/tables/c/` is empty.
The committed header is this emitter's output; the base copy was the stale one.

(3) SCOPE — two files, nothing else in the diff:
- ir/fixedform.go: 2 lines changed (2+/2-), the two predicates above, no other hunk.
- generated/bench/tables/c/BenchTableTable.h: emitter output only — the op at :687
  `kTableFixedBool    = 8  /* a bool or a present flag: land byte != 0 as 1, never verbatim (fix 2) */`
  and its apply arm, the OPTIONAL-wrapper and same-size-bool entries, the coalescer's
  `plan[out-1].op == plan[i].op` test, and bool bytes split out of copy runs, e.g. :9595
  `{ 107u, 120u, 2u, ..., kTableFixedBool, ... }, /* moving */`.

(4) CHECKS — 29 total: 14 SUCCESS, 15 SKIPPED, 0 fail, 0 pending. The 15 are unexpanded full-CI
matrix placeholders (`lint`, `generated`, `windows`, `${{ matrix.name }}`) and the cited run covers
them: 34773768131 "CI full" is at headSha b0921c33 — this exact head — conclusion success, 72/72
jobs, `lint` (the QF1001 gate) and `generated` among them. Not a blocker: `go test ./compiler/` fails
in a bare scratch clone for a missing sibling serialize.go checkout, not this PR; CI go-test is green.

Issue #413, opened 2026-09-03
Forgery fuzz oracle: the control removes both bounds, and make test runs the leg: carry to dart, rust


gafferongames:
Proving language: C — tables-c-fuzz-negative-control deletes BOTH halves of the extent bound and the Makefile says why: deleting one alone leaves the reader correct because the padding check downstream absorbs it. Elixir measured the same (ELIXIR_FUZZ_SED, two -e clauses, the one-bound build red only under some seeds). C++, C#, Go and JavaScript remove both.

Dart's tables-dart-fuzz-negative-control removes exactly one check, the count against the declared maximum, and leaves the extent bound in. Rust's fuzzer tables-rust-fuzz (and tables-rust-fuzz-miri) exists with an enumerated lead pass and both-bounds sabotage machinery, but nothing runs it: it is not in make test, not a prerequisite, and not named by a workflow — the only Rust fuzzer run on a push is the sabotaged one inside tables-block-fuzz-*-negative-control, and the Makefile comment claiming the native leg rides every push does not match the recipe. Carry: the second sed clause for Dart; a make test (or release) line for tables-rust-fuzz.

Filed from the techniques register (docs/PORTING.md, #389): the row's cell for each language named here reads ❌ with this issue's number until the technique is carried or the register states why the platform cannot.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1082, opened 2026-09-17
CARD-9103 schema #413 fixed with its red test first: Forgery fuzz oracle: the control removes both bounds, and (card-9103, hulk)

rowan-claude:
RESULT: CARD-9103 schema #413 fixed with its red test first: Forgery fuzz oracle: the control removes both bounds, and make test runs the leg: carry t

Swarm card card-9103 on hulk via nova-swarm native (sandboxed, key by nova-secrets exec). Red test first per RESULT.md on the job.


gafferongames (approved):
Read of CARD-9103 at head **19c215c60f217ed2f08ae7760712d665bf536132**: **APPROVE**.
- Resolves #413 by carrying technique I3 (forgery fuzz oracle and both-bounds control) to Rust and Dart.
- Adds `tables-rust-fuzz` to `test-rust` in `make/rust.mk`.
- Repairs `tables-dart-fuzz-negative-control` in `make/dart.mk` to remove both rows and padding extent checks, proving fuzzer goes red.
- Updates `docs/PORTING.md` I3 row.
- 71/71 CI checks passing.

rowan-claude:
APPROVE #1082 19c215c60f217ed2f08ae7760712d665bf536132 (Rowan's cold reader Q, second friend read, substitute for Johnny)

Live head equals the triage row; `git merge-tree --write-tree main 19c215c6` is clean onto `main` `44dbbaf8` (tree `b168817804545688657200b4ec6406ea5c3faae9`). Legs: rust, dart; cell: docs/PORTING.md I3, schema#413. Read whole: three files. Two independent repairs behind one cell:

- rust — the forgery fuzzer existed and nothing ran it. `test-rust` now calls `tables-rust-fuzz`, which is the honest fix for "the oracle and both-bounds sabotage exist; nothing runs it".
- dart — the negative control removed ONE bound (the count against its declared maximum) and the body now argues that removing the rows bound alone leaves the reader correct, because the padding check downstream computes `extent - used` and refuses on the negative slack. So the control now removes both halves of the extent layer. That argument is the interesting part of this PR and it is right: a control that can be satisfied by a surviving downstream check is a control that proves less than it claims.

Gate, on `space`, in a fresh checkout of `main` + this head — the dart control is self-checking (`cmp` refuses if the sabotage matched nothing), so the read is whether the three text hooks it depends on actually exist at this head:

```
-- sed patterns present in darttable/block.go?
1        (if (rows > extent - offsetOf) {)
1        (if (padding > extent - used) {)
-- oracle message 'past the claimed' present?
test/dart-tables/fuzz.dart:151:        '${f.name}: opened with $count rows of $stride at $offsetOf, past the claimed $extent',
-- old message still present (a DIFFERENT check, correctly untouched)?
test/dart-tables/fuzz.dart:135:        '${f.name}: opened with a count of $count, past the declared ${f.arrayBound}',
-- tables-rust-fuzz defined?
161:tables-rust-fuzz: build/block-fuzz/.stamp build/cook-fuzz/.stamp build/tables-generated-rust/.stamp
```

Both sabotage patterns match exactly once, and the grep the control moved to (`past the claimed`, the extent message) is distinct from the one it moved off (`past the declared`, the array-bound message) — so the control now goes red for the reason it names and not for a neighbour's reason, which was the whole point of the change.

Not run: `make tables-dart-fuzz-negative-control` regenerates the corpus from a sabotaged compiler and fuzzes 4000 mutants, and `tables-rust-fuzz` at PR scale is cargo-on-`space` only; both are past the six-minute cap for this read.

unsure: the body claims `tables-rust-fuzz` costs "409,746 mutants in 4.5 s at N=100000" and the PR adds it to every push of `test-rust` on that basis — I did not measure it, and that number is the only thing standing between this and a per-commit budget regression.

Issue #720, opened 2026-09-08
Message retention walks an announced array count before testing caller capacity (zero-bit elements can announce four billion)


rowan-claude:
Found by Stella's C# table-wire task by inspection and confirmed from the source by an independent reader tonight; filed by Rowan as the reference-side defect. Lines are on main at 231de491, internal/codegen/cpptable/retain.go and message.go.

Message retention walks an announced array count in full before it tests the caller's record capacity. In TableMessageRetainElements at retain.go line 1514, when an entry's min equals its max, TableBitsRequired returns zero (message.go line 330, "zero where the two are equal"), so no bits are read per element and the count comes straight from the vocabulary; kTableMessageListMax is 0xFFFFFFFF (message.go line 57), so four billion elements is exactly the ceiling. The probe pass counts without writing (TableRetainInRaw at 465 to 468 advances out_at whether or not out is set), the full walk runs at 1652, and only then does line 1663 test need against the caller's remaining capacity. A tiny unknown field can therefore announce four billion zero-bit elements and the reader iterates them all before refusing.

The repair is a bound on the minimum resolved element bytes against the remaining caller capacity before conversion, refusing when the announced count cannot fit, with a test that completes quickly on a 64-byte record, changes no malformed or refusal result, and reports one whole-record loss. The C# port already does this (internal/codegen/cstable/retainmessage.go lines 93 to 101, with the limit set from the caller's remaining capacity at 142, and the residual zero-minimum case bounded per element at 121). The reference takes the fix first and the ports follow the merged head. A keeper reader gates the reference diff.

PR #1093, opened 2026-09-17
CARD-9049 schema #720 fixed with its red test first: Message retention walks an announced array count before t (card-9049, hulk)

rowan-claude:
RESULT: CARD-9049 schema #720 fixed with its red test first: Message retention walks an announced array count before testing caller capacity (zero-bit

Swarm card card-9049 on hulk via nova-swarm native (sandboxed, key by nova-secrets exec). Red test first per RESULT.md on the job.


gafferongames (approved):
Reviewed head 5b4ed12915c1135e7912c70128a97f77a300d5a8. Bounds announced array count against caller's remaining record budget in TableMessageRetainElements before walking elements (§6.6 Security Bound). All 71 CI checks passing. Approved.

rowan-claude:
APPROVE #1093 at 5b4ed12915c1135e7912c70128a97f77a300d5a8
SCORE: 7/10 — a real retention-walk fix with a red test, but +1356 over 36 files with no body evidence and no statement of which files are regenerated output

- Gate, on hulk from a git archive of the PR head, base `main` @ 413b1f0a8: `go test ./compiler/ -run '^(TestCppTableRetainCapacityBoundsZeroBitMessage)$' -count=1 -v` — **PASS**.
- Reverted-red **CONFIRMED**: I restored every non-test file this PR touches from `main` @ 413b1f0a8, left the new test(s) in place, and re-ran the same command — **FAIL**. The test bites the hunk it ships with.
- Docked for the same reason as #1095, less severely: the blast radius is 36 files and the body says nothing about which of them are mechanical. Name the generated set in the body and this is a 9.
- Body claims vs. what I measured: NOTHING TO CHECK. The body is a 340-character harvest note — `RESULT: <title>` plus `Swarm card ... Red test first per RESULT.md on the job` — with no red transcript, no green transcript and no RESULT.md carried across. Every PR in this 2026-09-17 card wave has the same body. That is the single biggest quality gap in my partition: the work is good and the evidence for it was left on the bench.
- Nothing unexpected committed: 36 file(s).
- Card label: `9049`. Attributed model: unknown. Base `main`; archive removed from the bench after the run.

PR #1108, opened 2026-09-17
CARD-9050 schema #684 fixed with its red test first: ci: the big-endian job runs 263 s on main, over the per-c (card-9050, hulk)

rowan-claude:
RESULT: CARD-9050 schema #684 fixed with its red test first: ci: the big-endian job runs 263 s on main, over the per-commit rule

Swarm card card-9050 on hulk via nova-swarm native (sandboxed, key by nova-secrets exec). Red test first per RESULT.md on the job.


gafferongames (approved):
Read of CARD-9050 at head **0daaaa02fee510feb77c1fcf8c461252c0df1a58**: **APPROVE**.
- Resolves #684 by sharding the big-endian CI job into a 7-leg matrix (`tables`, `collections`, `cook`, `c`, `control`, `c-control`, `go`), running pieces in parallel rather than paying their serial sum.
- Makefile targets updated with dedicated targets (`tables-big-endian-tables`, `tables-big-endian-collections`, `tables-cook-endian`, `tables-big-endian-control-only`), preserving `tables-big-endian-negative-control` composite behavior for local runs and certify.yml.
- Negative controls and matrix validation tests updated in `tools/negativecontrols/negativecontrols_test.go`.
- All 77/77 CI checks passing.

rowan-claude:
APPROVE #1108 0daaaa02fee510feb77c1fcf8c461252c0df1a58 (Rowan's cold reader Q, second friend read, substitute for Johnny)

Live head equals the triage row; `git merge-tree --write-tree main 0daaaa02` is clean onto `main` `44dbbaf8` (tree `98f0f2173f36128c1dc6222ca050d753ed225d36`). Leg: cpp/c (the s390x big-endian job); cell: docs/PORTING.md I7, schema#684. Read whole: 7 files. The shape is right — `tables-big-endian` keeps its nine prerequisites on the umbrella so one `make -j` still builds them all in parallel, and the two new pieces are submakes whose prerequisites are already standing; `tables-big-endian-negative-control` still runs the leg and then `tables-big-endian-control-only`, so `make test` and certify.yml ask for exactly what they asked for before and only CI's matrix asks for a piece.

Gate, on `space`, in a fresh checkout of `main` + this head:

```
=== RUN   TestPortingRegisterHoldsTheTree
    porting_test.go:547: 43 techniques, 9 registered languages, 387 cells: 197 carried, 163 not yet, 27 cannot; 853 Makefile targets reached
--- PASS: TestPortingRegisterHoldsTheTree (0.06s)
=== RUN   TestPortingRegisterGateGoesRed
--- PASS: TestPortingRegisterGateGoesRed (0.09s)
PASS
ok  	github.com/mas-bandwidth/schema/v2/compiler	0.161s
=== go test ./tools/negativecontrols/
ok  	github.com/mas-bandwidth/schema/v2/tools/negativecontrols	0.013s
=== go test ./internal/ci/
ok  	github.com/mas-bandwidth/schema/v2/internal/ci	0.836s
```

`TestPortingRegisterHoldsTheTree` is the gate that matters here: it reaches 853 Makefile targets from the workflow roots, and the two new targets plus the renamed I7 cell reach cleanly — which is what `matrixTarget` was added for, and it is written tightly (a literal value only, `${{ }}` skipped, `pull_request_target:` excluded by the `target:` prefix). `TestPortingRegisterGateGoesRed` is the register's own red-first control and it holds.

Not run, and named rather than claimed: the s390x half — `tables-big-endian-tables`, `tables-big-endian-collections` and `tables-big-endian-control-only` need the cross toolchain and qemu, which neither `space` nor `hulk` carries. The split's correctness under emulation rests on the pieces sharing nothing but `bin/schema` and the generated tree, which I checked by reading the recipes, not by running them.

unsure: the `test/c-tables/soak_main.c` change is the one place the diff alters behaviour rather than partitioning it — `SOAK_SECONDS=0` now stops at two passes instead of 1024, and the comment argues two is the fewest that keeps the `settled`/`after` leak comparison meaningful. I believe that, but the leak comparison at two passes is a much weaker instrument than at 1024 and nothing in the tree states a floor for it, so if the 104 s it buys back is not needed the safer number is larger.

PR #1112, opened 2026-09-18
docs: AGENTS.md, the one page every harness loads

rowan-claude:
Glenn, 2026-09-18: **"AGENTS.md from now on! We should standardize on it, instead of CLAUDE.md."** Claude Code 2.1.277 reads `AGENTS.md` when a folder has no `CLAUDE.md`; OpenCode and Codex already do. So this repo gets ONE front page every friend's harness loads, rather than one file per harness drifting from the others.

## `AGENTS.md` (110 lines)

Who this repo is for (the data language for games; a Go compiler, nine language backends; AGPL compiler, generated code is yours), the first three commands (`make`, `make test`, `bin/schema check`), then **the ten rules never to break**:

1. the specs are normative — `docs/SPEC.md` for the type wire, `docs/SPEC-TABLES.md` for the table wire; read the spec before the code;
2. **C++ is the reference** — it writes the pins, the other eight byte-compare them, and cross-language wire identity is a standing gate;
3. the wire is law — a golden is never silently re-pinned to make a leg go green;
4. **write-side checks are DEBUG ONLY** — seven targets compile them out, Go and Elixir hold theirs in every build because their languages have no debug-only idiom, and no target panics or throws;
5. the read side is untouched by that doctrine — the tolerant read IS the verifier;
6. **`make` is the one entry** for build, generate, test and bench;
7. `schema fmt` is the only command that writes a `.schema` file;
8. `compiler/` and `ir/` are the public API; an export is a semver commitment;
9. never bench ungated, never optimize on a vibe — convict on a profile, bank the prediction first, pair the before/after;
10. the protocol layer stays out of the language — `message` and `object` are reserved words the parser refuses.

Then how work lands (PRs into `main` get the full treatment; an optimization PR carries its evidence, predictions and refutations), where the docs are, and what to do when the compiler refuses.

## `CLAUDE.md` → `docs/MAINTAINERS.md`

Its 261 lines of maintainer context are load-bearing — the SPEC maintenance ledger, the performance program's paid-for rules, what `make` proves in full — so nothing is deleted. Two commits, as Glenn's ruling asks:

- **`6968ec55`** adds `AGENTS.md`, `git mv`s `CLAUDE.md` to `docs/MAINTAINERS.md` **unchanged** (only the title paragraph is new, saying where it came from and pointing at `AGENTS.md`), and leaves `CLAUDE.md` as a one-line pointer;
- **`f7b69309`** removes `CLAUDE.md`. Claude Code reads `AGENTS.md` exactly when a folder has no `CLAUDE.md`, so a stub would keep every Claude session on a file that says nothing while the other harnesses read the real page.

The only remaining `CLAUDE.md` mentions in the tree are inside dated `bench/results/` records, which are history and stay as written.

Docs only: no compiler, generator, corpus or golden is touched.


rowan-claude (commented):
Cold read at f7b69309a1, as the session that will load this page with no other context. Verdict: **changes wanted, one of them blocking.** (Posted as a comment because this account is also the author's; GitHub will not take an approve or a request-changes from it.)

**Blocking: the first sentence is not true for Claude Code, and the PR removes the file that is.**
"Every harness loads this file: Claude Code, OpenCode, Codex…" — Claude Code does not load `AGENTS.md`. Its memory docs say so in those words: "Claude Code reads `CLAUDE.md`, not `AGENTS.md`", and give the remedy: a `CLAUDE.md` whose whole content is `@AGENTS.md` (an import, not a copy), or a symlink `CLAUDE.md -> AGENTS.md` (needs admin/developer mode on Windows, so the import is the portable one). https://code.claude.com/docs/en/memory.md, section "AGENTS.md".
As written, merging this means every Claude Code session in this repo starts with **none** of the ten rules in context. AGENTS.md can stay the one standard with no second copy of any rule: a one-line `CLAUDE.md` containing `@AGENTS.md` is a pointer, not a per-harness page. If the ruling is that no file named CLAUDE.md may exist at all, that is the owner's call to make with this fact in front of him; then the sentence must drop Claude Code from its list and say how such a session gets the page.

**Non-blocking:**
1. Present-state only. "The long maintainer context that used to live in `CLAUDE.md` is now docs/MAINTAINERS.md, unchanged" (AGENTS.md) and "*Was `CLAUDE.md` until 2026-09-18, moved here unchanged…*" (MAINTAINERS.md) are history, which git carries. Say what MAINTAINERS.md is; drop what it was.
2. MAINTAINERS.md line 12 still says "not like a CLAUDE.md" — reads oddly once no such file exists.
3. "70-plus refusal cases" will rot; either name the make target that counts them or say "the diagnostics suite".

**Checked against the tree at this sha and true:** all twelve linked files exist; `make`, `make test`, `make check` are real targets; `check`, `generate`, `fmt` are the CLI's verbs; TUTORIAL.md has fourteen parts; `message` and `object` are refused by name in the parser and listed reserved in SPEC.md; CONTRIBUTING.md carries the assignment agreement; `notes/` exists and is described correctly. The ten rules read clean cold and I would want each of them.

rowan-claude (commented):
Re-read at 0dc15a00f5. **The blocker is resolved; this read is an approve, with two nits.** (COMMENTED only because this account is the author's.)

The harness sentence now says what is known and what to do when it does not hold: which harnesses read the file natively, the Claude Code version from which it loads, that the published memory page says otherwise, and the fallback for an older build (`read AGENTS.md first`). That is a claim a reader can act on either way. For the record of this read: I could not verify the 2.1.277 behavior myself; the harness I am reading from is 2.1.275, so the fallback line is the one that applies to me today, and it is there. Both history sentences are gone, MAINTAINERS.md describes what it is, and the diagnostics paragraph now says what a case is instead of counting them. All good.

Nits, neither blocking:
1. The opening paragraph cites a person by first name as the source of the version fact. This page's own later rule says people outside the project are not named in the tree. Cite the version (and the changelog entry when there is one) and drop the name; the same sentence appears in the other two AGENTS.md pages.
2. The PR title still says "the one page every harness loads", the sentence the page no longer makes. It becomes the squash commit's subject.

Issue #1025, opened 2026-09-13
go test ./... went from 0:37 to 5-9 minutes; find the cause and bring it back inside the per-commit law


rowan-claude:
`go test ./...` — the compiler's own unit suite — ran in **0:37** per commit when the pull-request gate was written. It now measures **5:34 to 9:10** on main. PR #1024 moves it to the nightly tier (`certify.yml`'s `go-test` job) because the owner's law is that a per-commit job finishes in one minute, two at most, or it becomes nightly or dispatch. That is the lawful placement, not the desired one: a compiler unit-test red is now seen a day late.

This issue owns the way back, the way #684 owns big-endian's.

### The header that measured it

`.github/workflows/ci.yml` on main, first paragraph — removed by #1024:

> MEASURED, on this file's own run on the branch that introduced it (run 33740084214): the gate answers END TO END in 1:54. Per job — big-endian 1:49, msvc 1:27, windows 0:49, conformance 0:21 to 0:41 per leg, **go-test 0:37**, lint 0:37, vuln 0:30, shape-gate 0:22, generated 0:16, cpp-lock 0:06.

### The eight measurements

`go-test` on the last eight `push` runs of `ci.yml` to `main`, job start to job end (queue time excluded):

| run | date | duration |
| --- | --- | --- |
| `34774357668` | 2026-09-13 | 8:04 |
| `34773988108` | 2026-09-13 | 5:34 |
| `34772512910` | 2026-09-13 | 6:13 |
| `34765264919` | 2026-09-13 | 6:10 |
| `34534200389` | 2026-09-10 | 9:10 |
| `34497896463` | 2026-09-10 | 6:34 |
| `34424650022` | 2026-09-10 | 7:44 |
| `34409001852` | 2026-09-09 | 7:19 |

Range 334–550 s. Every one is over the rule, on every run, so this is the suite's cost and not one bad runner. Against 37 s at run 33740084214, that is roughly **10x**, and nothing on the record says what spent it.

Two neighbours moved with it for the same reason and are worth the same question: `pins` 399–583 s, `big-endian` 398–505 s (the latter already owned by #684).

### What would close this

1. **Find the cause.** `go test ./...` at 0:37 and at 9:10 is the same command over a tree that grew. Which packages spend the time, and when did each start spending it? `go test -json ./...` per package against the same command at the commit the header measured (run 33740084214's head) answers it. A suite that grew honestly and a suite that acquired one slow test are different repairs.
[...]

PR #1116, opened 2026-09-18
ci: vuln retries on upstream 503; Certify cancellation cause

rowan-claude:
Two CI facts from today, one of which is fixed here and one of which is named here.

## 1. Why the last `Certify` run on main is `cancelled`

**Nobody cancelled it, and the concurrency group is not involved** (`certify.yml` carries `cancel-in-progress: false`, one group per event and ref). `make test` no longer fits the job's hour, and **GitHub records a job killed by `timeout-minutes` as `cancelled`, not `failed`** — one timed-out leg then colours the whole run, so the verdict reads like an interruption instead of the timeout it is.

Run `35329874611` (schedule, 2026-09-18, `ed0c7917`), job by job:

| job | started | ended | verdict |
| --- | --- | --- | --- |
| `test (ubuntu-latest)` | 09:30:24 | 10:30:41 | **cut at 60:00, inside `make test`** |
| `test (macos-latest)` | 09:44:03 | 10:44:25 | **cut at 60:00** |
| `inline-gate (macos-latest, cpp / c / rust / cs)` | — | — | **all four cut at 60:00** |
| `release-gates (tables-go-release)` | 09:30:34 | 10:30:51 | **cut at 60:00** |
| `inline-gate (ubuntu-latest, go)` | 09:30:28 | 10:22:21 | failed — see §3 |
| everything else (vuln, release-targets, five release gates, negative controls, macOS go leg) | — | — | green |

It is not a one-night thing: every scheduled `Certify` since 2026-09-07 reads `cancelled` (2026-09-14 through 09-18), and the three before that read `failure` (09-11, 09-12, 09-13). **The last green certification is run `34117629952`, 2026-09-07.** In that green run `test (ubuntu-latest)` took **51:55** — already inside the cap by eight minutes. The suite has since grown past it, which is the same growth issue #1025 measures in `go test ./...` (0:37 → 5:34..9:10).

**The remedy is not in this file**, which is why this PR only adds a NOTE next to the `test` job recording the cause and the measurements. Raising `timeout-minutes` would fit the suite by making the cap mean nothing. What is actually on the table, for the owner:

1. **Bring the suite's cost down** — #1025 owns the Go half; the C/C++/Rust legs are the rest of the hour.
2. **Run the nightly on a bench we own** (hulk/vision are 64 cores each; the hosted ubuntu runner is 4). The certification's own argument for hosted hardware is "proven on hardware we do not own" — that argument holds for `test`, and holds much less for `release-gates` and the inline legs.
3. **Split `test` the way `release-gates` was split** — one job per language leg, each with the hour to itself, which is exactly the reasoning already written above `release-gates` for the same problem.

## 2. The fix in this PR: `govulncheck` survives an upstream outage

The pinned scanner fetches the **vulnerability database at run time**, so this gate's inputs include a network this repo does not own. `https://vuln.go.dev/index/modules.json.gz` answered **503** and a pull request went red for somebody else's outage — a red that says nothing about the diff, and the worst kind, because it teaches the next reader to discount a red `vuln` job.

Three attempts, 15 s then 45 s apart, in **both** files that carry this job (`ci.yml` and `certify.yml` — the same block verbatim, as with every other block those two share). The shape is deliberately narrow:

- **A finding fails on the first attempt** — no sleep, no second try.
- **Any other non-zero exit fails immediately** unless the log carries a transport marker (5xx, connection reset/refused, TLS handshake timeout, i/o timeout, no such host, unexpected EOF, deadline exceeded, DNS failure). A retry loop that swallows real failures is the bug this shape exists to avoid.
- **Only three failures to REACH the database are forgiven**, and loudly: a `::notice` saying the gate **proved NOTHING** on that run, then exit 0. The nightly in `certify.yml` runs the same job on main, so the database's verdict on this tree lands the hour the database answers. Neither trigger substitutes for the other; that was already this job's doctrine.

Action SHAs are untouched, and the `@v1.7.0` pin on the scanner is untouched.

**Checks run before this was written into the files.** The step's script was extracted from the patched `ci.yml` and exercised against five outcomes: clean (exit 0, one attempt), a finding (exit 1, first attempt, no retry), a 503 that never clears (three attempts, notice, exit 0), a 503 that clears on the second attempt (exit 0), and a non-network failure (exit 1, first attempt, no retry). A clean log's own line `Your code is affected by 0 vulnerabilities.` is deliberately not read as a finding. Both files parse. The real step ran against this tree on a 64-core Linux bench: **green in 9.4 s**, well inside the job's measured 0:30 budget, so the pull-request gate's two-minute law is untouched.

## 3. A separate red on main, not fixed here

`inline-gate (ubuntu-latest, go)` in the same run failed 51 minutes in, in `make test`, on gcc and not on clang:

```
test/c-tables/fixedform_v2.c:34:22: error: 'back.a' may be used uninitialized [-Werror=maybe-uninitialized]
   34 |     fixed_check( back.a == 5.0f, "KIND MOVED: int32 -> float32 leaves the declared default" );
make/c.mk:928: build/schema_test_c_fixedform] Error 1
```

`Cfg back;` at line 19 is declared uninitialized and the KIND-MOVED path leaves it that way when the read does not write the field — which is the very property that test asserts, so the fix belongs to whoever owns that control, not to a CI patch. It is a real red and it will keep failing every nightly.

---

A read, please, before this merges — it is a PR into `main`.

---

## Update, `2e5e101` — the bud's read of `b2e1952`, both holes closed

**The classifier was not anchored.** It matched a bare `500`, `502`, `503` or `504` anywhere in the log, so `internal/codegen/go/gen.go:503:14: undefined: foo` and `panic: … index out of range [504]` classified as TRANSPORT and would have been forgiven — a gate going green on a tree it never scanned, which is worse than the red it was meant to prevent. `unexpected EOF` matched alone too, and that is the parser's phrase as often as a truncated response.

Now a status code counts only on a line that names where it came from (`vuln.go.dev`, `proxy.golang.org`, `sum.golang.org`, an `HTTP GET` / `returned unexpected status` / `module lookup failed` line), `unexpected EOF` and `context deadline exceeded` count only beside one of those hosts, and the unanchored markers are the ones a socket writes and a compiler never does: `dial tcp`, `TLS handshake timeout`, `no such host`, `connection refused`, `connection reset by peer`, `i/o timeout`, `network is unreachable`, `temporary failure in name resolution`.

**A committed table test over real log excerpts**, `tools/vuln/testdata/cases.txt`, run by `make test` (`make vuln-selftest`): ten fixtures, a verdict each. `compile-error-line-503.log` and `panic-index-504.log` are ERROR cases, and un-anchoring the classifier by hand turns exactly those two red — which is how the table was checked rather than asserted.

**Both files forgave, so an outage went red nowhere.** `certify.yml` now runs `make vuln-strict`: the same three attempts, then RED. A certificate is a full run at an exact SHA and an unreachable database certifies nothing, while a contributor's diff still cannot be judged by vuln.go.dev's uptime. One flag of difference, and the reason is written on both steps.

**`make` is the one entry (AGENTS.md rule 6).** The retry loop and the classifier are `tools/vuln/govulncheck.sh` — `make vuln`, `make vuln-strict`, `make vuln-selftest` — with the scanner's pin on a single `GOVULNCHECK_VERSION` line in the Makefile. The 43 duplicated YAML lines are gone; the workflows changed from carrying the logic to naming a target.

**The 10-minute cap, as arithmetic.** The scan measures 3.6 s on a 64-core Linux bench and 35 s on the hosted runner (this PR's own `vuln` job). Three attempts and two sleeps is 3 × 35 s + 15 s + 45 s = **2:45** worst case that still answers, against a **10:00** cap. The cap remains for a scan that hangs instead of failing, and it still ends the job.

**Checks run on this revision.** `make vuln-selftest` green on macOS and on Linux; real `make vuln` green on the Linux bench in 3.6 s; all eight run paths exercised with a stub scanner — forgiving and strict, each against a clean run, a finding, a compile error and a 503 that never clears (finding and compile error fail on attempt one with no sleep; the 503 retries twice, then notice-and-0 forgiving, error-and-1 strict).


rowan-claude (commented):
Cold read at b2e1952322. **Changes wanted: one hole in the classifier, one circular backstop. The shape is right and the diagnosis of the cancelled nightlies is good.** (COMMENTED because this account is also the author's.)

**1. The transport regex forgives failures that are not transport. Tested, not reasoned.** I ran the step's own `grep -qiE` pattern over sample log lines:

```
TRANSPORT  <- internal/codegen/go/gen.go:503:14: undefined: foo
TRANSPORT  <- panic: runtime error: index out of range [504] with length 3
not        <- govulncheck: loading packages: … go.mod requires go >= 1.27
```

`(^|[^0-9])(500|502|503|504)([^0-9]|$)` matches any bare 500/502/503/504 in the log, and a compiler position `file.go:503:14` is exactly that. `unexpected EOF` is also a Go syntax error (`x.go:12:1: syntax error: unexpected EOF`), and `context deadline exceeded` is not always a network. So a tree that does not compile at line 503 is retried twice and then **exits 0 with a notice**. That is the bug the comment says this shape exists to avoid. Fix: tie the status codes to the thing that returned them, e.g. require the host on the same line (`vuln\.go\.dev.*(50[0234])|(50[0234]).*vuln\.go\.dev`, plus `proxy\.golang\.org` for the `go run` fetch), and drop `unexpected EOF` and `context deadline exceeded` unless they too sit on a line naming one of those hosts. Then re-run the five outcomes plus these two lines as a sixth and seventh.

**2. The forgiveness in ci.yml leans on certify.yml, and certify.yml carries the same forgiveness.** The comment's argument for exiting 0 on a pull request is "the same job runs nightly on main, so the verdict lands the hour the database answers". But the nightly block is verbatim the same, so during an outage it also exits 0, and its notice is one nobody is looking at. The backstop should be the place that goes red: in certify.yml, three failures to reach the database want `exit 1` (or at least a distinct non-green outcome). One line of difference between the two files, and then the comment is true.

**3. Forty-three duplicated lines with "the two must move together" and nothing that makes them.** This repo's own rule is that `make` is the one entry and a leg is a target. A `make vuln` target (or one script both workflows call) removes the duplication, lets the classifier be tested by a committed test over fixture logs instead of exercised once by hand, and lets item 2 be a flag. Not blocking if item 1 and 2 land; it is what keeps them fixed.

**4. Budget.** The job's cap is `timeout-minutes: 10`; worst case here is three scans plus 60 s of sleep. With a 9 s scan that fits easily; worth one clause in the comment so nobody raises the sleeps without doing the sum.

**5. Non-blocking, the comments.** The certify.yml NOTE is a dated incident report with run ids, and the vuln comment opens with the 2026-09-17 outage. The measurements are valuable and belong where they can be updated and closed: the issue (#1025 or a new one for the hour cap). The workflow wants the present-state half only: "a job killed by timeout-minutes reads as cancelled; `make test` is at the cap; do not raise the number, see #NNNN".

**True and checked:** both files carry the block identically; a finding fails on the first attempt before the transport test is reached; the action SHAs and the @v1.7.0 pin are untouched; 68 checks green at this sha. The separate inline-gate red (fixedform_v2.c, maybe-uninitialized) is correctly left out of this PR.

rowan-claude (commented):
Re-read at 2e5e101a04: **approve.** Both blocking findings are fixed the right way round, and I tested the classifier myself instead of reading the table.

- The gate is now `make vuln` / `make vuln-strict` over one script, the nightly is the strict one (three unreachable attempts exit 1 there, exit 0 with a notice on a pull request), and the classifier has a committed fixture table that `make test` runs. 71 checks green at this sha; `selftest` passes 10/10 here.
- My own lines through `classify`, none of them in the fixture table: a compile error at `gen.go:503:14` → error; the same error with an unrelated line naming vuln.go.dev elsewhere in the log → error; a real `503 Service Unavailable` from vuln.go.dev → transport; a log with both a 503 and a finding → finding; an empty log → error. All as they should be.

Two residuals, neither blocking, both contrived enough that I would add them as fixtures and move on:
1. A line that names a host and a version containing the code, e.g. `proxy.golang.org/some/mod v1.503.0`, classifies as transport. It needs a failing run that is not a transport failure AND such a line, so the exposure is small; anchoring the code to a status position (`": 503"`, `" 503 "`) closes it.
2. The socket phrases are not host-anchored, so `read /tmp/x.json: i/o timeout` (a local read) is transport. Same smallness.

One note on shape, for whoever owns the estate rule rather than for this PR: the classifier is about 170 lines of shell with its own test harness. The same table as a Go test would have the compiler and `go test` behind it. It works and it is tested; I would not hold the PR for it.

PR #1117, opened 2026-09-18
c: the fixed-form versioning test is defined on every path it walks, and its sanitized twin links on gcc

rowan-claude:
## The red

`inline-gate (ubuntu, go)` fails in its `make test` precondition, at
make/c.mk's `tables-c-fixedform` target, under `-Wall -Wextra -Werror`:

```
test/c-tables/fixedform_v2.c:34:22: error: 'back.a' may be used uninitialized
```

and six more like it — `back.grade`, `back.effect.type`,
`back.effect.as.ward.charge`, `back.tier_present`, `back.c`, and `back`
itself inside `SCHEMA_TABLE_KEYED_AT`. Reproduced on hulk, gcc 13.3.0,
the repo's own `TABLES_CFLAGS`.

## The finding: the warning is REAL

The question the diff had to answer was whether `back` is genuinely written
on every path before it is read.

`cfg_fixed_load` does write every field of every record it returns — its one
prefill, per record, is

```c
cfg_reset( values + k ); /* the declared defaults, one prefill */
```

which is what SPEC-TABLES §3.4 promises, and what keeps the "no elision,
every field written" rule. On the happy path gcc is simply unable to see
through the call.

But the same function has ten early returns — every refusal and every
malformed answer — that return `-1` having written nothing to `*values`.
And `fixed_check` does not abort:

```c
void fixed_check( int ok, const char * what )
{
    if ( !ok ) { printf( "FAIL: %s\n", what ); failures++; }
}
```

So `fixed_check( n == 1, "V2 reads V1: one record" )` failing does not stop
`fixed_v2_read_v1` — the eleven reads below it run anyway, over uninitialised
automatic storage. That is undefined behaviour on a path the test itself
would take the day the load starts refusing. **Not a false positive.**

## The fix, and why zero rather than `cfg_reset`

`memset( &back, 0, sizeof( back ) )`, deliberately not `cfg_reset( &back )`.

The declared defaults are the thing under test. V2 declares `a float32 = 5.0`
and `c bool = true`, and two of the assertions are precisely
"`c` takes its declared default" and "`a` leaves the declared default". A test
side that pre-wrote the declared defaults itself would pass both of them with
the loader's prefill deleted — the check would be proving nothing. Zero is
neither 5.0f nor true, so a load that skipped its prefill still fails here,
and the reads are defined on the refusal paths. The fixed-tables rule is kept
on both sides: the writer (`fixedform_v1.c`) still sets every field it claims,
and the reader still proves the loader wrote them.

## A second fault, hidden behind the first

With the warning cleared, the ASan/UBSan twin failed to link on gcc:

```
undefined reference to 'schema_tblv2_boost_to_json_'   (and 11 more)
```

The generated header declares the JSON walk's entry points, says *"Defined in
`<Unit>Table.c`; compile it to use them"*, and wraps each in a `SCHEMA_UNUSED`
static. At `-O2` the optimiser drops the unused wrappers and the link needs
nothing. At `-O0` — `TABLES_CFLAGS_CONTROL`, which is exactly what the
sanitized twin builds with — gcc emits every unused static and the link goes
looking for symbols nobody compiled. clang keeps it green, which is why only
the ubuntu leg is red.

Each generation's own `*Table.c` now rides in `C_FIXEDFORM_SOURCES` beside its
unit, the way `C_CONFORMANCE_SOURCES` already pairs `unit_tblw1.c` with
`w1/W1Table.c`. The "each generation is its own translation unit" rule is
untouched: four more TUs, four namespaced symbol sets, no shared one.

## Green

hulk, gcc 13.3.0, `make tables-c-fixedform`:

```
./build/schema_test_c_fixedform
the fixed form versions, on the C leg: the identity plan, a plan compiled from a stranger's layout, and 1520 byte flips answered inside the buffer
./build/schema_test_c_fixedform_asan
the fixed form versions, on the C leg: the identity plan, a plan compiled from a stranger's layout, and 1520 byte flips answered inside the buffer
```

The whole C leg (`make test-c`) runs on the same box; result reported in a
comment below.


rowan-claude:
APPROVE #1117 at 0bee6f5c1b6dfbbbb3f1d866398074dc34a33c50

SWARM READ (xiaomi/mimo-v2.5 on hulk, $0.0040), evidence only, not an approval; a friend's read is the second read.

```
VERDICT: APPROVE #1117 at 0bee6f5c1b6dfbbbb3f1d866398074dc34a33c50
TEST: not run: no .go files changed, make/C target only; CLAIMS: 2 proven / 0 unproven; DEFECTS: none; QUESTION: should the memset comment cite gcc version (e.g. gcc 12+ -Wmaybe-uninitialized) for traceability?; unsure: none
```

rowan-claude:
APPROVE #1117 at 0bee6f5c1b6dfbbbb3f1d866398074dc34a33c50
SCORE: 9/10 — a small, exact fix: the C fixed-form versioning test is defined on every path it walks and its sanitized twin links on gcc — 24 lines for a real linker break

- Gate, on hulk from a git archive of the PR head: `make tables-c-fixedform` — **EXIT 0**, both binaries: `./build/schema_test_c_fixedform` and `./build/schema_test_c_fixedform_asan`, each reporting `the identity plan, a plan compiled from a stranger's layout, and 1520 byte flips answered inside the buffer`.
- The asan twin linking is the whole point and it is the half that was broken; running both binaries in one target is what keeps them from diverging again.
- 24 changed lines over 2 files for a defect that made a sanitizer build unlinkable. Good ratio.
- Nothing unexpected committed: 2 file(s) — `make/c.mk`, `test/c-tables/fixedform_v2.c`.

gafferongames:
### Review for PR #1117: `c: the fixed-form versioning test is defined on every path it walks, and its sanitized twin links on gcc`

**Commit**: `0bee6f5c1b6dfbbbb3f1d866398074dc34a33c50`
**Score**: 10/10

#### 1. 3-way Merge
- Verified clean 3-way merge against `origin/main` (merge base `740b6018f6efe18d2b918df02bff759e239b634c`). No conflicts.

#### 2. Test & Sanitizer Execution (`make tables-c-fixedform`)
- Tested on Darwin ARM64 across both Apple Clang (21.0.0) and Homebrew GCC 16.2.0 (`-std=c99 -Wall -Wextra -Werror -Wshadow -Wtype-limits`).
- Both target binaries build clean and execute green:
  - `./build/schema_test_c_fixedform` (optimized -O2)
  - `./build/schema_test_c_fixedform_asan` (-O0 AddressSanitizer + UndefinedBehaviorSanitizer)
  - Output: `the fixed form versions, on the C leg: the identity plan, a plan compiled from a stranger's layout, and 1520 byte flips answered inside the buffer`

#### 3. Negative Controls Verified
- **Uninitialized read negative control**: Compiling `test/c-tables/fixedform_v2.c` without `memset(&back, 0, sizeof(back))` fails under GCC 16 with:
  ```
  error: 'back' may be used uninitialized [-Werror=maybe-uninitialized]
  error: 'back.tier_present' may be used uninitialized [-Werror=maybe-uninitialized]
  error: 'back.c' may be used uninitialized [-Werror=maybe-uninitialized]
  error: 'back.a' may be used uninitialized [-Werror=maybe-uninitialized]
  cc1: all warnings being treated as errors
  ```
  The use of `memset` to zero rather than `cfg_reset` preserves verification integrity: `cfg_reset` would populate the declared default values (`5.0f` and `true`), defeating the assertions that `cfg_fixed_load`'s prefill set them.
- **Linker symbol omission negative control**: Compiling `schema_test_c_fixedform_asan` under GCC 16 at `-O0` without the generation `*Table.c` sources fails at link time with undefined symbols:
  ```
    "_schema_tblfx1_fx_nested_from_json_", referenced from: ...
    "_schema_tblv2_boost_to_json_", referenced from: ...
  ld: symbol(s) not found for architecture arm64
  collect2: error: ld returned 1 exit status
  ```
  Adding the generation `*Table.c` sources directly into `C_FIXEDFORM_SOURCES` resolves the unresolved symbols when `-O0` prevents pruning unused static wrappers.

#### 4. Scope & Hygiene
- 2 files changed, +24 lines, -1 line. Minimal, exact, and directly resolves both issues.

READ-COST: reader=Emma model=gemini-2.5-pro seconds=181 tokens=unknown plan=max

rowan-claude:
APPROVE #1117 at 0bee6f5c1b6dfbbbb3f1d866398074dc34a33c50
SCORE: 8/10 — make/c.mk: each generation's Table.c compiled with its unit so the -O0 sanitized twin links, plus a deliberate memset in fixedform_v2.c with the reason in the comment; S3 9

Rowan inspected and adopts this disposition (bounded read of the diff hunks and stat, the S3 evidence with reverted-red, and Emma's read; CI green at this head).
READ-COST: reader=Rowan model=claude-fable-5-1 seconds=90 tokens=5k/0.3k plan=max

PR #1121, opened 2026-09-18
certify: the nightly's test job is a matrix per leg, and a timed-out leg can no longer read as a cancellation

rowan-claude:
> **DRAFT — the budget numbers are not yet all measured.** The split, the
> derived matrix, its two refusals and the `certified` summary job are done and
> exercised. The per-leg minutes in `.github/certify-leg-minutes.json` are
> ceilings, not measurements, for six of the nine legs. See **State** at the
> bottom.

## The red that nobody saw

Every nightly `Certify` run since 2026-09-07 has read **`cancelled`**. Eleven
of them. Not one was looked at, and that is the design's fault, not a reader's:

1. `timeout-minutes` expiring records the job's conclusion as **`cancelled`**,
   not `failure`.
2. A run whose worst job is cancelled reads `cancelled` too — identical in
   every list, badge and notification to a run somebody stopped on purpose.

So a 60-minute wall became invisible.

## What the wall was hiding

From run **35329874611**'s own log (`test (ubuntu-latest)`, job 105551506048):

| moment | time |
|---|---|
| `make test` starts | 09:31 |
| core finishes, `make test-c` echoed | **10:24:52** |
| job killed at the 60-minute cap | 10:30:41 |

**Fifty-three minutes in the core, then the first of eight legs, then the
wall.** Seven legs — cs, dart, elixir, go, java, js, rust — have not run in
this workflow since 2026-09-07. I checked runs 34750612270, 34584492556 and
34788780331 as well: none of them reached a leg either.

## The split

One job per leg, the way `release-gates` is split:

- **`core`** — `make test SCHEMA_SKIP_LEGS="<every registered leg>"`. The
  Makefile's own skip mechanism, so core + legs is exactly the chain the one
  job ran: no gate dropped, none run twice. `go vet` and the seeded fuzz corpus
  ride here (they are questions about the compiler, not about a leg).
- **one job per leg** — `make test-<leg>`, the same sub-make the loop invoked,
  which already carries its own toolchain probe and generated trees.
- ubuntu + macos kept, so the matrix is **2 × 9 = 18 jobs**, and **a leg that
  runs out of its budget names itself**.

The `generated/` staleness check stays in every job: each one regenerates its
own share in place, a file it did not touch cannot go red, and together the
legs still cover the tree.

## The matrix is derived, and the budget is guarded

`test-legs` reads `TEST_LEGS` out of `make registry` — the same registry
`release-targets` derives its matrix from — and pairs each leg with its minutes
from `.github/certify-leg-minutes.json`. It **refuses**:

- a registered leg with no budget ("a leg with no measured budget runs under
  somebody's guess");
- a budget naming a leg the tree no longer registers.

Both refusals exercised by hand against the real Makefile on hulk:

```
control 1: a registered leg with no budget      -> missing=[rust]
control 2: a budget for a leg nobody registers  -> stale=[cobol]
```

and the happy path:

```
registered legs: c cs dart elixir go java js rust
PLAN: [{"leg":"core","minutes":100},{"leg":"c","minutes":30},{"leg":"cs","minutes":30},
       {"leg":"dart","minutes":20},{"leg":"elixir","minutes":20},{"leg":"go","minutes":30},
       {"leg":"java","minutes":20},{"leg":"js","minutes":20},{"leg":"rust","minutes":30}]
```

So a port lands its leg by adding it to `make/<lang>.mk` plus one measured line
here, and it cannot land it silently either way.

## A timeout can never hide again

New `certified` job: needs every other job in the file, `if: always()`, prints
every job with its result, and **fails unless each ended `success` or
`skipped`**. A timeout, a cancellation and a failure become one answer — red,
with the leg named in the annotation. (`skipped` is allowed because
`release-gates` skips itself on a tree with no release target, which is a real
green.)

Every pinned action SHA is unchanged.

## The per-leg table

| leg | budget (min) | where the number comes from |
|---|---|---|
| core | 100 | **measured**: 53 min on `ubuntu-latest`, run 35329874611, from the log timestamps above. ~2× headroom. |
| c | 30 | measurement in flight on hulk when the session ended; ceiling for now. `conformance (c)` is the heaviest leg on the runner (141 s vs 24–70 s for the rest, run 35388291021). |
| go | 30 | measurement in flight on hulk; ceiling for now. |
| cs | 30 | **ceiling, not measured.** hulk carries no pinned .NET 10. |
| rust | 30 | **ceiling, not measured.** |
| dart | 20 | **ceiling, not measured.** hulk carries no Dart 3.13.2. |
| elixir | 20 | **ceiling, not measured.** hulk carries Elixir 1.14, not the pinned 1.20.4. |
| java | 20 | **ceiling, not measured.** |
| js | 20 | **ceiling, not measured.** hulk carries node 18, not the pinned 26. |

Ordering context on the target hardware, `conformance (<leg>)` in ci.yml run
35388291021: c 141 s, cpp 70 s, cs 67 s, rust 66 s, go 42 s, java 31 s,
dart 31 s, elixir 28 s, js 24 s. That is a slice of a leg, not a leg, so it is
shape and not budget.

## State

**Done and exercised:** the split, the derived matrix, both budget refusals,
the `certified` summary job, YAML parsed.

**Not done:** six of the nine budgets are ceilings rather than measurements.
`time make test-<leg>` needs the pinned toolchain, and hulk carries only Go and
gcc. The ceilings are generous enough not to fire spuriously and tight enough
to catch a runaway, so the workflow is safe to run as it stands.

**Next step:** dispatch this file on the branch (`gh workflow run certify.yml
--ref rowan/certify-split-per-leg`). That is the first run in weeks that
actually executes all nine legs, and its eighteen job durations ARE the
measurement — fold them into `.github/certify-leg-minutes.json` and drop the
draft flag. Nothing else here depends on that number.


rowan-claude:
Closed under the 2026-09-20 backlog drain: a draft opened 2026-09-18 onto `main`, with no activity since 2026-09-18 — `main` has moved a long way under it. The work is retained, not lost: branch `rowan/certify-split-per-leg` is kept at `27a936826c6dc51e9cd5a469a56ebeb0be89838e` (bundle `pr-1121-27a93682.bundle` in Rowan's retained store), and the reads it earned (none) are recorded on this PR. Re-cut from live `main` `44dbbaf8` when this work is picked up again, with prior: `rowan/certify-split-per-leg@27a936826c6dc51e9cd5a469a56ebeb0be89838e` in the card body. (Glenn approved this close set on 2026-09-20; triage in TRIAGE-schema.md.)

PR #1125, opened 2026-09-19
fixed: row 4 writer_bound_count on c — RED: the C leg lands the forged count, not the writer's bound

rowan-claude:
**DRAFT, and red on purpose: this test does not pass, and what it fails on is the finding.**

**row**: `writer_bound_count` — §5.8 row 4 of `docs/FIXED-FORM-VERSIONING-TESTS.md`, owed on every leg since 2026-09-11 (`docs/FIXED-FORM-OWED-ROWS.md` on `origin/reconcile-970`). It landed on `go` in #1122.
**leg**: `c` — `internal/codegen/ctable/fixedversioning_test.go`, one file, +57 lines.
**card**: `c-writer-bound-count`, card_sha256 `cd26bbc8c5a04ddfb49826051323ff25f50f3ee540fa418cd6c5b78677948662`, run on `vision` under `nova-sandbox` (landlock).

The probe forges `old_array_bounded_grow.bin`'s count word to `7` — between the writer's `4` and the reader's `8` — in the probe's own memory through `cRunVersionProbe`'s `pre` hook, locating the word by the needle `AA AA AA AA 04 00 00 00` and asserting it matches exactly once, so no byte offset is hard-coded. The NEW build reads it through the lineage, so the plan is a COMPILED one and the oracle says `vals_count == 4`, the WRITER's bound carried by the plan, with `clamped == 1` exactly.

**what the C leg actually does**:

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -run TestFixedVersioningWriterBoundCount -count=1
--- FAIL: TestFixedVersioningWriterBoundCount (0.39s)
    fixedversioning_test.go:427: writer_bound_count: exit status 1
        writer_bound_count: vals_count == 7, not the writer's 4
FAIL	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	0.399s
```

and with the counters printed beside it (a one-line edit to the same assertion, reverted afterwards):

```
writer_bound_count: vals_count == 7, not the writer 4 (clamped=0 unknown=0 widened=0 refused=0)
```

So the C reader takes the forged count whole — it neither clamps to the writer's bound nor counts the clamp — where the Go reader lands `4` with `Clamped == 1` (#1122, green). `7 <= 8` is the reader's own bound, which is what a leg lands when the bounds pass uses the READER's capacity instead of the plan's. Nothing in `internal/codegen/ctable/` probes a forged count today (`grep -rn "forged\|clamped == 1" internal/codegen/ctable/*_test.go` is empty), which is why this has been invisible.

**the baseline, so the red is not read as this PR breaking the leg**:

```
$ git checkout fixed-table-form && SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -run TestFixedVersioning -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	2.054s
```

The leg is green without this row and red with it, on the same tree, same corpus, same machine.

**the whole leg with this test in**:

```
--- FAIL: TestFixedVersioningWriterBoundCount (0.40s)
        writer_bound_count: vals_count == 7, not the writer's 4
FAIL	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	2.586s
```

**swarm receipt**: model `deepseek-v4-pro` (DeepSeek direct, Pro), 134.48 s wall, usd 0.0352, tokens 43,163 in / 4,533 out / 1,016,960 cache read / 10,121 reasoning.

gate run by the coordinator's child outside the wall (nova-tools#1465)

It stays a draft until somebody rules on which side moves: the C runtime's bounds pass, or the row's expectation for the C leg. If the runtime is the side that moves, this test is the probe that says so, unchanged.


rowan-claude:
Superseded by #1126, which carries this test plus the fix in the C bounds pass (the count entry was pushed with the reader's size; now the writer's bound, clamp counted). Closed with the pointer; the finding was this draft's, and it was right.

PR #1126, opened 2026-09-19
fixed: row 4 writer_bound_count on c — the compiled plan's count op takes the WRITER's bound (supersedes #1125)

rowan-claude:
ROUTE unit=c-writer-bound-count-fix rung=opus (nova-decide route, kind fix-with-red-test)

**Supersedes #1125** (the swarm card's RED-only draft). #1125 is left open for Rowan to close with a pointer here; its test is cherry-picked onto this branch unchanged (`670e0668`).

## the defect

`table_fixed_compile_entry`'s kind-14 arm in `internal/codegen/ctable/fixedruntime.go` pushed the `kTableFixedCount` entry with `e.size = my_n` — **THIS reader's** capacity. A forged count of `7` from a writer bounded at `4`, read by a reader grown to `8`, passes `7 <= 8`: it lands whole and moves no counter.

The Go leg has always pushed `n = min( their_n, my_n )` (`internal/codegen/gotable/fixedruntime.go`, kind 14), which is why #1122 lands `4` with `Clamped == 1`. The C leg computed the identical `n` **one line too late** — after the push — and used `my_n` instead. The `kTableFixedCount` op body itself is already byte-for-byte the Go one, including `(*clamped)++`, so the fix is the entry's `size`, not the loop.

The rule is the spec: `docs/FIXED-FORM-VERSIONING-TESTS.md` §5.8 row 4 `writer_bound_count`; §5.2 EMIT kind 14 and §5.3 step 11 (BOUNDS runs against the **PLAN's** bounds); Glenn's ruling of 2026-09-09/10 that the fixed table reads backward and the bounds pass clamps to the writer's bounds. **The ruling here: the C runtime moves; the row's expectation stands.**

The plan already carried the writer's bound for C — `their_n` is computed two lines above — so **no contract decision is owed**.

## the fix

`internal/codegen/ctable/fixedruntime.go`, in the C emitter's own idiom: `n` is hoisted above the `d->counted` push and the entry takes it, with the reasoning in a block comment beside it. The stale enumerator comment (`a count: clamp it to the reader's own bound`) is corrected to name the entry's bound. Read checks always run; nothing was added to a write path (Glenn's rule: write checks debug only, read checks always). No other leg is touched.

```
 internal/codegen/ctable/fixedruntime.go         | 13 ++++--
 internal/codegen/ctable/fixedversioning_test.go | 57 +++++++++++++++++++++++++
 2 files changed, 67 insertions(+), 3 deletions(-)
```

## the corpus vector

[...]

PR #1127, opened 2026-09-19
fixed: row 12 forged_ordinal_both_plans on c — an ordinal past the WRITER's set counts Clamped

rowan-claude:
ROUTE unit=c-ordinal-past-writer-set rung=opus (kind fix-with-red-test)

**STACKS ON #1126 AND LANDS SECOND.** The base stays `fixed-table-form`, but this branch is now rebased onto #1126's head, so review it after that one and merge it after that one. The two runtime hunks are independent (`table_fixed_apply`'s ordinal arm here, `table_fixed_compile_entry`'s kind-14 arm there); what forces the order is that **both re-pin the same twelve C goldens**, since `ctable/fixedruntime.go` is the C emitter's runtime and rides in every emitted `Table.h`. Landed out of order, the second one's pins would be stale and the block-gate would go red.

Pairs with #1123 (the same row on `go`), which this leg's test follows as its pattern.

## the defect

`kTableFixedOrdinal` in `internal/codegen/ctable/fixedruntime.go` remapped correctly — an ordinal past `table[0]`, the WRITER's top, lands `None` — and then **moved no counter**. The Go leg's `tableFixedOrdinal` has always done `report.Clamped++` on that arm (§5.4).

The consequence is the one §5.8 row 12 exists to forbid: the IDENTITY plan's generated bounds pass already counts `1` for the same forged byte, so the **same bytes landed `clamped == 1` on one plan and `0` on the other**. A forged ordinal crossed a compiled read silently — the value was safe, the report was not, and a caller watching the counters saw a clean read.

## the fix

One arm, in the C emitter's own idiom, mirroring Go exactly:

```c
if ( raw != 0 && raw <= (uint64_t) table[0] ) { v = table[raw]; }
else if ( raw != 0 ) { (*clamped)++; }
```

with the reasoning beside it: the test is the WRITER's top ordinal `table[0]`, while the pass that runs after the loop tests THIS reader's own top value — which a remapped `0` never trips. So the event is counted once and never twice (§5.4's "counts twice"), which is what makes the row's `== 1` hold on both plans rather than `2` on the compiled one.

Read checks always run; no write path is touched (Glenn's rule). No other leg is touched.

```
 internal/codegen/ctable/fixedruntime.go         |  8 +++
 internal/codegen/ctable/fixedversioning_test.go | 71 +++++++++++++++++++++++++
 2 files changed, 79 insertions(+)
```

## the test

`TestFixedVersioningForgedOrdinalBothPlans`, two subtests because a row on this leg is two probe binaries (C has no namespace, so the two generations have no spelling inside one binary):

- **compiled** — the NEW build reads the forged OLD file through the lineage.
- **identity** — the OLD build reads the same forged bytes on its own hash.

Corpus vector: `build/fixedform-corpus/old_enum_append.bin` (`make tables-fixedform-corpus`, manifest line 53):

```
file=old_enum_append.bin row=enum_append side=old root=Lineage records=1 values=r0.tier=3,r0.seq=9
```

`Tier { Bronze, Silver, Gold }` → `+ Platinum`. The forge sets `r0.tier` to `4`: past the writer's three variants, and a name only the NEW reader has. It is located by NEEDLE — the 5-byte record body `03 09 00 00 00`, asserted to occur exactly once AND to be the file's tail — so no byte offset is hard-coded. Both columns assert `tier == None` (never `Platinum`), `seq == 9`, `clamped == 1` exactly, every other counter `0`, and no refusal.

## RED (commit `741ce2b7`, the test alone)

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -run TestFixedVersioningForgedOrdinalBothPlans -count=1
--- FAIL: TestFixedVersioningForgedOrdinalBothPlans (0.00s)
    --- FAIL: TestFixedVersioningForgedOrdinalBothPlans/compiled (0.45s)
        fixedversioning_test.go:432: forged_ordinal_both_plans, the compiled plan: exit status 1
            forged_ordinal: clamped is the bounds pass's count, once per field: 0
FAIL
FAIL	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	0.714s
FAIL
```

The **identity** column is green in the red run — which is the finding stated precisely: the divergence is the compiled plan's alone, and the row's proof is the equality the two plans did not have.

## GREEN (with `d45e00f1`)

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -run TestFixedVersioningForgedOrdinalBothPlans -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	0.805s

ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	29.642s

$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	28.627s
```

gotable, unchanged and proven so:

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/gotable/ -run TestFixedVersioning -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/gotable	23.858s
```

All runs on the Studio, `GOMAXPROCS=8`, `-count=1`, the two packages only.

Not auto-merged, not merged.

---

## the goldens, re-pinned (`62b42671`)

The block-gate was red on the first push because the twelve committed C goldens carry the emitter's runtime verbatim (landing child's comment 5737947702). Re-pinned with `make update-goldens-c` (`make/c.mk`, `GOLDENS_LEGS`), never by hand, on top of #1126's re-pin.

The regenerated diff was read before pushing. All twelve files carry a **byte-identical** change, and the union of every changed line across them, measured against #1126's head, is exactly the new counter and its comment:

```
+            /* AN ORDINAL PAST THE WRITER'S SET LANDS None AND COUNTS CLAMPED
+               (§5.4, §5.8 row 12): ... */                      (6 comment lines)
+            else if ( raw != 0 ) { (*clamped)++; }
```

Nothing else moved.

```
$ make tables-block-zero-cost
block zero-cost gate: no Table source carries one symbol of the block form
block zero-cost gate: the C# Table sources declare no accelerator build-version constant
block zero-cost gate: 123 Table sources byte-identical to their pins
```

Re-run after the rebase, on the rebased tree:

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -run 'TestFixedVersioningForgedOrdinalBothPlans|TestFixedVersioningWriterBoundCount' -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	15.184s

$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	65.579s
```

The rebase kept **both** test blocks in `fixedversioning_test.go` — #1126's `TestFixedVersioningWriterBoundCount` and this one's `TestFixedVersioningForgedOrdinalBothPlans`, in that order, above the harness banner. The conflict was the shared insertion anchor and nothing semantic.


rowan-claude:
**Not merged — red on `block-gate`, and the red is a finding, not a flake.**

`make tables-block-zero-cost` fails deterministically on this head:

```
ZERO-COST GATE FAILED: testdata/golden/tables/examples-c/GuardedTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/examples-c/KeyedTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/examples-c/NestedTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/examples-c/PackTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/examples-c/RangesTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/examples-c/TablesTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/examples-c/WideTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/block-c/PaddedTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/block-c/RenderTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/pointers-c/GraphTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/pointers-c/MarksTable.h moved
ZERO-COST GATE FAILED: testdata/golden/tables/pointers-c/PartsTable.h moved
make: *** [tables-block-zero-cost] Error 1
```

**The cause.** Both this PR and its sibling edit `internal/codegen/ctable/fixedruntime.go` — the C
emitter's runtime — so the emitted `.h` for every C table changes, and the twelve checked-in golden
headers under `testdata/golden/tables/{examples-c,block-c,pointers-c}/` were not regenerated with it.
Every other leg's row tonight (#1123, #1124, #1128, #1129, #1130, #1131) is green on `block-gate`;
the two that touch the C emitter are the two that are red, and they are red on the same twelve files.
That is deterministic and the same on both, so it is neither a flake nor a cancelled leg.

`fixed form + versioning (c)` **passes** on this head — the row's own proof stands. What is owed is the
golden regeneration in the same PR, so the gate reads the emitter's new output as the intended one.

Skipped by the lander, not closed and not force-merged; red means stop. Re-push with the goldens
regenerated and it goes back in the queue.

PR #1141, opened 2026-09-19
fixed: row 11 unknown_census on c — RED: the C leg counts the census once per element (same as dart #1138)

gafferongames:
Row 11 `unknown_census` on the **c** leg. Worker commit `b42568f1`, one new file,
`internal/codegen/ctable/fixedversioning_unknown_census_test.go`, +45.

**DRAFT, and the red IS the finding: the C leg counts the unknown census once per ELEMENT, not once per field per peer.** This is the same divergence that #1138 has just fixed on dart.

The row's oracle (§5.8 row 11) is explicit: `Item.drop` is ONE field of ONE peer however many of the four
elements carry it, so `unknown == 1` exactly, and — in the card's own words — "`4` is the divergence this
row exists to catch". The pair is `[4]Item`. The C leg lands `4`.

Every gate below was run BY ME outside the wall, on the worker's own commit. The card's own gate rows all
read `BLOCKED-TOOLCHAIN: /home/glenn/go/bin/go: Permission denied` (nova-tools#1465 again) and it said so
rather than claiming a green — the honest shape.

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -run TestFixedVersioningUnknownCensus -count=1
--- FAIL: TestFixedVersioningUnknownCensus (1.63s)
    fixedversioning_unknown_census_test.go:43: unknown_census: exit status 1
        unknown_census: unknown == 4, not exactly 1
FAIL	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	3.577s

# the whole leg, with the new test in
--- FAIL: TestFixedVersioningUnknownCensus (0.39s)
        unknown_census: unknown == 4, not exactly 1
FAIL	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	16.328s

# THE BASELINE — same tree, same corpus, same machine, WITHOUT the new test
$ git checkout fixed-table-form && SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/ctable/ -run TestFixedVersioning -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/ctable	17.736s
```

So the leg is green before this row and red only on this row: the new test isolates the divergence and
breaks nothing else.

**The corpus** was rebuilt from the current dump for this run, because `old_unknown_census.bin` only exists
after #1129: 69 manifest lines, `row=unknown_census` present. Built on space from `fixed-table-form` plus
the `serialize` sibling.

**Left as a DRAFT on purpose.** The ruling owed is which side moves — the C runtime, as dart's did in
#1138, or the row's expectation for C. The probe is correct either way, and #1138 is the precedent that
says it is the runtime. Whoever owns the C runtime decides; this PR is the probe that proves it.

ROUTE: `nova-decide-tuned route --kind row-test --files 1 --packages 1 --lanes 1` -> rung **pro**,
confidence 0.88, ask=card. Ran on vision as one of a 20-card batch.


rowan-claude:
**Coordinator gate — this PR is a no-op against the current head; recommend CLOSE, do not merge.**

Rebased onto `fixed-table-form` @ 9748cc15:

```
$ git rebase origin/fixed-table-form
Rebasing (1/1)dropping b42568f1 fixed: row 11 unknown_census on c -- patch contents already upstream
Successfully rebased and updated detached HEAD.
$ git rev-parse HEAD
9748cc15...          # the base itself; zero commits, empty diff
```

`internal/codegen/ctable/fixedversioning_unknown_census_test.go` at this head is **byte-identical** to the blob already on `fixed-table-form` — #1156 (`census-once-c-elixir`) landed the C census test. Nothing here is new.

Worth flagging for the other census PRs: `git diff --name-only fixed-table-form...pr-1141` still lists the file, because the three-dot diff runs from the older merge-base. That makes an already-landed PR look live. Comparing the blob at the PR head against the blob at the base is the check that catches it — I ran it across all thirteen open row PRs and **only this one** is a no-op.

rowan-claude:
Closed: its file is byte-identical to fixed-table-form after #1156 (c census fixed there); the three-dot diff made it look live. The finding was this draft's and it was right.

PR #1152, opened 2026-09-19
fixed: row 11 unknown_census on cpp — RED: counts the census once per element (same class as #1138, #1140)

gafferongames:
Row 11 `unknown_census` on the **cpp** leg. Worker commit `e9e3bd89`, rebased by me onto
`fixed-table-form` at `5e8a35ab` (clean), one new file under `internal/codegen/cpptable/`.

**DRAFT, and the red IS the finding: the cpp leg counts the unknown census once per ELEMENT, not once per field per peer.**

The oracle (§5.8 row 11) is explicit — `Item.drop` is ONE field of ONE peer however many of the four
elements carry it, so `unknown == 1` exactly, and in the card's own words "`4` is the divergence this
row exists to catch". The pair is `[4]Item`. This leg lands `4`.

My gate, outside the wall, on the worker's commit rebased onto the current head:

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/cpptable/ -run TestFixedVersioningUnknownCensus -count=1
--- FAIL: TestFixedVersioningUnknownCensus (0.46s)
    fixedversioning_unknown_census_test.go:43: unknown_census: exit status 1
        unknown is not exactly 1: 4
FAIL	github.com/mas-bandwidth/schema/v2/internal/codegen/cpptable	0.464s
```

**This is not one leg's bug — it is six.** Measured this shift on one corpus, one tree, one machine:

| leg | unknown | state |
|---|---|---|
| go | 1 | correct (landed #1133) |
| js | 1 | correct (landed #1135) |
| dart | was 4 | **fixed** by #1138 |
| java | was 4 | **fixed** by #1140 |
| c | **4** | red, draft #1141 |
| cpp | **4** | red, this class |
| elixir | **4** | red, this class |
| rust | **4** | red, this class |

Six of the eight legs probed for this row were wrong the same way, and the fix has landed on only two of
them. The row is doing exactly the job it was written for, and the remaining four want the same runtime
change dart and java already took (the census flag carried once per field per peer, not per element).

Left as a DRAFT on purpose: the ruling owed is which side moves, and #1138 and #1140 are the precedent
that says it is the runtime. The probe is correct either way.

ROUTE: `--kind row-test --files 1 --packages 1 --lanes 1` -> rung **pro**, ask=card.


rowan-claude:
**Coordinator gate — outside the wall**, on this PR's commit rebased onto `fixed-table-form` @ 9748cc15, corpus built from that tree (`make tables-fixedform-corpus`).

```
SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/cpptable/ -run '^(TestFixedVersioningUnknownCensus)$' -count=1
unknown is not exactly 1: 4 -- FAIL  .../cpptable  1.168s
SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/cpptable/ -run TestFixedVersioning -count=1
cppLINE
```

**RED / RED — the red in the title is confirmed on the current head.** The cpp leg counts the census once per element (4) instead of once per field per peer (1) — the same class as #1138, #1140 and #1156. #1156 fixed c and elixir; cpp still needs its own fix. Do not merge as green..

The corpus requirement is load-bearing here: with `build/fixedform-corpus` moved aside the same test fails with `SCHEMA_REQUIRE_CORPUS is set and the reference's byte oracle is not in .../build/fixedform-corpus`, so the pass is not a skip.

rowan-claude:
Gate outside the wall, shift 4, on **space** at branch head `148611bf7f21ff742fc925d63ff96663e9806c2c`. **RED, and it is a real finding — the test is right and the cpp runtime is wrong.**

```
--- ROW TestFixedVersioningUnknownCensus ---
    fixedversioning_unknown_census_test.go:43: unknown_census: exit status 1
        unknown is not exactly 1: 4
--- FAIL: TestFixedVersioningUnknownCensus (0.45s)
FAIL  github.com/mas-bandwidth/schema/v2/internal/codegen/cpptable  0.456s
```

Same class as #1138 (dart), #1140 (java), #1156 (c + elixir): the census is counted **once per element** of `items [4]Item` instead of **once per field per peer** (ALG 5.4, §5.8 row 11).

**This is the same root cause as the branch-wide red.** `tools/fixedtwin` `TestPairedHeadersAgree` has been failing on every commit since #1156, because #1156 added `int census` to `internal/codegen/ctable/fixedruntime.go` (`table_fixed_compile_entry`, `table_fixed_match_children`, `if ( !named && census && c->report != NULL )`, and `census && i == 0` / `census && k == 0` at the array and keyed-array sites) and left `internal/codegen/cpptable/fixedruntime.go` unchanged. Fixing cpp fixes both this row and the pairing gate.

Keeping this draft open as the red that proves the fix. A cpp census runtime fix card is cut.

rowan-claude:
Superseded by #1163, which landed as `f2d33e802a4933b5e8212ce87c3fa67280faf98a`.

#1163 carries the cpp census runtime fix **and** an equivalent `TestFixedVersioningUnknownCensus` at the same path this draft uses, so the two cannot both land. The red this draft found is real and it is now green:

```
red first, this draft's assertion against the base runtime:  unknown is not exactly 1: 4
after #1163:  --- PASS: TestFixedVersioningUnknownCensus (0.46s)
              ok  .../internal/codegen/cpptable  1.817s        # whole leg
              ok  .../tools/fixedtwin            0.102s        # the pairing gate, red since #1156
```

Thank you for the finding — it was the same root cause as the branch-wide `fixedtwin` red, and naming it is what made the fix a ten-minute job. Closing as superseded, the #1141/#1153 pattern.

PR #1157, opened 2026-09-19
docs: a union's unselected arms after a read — NEEDS GLENN, leaning lawful as is

rowan-claude:
ROUTE (Jev, kind `cause-to-find`): opus ask=child. Docs only — **no runtime change**, deliberately.

The question came out of #1140: with the java probe's poison made TOTAL, `union_append` and `union_arm_payload_widen` leave a union's unselected arms carrying the caller's bytes. I declined to fold that into the fix PR and said it needed its own ruling. This is the ruling.

## The answer: NEEDS GLENN, leaning LAWFUL AS IS

Not because the documents are silent everywhere — they settle it clearly for the **variable** form — but because they are silent for **this** form, and §3.4 says §3.4 is the law here.

## The evidence, quoted

`docs/SPEC.md` §4.8, which names Java in the list:

> in mutable targets whose storage lays every arm out separately (Go, C#, JS, Dart, Java) an unselected arm keeps whatever it last held — the reused-storage discipline

and §5 repeats it as a carve-out from "read success fully initializes it".

**`docs/SPEC-TABLES.md` carries no "fully initializes" sentence of its own** — the phrase does not occur in the file (`grep -c` returns 0) — so the fixed form neither inherits that promise nor overrides §4.8's carve-out. But §3.4's own prefill line is still a bare `Reset( value )`, under which Java *would* restore every arm. Glenn's **fix 15** — "prefill the bytes the plan does not write; identity's list is empty" — is what opened the gap, and its cost to unions was never priced. That is a ruling whose side effect deserves to be seen, not a rule a leg applied wrongly.

**This page never claimed it.** Every sibling row that owes a default says so in its own cell — `field_append`'s "w = default", `array_bounded_grow`'s "slots 4..7 default", `array_fixed_grow`'s "4 at the ELEMENT DEFAULT". `union_append`'s NEW-READS-OLD cell says "an `a` record lands a; the tag width equal" and stops. Nor does the C++ reference assert it: `union_append_case()` reads `pick.type`, `pick.alpha.m`, the ordinal and `seq`, and never reads `beta` or `gamma`. **No assertion about an unselected arm exists anywhere in the tree.**

## The mechanism — the scatter, not the prefill

Every union entry is GUARDED, so `holes` leaves the whole union a hole and the record **image** does take the defaults arm by arm in declared order (§5.2, §5.9 #38). The image-to-value scatter is where it stops — the generated union decode is a switch on the landed tag, and its own emitted comment says it: `the tag, then the selected arm; an unselected arm keeps what it held`. The appended arm matches no writer arm and gets **no plan entry at all**. No prefill change reaches this.

## The legs already disagree, observably

| behaviour | legs |
|---|---|
| unselected arm gets its DECLARED DEFAULT every record | go, rust ("an unselected union arm keeps the default rather than the previous record") |
| unselected arm keeps the caller's bytes | java, dart, js, c# |
| indeterminate by construction (a real C union) | c, c++ |

Nine ports written from prose have come out three ways, which is the exact condition §3.4 exists to end.

## One caution about the evidence, which I want on the record

`pick.beta.n` is the **weaker** half of the red. The java probe poisons only the `reads` column — `probeSource(class, pkg, root.Name, side.key != "refuses")` — and `beta` is a SHARED field compared old-value against new-value, so the unpoisoned older build's constructor zero against the poisoned newer build's `0x5A` reds it *whatever the codec does*. **`pick.gamma.p` is the honest half**: a NEW-only field compared against the probe's own `FRESH` dump, red because the read genuinely never writes it. Anyone re-running this should poison both columns or compare only `gamma`.

## The proposed row line

Written out in the doc for the lawful answer:

| row | the rule | the oracle | which legs |
|---|---|---|---|
| `union_unselected_arm` | after a read that RETURNS, an arm the landed tag did not select is **UNSPECIFIED** on separate-storage targets and INDETERMINATE on overlaid ones; the selected arm and the tag are the whole of what a union read promises | `VOLD_/VNEW_union_append` as they stand, destination POISONED on BOTH columns: `pick.type` and `pick.alpha` land the writer's, `seq` stands, every counter `0` — and the probe **does not compare** `pick.beta` or `pick.gamma`, which is the row's whole content | every leg; no leg is red today, and the row exists to stop one being written |

**If Glenn rules it a defect instead**, the expectation is stated too: every arm holds either the writer's value (selected) or its own declared defaults (unselected), delivered in the scatter because the flat image provably cannot carry two arms' defaults at one byte (§5.9 #38). That is a scatter shape change in four legs, a per-record per-arm reset on the hot path, and unstatable for c and c++ — so it would have to be worded for separate-storage targets only, which is what §4.8 already does in the other direction.

Filed the way row 9's multi-record question is: **named here rather than answered**.


rowan-claude:
Decided 2026-09-19 by Rowan, under Glenn's live word tonight ("Do what you think is best. you have my permission. go"): **LAWFUL AS IS.** Glenn can reverse this in one line and the row comes out.

Why: §4.8 already says it for the variable form and names the same targets, so ruling the other way would make the fixed form the stricter of the two on exactly the path that exists to be the fast one. The defect reading costs a per-record, per-arm reset on the hot path in four legs and cannot be stated at all for c and c++, where the arms overlay. And a caller who reads an arm the tag did not select has a bug whatever those bytes hold; the promise worth making is the one every leg can keep: the tag and the selected arm.

What lands: the `union_unselected_arm` row exactly as proposed above (UNSPECIFIED on separate-storage targets, INDETERMINATE on overlaid ones; the probe poisons BOTH columns and does not compare `pick.beta` or `pick.gamma`). Go and rust keeping the declared default is allowed by "unspecified" and needs no change. Docs and probe only; no runtime change.

rowan-claude:
**The one red on this PR is the branch's, not this row's — named cause.**

`go-test-touched` fails `--- FAIL: TestPairedHeadersAgree` in `tools/fixedtwin`. That gate holds the C and C++ fixed-table runtimes line-paired. #1156 added the `census` parameter to `internal/codegen/ctable/fixedruntime.go` (`TableFixedCompileEntry(..., int census)`, `TableFixedMatchChildren(..., int census)`, `if ( !named && census && c.report != NULL )`) and did **not** make the matching change in `internal/codegen/cpptable/fixedruntime.go`. The pairing gate has been red on every commit since.

Receipt that it is pre-existing and not this PR's: the same `--- FAIL: TestPairedHeadersAgree  tools/fixedtwin` is red on the branch head itself (`da6cdc101ad6e390c578ff1621f3b935a4be78a9`, full lane), in `go-test (unit-rest)` and in `pins` (`make update-goldens` runs the same test). This PR touches no Go source that the pairing gate reads.

Fix owed and being cut this shift: the cpp census runtime fix (same class as #1156 for c/elixir, and the same root cause as the red draft #1152).

rowan-claude:
**The three-way table is a four-way table, and cs is not where this issue put it.** Measured tonight (2026-09-19) on `space` at `fixed-table-form` head `30c0c0d3`, poisoning **every** arm of the destination — the tag, `pick.alpha.m`, `pick.beta.n`, `pick.gamma.p` and `seq` — to `0x5A5A5A5A` before the load, on **both** columns, then reading what came back. The `union_unselected_arm` row landed on all nine legs tonight (#1218, #1223–#1230).

| leg | the APPENDED arm (`pick.gamma.p`) | the SHARED arm (`pick.beta.n`) |
|---|---|---|
| go | the declared default `0` | the declared default `0` |
| rust | **not addressable** — `pick.gamma()` is `None` for an alpha record, and the crate-internal overlay has no such field (`error[E0609]: no field 'gamma' on type 'PickRowArms'`) | same |
| elixir | **no caller destination exists at all** — `<row>_fixed_load/2` takes no image and the runtime assembles the record from the prefill | the declared default `0` |
| **cs** | **the declared default `0`** | **the caller's poison `0x5A5A5A5A`** |
| java | the caller's poison `1515870810` | the caller's poison `1515870810` |
| dart | the caller's poison `1515870810` | — |
| js | the caller's poison `1515870810` | — |
| c | the SELECTED arm's storage — reads `7`, which is `pick.alpha.m` | overlaid |
| cpp | the SELECTED arm's storage — reads `7` | overlaid |

## cs is a mixed case, and the two readings are each other's control

This issue's table says *"unselected arm keeps the caller's bytes: java, dart, js, c#"*. On the FIXED form that is half true of cs:

```
# the appended arm, asserted to be its declared default
ok  github.com/mas-bandwidth/schema/v2/internal/codegen/cstable  2.019s

# the complement — demand it still be the poison
ROW union_unselected_arm/new_reads_old: FAILED pick.gamma.p is NOT the poison any more, it is 0

# the same complement on the SHARED arm
ok  github.com/mas-bandwidth/schema/v2/internal/codegen/cstable  2.092s     (beta.n IS still 0x5A5A5A5A)
```

`beta.n` coming back `0x5A5A5A5A` is the proof that the poison reached the destination, so `gamma.p` coming back `0` is **the read resetting it**, not a poison that never landed. The arm the writer's generation also has keeps the caller's bytes; the arm only the reader's generation has is reset to its declared default.

**Nothing here is a defect and nothing was changed.** All four behaviours are inside the ruling — that is what "UNSPECIFIED on separate-storage targets and INDETERMINATE on overlaid ones" buys. But the ruling's evidence table is a statement of fact about the legs, it is quoted in `docs/FIXED-FORM-VERSIONING-TESTS.md`, and it was used to write nine swarm cards tonight: my own cs card told its worker to expect a RED and to treat a GREEN as proof the poison had failed. It would have been wrong, and the worker was blocked before it could be misled.

## Two other corrections for the record

- **elixir was not in the table at all.** It cannot be: there is no caller destination on the BEAM. The probe for that leg deliberately carries no poison, because a poison computed and discarded is exactly the vacuity removed from that leg tonight in #1212.
- **The poison receipt is the UNSELECTED arm, never the selected one.** #1217 landed the row saying the receipt was to demand `pick.alpha` be the poison; that control goes red whether the destination was poisoned or not, because the read overwrites the selected arm either way. #1222 corrected it. The real receipt is the temporary `pick.gamma.p == 0` assertion, and it is only available on a leg that keeps the caller's bytes — which, per the table above, now means java, dart and js, and **not** cs.

rowan-claude:
**Ruled, 2026-09-19, by Glenn, in his own word:**

> as designed it is "undefined"

So the ruling stands as LAWFUL AS IS, and the *word* is now one word for every target. The text this PR landed said an unselected arm is **UNSPECIFIED** on separate-storage targets and **INDETERMINATE** on overlaid ones — two words, split by target — and that split reads as a promise about which legs do what. It is not one.

**A union read DEFINES the tag and the SELECTED arm and nothing else. An unselected arm is UNDEFINED.** One word, all nine legs.

The per-leg behaviour stays written down, and is now explicitly marked as what it is: go and rust *happen to* land the declared default, java, dart, js and cs *happen to* leave the caller's bytes, c and cpp overlay so the question cannot be put to them — each with **"an observation and not a guarantee, and nothing may be relied on it"** beside it.

Landing in #1266: `docs/FIXED-FORM-VERSIONING-TESTS.md` (the ruling paragraph, the rows-table entry, the evidence paragraph and the poison receipt's "unspecified") and the nine per-leg `fixedversioning_union_unselected_arm_row_test.go` headers. Comments and text only — no probe, no assertion and no emitter changed, and the row still refuses to compare `pick.beta` or `pick.gamma`, which is its whole content. The reversal path in the doc is left exactly as written, so a reversal stays one edit.

— Rowan

PR #1159, opened 2026-09-19
roadmap data: reconcile the fixed-table rows that landed 2026-09-19 (evidence only, no cell turns green)

rowan-claude:
## What this is

The first reconciliation of `docs/roadmap.sexp` against the swarm rows that landed
on `fixed-table-form` overnight. **Evidence only — no cell turns green.**
`ROADMAP.md` is byte-for-byte unchanged, and that is the correct outcome: a cell
ticks only when *every* required leaf is `:done`, and none of the cells touched
here is complete. This moves the data layer, not the picture.

Written by hand from the wire, not by a card.

**This body describes the head after two rounds of repair.** Two independent cold
reads HELD it at `5af1bfb4` and both were right about something; the history is
kept rather than squashed so the corrections are auditable.

## `:source-revision`

`8ea5ed8e4656875088250f564e88a965a7135e7c` → `f2d33e802a4933b5e8212ce87c3fa67280faf98a`, which is #1163's own merge commit and carries #1162 as an ancestor, read off the wire at **2026-09-19T02:52:06Z**. It is deliberately NOT the current head: #1165 (row 4 with two lineage peers, on go) and #1169 (row 9 on elixir with a destination claim that can fail) landed after it and speak directly to notes in this PR, so pinning forward would make them stale on arrival. Every note here is true as of this pin; those two rows are the follow-up.

The schema manager is landing all night, so **expect the head to have moved again
by the time you read this.** Every cited merge sha is verified an ancestor of the
pin. The other occurrences of the old sha are evidence URLs, `:tested-revision`
values and one line of prose about PR836, all deliberately untouched — 19 full
40-char occurrences before, 18 after (and 20 → 19 if you count the short `8ea5ed8e`
prefix, which also appears in the PR836 line; both counts are right, they count
different things).

## Nodes set `:done` (13)

`<leg>/R11` — *"the plan carries the WRITER's bounds, per plan; the hostile pass
runs against the plan's bounds"* — from **§5.8 row 4 `writer_bound_count`**, which
forges the count 4 → 7 and asserts the writer's `4` exactly (never the reader's
`8`, never the forged `7`) with `clamped == 1`, never `>= 1`.

`<leg>/R12` — *"the per-record hash check is before the prefill: no_layout writes
nothing"* — from **§5.8 row 9 `refuse_writes_nothing`**: `no_layout`, `malformed`
FALSE, every counter `0`, and the caller's storage poisoned `0x5A` still `0x5A` in
every byte.

[...]

PR #1155, opened 2026-09-19
fixed: row 9 refuse_writes_nothing on elixir

gafferongames:
Row 9 `refuse_writes_nothing` on the **elixir** leg. Worker commit `086adca9`, rebased by me onto
`fixed-table-form` at `5e8a35ab` (clean), one new file under `internal/codegen/elixirtable/`.

Gate run BY ME outside the wall on the worker's own commit rebased onto the current head:

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/elixirtable/ -run TestFixedVersioningRefuseWritesNothing -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/elixirtable	0.927s
```

This card carried the poison warning from schema#1140 — the java false red where the shared probe's
`poison()` skipped final fields and a nested value read back as a prefill. The card was told to read this
leg's probe helper and confirm it poisons NESTED values, and to refuse to report a red it could not
attribute. It came back green, so the warning cost nothing here; it is the reason I trust the green.

Corpus rebuilt from the current dump for this run. ROUTE: `--kind row-test --files 1 --packages 1
--lanes 1` -> rung **pro**, ask=card.

PR #970, opened 2026-09-11
every row, every leg: the gate that keeps the edge cases held

gafferongames:
## The silence

The rows of `docs/FIXED-FORM-VERSIONING-TESTS.md` are the fixed form's versioning edge cases — one definition change per row, each read two ways — and each leg's harness probes them **by row name**: `{row: "field_append"}` in the Go-driven harnesses, `field_append_case()` in the C++ reference. CI's full lane runs all of them.

Nothing said a row is probed on **every** leg. A row could be written into the page, probed on four legs, and nobody would see the other five. Glenn, today:

> "The edge cases must be probed, tested and known to hold, and stay held."
> "Everything you decide, every edge case you fix, this creates new tests that enforce that decision moving forward across all languages."

## The gate

`compiler/everyrow_test.go`, two tests and a ledger.

**`TestEveryRowEveryLeg`** parses the page's row tables — a table whose first header cell is `row`, which is the main table of definition changes and the divergence table, and not the tables *about* the rows — greps each of the nine legs' harnesses for each row name, and fails naming every `(row, leg)` pair nobody probes. The match is anchored on both sides and allows only the reference's `_case` / `_hostile_case` suffixes, so `string_grow` takes no credit from `wstring_grow` and `fixed_I_grow` none from `fixed_I_grow_element`. A leg whose harness file does not exist yet probes nothing, which is the true state of the Java leg.

A row whose **NEW-READS-OLD column is `—`** is LOCK-only: no file exists for a leg to read, its proof is the lock's refusal in `internal/lockfile`, and the nine read harnesses do not owe it. That is read off the page rather than listed in Go, so a row that **gains** a read column starts being owed on nine legs without anyone editing this file.

**`docs/FIXED-FORM-OWED-ROWS.md`** is the exception list, and it is a ledger rather than an escape — row | leg | owed by | since. The gate also fails when a line is for a pair that **is** probed, for a row the page does not have, for a leg outside the nine, or for a row the page marks LOCK-only. **The list can only shrink.** A row added to the page is owed on nine legs the minute it is written; a row that lands on a leg is one line deleted.

**`TestEveryRowHasCorpusBytes`** is the second gate: a row with a read column needs bytes, and the dump writes one `row=` per row into `build/fixedform-corpus/manifest.txt`. The corpus is a build product, so the test skips when it is absent and fails under `SCHEMA_REQUIRE_CORPUS`. Its exceptions are the same ledger, under the target `corpus`.

## The counts, seeded with today's true state

| | |
[...]

PR #1176, opened 2026-09-19
roadmap data: pin forward to 801ab405, elixir/R12 earned on #1169, and R16's reason corrected

rowan-claude:
## What this is

The follow-up #1159 promised, in its own words: *"#1165 and #1169 are the first
content of the follow-up PR."* Same method, same discipline — every citation
re-read off the wire immediately before the push, and the pin set at a revision
**every note is true as of**.

**Still evidence only.** One node changes state; `ROADMAP.md` is byte-for-byte
unchanged, because no cell's leaves are all `:done`.

## 1. The pin moves forward, and two residuals close

`:source-revision` `f2d33e80` → **`801ab4055b429775cb9d36910777f63e71f61856`**,
`fixed-table-form`'s head read off the wire at **2026-09-19T03:18:35Z**.

### `go/R11`'s residual closes — #1165

#1165 (`4d4102e08695146ae09de507f5fae477137a41ee`) hands the reader **two** lineage
peers with different bounds and makes the distractor's bound the discriminator:

```go
out, err := runVersionProbe(t, newer, []string{writer, distractor}, src)
```

```go
if back[0].ValsCount != 4 {
    t.Fatalf("writer_bound_count: the count landed %d, not the WRITER's bound 4 " +
             "(never the distractor's 6, never the reader's 8, never the forged 7)", ...)
}
```

Its own header says why that is the test worth having: *a reader that clamps to its
own bound lands 8, a reader that clamps to whichever peer it saw last lands 6, and
only a plan that carries the WRITING peer's bound lands 4.* That is `per plan`,
distinguished — a runtime holding one writer bound globally now fails.

**#1165 is go-only.** So `go/R11`'s note drops the residual, and the other six R11
notes keep it, now naming where the closing row landed and that it is owed on their
leg. cs and rust still have no row-4 test at all.

### `elixir/R12` becomes `:done` — #1169

#1169 (`cce7d129535ebb3bb0ac3a4a780538fd3f2ef925`) **replaced** the tautology #1155
shipped — the `fresh == fresh_value()` comparison of two freshly built struct
literals that #1159 was held for. The new test reasons about the leg instead of
imitating the other six:

```elixir
# THE DESTINATION, WHICH IS THE ROW (§5.8 row 9). The BEAM has no caller-owned
# image to poison: a read's destination is the value it RETURNS ...
check(is_atom(why), "REFUSE wrote a destination value: #{inspect(why)}")
{ok_tag, ok_values, ok_report} = load(File.read!(file()))
check(ok_tag == :ok, "the unforged control read refuses: ...")
check(length(ok_values) == 1, "the unforged file builds one value, not #{length(ok_values)}")
```

It **can** fail: if the reader returned a built value in the refusal's value slot,
`is_atom(why)` is false. And the control read of the **same record, unforged,
through the same reader** returns one value — so the refusal is shown to withhold a
value this reader demonstrably can and does build. That is what makes the negative
mean something.

This is **different in kind** from the six legs that poison `0x5A` and sweep bytes,
and `elixir/R12`'s note says so rather than flattening it into them.

## 2. R16's reason was false — the cold reader's carried finding, confirmed

The note I wrote in #1159 said *row 11 is a runtime load, and R16's first clause is
about COMPILE time*. **That is wrong, and I verified it is wrong before rewriting:
the compile happens inside the load.**

- The census `Unknown++` lives in the plan **compiler**:
  `if !named && census && c.report != nil { c.report.Unknown++ }` in
  `tableFixedMatchChildren`, `internal/codegen/gotable/fixedruntime.go`, with the
  same line in the cpp twin's `TableFixedCompileEntry`.
- It is reached from the load: `tableFixedLineagePlans` calls `tableFixedCompile`,
  which walks the peers and censuses them.
- So **row 11 moves exactly the compile-time counter** the clause is about.

**The real ground for `:unknown` is the clause's second half — "and never per
record".** `old_unknown_census.bin` carries **one** record, so *once per peer at
compile* and *once per record* both predict `1`; the corpus cannot tell them apart.
And the runtime does have a genuinely per-record `Unknown++` of its own — the union
tag past the arm count, in the plan-execution loop — that row 11 never touches.

All eight notes now say that, and name what would close it: **a multi-record
`unknown_census` corpus.** `widened` (row 11 asserts `widened == 0`) and the text
lane of `clamped` remain unproved, as before.

## 3. Evidence strings made consistent

The first cold read's carried nit: some entries cited a file without its test
function. Every evidence list is now the same three-part shape —

1. the PR url and what it proves,
2. `internal/codegen/<leg>table/<file>: <TestFunc>`,
3. `merged into fixed-table-form at <sha>`

— and for R16, that triple twice, once per row. No entry names a file without its
test function any more.

## State change

| node | before | after | why |
|---|---|---|---|
| `elixir/R12` | `:unknown` | `:done` | #1169, a destination claim that can fail, with a control |

That is the only one. `:done` 195 → **196**, `:unknown` 826 → **825**, `:todo` 23,
total 1044 unchanged. `go/R11` was already `:done`; what changed there is the note
and a second evidence PR.

## Verification

- `render.lisp --test` → **21 passed, 0 failed**
- `go test ./tools/roadmap/` → **ok**
- `ROADMAP.md` diff **empty**
- Every cited sha an ancestor of the pin
- **Every PR cited anywhere in the file re-read with `gh pr view` immediately
  before the push.** All MERGED into `fixed-table-form` at the shas cited, with
  three exceptions that are **pre-existing prose in nodes this change does not
  touch**: #836 (OPEN, the integration PR), #970 (OPEN, an acceptance gate) and
  #1000 (merged to `emma/fixed-form-layout-validation-992`). Flagging them rather
  than quietly leaving them: they are worth a look by someone who owns that prose.

## What I want

**A read from the cold reader and from a friend.** The two judgement calls are
whether #1165 really closes `per plan` on go, and whether #1169's return-shape
claim plus its control is enough to earn `elixir/R12` on a leg with no mutable
destination. **Do not merge on my say-so.**

Still owed, and not in this PR: the `two_peers` row on the other six legs; a
multi-record `unknown_census` corpus to close R16's first clause; and the
`cfloat_res_refine` rows (#1166–#1175) which landed in the pin's range and are a
different row family, unreconciled here.


rowan-claude:
Cold read at 3d4a349: APPROVE

Same method as the three reads on #1159, at `3d4a34927f2055942dfc8b21b22ffca6fb1647fe`, fresh fetch, 2026-09-19T03:27:38Z. The carried finding is adopted with a correction I did not ask for and would not have thought to demand. Three non-blocking findings, all of them shavings.

## (2) `go/R11` — #1165 really does distinguish "per plan"

I went looking for the hole and the runtime closed it. `internal/codegen/gotable/fixedform.go:689-691`:

```
// One plan per lineage entry, laid down from THE LOCK's bytes: nothing
var <T>FixedLineagePlans = tableFixedLineagePlans(<T>FixedKnown, <T>FixedLayout, <T>FixedDst, <T>FixedHash)
```

and `fixedruntime.go:1151-1152`, `out := make([]tableFixedLineagePlan, len(known))`, compiling each peer's plan and storing its own counters on the lane. So two peers means **two plans compiled and held at once**, each carrying its own writer's bound, and the load picks between them by hash (`fixedform.go:769`, `lane := &<T>FixedLineagePlans[pick]`). #1165 hands the reader exactly that — `runVersionProbe(t, newer, []string{writer, distractor}, src)` with VOLD at `[..4]` (which wrote the file) and VMID at `[..6]` (which wrote nothing) — forges the count to 7 and asserts:

```
	if back[0].ValsCount != 4 {
		t.Fatalf("writer_bound_count: the count landed %d, not the WRITER's bound 4 (never the distractor's 6, never the reader's 8, never the forged 7): %+v", back[0].ValsCount, back[0])
	}
```

That is the exact hole I named at `5af1bfb`: *"a runtime that stashed one writer bound globally would pass."* It no longer would — a global stash takes whichever peer compiled last and lands 6. The new `VMID_array_bounded_grow.schema` is disciplined about it too: it differs from VOLD in the bound and nothing else, *"because a distractor that differs in any other way is not a distractor, it is a different row."* `docs/roadmap.sexp:2752` says *"No residual on go"* and I agree with that. The other six R11 notes still carry the residual verbatim — `:2703 :2719 :2782 :2798 :2814 :2830`, six of six, go the only one dropped.

**Finding 1 (cosmetic).** `:2752` and #1165's own comment both say *"A reader clamping to its own bound lands 8."* It lands **7**: the forge writes 7, the reader's bound is 8, so no clamp fires and the forged value survives unclamped. One of the three failure modes is misdescribed. Nothing is missed by it — the assertion is `ValsCount != 4`, which catches 6, 7 and 8 alike, and `Clamped != 1` catches the no-clamp case a second time — but a note that enumerates what each wrong answer would look like should get all three right.

What #1165 does not do, and nothing claims it does: VMID writes no corpus file, so no load ever goes through the VMID plan to show that *it* lands 6. The strongest form of "per plan" — both plans exercised, each landing its own bound — is still one corpus file away. I raise it for the record, not as an objection; the discriminating power that matters is there. And I withdraw the identity-plan half of my original residual: on the identity plan the writer is the reader, so there is no bound to get wrong, and row 12 exercises that plan anyway.

## (3) `elixir/R12` — it can fail, and "different in kind" is said truthfully

#1169 deletes the tautology and puts a control in its place:

```
-  check(fresh == fresh_value(), "REFUSE wrote destination values")
+  check(is_atom(why), "REFUSE wrote a destination value: #{inspect(why)}")
+  {ok_tag, ok_values, ok_report} = load(File.read!(file()))
+  check(ok_tag == :ok, "the unforged control read refuses: #{inspect(ok_tag)} #{why(ok_report)}")
+  check(length(ok_values) == 1, "the unforged file builds one value, not #{length(ok_values)}")
```

**Can it fail: yes**, and the part that can is the control. A reader that refused the unforged file, or built no value from it, goes red on `ok_tag == :ok` or `length(ok_values) == 1`, and neither is implied by anything above it. That is a real assertion about a real capability, and it is what makes the refusal mean something: the reader is shown to withhold a value it demonstrably can build.

**Finding 2 (non-blocking).** The evidence string at `:1135` points at #1169 as *"a destination claim that CAN fail"*, and the note at `:1141` leads with *"#1169 asserts the refusal tuple second element is the reason ATOM and never a built value"*. Taken on its own that check is dominated: three lines above it sits `check(why == :no_layout, ...)`, and every state in which `is_atom(why)` is false is a state in which that one has already gone red. `check/2` does not halt, so it is not dead code, but it cannot be the sole cause of a failure. The row rests on the control, not on the atom check, and the note leads with the weaker of the two.

The second thing, which the note is honest about in kind but does not record as a residual: the node's first half is *"the per-record hash check is before the prefill"*, and on the BEAM that **ordering is unobservable in principle**. The six poison legs genuinely see it — `Vec.w = 88` is a nonzero prefill, so a prefill that ran before the check shows up in the sweep. Elixir can only observe that no value comes back, which is equally consistent with a prefill that ran and was discarded. So this `:done` rests on a strictly weaker proof than its six siblings. `:1141` says so in kind — *"Different IN KIND from the six legs that poison 0x5A and sweep bytes, and recorded as such rather than flattened into them"* — which is truthful and is the right instinct. What it does not do is carry a `RESIDUAL:` clause the way the six R11 notes do, and this is the same shape of gap those clauses exist for. Given the state is earned in the only way that runtime permits, and given the note refuses to pretend otherwise, I am not holding on it.

## (4) The R16 rewrite — checked against the code, and it is right

`docs/roadmap.sexp:2582` and its seven twins now say:

> "the counter it moves IS the compile-time one, the census Unknown++ inside the plan compiler (tableFixedMatchChildren in internal/codegen/gotable/fixedruntime.go, TableFixedCompileEntry in the cpp twin), which the load reaches through tableFixedLineagePlans. CORRECTION to an earlier note on this task: the claim that row 11 is a runtime load and therefore cannot touch a COMPILE-time clause was FALSE — the compile happens inside the load."

Every mechanism named is real at the pin. `gotable/fixedruntime.go:810-811`:

```
		if !named && census && c.report != nil {
			c.report.Unknown++
```

inside `tableFixedMatchChildren`; `tableFixedLineagePlans` (`:1151`) calls `tableFixedCompile(parsed, myLayout, dst, plan, &census)` and stores `Unknown: census.Unknown` on the lane; the generated load copies it — `censusUnknown, censusKind = lane.Unknown, lane.KindMismatch`, under the emitted comment *"THE CENSUS IS ONCE PER PEER and never per record (§5.4), and it lands only on a read that returns: REFUSE moves no counter."* I had doubted `tableFixedLineagePlans` as the route and was wrong: on go the lineage plans are laid down once and the load reads a precomputed per-peer number, which is a better answer than the one I carried.

The rest holds too. *"old_unknown_census.bin carries ONE record, so once per peer at compile and once per record predict the same 1"* — the corpus is one record. *"the runtime does have a per-record Unknown++ of its own (the union tag past the arm count, in the plan-execution loop)"* — `fixedruntime.go:272-277`, `if tag > uint64(p.Aux) { tag = 0; report.Unknown++ }`. *"row 11 asserts widened == 0"* — it does. And the node stays `:unknown` with the closing condition named: *"a multi-record unknown_census corpus would close the first clause."* Correcting a note by writing the correction into the note, rather than quietly editing it, is the behaviour I would want from this file.

## (1), (5), (6), (7) — mechanical, all clean

**Pin.** `docs/roadmap.sexp:12` is `801ab4055b429775cb9d36910777f63e71f61856`, an ancestor of `origin/fixed-table-form` (now `88ca0cfd`). **All 32 `(node, PR, sha)` citations** across 29 distinct PRs read back `MERGED`, `baseRefName=fixed-table-form`, sha exact — zero mismatches — and **all 29 merge shas are ancestors of the pin**, including the two new ones, #1165 at `4d4102e0` and #1169 at `cce7d129`, which were the post-pin exclusions I checked at `6c006c2`. Every note is true as of the pin; I found no claim that needs a later commit to be true.

**Evidence strings.** Eight sampled, each resolved at the pin by `grep -c "func <fn>("`: the two-peers go test, elixir refuse, cpp census, rust forged-ordinal, js census (`TestJSFixedVersioningUnknownCensus`), java writer-bound, dart refuse, js writer-bound. Eight of eight present. The `file: TestName` shape is now uniform, which is what made this sample cheap — at `5af1bfb` three of them named a file and no function.

**One flip.** A per-node state diff against `main` returns exactly one line, `elixir/R12 unknown → done`. `:done` 195 → 196, `:unknown` 826 → 825, `:todo` 23 → 23, total 1044. Nothing else moved.

**What main runs.**
```
rc=0
$ go test ./tools/roadmap/
ok  	github.com/mas-bandwidth/schema/v2/tools/roadmap	0.304s
```
Regeneration leaves `ROADMAP.md` byte-identical; `git diff origin/main...HEAD -- ROADMAP.md` is empty. No invented key — the keyword set is identical to `main`'s but for `:ok`, which occurs once, inside the `elixir/R12` note string at `:1141`.

**CI**, completed checks only: **68 COMPLETED with conclusion SUCCESS, nothing failed.** `go-test`, `pins` and `big-endian` are IN_PROGRESS and `mergeStateStatus` is `UNSTABLE` for that alone. I am not counting those three.

## (8) The three pre-existing citations — agreed, not this PR's to fix

All three are untouched: `git diff origin/main...HEAD` changes **zero** lines mentioning them, and their occurrence counts are identical in `main` and here.

- **#836** — `OPEN`, draft, base `main`. Cited on `integration` (`:4159`), which is `:state :todo`. `render.lisp` requires `:evidence` only on `:done`, and the node's own note already reads *"PR836 remains draft/open"*. Nothing green rests on it.
- **#970** — `OPEN`, draft, base `fixed-table-form`. Cited under `:acceptance-source`, not `:evidence`, on `gates/row-coverage/pairs` and `.../corpus` (`:4075`, `:4079`), both `:unknown` with `:evidence nil`, and both notes already say it is draft/open. Nothing green rests on it.
- **#1000** — the only one worth a second look, because it is cited in the `:evidence` of seven `:done` nodes, `shared/LOCK-L1` through `L7` (`:4042`–`:4066`). It is `MERGED`, but into `emma/fixed-form-layout-validation-992`, a stacked branch. **The work landed anyway**: its merge commit `f2501193` is an ancestor of the pin, carried in by #999, which is cited beside it on every one of those seven nodes and merged into `fixed-table-form` at `ae4e935e`. So this is a citation whose *form* is unusual, not a green resting on unlanded code.

**So: no, this should not hold this PR**, and I agree with the lean. It is a pre-existing convention question — whether `:evidence` should name the stacked PR or the one that actually reached the branch — across seven nodes none of which this PR touches. A separate issue, and I would rather it were filed against the seven nodes with `f2501193`'s ancestry written down than folded into a reconciliation PR that has nothing to do with locks.

## Verdict

**APPROVE at `3d4a349`.** One state moves and it is earned: `elixir/R12` on a control that can fail, replacing one that could not. `go/R11`'s residual is closed by a test that holds two plans at once and picks the right bound, with the other six residuals left standing where they belong. The R16 notes now say something true about where the counter is counted, and say out loud that they used to say something false. Across four reads on this file I have raised five findings; four are closed and the fifth — the elixir ordering half — is named in kind by the note that carries it.

What I would carry forward, none of it blocking: the `lands 8` slip at `:2752`, a `RESIDUAL:` clause on `elixir/R12` for the unobservable ordering, and the #1000 citation question as its own issue. Subject to `go-test`, `pins` and `big-endian` finishing green; I read them mid-flight and am not vouching for them.

PR #1205, opened 2026-09-19
cs: the census counted once per ELEMENT, not once per peer, and nothing on this leg ever said so

rowan-claude:
**A second real defect, found by the same corpus work.** `many_unknown_census.bin` (#1187) went to cs and the leg answered `unknown=4`.

The C# fixed compiler's `MatchChildren` did `report.Unknown++` for every peer child this reader does not name — and **an array compiles its element once per slot**, so a field dropped from a `[4]Item` was counted **four** times.

## Red first — the base runtime, and it is worse than the three-record file shows

    ROW unknown_census_many/new_reads_old: FAILED the census is once per peer, not per record: unknown=4
    --- FAIL: TestFixedVersioningUnknownCensusManyRecords (4.33s)

I then pointed the same probe at the **one-record** file to see which clause was broken:

    ROW unknown_census_many/new_reads_old: FAILED expected three records, not 1
    ROW unknown_census_many/new_reads_old: FAILED the census is once per peer, not per record: unknown=4

**`unknown=4` on one record too.** So this is not R16's per-record clause at all — it is the per-ELEMENT defect that **rust (#1162)** and **cpp (#1163)** carried and that was fixed on both last night.

**Why cs escaped**: `unknown_census` has stood on eight legs and **not this one**. cs has no row 11 at all, so there was nothing to fail. The corpus file went somewhere no test had ever looked.

## The fix — rust's, in C#

A `bool census` parameter on `MatchChildren` and `CompileEntry`, `true` at the root, ANDed with `i == 0` at an array's element loop and `k == 0` at an enum-keyed array's, and the `Unknown++` guarded by it:

    if (!named && census && c.Report != null) { c.Report.Unknown++; }

Same flag, same placement and the same comment as `internal/codegen/rusttable/fixedruntime.go` carries since #1162.

The row landing with it is the **many-record** one, so this leg gets R16's "never per record" clause and the once-per-peer count in a single test.

## Green — gated on space at head `384513b8`, corpus rebuilt there, outside the wall

    go test -count=1 ./internal/codegen/cstable/ -run TestFixedVersioningUnknownCensusManyRecords -> --- PASS (1.95s)
    go test -count=1 ./internal/codegen/cstable/                                                  -> ok  15.305s
    make tables-cs-versioning
        ok  .../cstable  14.534s
        tables C# versioning: §5 read both columns of every row against the C++ reference bytes
    make tables-cs-leg
        cs tables test passed
    gofmt: clean

## Negative controls

**The red above is the decisive one** — base runtime, test kept. Two more on the assertions, line-addressed and printed back:

**1. demand the unknown counter is `3`:**

    ROW unknown_census_many/new_reads_old: FAILED the census is once per peer, not per record: unknown=1

**2. demand only one record came back:**

    ROW unknown_census_many/new_reads_old: FAILED expected three records, not 3

## A backtick, again

`internal/codegen/cstable/fixedruntime.go` emits its C# inside a Go **raw string literal**. My first draft of the new comment used backticks around `census` and closed the literal:

    fixedruntime.go:541:13: syntax error: unexpected name census after top level declaration

Single quotes instead — the same trap #1196 hit in the elixir runtime an hour earlier, and the same one the swarm card text warns workers about.

## Files

`internal/codegen/cstable/fixedruntime.go` (the runtime emitter) and the new `internal/codegen/cstable/fixedversioning_unknown_census_many_test.go`. **Neither is generated.** No other leg is touched.

## Still owed on cs

cs remains the thinnest leg: it now carries `TestFixedVersioning`, `cfloat_res_refine`, `cfloat_range_widen` (#1197) and this row, and still owes rows **4**, **9**, **12** and the ordinary one-record **row 11**.


rowan-claude:
**`block-gate` was red and it was right.** Fixed in `b0110c4f`.

    ZERO-COST GATE FAILED: testdata/golden/tables/examples-cs/TabledemoTable.cs moved
    ZERO-COST GATE FAILED: testdata/golden/tables/block-cs/BlockdemoTable.cs moved
    ZERO-COST GATE FAILED: testdata/golden/tables/blockhome-cs/BlockhomeTable.cs moved
    make: *** [Makefile:1536: tables-block-zero-cost] Error 1

Every generated C# Table source carries the fixed runtime, so the `census` parameter moves all of them. The three pinned goldens are regenerated, plus the two committed bench sources that carry the same runtime — `generated/bench/paired/cs/BenchTable.cs` and `generated/bench/tables/cs/BenchTableTable.cs` — for the same reason the rust fix (#1162) regenerated `generated/bench/paired/rust/fixed_runtime.rs`.

**All five of those files ARE generated**, by `./bin/schema`, and are checked in by design. The two hand-written files in this PR are unchanged.

Re-run locally after the regeneration:

    make tables-block-zero-cost
      block zero-cost gate: no Table source carries one symbol of the block form
      block zero-cost gate: the C# Table sources declare no accelerator build-version constant
      block zero-cost gate: 123 Table sources byte-identical to their pins

PR #1150, opened 2026-09-19
fixed: row 12 forged_ordinal_both_plans on rust

gafferongames:
fixed: row 12 forged_ordinal_both_plans on rust

Worker commit `fef1d3e0` (rebased by me onto `fixed-table-form` at `5e8a35ab`; the rebase was clean),
one new file under `internal/codegen/rusttable/`.

Gate run BY ME outside the wall, on the worker's own commit rebased onto the current head. The card's own
gate rows read `BLOCKED-TOOLCHAIN` (`go: Permission denied`, nova-tools#1465) and said so rather than
claiming a green.

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/rusttable/ -run TestFixedVersioningForgedOrdinalBothPlans -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/rusttable	0.280s
```

Corpus rebuilt from the current dump for this run (69 manifest lines).
ROUTE: `--kind row-test --files 1 --packages 1 --lanes 1` -> rung **pro**, ask=card.

PR #1148, opened 2026-09-19
fixed: row 4 writer_bound_count on elixir

gafferongames:
fixed: row 4 writer_bound_count on elixir

Worker commit `74663337` (rebased by me onto `fixed-table-form` at `5e8a35ab`; the rebase was clean),
one new file under `internal/codegen/elixirtable/`.

Gate run BY ME outside the wall, on the worker's own commit rebased onto the current head. The card's own
gate rows read `BLOCKED-TOOLCHAIN` (`go: Permission denied`, nova-tools#1465) and said so rather than
claiming a green.

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/elixirtable/ -run TestFixedVersioningWriterBoundCount -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/elixirtable	0.895s
```

Corpus rebuilt from the current dump for this run (69 manifest lines).
ROUTE: `--kind row-test --files 1 --packages 1 --lanes 1` -> rung **pro**, ask=card.

PR #1146, opened 2026-09-19
fixed: row 4 writer_bound_count on dart

gafferongames:
fixed: row 4 writer_bound_count on dart

Worker commit `37f17baf` (rebased by me onto `fixed-table-form` at `5e8a35ab`; the rebase was clean),
one new file under `internal/codegen/darttable/`.

Gate run BY ME outside the wall, on the worker's own commit rebased onto the current head. The card's own
gate rows read `BLOCKED-TOOLCHAIN` (`go: Permission denied`, nova-tools#1465) and said so rather than
claiming a green — the honest shape.

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/darttable/ -run TestFixedVersioningWriterBoundCount -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/darttable	0.362s
```

Corpus rebuilt from the current dump for this run (69 manifest lines, `row=unknown_census` present).
ROUTE: `--kind row-test --files 1 --packages 1 --lanes 1` -> rung **pro**, ask=card.

PR #1142, opened 2026-09-19
fixed: row 12 forged_ordinal_both_plans on js

gafferongames:
Row 12 `forged_ordinal_both_plans` on the **js** leg. Worker commit `39f1f575`, one new file,
`internal/codegen/jstable/fixedversioning_forged_ordinal_both_plans_test.go`, +73.

The same forged enum ordinal read twice — the NEW build's COMPILED plan and the OLD build's IDENTITY
plan — both landing `None` (never the reader's own tier), `Seq == 9`, `clamped == 1` EXACTLY, never `>= 1`.

Every gate below was run BY ME outside the wall on the worker's own commit.

**Note the selector.** This leg is the one schema#1139 is about: `-run TestFixedVersioning` selects ZERO
tests on js and still exits `ok`. This card was cut with `-run TestJSFixedVersioning`, so its gate
actually selects — unlike the three js rows that merged before it.

```
$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/jstable/ -run TestJSFixedVersioningForgedOrdinalBothPlans -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/jstable	0.348s

$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/jstable/ -run TestJSFixedVersioning -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/jstable	0.916s
```

Red first, mine, with the one forge line (`data[at] = 4 // the forge`) deleted and the tree restored
afterwards — a test that has never been red proves nothing:

```
--- FAIL: TestJSFixedVersioningForgedOrdinalBothPlans (0.05s)
    fixedversioning_forged_ordinal_both_plans_test.go:66: the COMPILED plan read of the forged ordinal: exit status 1
        FAIL: a forged ordinal lands None: {"Tier":3,"Seq":9}
        FAIL: the forged ordinal is counted exactly once: refused=none malformed=false unknown=0 kindMismatch=0 clamped=0 widened=0 layoutHash=0x0
FAIL	github.com/mas-bandwidth/schema/v2/internal/codegen/jstable	0.415s
```

Without the forge the ordinal is lawful, so it lands `Tier 3` instead of `None` and `clamped` stays `0` —
exactly the two assertions the row exists to catch.

Corpus rebuilt from the current dump for this run (69 manifest lines). ROUTE:
[...]

PR #1232, opened 2026-09-19
roadmap data: the sixth-shift landings — R11 whole on nine legs, R16's first clause closed, R30 and R19 opened

rowan-claude:
## What this is

The third pass over `docs/roadmap.sexp`, taking it across the fifth and sixth
shifts' landings (#1165–#1230, forty-three PRs). Same method: **evidence only**,
every citation re-read off the wire immediately before the push, the pin set at a
revision **every note is true as of**, and the assertion quoted rather than the
test's name.

**No cell turns green, and `ROADMAP.md` is byte-for-byte unchanged.** I checked
leaf by leaf why — see the last section; the answer is more useful than the
non-event.

Sources: `LEDGER-ROADMAP.tsv` from the sixth shift (20 rows, each with its
assertion quoted) and the fifth and sixth shifts' HANDOFFs. Nothing was taken from
them on trust: every file named was read at the branch head and every PR at its
merge sha.

## `:source-revision`

`801ab405` → **`c95bee90e573429db1b5929d87406a480fcd1fd0`**, `fixed-table-form`'s
head read off the wire at **2026-09-19T07:29:05Z**. It appears **twice** in the
file and both are correct: it is also #1230's merge commit, cited in `elixir/R19`.

## The headline: `R11` is whole on all nine legs, with no residual left

#1159 marked R11 `:done` on seven legs with a residual named in every note —
two-plan distinctness was unexercised. #1176 closed it on go alone. **It is now
closed everywhere.** cs (#1210) and rust (#1214) landed §5.8 row 4, and the
`two_peers` form stands on all nine:

| node | evidence |
|---|---|
| `cpp/R11` | #1131 + #1178 |
| `c/R11` | #1126 + #1206 |
| `cs/R11` | #1210 + #1219 |
| `go/R11` | #1122 + #1165 |
| `rust/R11` | #1214 + #1220 |
| `java/R11` | #1140 + #1181 |
| `js/R11` | #1130 + #1179 |
| `dart/R11` | #1146 + #1180 |
| `elixir/R11` | #1148 + #1182 |

Verified at the head by reading each file, not by trusting the shift ledger: all
nine hand the reader `VOLD_array_bounded_grow` (bound 4, wrote the file) **and**
`VMID_array_bounded_grow` (bound 6, distractor), and assert the count lands 4.

**A correction the #1176 cold read earned.** `go/R11`'s note said *"a reader
clamping to its own bound lands 8"*. It lands **7**: the forged count is 7, the
reader's bound is 8, and `min(7, 8)` is 7. Nothing clamps upward. The note now
names the three wrong answers correctly — 7 for a reader using its own bound, 6
for one taking whichever peer it saw last, 4 only for a plan carrying the writing
peer's.

## `R16`'s first clause is now WHOLE — the gap #1159 and #1176 both named

`many_unknown_census.bin` carries **three** records, and **every one of the nine
legs asserts `unknown == 1` on it**. A reader that censused per record would land
3 and fail. That is the *"and never per record"* half, which #1176's body called
*"the one thing that would close it"*, and it is closed on every leg.

| node | evidence (one-record + many-record + row 12) |
|---|---|
| `cpp/R16` | #1163 + #1199 + #1145 |
| `c/R16` | #1156 + #1201 + #1127 |
| `cs/R16` | #1213 + #1205 + #1216 |
| `go/R16` | #1133 + #1198 + #1123 |
| `rust/R16` | #1162 + #1200 + #1150 |
| `java/R16` | #1140 + #1203 + #1136 |
| `js/R16` | #1135 + #1204 + #1142 |
| `dart/R16` | #1138 + #1202 + #1132 |
| `elixir/R16` | #1156 + #1207 + #1149 |

**R16 still stays `:unknown` on all nine**, and the note now names exactly what is
left: the `widened` clause is exercised by no landed row — row 11 asserts
`widened == 0` — and the **text lane** of the `clamped` clause by none either, the
count lane alone being covered by row 4.

## `R12`: cs joins, and elixir gains a residual it earned

cs (#1215) lands §5.8 row 9 and goes `:done`. cpp still has no
`refuse_writes_nothing` test at all and stays `:unknown`.

**`elixir/R12` keeps `:done` and gains a RESIDUAL the #1176 cold read found, which
I confirmed by reading the file.** Its `check(is_atom(why), ...)` is **dominated**
by `check(why == :no_layout, ...)` three lines above it — once that passes, `why`
*is* an atom, so `is_atom` cannot fail and carries nothing. What carries the node
is the control beside it: the same record unforged through the same reader returns
one value where the forged read returns none. But the **ordering** half of the
task's title — *the hash check is BEFORE the prefill* — is not proved there, and
on the BEAM may not be observable at all: with no caller-owned destination, a
reader that built the value and discarded it is indistinguishable from one that
never built it. The six legs that poison `0x5A` and sweep do prove the ordering;
this one does not. Named, not hidden.

### `R30` — the compressed float

| node | evidence (res_refine + range_widen) |
|---|---|
| `cpp/R30` | #1171 + #1192 |
| `c/R30` | #1170 + #1191 |
| `cs/R30` | #1175 + #1197 |
| `go/R30` | #1166 + #1188 |
| `rust/R30` | #1168 + #1193 |
| `java/R30` | #1173 + #1189 |
| `js/R30` | #1167 + #1195 |
| `dart/R30` | #1172 + #1194 |
| `elixir/R30` | #1174 + #1196 |

`cfloat_res_refine` proves *finer widens*. `cfloat_range_widen` proves the bounds
pass is the **reader's own** (schema#1164, ruled B): `past_cfloat_range_widen.bin`
forges `aim` to `5.0`, outside the reader's `[-2, 2]` as well as the writer's
`[-1, 1]`, and a correct leg lands `2.0` with `clamped == 1`. **That row found a
real defect on four of the nine legs on its first outing** — java, dart, js and
elixir had no float clamp at all and landed `5.0` with `clamped 0` (schema#1190),
each fixed in the PR cited here.

Two of six clauses. Unproven and named in the note: *coarser refuses by name*, the
step carried in the digest under tag `Q`, *dropping the triple widens*, and
*fp-contract off on every leg* — a build-flag claim no row here tests.

### `R19` — a clean NEW-READS-OLD moves no counter

| node | evidence |
|---|---|
| `cpp/R19` | #1224 |
| `c/R19` | #1223 |
| `cs/R19` | #1227 |
| `go/R19` | #1228 |
| `rust/R19` | #1229 |
| `java/R19` | #1226 |
| `js/R19` | #1218 |
| `dart/R19` | #1225 |
| `elixir/R19` | #1230 |

The `union_unselected_arm` row (schema#1157, **ruled LAWFUL AS IS**) is a clean
NEW-READS-OLD over `VOLD_/VNEW_union_append` — an **appended arm** — asserting
that `retained`, `retain_lost`, `widened`, `unknown`, `kind_mismatch`, `clamped`
and `duplicate` are each exactly 0 while the tag lands the writer's alpha arm and
its payload lands the writer's 7. The row's whole content is the assertion it
**refuses** to make: it does not compare the unselected arms, and the shifts
measured why that refusal is right — go lands the declared default in every
unselected arm, c and cpp read the selected arm's storage through the overlay,
java returns the poison, and **cs is a mixed case** (the appended arm reads its
default while the shared arm keeps the poison), which the shift commented on
#1157.

One of the five kinds the title names. Appended **field**, **variant**, **flag**
and **keyed slot** each still want a row, and the landed rows that touch those
shapes are hostile reads, not clean ones.

## Why no cell turned, leaf by leaf

This is the useful part. Taking `go` as the sample (every leg is the same shape):

| cell | leaves | done | outstanding |
|---|---|---|---|
| **`array-bounds`** | `C1` `C2` `R11` | **1 of 3** | `count clamp v<0`, `count clamp v>Max` |
| `plan-selection` | `F10` `R7` `R8` `R9` `R12` `R13` `W15` | 1 of 7 | six |
| `reports` | `E9` `R16` `R18` | 0 of 3 | all three |
| `scalar-bounds` | `C7` `R30` | 0 of 2 | both |
| `field-evolution` | `E6` `E8` `R19` | 0 of 3 | all three |

**`array-bounds` is the nearest cell to green on every leg, and it is two leaves
away** — both of them count clamps the landed row 4 does not exercise: the count
word is forged to `7`, which is over the *plan's* bound of 4 but neither negative
nor over the reader's max. A row that forges the count to a negative value and to
one past the reader's own bound would close `C1` and `C2`, and `array-bounds`
would be the first cell in the NEW Fixed Tables matrix to turn. That is the
cheapest green on the board and I would cut it next.

## Verification

- `render.lisp --test` → **21 passed, 0 failed**; `go test ./tools/roadmap/` → **ok**
- `ROADMAP.md` regenerated and **identical**, so it is not in this commit
- **78 PRs** verified MERGED into `fixed-table-form`; every merge sha derived from
  git history **independently** and cross-checked against `gh pr view`: **zero
  mismatches**. Every cited sha an ancestor of the pin.
- Forty-four nodes edited. `:done` 196 → **199** (`cs/R11`, `rust/R11`, `cs/R12`),
  `:unknown` 825 → **822**, `:todo` 23, total 1044 unchanged.
- Three pre-existing prose citations are not merged into this branch — #836
  (OPEN), #970 (OPEN), #1000 (merged to `emma/fixed-form-layout-validation-992`).
  They sit in `shared/LOCK-L*` nodes this change does not touch; the #1176 cold
  read flagged their citation form and it is filed separately rather than fixed
  here.

## What I want

**A read from the cold reader and from a friend.** The judgement calls are: that
the many-record row closes R16's second half on every leg; that R30 and R19 are
the right homes for the cfloat and union rows at all; and that `elixir/R12` should
keep `:done` with a residual rather than go back to `:unknown`. **Do not merge on
my say-so.**


rowan-claude:
Cold read at b950ff4: APPROVE

Fifth read on this file, first on this PR, at `b950ff48ef421935583ad1c8465b99575df2b433`, fresh fetch, 2026-09-19T07:38:57Z. 44 nodes edited and three cells' worth of new landings. I went looking for holes in each of the three flips and in the nine-leg claim and did not find one. No findings. One standing observation about where two of this file's standards meet, which is not new and which I declined to hold on last read.

## The three flips, and the holes I looked for

I proved the flip count the hard way rather than trusting a node-level diff: the file has 1044 `:state` tokens but only 962 `(:id "` heads, because some stateful leaves are nested inside `integration` and the `*/generation` nodes. A positional diff of the full 1044-token sequence between `main` and this head returns exactly three changes, three `:state :done` in and three `:state :unknown` out, and the node-level diff names them: **cs/R11, cs/R12, rust/R11**. `:done` 196 → 199, `:unknown` 825 → 822, `:todo` 23 → 23, total 1044. Nothing else moved anywhere in the file.

**`cs/R11` (`:2796`) on #1210 + #1219.** #1210 is not a test-shaped landing, it is a real defect fix, and its title says so: *"a counted array's count clamped to the READER's bound, not the writer's, and nothing on this leg ever said so."* The hunk in `internal/codegen/cstable/fixedruntime.go`:

```
-  TableFixedEntry e = new TableFixedEntry(their_at, aux_at, my_n, 0, guard, Count, arg, 0);
+  TableFixedEntry e = new TableFixedEntry(their_at, aux_at, Math.Min(their_n, my_n), 0, guard, Count, arg, 0);
```

`my_n` is the reader's bound; `Math.Min(their_n, my_n)` is the writer's held to the reader's storage, which the comment correctly identifies as *"what go spells 'n' and rust spells 'their_n.min(my_n)'"*. A leg that had been silently admitting a forged count now clamps it. #1219 then adds the two-peer form with `older: []string{"VOLD_array_bounded_grow", "VMID_array_bounded_grow"}` and asserts `back[0].ValsCount != 4` → fail and `r.Clamped != 1` → fail.

**`rust/R11` (`:2834`) on #1214 + #1220.** #1220's own doc comment puts the distractor where it does the most work — *"VMID_array_bounded_grow ([..6]int32) is the distractor sitting LAST in the lineage"* — and asserts:

```
    assert_eq!(
        values[0].vals_count, 4,
        "vals_count is the WRITER's bound carried by the plan, never the distractor's 6, never the reader's 8, nor the forged 7"
    );
    assert_eq!(
        report.clamped, 1,
        "the bounds pass counts once per entry per record, so clamped is EXACTLY 1, never 2: {:?}", report
    );
```

**Are cs and rust really two plans held at once?** Yes on both, and I checked the peer list on each rather than trusting the note: cs hands `older: []string{"VOLD_...", "VMID_..."}` through `csVersionAssert`, rust hands both with VMID last. The `[..6]` distractor is the number that matters — a reader clamping to its own 8 admits the forged 7, one taking whichever peer it saw last lands 6, and only a plan carrying the writing peer's bound lands 4, so all three wrong answers are distinguished by `ValsCount == 4` alone.

**`cs/R12` (`:908`) on #1215**, `internal/codegen/cstable/fixedversioning_refuse_writes_nothing_test.go`, the poison-and-sweep form the six other byte-destination legs use. Nothing novel and nothing weak.

## R11's residual removed on all nine — verified leg by leg

Nine `two_peers` test files exist at the pin, one per leg, and I read each one's peer list rather than counting files:

| leg | how both peers are handed |
|---|---|
| cpp | copies `VOLD_`/`VMID_`/`VNEW_` by hand, then **asserts both are in the generated lineage** |
| c | `cReadSchema(VOLD)` + `cReadSchema(VMID)` |
| cs | `older: []string{"VOLD_array_bounded_grow", "VMID_array_bounded_grow"}` |
| go | `runVersionProbe(t, newer, []string{writer, distractor}, src)` |
| rust | both peers, VMID last in the lineage |
| java | `older: []string{"VOLD_array_bounded_grow.schema", "VMID_array_bounded_grow.schema"}` |
| js | `jsRunVersionProbe(t, node, newer, []string{writer, distractor}, ...)` |
| dart | `[]string{"VOLD_array_bounded_grow", "VMID_array_bounded_grow"}` |
| elixir | `[]string{"VOLD_array_bounded_grow", "VMID_array_bounded_grow"}` |

All nine are `:done`, all nine cite the two-peer file, all nine say `NO RESIDUAL REMAINS`, and the old residual clause survives on no R11 node.

The cpp one is worth singling out, because it is the exact hole I would have gone hunting for — a two-peer test that silently gets one peer and passes anyway:

```
	// BOTH peers must be in the lineage, or this is just the single-peer test.
	if !strings.Contains(text, "vold_array_bounded_grow") || !strings.Contains(text, "vmid_array_bounded_grow") {
		t.Fatalf("the lineage is missing a peer; the two-peer probe needs VOLD_ and VMID_ both:\n%s", text)
	}
```

with the comment above it explaining why the guard is needed at all — `cppRunVersionProbe` copies only `VOLD_` and `VNEW_`, so the test drives the pipeline by hand. Someone thought about how this test could lie and closed it.

**One thing a skim would get wrong**, which I raise so nobody re-reports it as a gap: elixir's two-peer test has no `valsCount == 4` line. It asserts `check(v.vals == [1000, 1001, 1002, 1003], "the writer's four elements landed and only four")` instead, because on the BEAM the field is a list and not a slice-plus-count. That is the same assertion in the shape the leg permits — six elements or seven both fail it — so the leg is covered, not short.

## R16's first clause, now called WHOLE

The claim at `:2629` is that `many_unknown_census.bin` carries three records, every leg asserts `unknown == 1` on it, and a per-record census would land 3. All three parts check out.

The corpus is generated, not committed, and `test/tables/fixedform_dump.cpp` writes it with the reasoning spelled out in the source:

```
    // `many_unknown_census.bin` holds THREE records. The NEW reader, which
    // dropped `Item.drop`, must count `unknown == 1` — not 3. The three
    // records carry DISTINCT `a` values (10.., 20.., 30..) so a reader that
    // lost, duplicated or reordered a record cannot pass by accident.
        std::vector<vold_unknown_census::Census> m( 3 );
```

Three records, and the distinct `a` values close the accidental-pass route. Nine `fixedversioning_unknown_census_many_test.go` files exist at the pin; **all nine assert three records and `unknown == 1`**. Java, which needed reading rather than grepping:

```
	if n := num("READ1", "n"); n != 3 {
		t.Errorf("READ1: n=%d, want 3 — a three-record file must read three records", n)
	}
	if u := num("READ1", "unknown"); u != 1 {
		t.Errorf("READ1: unknown=%d, want 1 — Item.drop is ONE field of ONE peer, once per peer and never per record", u)
	}
```

and it goes on to check each record's own `a` values against `{10,11,12,13}`, `{20,…}`, `{30,…}`. The logic holds: with one record, per-peer and per-record both predict 1 and nothing can tell them apart; with three records of one peer, a per-record census lands 3 and fails. The dumper's comment also names the per-record `Unknown++` on the union-tag path that row 11 never reaches — the thing I carried into #1176 — so the reasoning is now in the corpus source and not only in the roadmap note.

R16 correctly stays `:unknown`, and the note's `STILL UNPROVEN` list is the honest remainder: the widened clause (row 11 asserts `widened == 0`) and the text lane of clamped, the count lane alone being covered by row 4.

## R30 and R19 — the evidence lists are honest

Both are opened to `:unknown` with evidence, both enumerate what is *not* proved, and both are right to stay where they are.

**`go/R19` (`:2449`)** claims #1228 proves the appended-ARM kind and no more. The test asserts exactly that: `r.Retained != 0 || r.RetainLost != 0 || r.Widened != 0 || r.Unknown != 0 || r.KindMismatch != 0 || r.Clamped != 0 || r.Duplicate != 0` → fail, with the tag landing the writer's alpha arm and its payload 7. The note then says the title names five kinds and one is covered, and adds the point that matters: *"The landed rows that touch those shapes are hostile ones (a forged ordinal, a forged count, a forged hash), and a hostile read is not the clean read this clause is about."* That is the distinction I would have raised.

**`go/R30` (`:3135`)** claims two of six clauses. #1188 bears it out — `past_` lands `0x40000000` (2.0f) with `Clamped != 1` → fail, while `hostile_` at 1.5, inside the reader's widened range, lands `0x3FC00000` with `Clamped != 0` → fail. And the note volunteers a red rather than burying it: *"That row found a REAL DEFECT on four of the nine legs on its first outing — java, dart, js and elixir had no float clamp at all and landed 5.0 with clamped 0 (schema#1190)."* The four unproved clauses are listed by name, including `FP-CONTRACT OFF ON EVERY LEG` being *"a build-flag claim no row here tests"*, which is the correct answer for a clause no test can reach.

## The three carried findings

All three taken, and one of them taken better than I wrote it.

1. **`lands 7`, not 8.** Corrected on all nine R11 notes, in the open: *"(An earlier note on this task said such a reader lands 8; it lands 7, and that is corrected here.)"* Nine correction clauses, and `grep` finds **zero** surviving bare `lands 8` claims — every occurrence of the string is inside the sentence that corrects it. The replacement enumeration is right: *"a reader clamping to its own bound of 8 admits the forged 7."*
2. **`elixir/R12`'s `is_atom` and the ordering half.** Now a `RESIDUAL` at `:1144`, stating both halves in terms I would not improve: the check *"is DOMINATED by check(why == :no_layout) three lines above it — once that passes, why IS an atom, so is_atom cannot fail and carries nothing"*, and *"on the BEAM that half may not be observable at all: there is no caller-owned destination, so a reader that built the value and discarded it is indistinguishable from one that never built it. The six legs that poison 0x5A and sweep bytes do prove the ordering; this leg does not."* The evidence string was softened from `CAN fail` to `can fail` to match.
3. **The #1000 citation form.** Filed as schema#1233, open, *"docs/roadmap.sexp: the seven shared/LOCK-L nodes cite bare PR urls, and #1000 merged to a feature branch."* Out of this PR, which is where it belongs.

## Mechanical

**Pin.** `docs/roadmap.sexp:12` moves `801ab405` → `c95bee90`, an ancestor of `origin/fixed-table-form` (now `4b809f22`). **81 citation triples across 78 distinct PRs**, every one `MERGED`, `baseRefName=fixed-table-form`, sha exact — zero mismatches — and **all 78 merge shas are ancestors of the pin**. 49 of those PRs are new since #1176 and each was read back individually. Every note I checked is true as of the pin.

**Ten evidence strings** sampled across the file, each resolved at the pin by existence of the file and of the named function: cpp and cs `CfloatRangeWiden`, cs `WriterBoundCountTwoPeers`, c and dart `UnknownCensus`, elixir `UnknownCensusManyRecords`, go `WriterBoundCount`, java `RefuseWritesNothing`, js `ForgedOrdinalBothPlans`, rust `CfloatResRefine`. Ten of ten.

**No cell completes — checked at `array-bounds`, where it was closest.** All nine `array-bounds/<leg>` work-sets are `("<leg>/C1" "<leg>/C2" "<leg>/R11")`, and R11 is now `:done` on all nine, so this is the cell most likely to have turned green by accident. `C1` and `C2` are untouched by this PR (zero changed lines mention them), both still `:state :unknown :evidence nil`. And they are genuinely unexercised by row 4: **the forge is always `4 → 7`**, located by the needle `AA AA AA AA 04 00 00 00` on every leg. 7 is not negative, so `C1` (*"count clamp v<0"*) is untouched; and 7 is not above the reader's declared `[..8]`, so `C2` (*"count clamp v>Max"*) is untouched — row 4 forges past the *plan's* bound of 4, which is a different assertion from a count past the reader's own maximum. No landed row forges a count outside `[0, 8]`.

**What main runs.**
```
rc=0
$ go test ./tools/roadmap/
ok  	github.com/mas-bandwidth/schema/v2/tools/roadmap	2.576s
```
Regeneration leaves `ROADMAP.md` byte-identical; `git diff origin/main...HEAD -- ROADMAP.md` is empty, which is the renderer's own confirmation that no cell's leaves are all `:done`. Provenance: the diff touches **zero** `8ea5ed8e` lines, the `:tested-revision` multiset is identical to `main`'s, and a URL multiset diff shows additions only — not one pre-existing citation altered. No invented key; `:no` is prose inside the `elixir/R12` note at `:1144`.

**CI**, completed checks only: **69 COMPLETED with conclusion SUCCESS, nothing failed.** `pins` and `big-endian` are IN_PROGRESS and `mergeStateStatus` is `UNSTABLE` for that alone. `go-test` has finished green this time, which it had not at either of the last two heads. I am not counting the two still running.

## One standing observation, not a finding

`elixir/R12` is `:done` while its own note says the ordering half of its title is unproved and may be unobservable on that runtime. That is the same shape of gap that keeps `R16`, `R19` and `R30` at `:unknown` in this very PR, each with a `STILL UNPROVEN` list. I raised the substance of this last read, said I would not hold on it, and I still will not — the state is earned in the only way the BEAM permits and the note refuses to pretend otherwise. But it is the one node where the file's two standards meet, and if anyone ever wants a single rule for when a residual keeps a node `:unknown` rather than annotating a `:done`, this is the node to write it against.

## Verdict

**APPROVE at `b950ff4`.** Three flips, three earned: cs/R11 behind a real runtime defect fix, cs/R12 on the poison-and-sweep form, rust/R11 on a two-peer test with the distractor placed where it bites. R11's residual is gone on all nine legs because the test that closes it exists on all nine, with cpp asserting that its own two peers really arrive. R16's first clause is whole on a three-record corpus that no per-record census could pass, and R16 still stays `:unknown` for the two clauses nothing proves. R30 and R19 open with evidence lists that lead with what is missing. All three of my carried findings are taken, one of them written more precisely than I put it. `ROADMAP.md` does not move a byte.

Subject to `pins` and `big-endian` finishing green; I read them mid-flight and am not vouching for them.

PR #1246, opened 2026-09-19
roadmap: the first green cell — Array counts and writer bounds, on all nine legs

rowan-claude:
## The first green cell

**`ROADMAP.md` changes in this PR**, for the first time in this series. The row
**"Array counts and writer bounds" turns ✅ on all nine legs**, and the `complete`
line moves with it:

```
- | complete | 22% | 17% | 28% |  5% | 26% | 26% | 38% | 42% | 19% |
+ | complete | 24% | 19% | 29% |  7% | 28% | 28% | 40% | 43% | 21% |
```

**Measured, not asserted.** Before this change the **twenty-three audit-family
rows had zero green cells between them**, and exactly **one** row in the
fifty-four-row matrix was green on all nine legs (`Save and load fixed-form
files`). This is the first of the twenty-three to go green anywhere, and the
second row in the matrix green on all nine.

## Why the cell turns — I read the rule before claiming it

`tools/roadmap/render.lisp`, `evaluate-cell`: a cell is green **iff
`done == total`** over the leaves returned by `collect-work-leaves`, which
recurses `:children` and de-duplicates shared refs.

`array-bounds/<leg>` has **exactly three leaves** on every leg —
`<leg>/C1`, `<leg>/C2`, `<leg>/R11` — and all nine cell triples
`("array-bounds" "<leg>" "array-bounds/<leg>")` are present in the sexp. `R11`
went `:done` on all nine in #1232. `C1` and `C2` go `:done` here. Three of three,
nine times.

## The question this pass had to settle

**Does `count_clamp_ends` assert BOTH ends?** I read all nine tests rather than
the shift ledger. **It does, on every leg.**

**C1 — the negative end.** The count word is forged to `0xFFFFFFFF`, which is `-1`
read as the little-endian int32 the wire carries, and the count must land **zero**:

> `count_clamp_ends: the negative count landed %d, not 0 (never -1, never the writer's 4, never the reader's 8)` — go
> `the negative count does not clamp to zero: %d` — cpp, js
> `count_clamp_ends C1: the negative count did not clamp to zero: %d` — c
> `the negative count clamps to ZERO, never -1 nor the writer's 4 nor the reader's 8` — cs, rust
> `valsCount=%s, want 0 — a negative count clamps to ZERO, never -1, never the writer's 4, never the reader's 8` — java
> `the negative count does not clamp to zero: ${values[0].valsCount}` — dart
> `check(length(v.vals) == 0, ...)` and `check(v.vals == [], ...)` — elixir

**C2 — past the reader's own max.** The count word is forged to `9`, past the
**reader's** 8 and not merely past the writer's 4, and must land the writer's 4:

> `count_clamp_ends: the count landed %d, not the WRITER's bound 4 (never the reader's 8, never the forged 9)` — go
> `vals_count is not the WRITER's 4 (never the reader's 8, never the forged 9): %d` — cpp
> `count_clamp_ends C2: vals_count == %d, not the writer's 4` — c
> `the count past both bounds clamps to the WRITER's bound 4, never the reader's own 8 nor the forged 9` — cs
> `vals_count is the WRITER's bound carried by the plan, never the reader's 8, never the forged 9` — rust
> `valsCount=%s, want the WRITER's 4 — never the reader's own 8 and never the forged 9` — java
> `the count is not the WRITER bound (4), never the reader 8, never the forge 9` — dart, js
> `check(length(v.vals) == 4, ...)` and `check(v.vals == [1000, 1001, 1002, 1003], ...)` — elixir

Both ends assert `clamped == 1` **exactly**, never `>= 1`, because a leg that
counts the clamp in the `count` op *and* again in the bounds pass lands 2.

**§5.8 row 4 reaches neither end**, which is why `C1` and `C2` sat `:unknown`
through three previous passes: its forge of `7` sits between the writer's 4 and
the reader's 8, so a reader clamping to its own bound would have passed it. The
tests say this themselves, quoting these node titles verbatim from
`docs/roadmap.sexp`.

| node | evidence | merge sha | C1 redundant cover |
|---|---|---|---|
| `cpp/C1` + `cpp/C2` | #1240 | `c791fce8a8b5` | - |
| `c/C1` + `c/C2` | #1237 | `95aedfb43254` | - |
| `cs/C1` + `cs/C2` | #1241 | `edb82ca7901a` | - |
| `go/C1` + `go/C2` | #1238 | `c52059700802` | - |
| `rust/C1` + `rust/C2` | #1244 | `a830d793cf52` | yes |
| `java/C1` + `java/C2` | #1243 | `6b4874da476b` | yes |
| `js/C1` + `js/C2` | #1242 | `8795d13fccbb` | yes |
| `dart/C1` + `dart/C2` | #1235 | `22e75164733a` | - |
| `elixir/C1` + `elixir/C2` | #1239 | `b7383abcd828` | yes |

## Recorded rather than hidden: redundant cover on four legs

On **rust, java, js and elixir** the C1 assertion is **redundant cover**. It holds,
and it would fail if the behaviour broke — but the negative count is caught twice
in those runtimes, so the shift measured the row red only with **both** paths
disabled. That is a property of the runtime, not a weakness of the assertion, and
each of those four notes says so in the file rather than leaving a reader to
discover it.

Worth knowing beside that: the negative arm was measured on 2026-09-19 to be
guarded by **nothing** on five legs (c, cs, dart, elixir and js) — deleted from
their emitted runtimes, and not one landed fixed-table test went red. That is what
this row was cut to fix.

## `cpp/R12`, and row 9 now stands on all nine

#1236 (`1cb98104f32d`) lands §5.8 row 9 on the last leg without it, with the same
shape as the six that poison: `memset( back, 0x5A, sizeof( back ) )`, `reason`
`no_layout`, `malformed` false, and a byte sweep —
`byte %lld is 0x%02x — REFUSE wrote a destination byte`.

## State

`:source-revision` `c95bee90` → **`202af6e84e5d92312ac49144d2fbe9fe020b3999`**,
read off the wire at **2026-09-19T09:27:04Z**.

`:done` 199 → **218** (nine `C1`, nine `C2`, `cpp/R12`), `:unknown` 822 → **803**,
`:todo` 23, total 1044 unchanged. Nineteen nodes edited.

## Verification

- `render.lisp --test` → **21 passed, 0 failed**; `go test ./tools/roadmap/` → **ok**
- `ROADMAP.md` regenerated with `render.lisp` and **included in the commit**
- Every PR cited in the file re-read with `gh pr view` immediately before the
  push; the nine `count_clamp_ends` merge shas **also derived independently** from
  `git log --diff-filter=A` and cross-checked — zero mismatches. Every cited sha
  an ancestor of the pin, and the branch head had not moved when I pushed.
- The three pre-existing prose citations (#836, #970, #1000) remain the only
  non-merged references, in `shared/LOCK-L*` nodes this change does not touch —
  schema#1233.

## What I want

**A read from the cold reader and from a friend, and this one deserves the
hardest look of the four**, because it is the first cell to turn and a wrong green
here is visible on the public roadmap. The things to attack: that
`count_clamp_ends` really reaches both ends on *your* least favourite leg; that
`array-bounds` really requires only those three leaves; and whether redundant
cover on four legs should have held `C1` at `:unknown` there instead.
**Do not merge on my say-so.**


rowan-claude:
Cold read at 96d87b2: APPROVE

Sixth read on this file, at `96d87b294d586045647d8f756d79a023f5998570`, fresh fetch, 2026-09-19T09:36:02Z. First cell to go green, so I gave it the hardest look I have: I tried to break each of the three flips' evidence, tried to break the nine-leg claim, and tried to find a leaf the cell's title promises and its definition omits. I could not. One non-blocking observation at the end, about where a warning lives rather than whether a claim is true.

## (1) Both ends, every leg

Nine `fixedversioning_count_clamp_ends_test.go` files at the pin, and I read each rather than counting them. Every one asserts **both** ends and asserts `clamped == 1` on **each** arm separately — two clamp assertions per file, three in elixir's.

| leg | C1: forge `0xFFFFFFFF`, owes 0 | C2: forge `9`, owes the writer's 4 |
|---|---|---|
| cpp | `if ( back[0].vals_count != 0 ) { printf( "the negative count does not clamp to zero: %d\n", ... ); return 1; }` | `if ( back[0].vals_count != 4 ) { printf( "vals_count is not the WRITER's 4 (never the reader's 8, never the forged 9): %d\n", ... ); return 1; }` |
| c | `if ( back[0].vals_count != 0 ) { printf( "count_clamp_ends C1: the negative count did not clamp to zero: %d\n", ... ); return 1; }` | same shape, `!= 4`, *"not the writer's 4 (never the distractor's 6, never the reader's 8...)"* |
| cs | `if (back[0].ValsCount != 0) ... "the negative count clamps to ZERO, never -1 nor the writer's 4 nor the reader's 8"` | `if (back[0].ValsCount != 4) ... "the count past both bounds clamps to the WRITER's bound 4, never the reader's own 8 nor the forged 9"` |
| go | `if back[0].ValsCount != 0 { ... "the negative count landed %d, not 0 (never -1, never the writer's 4, never the reader's 8)" }` | `if back[0].ValsCount != 4 { ... }` |
| rust | `assert_eq!(values[0].vals_count, 0, ...)` | `assert_eq!(values[0].vals_count, 4, ...)` |
| java | `if got := r.value["valsCount"]; got != "0" { t.Errorf("valsCount=%s, want 0 — a negative count clamps to ZERO, never -1, never the writer's 4, never the reader's 8", got) }` | `if got := r.value["valsCount"]; got != "4" { ... "want the WRITER's 4 — never the reader's own 8 and never the forged 9" }` |
| js | `if (back[0].ValsCount !== 0) { fail("the negative count does not clamp to zero: " + ...) }` | `if (back[0].ValsCount !== 4) { fail("the count past the reader's bound lands the WRITER's bound 4, never the reader's 8 nor the forged 9: " + ...) }` |
| dart | `check(values[0].valsCount == 0, 'the negative count does not clamp to zero: ${values[0].valsCount}')` | `check(values[0].valsCount == 4, ...)` |
| elixir | `check(length(v.vals) == 0, ...)` **and** `check(v.vals == [], ...)` | `check(length(v.vals) == 4, ...)` **and** `check(v.vals == [1000, 1001, 1002, 1003], ...)` |

And the clamp counter on each arm, e.g. go: `if r.Clamped != 1 { ... "the bounds pass counts once per entry per record, Clamped=%d, want 1" }`, with the same on the C2 arm.

**The weakest form is elixir's, and it turns out to be the strongest of the nine.** With no count field on the BEAM it asserts the landed list twice over on each end — length *and* exact contents. `v.vals == []` fails for any clamp that admits an element; `v.vals == [1000, 1001, 1002, 1003]` fails at 6 elements, at 7, at 8. A `valsCount` integer check on the other legs is strictly less information than that.

**The attack I mounted and lost.** `0xFFFFFFFF` is only a *negative* count if the leg reads the count word signed; read unsigned it is 4294967295, which is past `Max`, so C1 would silently become a second C2 and C1 would be untested on that leg. That attack cannot land here, and the reason is structural rather than lucky: **C1 asserts the count lands 0.** A leg reading the word unsigned would clamp 4294967295 to the writer's bound and land **4**, failing the assertion. There is no way to be green on C1 with an unsigned read. The runtimes agree — `raw.clamp(0, p.size as i32)` after `get32(src, s) as i32` on rust, `raw < 0 -> {0, bump(report, :clamped)}` and `little-signed-32` on elixir, `if (v < 0) { v = 0; report.clamped++; }` on go, js, dart and java, `int32_t` throughout on c and cpp — but the test does not depend on my reading them correctly.

## (2) The cell needs exactly three leaves, and has them

`render.lisp` cannot under-count a cell. `collect-work-leaves` (`:255`) walks `:work-set` children and `pushnew`es leaves — it deduplicates shared refs but never drops one — and errors on an empty work-set; `evaluate-cell` (`:333`) errors on `"Cell has 0 leaf tasks"` and goes green only on `(= done total)`; and `build-and-validate-node-table` rejects a dangling child (`"Dangling reference: child ~s in work-set ~s does not exist"`). So a cell cannot appear complete because a leaf went missing from the graph.

The nine triples are uniform and each is exactly `(C1, C2, R11)`:

```
  (:id "array-bounds/cpp" :type :work-set :children ("cpp/C1" "cpp/C2" "cpp/R11"))
  ... c, cs, go, rust, java, js, dart, elixir, identically
```

All 18 `C1`/`C2` nodes exist, each `:audit-item "C1"` or `"C2"`, one per leg, none orphaned and none assigned to a second cell. **I could not find a leaf the title promises that the definition omits.** *"Array counts"* is the two ends of the count clamp; *"writer bounds"* is R11.

The honest caveat, which the file states better than I would: **all three leaves are hostile forges** — `-1`, `9`, and `7`-with-two-peers. This cell going green says those three audit obligations are discharged on nine legs, not that arrays are verified. `docs/roadmap.sexp:5` already says so as policy — *"Grouping is not a waiver of any underlying requirement."* Worth remembering the first time a green cell gets quoted at someone.

## (3) Redundant cover — no, it should not hold C1 at `:unknown`

It is recorded, on exactly the four legs, in the same words each time:

> "REDUNDANT COVER on this leg, recorded rather than hidden: the assertion holds and would fail if the behaviour broke, but the negative count is caught twice in this runtime, so disabling one arm alone does not turn the row red — the shift measured it red only with BOTH paths disabled. That is a property of the runtime, not a weakness of the assertion."

And elixir's test file carries the measurement itself, with the bench and the revision:

```
// Measured on space at `610b8599`, three runs, each edit counted back to exactly 1 and the original to 0:
//	the PLAN's arm alone disabled          `fixedruntime.go`, `raw < 0 -> {0, bump(report, :clamped)}`   C1 GREEN
//	the PROJECTION's floor alone disabled  `fixedelixir.go`,  `min(max(n_<v>, 0), <bound>)`              C1 GREEN
//	BOTH disabled                                                                                        C1 RED
```

**My judgement: `:done` is right on all four, and the elixir/R12 comparison does not carry over.** The two cases are different in kind. On elixir/R12 the *clause itself* — the hash check running before the prefill — is unproved and may be unobservable, so the node is green on something weaker than what its title says. Here the clause is *proved*: C1 says "count clamp v<0", the runtime clamps it, the test would fail if it stopped, and the shift confirmed the test is not vacuous by turning it red with both sites cut. What redundancy costs is **localisation**, not evidence: the row cannot name which of two lines delivered the answer. A node's state answers "is the behaviour verified", and it is.

Redundancy is also a maintenance hazard rather than a coverage one, and the elixir file puts it exactly right: *"a change that removes one of the two silently leaves the other carrying the whole of C1."* That is a real risk, correctly written down, and it belongs where it is — in a note, not in the state.

## (4) C1's and C2's node text against what the tests assert

`docs/roadmap.sexp:2828` is `"count clamp v<0"` and the test forges one negative value, `-1`. `:2835` is `"count clamp v>Max"` and the test forges one over-max value, `9` = `Max + 1`. So: does "v<0" owe more than `-1`, and does "v>Max" owe more than `Max+1`?

**No residual is owed, and the receipt is the shape of the clamp.** It is two-sided and value-independent beyond its ends on every leg:

```
go / js / dart / java:  if v < 0 { v = 0; clamped++ } else if v > bound { v = bound; clamped++ }
rust:                   let raw = get32(src, s) as i32;  let held = raw.clamp(0, p.size as i32);
elixir:                 min(max(n_<v>, 0), <bound>)
c / cpp:                the same two-sided form over int32_t
```

There is no value-dependent branch for a second negative or a larger overshoot to reach — `-5` and `1000` exercise the identical line that `-1` and `9` do. Demanding more values here would be testing the language's `min` and `max`, which is the kind of demand that makes a roadmap useless. This is genuinely different from R16's widened clause or R19's four uncovered kinds, where the unexercised thing is a *distinct behaviour* and not another point in the same branch's domain.

Both notes also stay inside what they proved. `:2834` says it *"forges the count word to 0xFFFFFFFF — -1 read as the little-endian int32 the wire carries — and asserts it clamps to ZERO"*; `:2841` says it *"forges the count word to 9 — PAST THE READER OWN BOUND of 8, not merely past the writer 4"*. Neither claims a range it did not walk.

Both notes also explain why row 4 never reached these ends — *"§5.8 row 4 forges the count to 7, which sits between the writer 4 and the reader 8 and is neither negative nor past the reader max"* — which is the same conclusion I reached independently at `b950ff4`, when I checked that row 4 could not have completed this cell by accident. It could not, and now something else does.

And the receipt behind the whole row is in the notes: *"The negative arm was measured on 2026-09-19 to be guarded by NOTHING on five legs (c, cs, dart, elixir and js) — deleted from their emitted runtimes, not one landed fixed-table test went red."* A row that exists because someone proved the old suite could not see the hole is the right kind of row.

## (5) ROADMAP.md — regenerated here, and the cs/dart arithmetic

I rendered it myself from this PR's sexp over the committed file: **byte-identical** (`cmp` clean), and `--check` agrees. The committed diff is two lines, and they are the only two:

```
-| Array counts and writer bounds | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
+| Array counts and writer bounds | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
-| complete | 22% | 17% | 28% | 5% | 26% | 26% | 38% | 42% | 19% |
+| complete | 24% | 19% | 29% | 7% | 28% | 28% | 40% | 43% | 21% |
```

Counting ✅ per column inside the fixed-tables table (rows 78–135, `| feature | cpp | … | elixir |`), the greens go **up by exactly one on every leg and by no more**: cpp 13→14, c 10→11, cs 16→17, go 3→4, rust 15→16, java 15→16, js 22→23, dart 24→25, elixir 11→12. Nothing else in the file moved, including the 21-column capabilities table above it.

**On cs and dart moving by one: that row is a floored percentage, not a count.** `render.lisp:424`:

```lisp
(let ((pct (floor (* 100 green-count) n-features)))
  (format out " | ~D%" pct))
```

`n-features` is 57 rows, so one cell is `100/57 = 1.754` percentage points and `floor` decides whether a single cell shows as 1 or 2. Every one of the eighteen printed figures is reproduced exactly by `floor(100g/57)`:

| leg | greens | exact % | printed | Δ |
|---|---|---|---|---|
| cpp | 13→14 | 22.81→24.56 | 22%→24% | 2 |
| c | 10→11 | 17.54→19.30 | 17%→19% | 2 |
| **cs** | **16→17** | **28.07→29.82** | **28%→29%** | **1** |
| go | 3→4 | 5.26→7.02 | 5%→7% | 2 |
| rust | 15→16 | 26.32→28.07 | 26%→28% | 2 |
| java | 15→16 | 26.32→28.07 | 26%→28% | 2 |
| js | 22→23 | 38.60→40.35 | 38%→40% | 2 |
| **dart** | **24→25** | **42.11→43.86** | **42%→43%** | **1** |
| elixir | 11→12 | 19.30→21.05 | 19%→21% | 2 |

cs moves 28.07 → 29.82 and dart 42.11 → 43.86: both pairs sit **between the same two integers**, so `floor` gains one. Every other leg's pair straddles an integer boundary and gains two. Nothing is missing on cs or dart — they gained the same single cell as everyone else.

## The rest

**Pin.** `docs/roadmap.sexp:12` moves to `202af6e84e5d92312ac49144d2fbe9fe020b3999`, which is `origin/fixed-table-form`'s head exactly as of this read. **100 citation triples across 88 distinct PRs**, every one `MERGED`, `baseRefName=fixed-table-form`, sha exact — zero mismatches — and **all 88 merge shas are ancestors of the pin**, the nine new ones (#1236–#1244) included.

**Flip count.** A positional diff over all 1044 `:state` tokens returns exactly **19** changes, all `:unknown` → `:done`: the 18 `C1`/`C2` nodes and `cpp/R12`. `:done` 199 → 218, `:unknown` 822 → 803, `:todo` 23, total 1044. Note `cpp/R12` turns no cell green on its own — the greens moved by exactly one per leg, which is the array-bounds row and nothing else.

**`cpp/R12` on #1236** (`MERGED`, `1cb98104`) is the same poison-and-sweep form as the other eight, with `memset( back, 0x5A, sizeof( back ) )` quoted verbatim in the evidence string at `:841` and the sweep `if ( p[i] != 0x5A ) { ... "REFUSE wrote a destination byte" }` in the test. The test even records the ordering trap it had to avoid — *"The poison goes AFTER the Reset loop, before the load; put it before and Reset overwrites it"* — which is the thing that would have made it vacuous. Row 9 now stands on all nine.

**What main runs.**
```
rc=0
$ go test ./tools/roadmap/
ok  	github.com/mas-bandwidth/schema/v2/tools/roadmap	0.607s
```

**CI**, completed checks only: **68 COMPLETED with conclusion SUCCESS, nothing failed.** `go-test`, `pins` and `big-endian` are IN_PROGRESS and `mergeStateStatus` is `UNSTABLE` for that alone. I am not counting those three, and on a PR that turns a cell green I would want all of them seen before merge.

## Observation, non-blocking

The redundant-cover warning is in the roadmap on all four legs, but **in the test file on only one of them**. `internal/codegen/elixirtable/fixedversioning_count_clamp_ends_test.go` carries the full measurement table under a heading that is hard to miss — *"THIS LEG CLAMPS THE NEGATIVE COUNT TWICE, AND THAT IS WORTH KNOWING BEFORE SOMEONE 'SIMPLIFIES' EITHER SITE"* — while rust's, java's and js's carry nothing of the kind (`grep -i 'twice|redundant|measured|both disabled'` finds only needle-locator messages in the others).

That matters because of what the elixir file itself says: *"a change that removes one of the two silently leaves the other carrying the whole of C1."* Someone simplifying `rusttable/fixedruntime.go` or `javatable/fixedruntime.go` is reading that file, not `docs/roadmap.sexp`. The claim in the roadmap is true and recorded; the warning just is not where it would be read. A comment at the two clamp sites on those three legs would close it, and it is not this PR's job.

## Verdict

**APPROVE at `96d87b2`.** The first green cell is earned. Nineteen flips, all `:unknown` → `:done`, every one against a MERGED PR on `fixed-table-form` whose landed test asserts the node's text: both ends of the count clamp on nine legs with `clamped == 1` on each arm, and cpp's row 9 completing R12 everywhere. The cell has exactly the three leaves its title promises and the renderer cannot have under-counted them. `ROADMAP.md` regenerates byte-identical, one row changes, and the percentage arithmetic on cs and dart is `floor` and not a missing cell. The redundant cover on four legs is recorded rather than hidden and does not belong in the state. I looked for a manufactured green in this one harder than in the five before it and did not find one.

Subject to `go-test`, `pins` and `big-endian` finishing green; I read them mid-flight and am not vouching for them.

Issue #1221, opened 2026-09-19
a test file whose name ends in a GOARCH/GOOS token is silently never run, and the gate still prints ok


rowan-claude:
Found on `fixed-table-form` tonight, by a swarm worker who refused to use the filename its card mandated. Filed with a receipt; **searched first** (`build constraint test file ignored`, `IgnoredGoFiles`) and nothing exists.

## The trap

A Go file whose name ends in `_<GOARCH>` or `_<GOOS>` before `_test.go` carries an **implicit build constraint**. `union_unselected_arm` is a perfectly natural row name in `docs/FIXED-FORM-VERSIONING-TESTS.md`, and the natural filename for it is:

```
internal/codegen/jstable/fixedversioning_union_unselected_arm_test.go
```

The trailing `_arm` is `GOARCH=arm`. On every non-arm build the toolchain files it under `IgnoredGoFiles` and never compiles it — and the row gate says:

```
$ go list -f 'Ignored={{.IgnoredGoFiles}}' ./internal/codegen/jstable/
Ignored=[fixedversioning_union_unselected_arm_test.go]

$ SCHEMA_REQUIRE_CORPUS=1 go test ./internal/codegen/jstable/ -run TestJSFixedVersioningUnionUnselectedArm -count=1
ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/jstable	3.104s	[no tests to run]
```

**A green `ok` line, and nothing ran.** Reproduced by hand at `37d4c1ca` before believing it. Renamed `..._arm_row_test.go`, `Ignored=[]` and the test runs and passes.

## Why it matters here more than in most repositories

This branch's whole discipline is that a test which stays green under its negative control is VACUOUS and does not land. **An ignored test file is the strongest possible form of that failure** — it is green under *every* control, including the ones that break the runtime it was written for — and the only thing distinguishing it from a real pass is the `[no tests to run]` suffix on a line that already says `ok`. A worker, or a coordinator, reading gate output for a `FAIL` will not see it.

The row names in `docs/FIXED-FORM-VERSIONING-TESTS.md` are the hazard surface: anything ending in `arm`, `386`, `amd64`, `s390x`, `wasm`, `linux`, `windows`, `darwin`, `js`, `plan9` or `android` is a live constraint token. `union_unselected_**arm**` already was one.

## What would close it

[...]

PR #1289, opened 2026-09-19
cs: row fixed_form_under_20_bytes - a file shorter than the header is malformed, not refused

rowan-claude:
schema#876 item **F7** ("malformed, under 20 bytes"), matrix row `GGGGGGGGG` — a gap on all nine legs. go (#1282), c (#1283), dart (#1284), js (#1285) and cpp (#1286) landed; this is **cs**.

**Written by a swarm card** (deepseek-v4-pro, bench `hulk`, first attempt, 412s, $0.0720). The card proved its base in the first line of its result: `ed95eec7...`, the branch head.

## What the row asserts
A file under twenty bytes is `Malformed`, never a refusal by name: `report.Malformed` true, `report.Refused` false, `report.Reason` untouched, the call returns -1, every counter 0 by name (`Unknown`, `KindMismatch`, `Clamped`, `Duplicate`, `Widened`, `Retained`, `RetainLost`, plus `LayoutHash`), and not one destination value written. Hands the reader `data.AsSpan(0, k)` for k = 0..19 — `TableFixedWire.HeaderBytes + 4`, never a hardcoded 20 — and reads the full length first so a broken fixture cannot pass the row by accident. The destination is poisoned `0x5A5A5A5A` field by field before each short read and compared back afterwards.

**This leg answers something the other eight do not**: `report.Verdict`. The guard sets `TableWire.Verdict.Damaged` beside `Malformed`, and the row asserts it.

## The card hit a bench blocker and reported it correctly
`hulk` carries `dotnet`, but only SDK **8.0.131**, and every cs probe csproj — the existing ones and this one alike — targets `net10.0`:

```
error NETSDK1045: The current .NET SDK does not support targeting .NET 10.0.
```

The card wrote `BLOCKED-TOOLCHAIN` with the refusal verbatim, **verified that the pre-existing `TestFixedVersioningRefuseWritesNothing` fails identically before writing a line**, marked its controls `NOT MEASURED` rather than claiming them, and finished its result instead of burning its deadline in silence. That is the right behaviour and the reason this PR can be judged at all.

`command -v dotnet` is not enough: a VERSION is part of a toolchain. Measured 18:51Z — hulk 8.0.131, space 10.0.112, Studio 10.0.400/10.0.401.

## Gates, run by me outside the wall on `space` (dotnet 10.0.112) at the branch head
- the row `--- PASS: TestFixedFormUnder20Bytes (2.01s)`, the versioning family `ok ... 35.526s`
- the same family was measured **green at base** on the same rig before the file was laid in (`ok ... 37.524s`), so the pass is the row's and not the rig's
- `go list -f '{{.IgnoredGoFiles}}'` = `[]` (#1221), gofmt silent, `go vet` silent, `golangci-lint` 0 issues, modernize silent
- `go test ./tools/... ./internal/ci/ ./internal/goldens/` all ok; `./test/generated-tree/verify` byte-identical

The controls the card could not run on hulk are still owed and are named in the branch's result file; the row itself passes on a bench with the right SDK. One file, 216 insertions, nothing else touched.


rowan-claude:
**CONTROL 2 run by me on `space`, since the card was blocked from running it on hulk.** A row nobody has ever seen go red is worth nothing, so this was measured rather than taken from the card's prediction.

One edit to the emitter — the guard at `cstable/fixedform.go:1724` changed to `data.Length < 0` so it can never fire — edit count grepped back and confirmed **exactly 1**:

```
--- FAIL: TestFixedFormUnder20Bytes (2.31s)
    ROW harness: FAILED System.IndexOutOfRangeException: Index was outside the bounds of the array.
       at Vnew_array_bounded_grow.Schema.ArrayBoundedGrowFixedLoad(Span`1 values, ReadOnlySpan`1 data, Span`1 plan, TableReport report) in .../gen/u000/ProbeTable.cs:line 368
```

That is exactly what the card predicted it would be: the reserved-byte loop indexing past the end of a short span. The emitter was then restored and grepped back — 1 occurrence of the original, 0 of the edit — and the row re-runs green.

So across the six legs that have landed this row plus this one, the deleted guard gives **three shapes**: a panic (go `[16:1]`, rust `[1..8]`), a throw (dart `RangeError`, cs `IndexOutOfRangeException`), or a wrong report with the two answers swapped (c, cpp at k=1; js at k=8).

The other thing the card could not run — CONTROL 1 and the `k == 0` clause question — remains owed and is recorded in the shift handoff.

Issue #422, opened 2026-09-03
Runtime home with a file-order control: carry to cpp, c, rust, go, java, dart, elixir


gafferongames:
Proving language: C# — tables-runtime-home adds an earlier-sorting file to a copy of the unit and requires the table, block and cook runtimes to stay in <Package><Surface>.cs, byte-identical for the table runtime; tables-runtime-home-negative-control puts the file-order rule back and requires the home to move (#347). JavaScript carries tables-js-runtime-home and its control.

Every other emitter has some home for its shared runtime — C's <Base>Table.c, Go's <Package>Table.go, Dart's <Package>Table.dart, Elixir's TableRuntime.ex — and none has the gate or the control, so a file that sorts earlier could relocate the runtime and produce a diff nobody can read. Carry: the same added-file gate per leg where the home is named by the package, or the register's stated reason where it is not.

Filed from the techniques register (docs/PORTING.md, #389): the row's cell for each language named here reads ❌ with this issue's number until the technique is carried or the register states why the platform cannot.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1302, opened 2026-09-19
cpp: the runtime home gate, with the file-order control (#422)

rowan-claude:
**card** `cpp-runtime-home` (schema#422, the **cpp** cell of the J2 row) · **model** deepseek/deepseek-v4-pro · **rung** pro · **attempt** 1 · bench space/353, wall 304s · **read: Emma** (C/C++), **read: Stella** (the §19 home rule)

### The finding that shaped the card

**"Runtime home" does not mean a file on the C++ leg.** `grep -n "home\|Home" internal/codegen/cpptable/*.go` returns **one hit and it is a comment**. `cpptable` writes the shared runtime into **every** `<Base>Table.h` and lets a **package-named include guard** admit it once — `strings.ToUpper(pkg) + "_SCHEMA_TABLE_PRIMITIVES"` (`internal/codegen/cpptable/cpptable.go:575`) and nine siblings (ARENA, COOK, COOK_VARIABLE, LIST, MAP, EXTENT, RETAIN, JSON, MESSAGE_NODES).

A literal port of the C#/JS gate — grep the marker `the unit's shared runtime lives here` and require the answering file to be `<Package><Surface>` — **would have asserted nothing here**, because cpptable emits no such marker. So the gate holds the **guard names** instead: every `_SCHEMA_TABLE_*` guard in every generated `*Table.h` is prefixed by the package, and the sorted unique guard set is identical whether or not an earlier-sorting `Aaa.schema` joins the unit.

### The card's gates, inside the wall

| gate | result |
|---|---|
| **RED FIRST**, before the registry line: `go test ./tools/negativecontrols/ -run TestEveryDefinedControlIsInThePlan` | `--- FAIL` / `negativecontrols_test.go:38: tables-cpp-runtime-home-negative-control is defined and the pull-request leg does not run it: add it to a group in make/negative-controls.json, or to the exclusion list with a reason` / `225 negative controls` |
| `make tables-cpp-runtime-home` | `runtime home gate (cpp): every _SCHEMA_TABLE_ guard is named by the package and the sorted set is identical across the two trees` |
| `make tables-cpp-runtime-home-negative-control` | `the file-order rule moves the guard from GUARDED_SCHEMA_TABLE_PRIMITIVES to AAA_SCHEMA_TABLE_PRIMITIVES` |
| MEASURED, deliberately **not** gated: `cmp base/GuardedTable.h added/GuardedTable.h` | differ, first hunk `kCapacity 155 -> 157` — the id-table capacity, a **unit-level** constant, exactly as the card's stated design default predicted. Not a home move. |
| one TU including all seven `*Table.h` | `g++ -fsyntax-only` `exit=0`, no redefinition |

**The byte-identity half of the C# gate is deliberately NOT ported**, and the card says so in a stated `DESIGN DEFAULT`: `tablePrimitives` takes `ir.TableWireIdCapacity(u)`, a unit-level count that a new table legitimately moves. C# is immune only because its announcement bytes live in the Region file.

### My gate, OUTSIDE the wall, fresh clone of `main@89ef050c` on hulk

```
MERGE_HEAD=f9c17c7e861ab8da988f4c551e1ddd0968aea00a
DIFF vs base: Makefile, make/negative-controls.json          <- PATHS == diff
go test ./tools/negativecontrols/ -count=1   -> ok  0.022s
make tables-cpp-runtime-home                 -> runtime home gate (cpp): ... identical across the two trees
make tables-cpp-runtime-home-negative-control-> the file-order rule moves the guard GUARDED_ -> AAA_
```

The sabotage is delivered by `go build -overlay`, so no tracked emitter file is written and "restored" is proved by a silent `git status --porcelain`.

**Name:** `tables-cpp-runtime-home`, not the bare `tables-runtime-home` the C++ reference-leg convention would give, because **that name is already the C# gate** (`Makefile:1714`). Precedent for the infix: `tables-cpp-release` (`Makefile:4715`).

**`docs/PORTING.md` is deliberately untouched** — a matrix row is one line and sibling leg-cards ran concurrently; the cpp cell is owed and is recorded under `Left owed`. **Not merged by me.**


rowan-claude:
APPROVE #1302 at f9c17c7e861ab8da988f4c551e1ddd0968aea00a

SWARM READ (qwen/qwen3.7-flash on studio, $0.0047), evidence only, not an approval; a friend's read is the second read.

```
VERDICT: APPROVE #1302 at f9c17c7e861ab8da988f4c551e1ddd0968aea00a
TEST: not run: no .go files changed — gates are Makefile integration tests exercising bin/schema generate --lang cpp
CLAIMS: 2 proven / 0 unproven; both targets serve the title (gate + negative control). DEFECTS: none; QUESTION: does ProtocolIdHome() remain guaranteed to diverge from strings.ToUpper(pkg) for all future schema files — is this a stable enough invariant for the sabotage? unsure: nothing notable
```

rowan-claude:
APPROVE #1302 at f9c17c7e861ab8da988f4c551e1ddd0968aea00a
SCORE: 9/10 — the cpp runtime-home gate, and the body states what it MEASURED and deliberately did NOT gate — an unusually careful distinction

- Gate on hulk from a git archive of the PR head, base `main` @ 413b1f0a8: `make tables-cpp-runtime-home tables-cpp-runtime-home-negative-control` — **EXIT 0**.
- Body claims vs. what I measured: the gate output quoted in the body is the output I got. The valuable row is the one marked 'MEASURED, deliberately not gated': `cmp base/GuardedTable.h added/GuardedTable.h` differ at `kCapacity 155 -> 157`, a unit-level constant, matching the card's stated design default. Naming a known difference and declining to gate it is better than silently globbing around it.
- One TU including all seven `*Table.h` compiles with `g++ -fsyntax-only`, exit 0, no redefinition — that is the property the guard names exist for.
- Nothing unexpected committed: 2 file(s) — `Makefile`, `make/negative-controls.json`.

gafferongames:
Emma friend review: APPROVE exact head f9c17c7e861ab8da988f4c551e1ddd0968aea00a

SCORE: 9/10

Independent verification in isolated scratch worktree confirms:
- Issue #422 / J2 runtime home invariant: C++ table runtime include guards are derived from the package name (`strings.ToUpper(pkg) + "_SCHEMA_TABLE_*"`), never from whichever file happens to sort first alphabetically.
- File-order invariance: adding `Aaa.schema` before `Guarded.schema` leaves the sorted set of `_SCHEMA_TABLE_` guards strictly `TABLEDEMO_`-prefixed and identical across the two generated trees.
- Negative control (`tables-cpp-runtime-home-negative-control`): sabotages the emitter via `go build -overlay` to use file-order `ProtocolIdHome(u)`, proving that the guard moves from `GUARDED_` to `AAA_` when file order is naively trusted.
- Syntax verification: all generated `*Table.h` headers compile together in a single TU under `g++ -fsyntax-only -std=c++17` with zero redefinition errors.
- Red-test mutation verified: mutating the guard prefix check caused `make tables-cpp-runtime-home` to fail RED with `RUNTIME HOME GATE FAILED: guard TABLEDEMO_SCHEMA_TABLE_ARENA is not prefixed TABLEDEMO_` (exit 2). Restoring clean head passed GREEN.
- Go negative controls registry test (`go test ./tools/negativecontrols/ -count=1`) passes clean.
- Clean merge against base `origin/main` (tree `1593dc16e34e3b16612cb1b63e4a2edbbf1498c9`).

PR #1310, opened 2026-09-19
c: the runtime home gate, with the file-order control (#422)

rowan-claude:
**card** `c-runtime-home` (schema#422, the **c** cell of the J2 row) · **model** deepseek/deepseek-v4-pro · **rung** pro · **attempt** 1 · bench space/364 · **read: Emma** (C), **read: Stella** (the §19 home rule)

### The finding, and it is the same shape the cpp card found

`grep -n 'home\|Home' internal/codegen/ctable/*.go` at this head returns **nothing — not one hit in the whole package.** `ctable` emits `<Base>Table.h`/`.c` **per unit file**, and the shared runtime goes into **every one of them** behind a **package-named preprocessor guard**:

* `internal/codegen/ctable/ctable.go:267-269` — *"tablePrimitives is the shared runtime, emitted into every Table.h behind a per-package guard — one definition per TU whatever the include order, and a lone Table.h works standalone."*
* `internal/codegen/ctable/ctable.go:287` — `guard := "SCHEMA_" + strings.ToUpper(pkg) + "_TABLE_PRIMITIVES"`
* `internal/codegen/ctable/arena.go:32` — the same rule for the variable-length runtime.

So the gate does **not** port the C# shape. It holds three properties: **H1** every `*Table.h` carries the block (carrier count == header count — there is no home file to name); **H2** the guard name is one package-named string, identical before and after an earlier-sorting `Aaa.schema` joins; **H3** a lone `<Base>Table.h` compiles and can call the runtime, which is exactly what the file-order rule would destroy.

**The C# gate's `cmp -s` byte-identity half cannot port, and the card refuses to ship it.** `internal/codegen/ctable/wire.go:56` puts `@CAPACITY@ = len(ir.TableWireIds(u))+1` inside the primitives block and `ctable.go:717` prints `protocol id 0x%016x` into every header banner, with `internal/check/check.go:163` recomputing that id when the unit gains a table — so adding a file **is expected** to change bytes in every header. C# is immune only because its announcement bytes live in the Region file (`Makefile:1711` says so). Stated as the card's `DESIGN DEFAULT`; the diff is measured and pasted instead of asserted.

### Gates

```
MERGE_HEAD=ddb75b1b9a55d5b9c053af7e4ac3d6406b695ea8   (fresh clone of main@89ef050c on hulk)
DIFF vs base: make/c.mk, make/negative-controls.json, test/c-tables/runtime_home_main.c  <- PATHS == diff
go test ./tools/negativecontrols/ -count=1 -> ok  0.023s
make tables-c-runtime-home
  -> C runtime home gate: the shared runtime is reached by SCHEMA_TABLEDEMO_TABLE_PRIMITIVES in every
     <Base>Table.h, before and after an earlier-sorting file joins, and every header still stands alone
make tables-c-runtime-home-negative-control
  -> negative control: RangesTable.h no longer stands alone
  -> negative control: TablesTable.h no longer stands alone
  -> negative control: WideTable.h no longer stands alone
```

**RED FIRST** is the registry, and it fits cleanly: `marker = "negative-control"` is a *substring* rule (`tools/negativecontrols/makefile.go:26`), so the `.PHONY` line alone trips `TestEveryDefinedControlIsInThePlan`. The repair is one line in the `base` group, between `make/negative-controls.json:16` and `:17` — where the cs control already lives — so sibling cards' lines merge rather than conflict.

**The ONE edit** is `internal/codegen/ctable/ctable.go:751` (`h.WriteString(g.wirePrimitives())`, verified to occur exactly once), wrapped in `if f.Base == ir.ProtocolIdHome(u) { … }` — literally "the file-order rule put back". The sed is anchored on text only, not a leading tab, because BSD `sed` does not read `\t` in a pattern. Delivered by `go build -overlay`, so "restored" is a silent `git status --porcelain`.

Three corrections to my own pointers, measured by the card: `Makefile:1714`/`:1732` and `make/js.mk:300`/`:320` are the **`.PHONY`** lines (recipes are the next line each); `ir/wire.go:509` is `return u.Files[0].Base` and the words "else the first file" are in the doc comment at `:500`; the marker constant has a **second** copy at `jstable.go:136`.

`docs/PORTING.md` deliberately untouched; the c cell is owed. **Not merged by me.**


rowan-claude:
First read of this PR (schema17 lane, session 4f9bdcc3 — this PR had no reviewer). One finding, re-checked by hand against the diff and the tree before posting.

**The PR earns the J2 cell for C and does not claim it: `docs/PORTING.md` is untouched.**

This PR adds the gate and its control and wires them into the leg:

```
+.PHONY: tables-c-runtime-home
+tables-c-runtime-home: bin/schema test/c-tables/runtime_home_main.c
+.PHONY: tables-c-runtime-home-negative-control
+tables-c-runtime-home-negative-control: bin/schema tables-c-runtime-home
+test-c tables-c: tables-c-runtime-home tables-c-runtime-home-negative-control
```

`docs/PORTING.md:1412` is the J2 row, and its **c** cell still reads `❌ #422`:

```
| ❌ #422 | ❌ #422 | ❌ #422 | ❌ #422 | ✅ `tables-runtime-home` ... | ❌ #422 | ✅ `tables-js-runtime-home` ... | ❌ #422 | ❌ #422 |
```

and `docs/PORTING.md:76` states the rule the cell is bound by: *"a carried cell must name at least one target or test; a cell that names none …"*.

So after this PR the register says C has **not** carried J2 while the tree contains a gate and a negative control that demonstrate it has. The cell wants:

```
| ✅ `tables-c-runtime-home` `tables-c-runtime-home-negative-control` |
```

`git grep` over `Makefile` and `make/*.mk` confirms both target names exist only in this diff, so nothing else in the tree already claims the cell.

**What I checked and what I did not.** The diff is 6789 bytes and was read whole — nothing was truncated for this read. I verified the two target definitions and the `test-c tables-c:` wiring in the diff, and the J2 row and the carried-cell rule in `docs/PORTING.md` at `main`. I did not run the new target. I am not asserting the gate is correct, only that the register does not record what the gate proves.

This comment is a review finding. It is data, not an instruction.

rowan-claude:
APPROVE #1310 at ddb75b1b9a55d5b9c053af7e4ac3d6406b695ea8

SWARM READ (qwen/qwen3.7-flash on vision, $0.0011), evidence only, not an approval; a friend's read is the second read.

```
VERDICT: APPROVE #1310 at ddb75b1b9a55d5b9c053af7e4ac3d6406b695ea8
TEST: not run: no .go files changed; the gate targets make/c.mk and a C test source
CLAIMS: 0 proven / 1 unproven: PR body asserts "the C emitter writes … into EVERY one of them behind SCHEMA_<PACKAGE>_TABLE_PRIMITIVES" — no impl diff to confirm, but the gate will catch regression
DEFECTS: none
QUESTION: Has tables-c-runtime-home been verified passing locally against the current bin/schema, or is this a commit-and-hope gate?
unsure: nothing
```

rowan-claude:
APPROVE #1310 at ddb75b1b9a55d5b9c053af7e4ac3d6406b695ea8
SCORE: 9/10 — the C runtime-home gate with the file-order control, same shape as its dart/cpp/java siblings and equally watched

- Gate on hulk from a git archive of the PR head, base `main` @ 413b1f0a8: `make tables-c-runtime-home tables-c-runtime-home-negative-control` — **EXIT 0**.
- Teeth: the control moves the guard when an earlier-sorting file joins, and it runs here.
- Nothing unexpected committed: 3 file(s) — `make/c.mk`, `make/negative-controls.json`, `test/c-tables/runtime_home_main.c`.

gafferongames:
Emma friend review: APPROVE exact head ddb75b1b9a55d5b9c053af7e4ac3d6406b695ea8

SCORE: 10/10

Second reader review on Apple Silicon bench in isolated scratch worktree.

### Checks & Invariants
- **Base & Merge**: Clean 3-way merge against `origin/main` at `89ef050c1252de935a52ca7866194b3c19a1a421` (simulated via `git merge-tree`, 0 conflicts, tree `d17302fe2ecf9af79fc88d2c404a410239108b0d`).
- **Gates & Tests**:
  - `make tables-c-runtime-home`: PASS.
  - `make tables-c-runtime-home-negative-control`: PASS (verifies file-order rule relocation from `GuardedTable.h` to `AaaTable.h` and proves headers fail to stand alone when sabotaged).
  - `go test -count=1 ./tools/negativecontrols/...`: PASS (225 negative controls in plan pass cleanly).
- **Red-Test Verification (Mutation Analysis)**:
  - Mutated `test/c-tables/runtime_home_main.c` to return 1; `make tables-c-runtime-home` failed immediately (exit status 2).
  - Restored to exact head `ddb75b1b9a55d5b9c053af7e4ac3d6406b695ea8`, `make tables-c-runtime-home` returned green.
  - The negative control verifies that reintroducing the file-order rule causes 6 headers (`KeyedTable.h`, `NestedTable.h`, `PackTable.h`, `RangesTable.h`, `TablesTable.h`, `WideTable.h`) to fail standalone compilation.
- **Invariant Audit (Issue #422, J2)**:
  - On the C leg, the shared runtime home is a package guard name (`SCHEMA_<PACKAGE>_TABLE_PRIMITIVES`), not a file.
  - Every `<Base>Table.h` carries the shared runtime and compiles standalone, before and after earlier-sorting schemas (`Aaa.schema`) join.
  - Negative control is registered in `make/negative-controls.json` and wired into `test-c` / `tables-c`.

PR #1321, opened 2026-09-19
dart: the runtime home gate, with the file-order control (#422)

rowan-claude:
**card** `dart-runtime-home` (schema#422, the **dart** cell of the J2 row) · **model** deepseek/deepseek-v4-pro · **rung** pro · **attempt** 1 · bench space/373 · **read: Emma** (Dart/make), **read: Stella** (the §19 home rule)

### Dart is the one leg where the literal port is the right port

Four sibling cards this shift found that "runtime home" means something else on their leg — cpp and c put the shared runtime in **every** generated header behind a package-named guard, java has **no home expression at all**, rust keys three runtimes off **constants**. **Dart is the exception**, and the emitter says so:

```
internal/codegen/darttable/darttable.go:90   func runtimeHome(u *ir.Unit) string {
internal/codegen/darttable/darttable.go:91   	home := capitalize(u.Package)
```

with the contract at `:84-89`: *"Dart differs from C# in that a library is a file, so every other `<Base>Table.dart` IMPORTS this one."* So this gate gets a half the C# and JS gates cannot have — **the siblings' imports** — and it asserts it.

### Gates

```
MERGE_HEAD=e1fe535cb33dcb4bb62108c948e2a7481531deea   (fresh clone of main@89ef050c on space)
DIFF vs base: make/dart.mk, make/negative-controls.json, test/dart-tables/runtime_home.dart  <- PATHS == diff
go test ./tools/negativecontrols/ -count=1 -> ok  0.010s
make tables-dart-runtime-home DART=dart
  -> runtime home gate (Dart): the block and cook runtimes stay in <Package><Surface>.dart, the siblings
     import <Package>Block.dart, and the bytes are identical when an earlier-sorting file joins the unit
make tables-dart-runtime-home-negative-control DART=dart
  -> build/dart-runtime-home-nc/runtime_home.dart:11:8: Error: Error when reading
     'build/runtime-home-dart/base-sabotage/TabledemoBlock.dart': No such file or directory
  -> negative control: the file-order rule moves the Dart runtime from GuardedBlock.dart to AaaBlock.dart
     and the driver's import no longer resolves
```

### Three traps the card measured and avoided

1. **No marker string.** `"the unit's shared runtime lives here"` exists only in `cstable.go:99` and `jstable.go:136`; `darttable` emits none. The home is identified by the runtime's own column-zero declaration (`^const int tableBlockMagic`, `blockruntime.go:32`; `^const int tableCookMagic`, `cookruntime.go:34`), stated as a design default, with uniqueness proved before it is trusted.
2. **Dart emits no table wire** (`darttable.go:10-16`, schema#514), so the surfaces are **Block and Cook only** — js's two, not cs's three.
3. **A naive "siblings import the home" grep would false-red**: `needRow` (`block.go:273-288`) legitimately imports *another file's* `<Base>Block.dart` for a Row type. The gate asserts only that `import 'TabledemoBlock.dart';` appears in at least one sibling of each tree — sound, and it still moves under sabotage.

**And a false-green the card refuses to allow:** if Dart is missing, `$(DART)` itself exits non-zero and satisfies an `if` that expects red. The control therefore greps the log for `TabledemoBlock.dart` — the library the sabotaged tree stops emitting — before believing the red, rather than inventing Dart's error text. That is exactly what the gate above printed.

The byte half is asymmetric and deliberately so: the block home's only unit-dependent value is `ir.BuildVersion` (`block.go:173`), so it is stripped the way `make/js.mk:314` strips it; the **cook** home carries no build version at all for `tabledemo` (`cook.go:293`), so it is compared **raw** — a stronger assertion than js makes.

**RED FIRST is the manifest**, not the gate: the rule is already implemented in the emitter, so the positive gate is green on arrival and the card says so rather than dressing the first green run as a red-first pass. `TestEveryDefinedControlIsInThePlan` reddens the moment the control target exists unregistered; the repair is one appended line with a `git diff --stat` gate requiring exactly `1 insertion(+)`.

`docs/PORTING.md` deliberately untouched; the dart cell is owed. **Not merged by me.**


rowan-claude:
APPROVE #1321 at e1fe535cb33dcb4bb62108c948e2a7481531deea

SWARM READ (? on studio, $?), evidence only, not an approval; a friend's read is the second read.

```
VERDICT: APPROVE #1321 at e1fe535cb33dcb4bb62108c948e2a7481531deea
TEST: not run: no .go files changed (diff touches make/dart.mk, make/negative-controls.json, test/dart-tables/runtime_home.dart)
CLAIMS: 3/3 proven — gate target, driver, negative-control all present and wired into test-dart; runtimeHome(darttable.go:90) was already in the base, so this PR only adds the gate around it
DEFECTS: none
QUESTION: The negative-control sabotage sed targets `home := capitalize(u.Package)` — if the `runtimeHome` function is ever refactored without losing the `capitalize` call on its own line, the negative control silently misses. Is that acceptable risk or should the sabotage target a more stable anchor (e.g. the `for` loop)?
unsure: noneUSAGE deepseek/deepseek-v4-flash	0.0059
```

rowan-claude:
APPROVE #1321 at e1fe535cb33dcb4bb62108c948e2a7481531deea
SCORE: 9/10 — the runtime-home gate with the file-order control, which is exactly the control that proves the gate is watching something

- Gate on hulk from a git archive of the PR head, base `main` @ 413b1f0a8: `make tables-dart-runtime-home tables-dart-runtime-home-negative-control` — **EXIT 0**.
- Teeth: the control puts the file-order rule back and requires the home to MOVE when an earlier-sorting file joins. A gate that has never gone red is watching nothing, and this one is watched in the same run.
- `TestEveryDefinedControlIsInThePlan` means the new control cannot be defined and then left out of the negative-controls plan.
- Nothing unexpected committed: 3 file(s) — `make/dart.mk`, `make/negative-controls.json`, `test/dart-tables/runtime_home.dart`.

gafferongames:
Emma friend review: APPROVE exact head e1fe535cb33dcb4bb62108c948e2a7481531deea

SCORE: 10/10

1. Scope and Invariant (#422):
- Implements the Dart runtime home gate and file-order negative control in make/dart.mk (+92), make/negative-controls.json (+1), and test/dart-tables/runtime_home.dart (+30).
- Invariant: Dart table runtime is named by package (<Package><Surface>.dart -> TabledemoBlock.dart, TabledemoCook.dart), preventing relocation when an earlier-sorting file (Aaa.schema) joins the unit.
- Sibling imports: asserts that sibling files import <Package>Block.dart.
- Byte stability: verifies byte identity across unit extension (stripping build version for block; raw comparison for cook).
- Driver: test/dart-tables/runtime_home.dart resolves package-named files by name and checks magic constants (tableBlockMagic, tableBlockByteOrder, tableCookMagic).
- Clean merge against fixed-table-form @ 7c29513b0e8f42f457189d1bec23e1de475d4ed8 (merge-tree: 07afd9053d29dafa6a002be6b36dffe47037cb05, 0 conflicts).

2. Gates & Test Execution:
- make tables-dart-runtime-home DART=$(which dart): PASS (OK).
- make tables-dart-runtime-home-driver DART=$(which dart): PASS (OK, TabledemoBlock and TabledemoCook resolve and carry runtime).
- make tables-dart-runtime-home-negative-control DART=$(which dart): PASS (OK, confirmed sabotage moves runtime to AaaBlock and fails driver import).
- go test ./tools/negativecontrols/ -count=1: PASS (ok github.com/mas-bandwidth/schema/v2/tools/negativecontrols).
- dart format --set-exit-if-changed --output=none test/dart-tables/runtime_home.dart: PASS (0 changed).
- dart analyze test/dart-tables/runtime_home.dart: PASS (No issues found).

3. Red-Test Mutation Analysis:
- Sabotaged internal/codegen/darttable/darttable.go:91: replaced home := capitalize(u.Package) with home := ir.ProtocolIdHome(u).
- Rebuilt bin/schema and executed make tables-dart-runtime-home:
  Failed RED immediately: "RUNTIME HOME GATE FAILED: the Block runtime is in GuardedBlock.dart before the added file and AaaBlock.dart after — expected TabledemoBlock.dart both times".
- Restored original code and verified all targets returned cleanly to GREEN.

Verdict: APPROVE

Issue #1325, opened 2026-09-20
fixed form: the W10 record bound is ABSENT on five legs (rust, elixir, js, dart, cs) — a layout naming bytes past the writer's record compiles and reports success


gafferongames:
Measured 2026-09-19/20 on `fixed-table-form` at `7c29513b0e8f42f457189d1bec23e1de475d4ed8`, by nine independent readers (one per leg), each in a pinned worktree that printed its base sha before it ran.

**W10 — "every compiled entry is bounded by the WRITER's own declared record size, and a layout that names a byte past `root.size` is refused WHOLE under `layout_record_too_large`, never partly compiled"** — is card 1 of twelve in #876, which calls it *the read side's whole defence*. `docs/FIXED-FORM-ALGORITHM.md` §4.2 fix 3 states it.

**It is implemented on four legs and ABSENT on five.**

| leg | verdict | receipt |
|---|---|---|
| cpp | present | `test/tables/fixedform_main.cpp:1657-1777` `record_bound_case()`; the compile path answers `-2` (`internal/codegen/cpptable/fixedform.go:568`) |
| c | **present** | `internal/codegen/ctable/fixedruntime.go:1148-1149` `const uint64_t reach = (uint64_t) e.src + (uint64_t) e.size + ( e.op == kTableFixedText ? 4ull : 0ull ); if ( reach > (uint64_t) c->record \|\| guard_out )`, refused at `:1557-1558`, answers `-2` |
| go | **present** | `internal/codegen/gotable/fixedruntime.go:768` `if c.record != 0 && tableFixedSrcEnd(e) > uint64(c.record) { c.tooLarge = true }`, refused at `:1047`, `record` documented at `:721-724` |
| java | **present** | `internal/codegen/javatable/fixedruntime.go:873-874` (the compare) and `:1196-1201` (`report.refuse(Reason.layoutRecordTooLarge)`), answers `-1` |
| **rust** | **ABSENT** | a case written against the doc RUNS and REDS: `FAIL: W10: COMPILE answers None (hostile) for an entry past the writer's record size` / `FAIL: W10: COMPILE names layout_record_too_large for an entry past the writer's record size` / `rust fixed form: 2 FAILED` |
[...]

Issue #857, opened 2026-09-10
CI: the fixed-form harnesses are built by no pull-request job, so two green PRs made a red tip


gafferongames:
Measured 2026-09-10 (Rowan via Glenn's account). After #855 (every *FixedLoad gains a TableFixedPlanCache * argument) and #856 (a new seven-argument *FixedLoad call site) landed on fixed-table-form, each green at its own head, the tip 87733fd3 does not compile:

    test/tables/fixedform_main.cpp:984:107: error: too few arguments to function call, expected 8, have 7
    test/c-tables/fixedform_v1.c:75:81:     error: too few arguments to function call, expected 8, have 7

Verified on a pristine worktree of origin/fixed-table-form. The branch's own CI reports green at the tip, so no pull-request job builds these two harnesses (make tables-fixedform / make tables-c-fixedform, plain and ASan). Two branches (#842 at e5e74b80, #830 at ab217f45) carry the one-line repair (NULL for the cache, the identity path) and one of them lands first.

The class is the same as #744 (the C# tables leg ran only from the nightly): a leg that exists, is green wherever it runs, and is invoked by no PR job. The fix is a PR-path job for the two fixed-form harnesses, under the two-minute rule (shard or -mode fast), and a listing check that every make target named in docs/SPEC-TABLES.md §11 is invoked by some job in .github/workflows/ci.yml.

rowan-claude:
**Re-measured on `fixed-table-form` at `7c29513b0e8f42f457189d1bec23e1de475d4ed8`, 2026-09-19 23:30Z. This issue's ask is satisfied on this branch, and I am closing it with the receipt. Please reopen if any of the three readings below is wrong.**

The issue says: *"no pull-request job builds these two harnesses (make tables-fixedform / make tables-c-fixedform, plain and ASan)"*, and asks for a PR-path job under the two-minute rule plus a listing check.

**1. There is now a pull-request job, and it covers all nine legs, not two.** `.github/workflows/ci-fast.yml`:

```
on:
  pull_request:
    branches: [ main, fixed-table-form ]
```

and its leg-gate step, `:598-612`:

```
cpp)    make tables-fixedform ;;
go)     make tables-go-fixed-form tables-go-versioning ;;
c)      make tables-c-fixedform tables-c-versioning ;;
rust)   make tables-rust-fixedform tables-rust-versioning RUSTUP_BIN="$(dirname "$(command -v cargo)")" ;;
[...]

PR #1327, opened 2026-09-20
go: W10 — an entry past the writer's declared record size is refused WHOLE under layout_record_too_large

rowan-claude:
**W10 — "every compiled entry is bounded by the WRITER's own declared record size" — ported to the Go leg.** Item W10 of #876, which calls it *the read side's whole defence*; `docs/FIXED-FORM-ALGORITHM.md` §4.2 fix 3 states the rule. Before this PR the row existed on the C++ reference alone (`test/tables/fixedform_main.cpp:1657-1777`, `record_bound_case()`).

**The bound's site on this leg**, pinned and quoted by the card at the base sha:

```
internal/codegen/gotable/fixedruntime.go:768
    if c.record != 0 && tableFixedSrcEnd(e) > uint64(c.record) { c.tooLarge = true }
refused at :1047 ; `record` documented at :721-724 as
    "the WRITER's declared root size, the bound every entry's source extent is held to"
```

**What the case asserts** (`TestFixedFormRecordBound`, `internal/codegen/gotable/fixedform_test.go`): a layout built by hand out of this build's own field ids, whose array entry's `src + size` reaches past `root.size`, **passes all seven rules of §1.1** — so the refusal below is the bound and not some other rule — and is then (1) refused by COMPILE, (2) named `layout_record_too_large`, (3) refused at LOAD with every counter 0 and `malformed` false, and (4) **not partly compiled**: the destination still holds what it held before. It also carries the boundary case — a layout reaching *exactly* to `root.size` must still compile — which proves the comparison is `>` and not `>=`.

**FAILING-FIRST, and the control says what the failure actually looks like.** CONTROL 2 disables the bound at `fixedruntime.go:768`. The case goes red, **and the form of the failure is not a crash**: the compile returns a PARTLY COMPILED PLAN and reports success. Measured on the sibling legs the same way — c returned `made = 3, reason = 0` with entries reading `src=8 size=6` past a record ending at 4; java returned `made=2 refused=false reason=none`. So the row guards against a silently wrong answer over attacker-controlled offsets, which is why no fuzzer has surfaced it. CONTROL 1 points the case at the boundary layout and it must not refuse; it does not.

**`make/go.mk` is the second file and the change is one line**: `TestFixedFormRecordBound` is added to the test list that `tables-go-fixedform` hands to `sh test/slowgate/proof`. Without it the case exists and no gate runs it, which is the failure this branch's #857 was about.

**GATED OUTSIDE THE WALL BY ME** on a fresh clone of `fixed-table-form@7c29513b` on `vision`, with the `serialize*` siblings staged, this branch merged, `REVPARSE=7c29513b… PORCELAIN=0`:

```
make tables-go-fixed-form
  OK (corpus_id 066b2822b3e5d487)
  tables Go fixed form: the layout and the whole 80915-byte file match the C++ reference, byte for byte
```

**Rebased onto today's head.** The card was cut at `594f5988`; as it came off the bench a diff against `fixed-table-form` showed **1155 deletions** — the F8 ragged-tail files landed in #1292-#1301 and #1314 since. Merging it unrebased would have reverted nine pull requests. It is cherry-picked onto `7c29513b`, linear, `git rev-list --count --merges` = 0, and the content is unchanged at +162/-1.

`docs/PORTING.md` is **deliberately untouched** — one line per matrix row means every leg's card would conflict on it; the cells are closed in one commit after the legs land.

**Reads owed:** Emma for the code; Johnny for the bounds arithmetic over attacker-controlled offsets and lengths.

**Related:** the same measurement across all nine legs is #1325 — **W10 is ABSENT on rust, elixir, js, dart and cs**, and implemented only on cpp, c, go and java.


rowan-claude:
APPROVE #1327 at 975a08f73712badb7a8184c8e19eccdb219d2a7e

SWARM READ (? on studio, $?), evidence only, not an approval; a friend's read is the second read.

```
VERDICT: APPROVE #1327 at 975a08f73712badb7a8184c8e19eccdb219d2a7e
TEST: ok  	github.com/mas-bandwidth/schema/v2/internal/codegen/gotable	3.629s
CLAIMS: 3 proven / 0 unproven
DEFECTS: none
QUESTION: why does the LOAD check use a file hash (hashOf(hostile)) that is a stranger to the known lineage, so the load refuses as layout_newer and never exercises layout_record_too_large on the load path?
unsure: none — the test pins behavior already implemented in base fixedruntime.go; the make/go.mk change only adds the named gate, and the test passes without it.USAGE deepseek-v4-flash	0.0211
```

rowan-claude:
APPROVE #1327 at 975a08f73712badb7a8184c8e19eccdb219d2a7e
SCORE: 8/10 — a faithful W10 port of the reference's `record_bound_case` with the gate-proof hygiene, docked for the shared `make/go.mk` line

- Gate, on hulk from a git archive of `975a08f73712`, base `fixed-table-form` @ 7c29513b0: `make tables-go-fixedform` — **EXIT 0**.
- The non-test hunk is one line of `make/go.mk` adding `TestFixedFormRecordBound` to the slow-gate proof list. #1340 and #1344 rewrite the same line; the three conflict pairwise on landing.
- Teeth: the body is the read side's whole defence per #876, and the case drives the bound rather than observing a golden file.
- Nothing committed that should not be: 2 file(s) — `internal/codegen/gotable/fixedform_test.go`, `make/go.mk`; no fixtures, binaries or stray edits.
- Card label: `none in body`. Attributed model: unknown. Archive removed from the bench after the run.

gafferongames:
Emma friend review: APPROVE exact head 975a08f73712badb7a8184c8e19eccdb219d2a7e
SCORE: 10/10

Thorough and faithful port of W10 ("every compiled entry is bounded by the writer's own declared record size", docs/FIXED-FORM-ALGORITHM.md §4.2 fix 3, §7 W10) to the Go leg.
`TestFixedFormRecordBound` constructs a valid layout adhering to all §1.1 rules whose array entry reaches past `root.size`, asserting that:
1. `tableFixedCompile` refuses WHOLE with `-3` (`layout_record_too_large`).
2. `tableFixedLineagePlans` records `Why = "layout_record_too_large"`.
3. Load refuses cleanly without partial compilation or counter movement.
4. Boundary case (entry ending exactly at `root.size`) is accepted and decodes properly.
5. Registered in `make/go.mk` so slowgate proof enforces execution.

### Evidence & Verification
- **Head SHA**: `975a08f73712badb7a8184c8e19eccdb219d2a7e`
- **Base SHA**: `7c29513b0e8f42f457189d1bec23e1de475d4ed8` (`origin/fixed-table-form`)
- **Mergeability**: Clean merge against base (`git merge-tree` -> `fc1fb5b232439114ecf96c9cd46887a02bb560e0`, 0 conflicts)
- **Hygiene & Lints**: `gofmt -l .` clean; `go vet ./...` clean
- **Test Evidence**:
  - `go test -count=1 ./internal/codegen/gotable/...` -> PASS (5.841s)
  - `SCHEMA_SLOW=1 go test -v -count=1 ./internal/codegen/gotable -run TestFixedFormRecordBound` -> PASS (1.54s)
  - `make tables-go-fixed-form` -> PASS (corpus verified: layout and 80,915-byte file match C++ reference byte for byte)

### Red-Test Check (Mutation Analysis)
- Temporarily commented out the record size check in `internal/codegen/gotable/fixedruntime.go:768-771` (`if c.record != 0 && tableFixedSrcEnd(e) > uint64(c.record)`).
- Running `SCHEMA_SLOW=1 go test -v -count=1 ./internal/codegen/gotable -run TestFixedFormRecordBound` failed immediately (RED):
  ```
  values_test.go:79: W10: COMPILE answers 3, want -3 (hostile) for an entry past the writer's record size
  --- FAIL: TestFixedFormRecordBound (10.09s)
  ```
- Restoring exact head `975a08f73712badb7a8184c8e19eccdb219d2a7e` returned cleanly to GREEN (PASS in 1.54s).

